PR #24, opened 2014-08-30
Remove unnecessary links against various libraries.

berenm:
This removes some unnecessary linkage between gemrb executable and CMAKE_DL_LIBS / CMAKE_THREAD_LIBS_INIT on Linux, and adds required linkage between gemrb_core and CMAKE_DL_LIBS.

The patch was written for debian packaging, to fix some lintian warnings about unnecessary linkage, but may not be correct for other platforms. Feel free to adapt / ignore / comment.


lynxlynxlynx:
This wasn't merged because the first part looks wrong and the second part isn't necessary (already done). Today a new version was released, so please recheck with lintian when packaging.

PR #30, opened 2014-10-02
Font/Text system rewrite

bradallred:
obviously this is unreviewable... and extends to many "subcomponents" of the text system such as dialog, several controls (especially TextArea).

Major changes:
- Font/Controls use `std::wstring` now.
- `Font` uses an "Atlas" instead of individual sprites for characters. TTF plugin will dynamically add to its atlas as characters are needed to avoid building a ton of worthless pages. Additionally, fonts (especially BAM based fonts) use "alias" entries to avoid making duplicate glyphs.
- Encoding (i18n) is no longer handled by `Font`. We were previously doing this *every* `Font::Print` pass.
- Python feedback in `Console`
- GUIScript interface for TextArea is radically slimmer
- `Content` class (and subclasses) added for laying out content once instead of doing (clumsy, broken) layout every `Font::Print` pass. 
- Helpful debug flags for `Font` and `Container` classes (rect drawing to visually see how content is being laid out and printed).
- This code is much easier to read/maintain/extend
- Comments (sad that this is a *major* change)

Lots of little text related things we were missing or doing wrong before have been added/fixed (various alignment issues, console feedback, journal format, MTA missing colors, tons of MTA issues, etc)

Hopefully with the "one pass" layout and encoding performance will see a minor boost. havent really checked.

I tested with BG 2 *only*. we should test it in some other games (and more throughly) before merging.


lynxlynxlynx:
the final implementation has been merged. :)

PR #32, opened 2015-01-08
Fixed party AI button to display correct status

DriimGit:
Party AI button was lit when AI was off and unlit when AI was on, switched that.


lynxlynxlynx:
When checking if all this is correct, I found that the problem was in the ai check in core, which is now fixed. This patch turned out to be redundant in the end, but only partly, since it led to a better fix. Thanks!

DriimGit:
Mission accomplished.

PR #33, opened 2015-02-02
Improved castign on rest script

DriimGit:
The script now tries to heal everyone to full instead of just the first injured character.

This request includes a helper function that I ended up not using after revising my script, but left it there for consideration. Excluding that, all releveant changes are in Game.cpp.


gemrb/core/Game.cpp
```diff
@@ -40,8 +40,27 @@
...
 namespace GemRB {
+	
+	
+struct HealingResource {
+	ieResRef resref;
+	Actor* caster;
+	ieWord amount;
+};
+
+struct Injuree {
```
bradallred: "Injuree" isn't a word... you mean Injured

DriimGit: You're right, I slavishly copied it from the original Game.cpp. In fact, if it were a word it'd be the one administering the injuring. I'll change it once with the other yet to be uncovered errors.

gemrb/core/Game.cpp
```diff
@@ -1737,44 +1756,100 @@ void Game::CastOnRest()
...
+					healingspells.erase(spell);
+				}
+			// Here is the place to write the loop for non hp healing spells.
 			}
 		}
 	}
 }
 
+//returns a vector of character with requested state. Container will be empty if none
+bool Game::FindPCs(std::vector<Actor*>& characters, ieDword state)
```
bradallred: why not just build and return the vector here instead of passing a reference and returning bool? NRVO will optimize away the copy.

gemrb/core/Game.cpp
lynxlynxlynx: On Monday 02 of February 2015 07:23:39 Brad Allred wrote:
> >  namespace GemRB {
> > 
> > +
> > +
> > +struct HealingResource {
> > +	ieResRef resref;
> > +	Actor* caster;
> > +	ieWord amount;
> > +};
> > +
> > +struct Injuree {
> 
> "Injuree" isn't a word... you mean Injured
I made it up, so it's clear it is not an adjective with ellipsis. I think it's 
a french pattern originally.

LP

gemrb/core/Game.cpp
DriimGit: Either which way. This way you can use it straight away in a condition. Other option is just to check if the vector is empty.

bradallred: well where is it used? you typically dont need to check if a vector is empty you just iterate it regardless.

DriimGit: I noted in the inital request it is not used at the moment. In chat you previously longed for such a function, as far as I understood, and thus I left it in there.
 
It was used in a previous iteration when a condition of a loop was that there were any injured individuals at all.

gemrb/core/Game.cpp
```diff
@@ -1737,44 +1756,100 @@ void Game::CastOnRest()
...
-						// we have the spell, so cast it on most injured
-						// NOTE: no distinction is made about the potency of spells
-						Actor *injuree = injurees[0];
-						tar->DirectlyCastSpell(injuree, special_spells[specialCount].resref, 0, 1, true);
-						if (injuree->GetStat(IE_HITPOINTS) == injuree->GetStat(IE_MAXHITPOINTS)) {
-							if (!injurees.empty()) {
-								injurees.erase(injurees.begin());
+		if (specialCount != -1 ) {
+			std::sort(wholeparty.begin(),wholeparty.end());
+			specialCount = core->GetSpecialSpellsCount();
```
bradallred: this line looks redundant. you just checked specialCount so it had better be set to the correct value already.

gemrb/core/Game.cpp
```diff
@@ -1737,44 +1756,100 @@ void Game::CastOnRest()
...
+							healingspells.insert(std::pair<ieWord,HealingResource>(special_spells[specialCount].amount,resource));
+						}
+					}
+					ps=ps2;
+				}
+			}
+			std::sort(wholeparty.begin(),wholeparty.end());
+			// Heal who's still injured
+			while (!healingspells.empty() && wholeparty.back().hpneeded > 0) {
+				std::multimap<ieWord,HealingResource>::iterator spell=healingspells.end();
```
bradallred: typedeffing `std::multimap<ieWord,HealingResource>` and other wordy types would be helpful maybe.

gemrb/core/Game.cpp
```diff
@@ -40,8 +40,27 @@
...
 #include "System/DataStream.h"
 #include "System/StringBuffer.h"
+#include <iterator>
+#include <vector>
+#include <algorithm>
 
 namespace GemRB {
+	
+	
+struct HealingResource {
```
bradallred: I dont see why either of these 2 structs are declared here. Their use appears to be completely localized to `CastOnRest()`

gemrb/core/Game.cpp
bradallred: sure, some kind of function would be potentially useful (this isnt exactly what I had in mind), but if its not being used anywhere... plus this is rather incomplete since it would only get injured PCs. the function I had in mind was much more robust. go ahead and remove any code that isnt used by the patch.

gemrb/core/Game.cpp
DriimGit: Nest levels reduced.

gemrb/core/Game.cpp
DriimGit: Will do, it was a skeleton by design since I did not know what else it'd be used for, just it was wanted.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
 		for (int idx = 1; idx <= ps; idx++) {
 			Actor *tar = FindPC(idx);
-			if (tar && tar->GetStat(IE_HITPOINTS) < tar->GetStat(IE_MAXHITPOINTS)) {
-				injurees.push_back(tar);
+			ieWord hpneeded=tar->GetStat(IE_MAXHITPOINTS) - tar->GetStat(IE_HITPOINTS);
+			if (tar && hpneeded > 0) {
+				wholeparty.push_back(Injured(hpneeded,tar));
+			} else {
+				wholeparty.push_back(Injured(0,tar));
 			}
```
lynxlynxlynx: this if check is redundant the way it is now. However, you're not dealing with tar possibly being NULL later, so better not add such "characters" to the vector at all.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
 void Game::CastOnRest()
 {
+	typedef std::multimap<ieWord,HealingResource> RestSpells;
+	typedef std::vector<Injured> RestTargets;
+
 	ieDword tmp = 0;
 	core->GetDictionary()->Lookup("Heal Party on Rest", tmp);
 	int specialCount = core->GetSpecialSpellsCount();
-
+	RestTargets wholeparty;
```
lynxlynxlynx: this name is misleading, since you're later adding only injured party members to it.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
 		SpecialSpellType *special_spells = core->GetSpecialSpells();
+		std::sort(wholeparty.begin(),wholeparty.end());
+		specialCount = core->GetSpecialSpellsCount();
+		RestSpells healingspells;
 		while (specialCount--) {
-			if (special_spells[specialCount].flags & SP_REST) {
-				if (injurees.empty()) injurees = PCs; // enable some nonhealing magic too
-				// check each party member for the spell
+			// Cast multi-target healing spells
+			if (special_spells[specialCount].flags & SP_HEAL_ALL) {
```
lynxlynxlynx: please include SP_REST too, as this bit may get other meanings in other combinations sometime.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
+						for (RestTargets::iterator injuree=wholeparty.begin();injuree != wholeparty.end() ; ++injuree) {
+								injuree->hpneeded-=special_spells[specialCount].amount;
 						}
 					}
 				}
-				ps = ps2;
+				ps=ps2;
+			}
+			// Gather the regular healing spells
+			else if (special_spells[specialCount].flags & SP_REST) {
```
lynxlynxlynx: this will catch all spells that are to be cast on rest (minus the heal-all ones from above), so perhaps everything works fine already (wrt your last comment), even though the naming is confusing, since it only talks about healing. The possible mistargetting when casting shouldn't have much of an effect, as we weren't guaranteeing anything before either and self-targetting spells won't be affected.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
-								injurees.erase(injurees.begin());
-							}
+						tar->DirectlyCastSpell(tar, special_spells[specialCount].resref, 0, 1, true);
+						for (RestTargets::iterator injuree=wholeparty.begin();injuree != wholeparty.end() ; ++injuree) {
+								injuree->hpneeded-=special_spells[specialCount].amount;
 						}
 					}
 				}
-				ps = ps2;
+				ps=ps2;
```
lynxlynxlynx: please use spacing consistently, it makes the code more readable. We like spaces around operators for the same reason.

gemrb/core/Game.cpp
```diff
@@ -1734,43 +1753,78 @@ bool Game::RestParty(int checks, int dream, int hp)
...
+						resource.amount=tar->spellbook.CountSpells(special_spells[specialCount].resref,0,0);
+						healingspells.insert(std::pair<ieWord,HealingResource>(special_spells[specialCount].amount,resource));
+					}
+				}
+				ps=ps2;
+			}
+		}
+		std::sort(wholeparty.begin(),wholeparty.end());
+		// Heal who's still injured
+		while (!healingspells.empty() && wholeparty.back().hpneeded > 0) {
```
lynxlynxlynx: ... except for this hp check, which seems redundant.

gemrb/core/Game.cpp
DriimGit: True, moved that NULL check outside the hpneeded classification

gemrb/core/Game.cpp
DriimGit: All party members are added, even with the previously dangerous tar check.

gemrb/core/Game.cpp
DriimGit: IfI put them inside CastOnRest() I just get a bunch of errors. Possibly related that I put them in a container.

bradallred: Its fine it was just a style suggestion anyway :)

lynxlynxlynx:
nice cleanup! Seems like only the /rest/ of the spells remain.

DriimGit:
Was hesitant to add that, but I gues there's no harm even if teh targets are basically random.
Added casting the non-healing spells as well.

lynxlynxlynx:
Sorry for the delay, I'll try to do a final test this weekend.

lynxlynxlynx:
This has now been merged and polished up. Thanks!

PR #34, opened 2015-05-08
Fix the tooltip delay not being restored correctly when the game restarts

scrawl:
Fixes [bug 273](https://sourceforge.net/p/gemrb/bugs/273/)


scrawl:
Hmm... this change unfortunately breaks interoperability with baldur.ini.

lynxlynxlynx:
This one is historically trickier, since at first it was settable from our config (still is, but discouraged). Is the problem now that we store it decoded?

scrawl:
@fizzet seems to have found the correct fix in 699c83fc4fb4e831a3ae868b :)

PR #36, opened 2015-09-09
Fix crash when unsing custom fonts (TTF).

romgen:
I was using "DejaVuSans.TTF" as substitue for "NORMAL", as a test case.
I got a crash during loading (referencing null pointer...).

### DIFF
--- a/gemrb/core/Palette.h
+++ b/gemrb/core/Palette.h
@@ -49,9 +49,10 @@ public:
 	Palette(const Color &color, const Color &back);
 
 	Palette(const Color* colours, bool alpha_=false) {
-		for (int i = 0; i < 256; ++i) {
-			col[i] = colours[i];
-		}
+		if(colours)
+			for (int i = 0; i < 256; ++i) {
+					col[i] = colours[i];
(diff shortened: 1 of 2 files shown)


bradallred:
this should be fixed in master now

romgen:
You're way too hostile. Chill down dude.

But thanks für the commit.

bradallred:
@romgen I can see how I cam off that way and it was "completely inappropriate" for me to respond that way :) Please accept my sincere apologies. I hope I didnt discourage you from reporting future issues; honestly.

You are also correct that I missed one of the NULL palette calls which is now fixed (unless the palette member is actually NULL itself) and *should* fix the crash. I *did* test this, just not with the OpenGL driver, because:

# it is incomplete and (IMO) still unusable for playing
# it doesnt currently build on my system so I can't

Please let me know if it still crashes and I can work out whatever dependency issues are preventing me from building the driver and can debug.

romgen:
Thanks for the quick answer & fix. 

Also good to know about the opengl status. I will recompile without it!

PR #37, opened 2015-09-12
Fix build error on OS X.

jdm:


bradallred:
good catch, merged.

bradallred:
looks like `NSRunAlertPanel` is now deprecated anyway... the whole pre launch GUI needs a modernization; ideally code unification with iOS too.

PR #38, opened 2015-11-03
CMakeLists: Use find_package() instead of include()

heirecka:
That way you can disable searching for SDL{,2}_mixer and libvlc when
you don't want/need that functionality via
-DCMAKE_DISABLE_FIND_PACKAGE_foo (You'll need cmake >= 2.8.6 to pass
this, though).


lynxlynxlynx:
Hmm, is the search problematic in some cases? Or is this for intentionally leaner builds (dissable vlc plugin)? We look for both sdls, but only use one.

heirecka:
I noticed this when looking at my distro's package, which incorrectly tried to disable those deps via CMAKE_DISABLE_FIND_PACKAGE.
But yes, libvlc is a rather huge depency and I'm using openal for sound.

lynxlynxlynx:
Ok, thanks!

PR #39, opened 2016-03-25
Sdl2 hacking

kujeger:
Trying to bring the SDL2 version up and above the SDL1 version:
- now uses windowed fullscreen, using SDL2's own scaling
- window can be resized, also using SDL2's own scaling
- cursor trails gone

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -150,6 +150,8 @@ doneFormat:
 		width, height, SDL_GetPixelFormatName(format));
 	backBuf = SDL_CreateRGBSurface( 0, width, height,
 									bpp, r, g, b, a );
+    // tmpBuf is here as a truly ugly hack, so we can copy backBuf to tmpBuf before blitting cursors, and then back again after the screen is presented.
+    tmpBuf = SDL_CreateRGBSurface( 0, width, height, bpp, r, g, b, a );
 	this->bpp = bpp;
 
 	if (!backBuf) {
@@ -329,9 +331,12 @@ void SDL20VideoDriver::showYUVFrame(unsigned char** buf, unsigned int *strides,
 
 int SDL20VideoDriver::SwapBuffers(void)
 {
-	int ret = SDLVideoDriver::SwapBuffers();
+    //this is not pretty. We make a complete copy of the backBuf into tmpBuf, then copy it back after SDLVideoDriver::SwapBuffers has blitted cursors and tooltips, and SDL_UpdateTexture has copied it to the screentexture.
+    SDL_BlitSurface(backBuf, NULL, tmpBuf, NULL);
+    int ret = SDLVideoDriver::SwapBuffers();
 
 	SDL_UpdateTexture(screenTexture, NULL, backBuf->pixels, backBuf->pitch);
+    SDL_BlitSurface(tmpBuf, NULL, backBuf, NULL);
 	/*
 	 Commenting this out because I get better performance (on iOS) with SDL_UpdateTexture
 	 Don't know how universal it is yet so leaving this in commented out just in case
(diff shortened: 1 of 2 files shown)


bradallred:
I'm super reluctant to merge anything for SDL 2 without confirming iOS and Android still work (the changes dont look like they would break anything tho). Mobile devices are the primary target for SDL2 at the moment. Desktops dont benefit from our SDL 2 much if at all so anything that diminishes the mobile experience here will be rejected.

Im also reluctant because compositing another screen size buffer might have a noticeable performance impact on mobile devices (maybe not since its mostly transparent?). 

Perhaps you should add some checking for the cursor drawing (if you have the cursor disabled it is a completely wasteful operation) then we dont need to worry about any potential performance hit.

Hopefully over the summer we will be merging a branch that completely rewrote the cursor drawing (among a great many other things) so nasty drawing hacks will be moot.

kujeger:
Thanks for having a look! I added a check for if the cursor is visible as you suggested.

I do think that SDL2 adds good benefit compared to SDL1 for desktops, particularly SDL_WINDOW_FULLSCREEN_DESKTOP (and SDL_WINDOW_RESIZE) is very handy as it lets SDL2 transparently rescale the game without changing the actual resolution of the display, and keeps the proper aspect ratio (e.g. 4:3 if running the default resolutions).
SDL2 should also be somewhat faster in theory, as it does more stuff hardware accelerated (although I have not done any benchmarks).

I don't have any iOS devices to test with currently, nor have I ever tried building anything for Android, so I can't promise I'll be able to test those in the very near future. Are there any buildbots for android/ios available that gemrb uses?

(The big rewrite over summers sounds very nice, looking forward to it!)

kujeger:
recording demonstrating the advantages SDL2 and SDL_WINDOW_RESIZABLE / SDL_WINDOW_FULLSCREEN_DESKTOP gives:

https://kumini.kujeger.net/dump/gemrb_sdl2_bg2.webm

bradallred:
SDL 2 is *not* faster the way GemRB is using it (except for maybe on newer OS X where SDL 1.2 does some really horrible inefficient things). We are still doing all rendering in software in our code (see SpriteRenderer.inl) and only transforming the final result to an SDL_Texture before rendering. To use SDL hardware acceleration we would have to do the compositing in hardware using SDL_Textures the whole way. We cant do that with SDL alone because there is no palette manipulation capabilities with textures which we rely heavily on. We also do several manual pixel changes that, while possible, are slow when using textures due to the need to fetch, edit via software, then reupload to vram.

The rewrite I mentioned earlier alleviates this somewhat because each window has its own buffer now and redrawing only happens when needed (cursors and tooltips no longer affect this either). This means that windows using a texture for their buffer can use hardware acceleration during the window manager compositing phase and reduce or skip the drawing phase entirely. Unfortunately most of the expensive drawing is done on the game window itself which is still entirely software until the very end and is always completely redrawn each frame.

Somebody would have to step in and finish the OpenGL driver for us to ever have true hardware acceleration (without adding more dependancies that is).

lynxlynxlynx:
I've made an android build, but can't test it:
https://sourceforge.net/projects/gemrb/files/Other%20Binaries/android/0.8.4/

The scaling works fine for me on linux. And it's less of a user problem than I worried — just restart to get the original size back. It's nice that also downscaling works.

Unrelated, keyboard shortcuts with modifiers don't work for me, so I had to use the options window to go fullscreen. Nothing seems to have changed in sdl2 in this regard.

lynxlynxlynx:
Here are some extra details about the broken keys. We end up not calling KeyRelease, because the event doesn't contain the composited key, but just the modifier (as a key, not mod).

One other thing that inhibits SDL2 as a default is the per-pixel gamecontrol scrolling — much too slow for a mouse.

kujeger:
I've tested the apk you built, @lynxlynxlynx: it seems to work quite well on my old Nexus 4.

So at least these patches don't break Android in any obvious way.

lynxlynxlynx:
Thanks. The SDL2 issues are now noted on the todo (end of Unsorted section):
http://www.gemrb.org/wiki/doku.php?id=todo#unsorted

PR #40, opened 2016-04-08
Fix "reverse" scroll in SDL2 when too tall or wide

kujeger:
when SDL2 scales the window to a different aspect ratio, it will pad
sides with black.
Since mouse cursor position was passed unsigned, the negative x and/or y
values generated when in the left and/or top padded areas was wrapped,
and scrolling went the wrong way.


bradallred:
I have no problem merging this. `OnGlobalMouseMove` is completely removed by the branch I've been working on, however. Hopefully the refactoring I'm upto makes this moot, but if not be aware it might break in the future.

PR #41, opened 2016-04-09
set blend mode in SetSurfaceAlpha

kujeger:
SDL_SetSurfaceBlendMode needs to be SDL_BLENDMODE_BLEND on the surface
in order for it to, well, blend.


lynxlynxlynx:
Thanks!

The travis failures are unconnected — it seems it doesn't add the secrets when testing pull requests.

PR #42, opened 2016-04-21
Sdl2 input hacking

kujeger:
The way the SDL2 code handles keypresses and so on interacts in some weird ways with the rest of the system. It is a bit muddled, I think, because so much is "shared" between the SDL1 and SDL2 paths (through SDLVideo), even when they way of doing things in SDL2 does not neatly map to SDL1.

This fixes the problems of strange/double input in text fields, and the fact that neither the console nor fullscreen could be toggled through keypresses ingame.

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -737,7 +737,7 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
 		case SDL_KEYUP:
 			{
 				SDL_Keycode key = SDL_GetKeyFromScancode(event.key.keysym.scancode);
-				if (key > 32 && key < 127) {
+				if ((key >= 32 && key < 127) || key == 1073742053) {
 					// ignore keys that generate text. handeled by SDL_TEXTINPUT
 					break;
 				}
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -745,7 +745,7 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
 					// if key is literally just ctrl or shift -- skip it.
 					break;
 				}
-				if (SDL_GetModState() & KMOD_NUM && key >= 1073741908 && key <= 1073741927 && key != SDLK_KP_ENTER) {
+				if (SDL_GetModState() & KMOD_NUM && key >= SDLK_KP_DIVIDE && key <= SDLK_KP_EQUALS && key != SDLK_KP_ENTER) {
 					// ignore numpad keys (handled by SDL_TEXTINPUT) if KMOD_NUM. Always ignore numpad enter.
 					break;
 				}
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -734,10 +734,23 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
 				break;
 			}
 		case SDL_KEYDOWN:
+			{
+				SDL_Keycode key = SDL_GetKeyFromScancode(event.key.keysym.scancode);
+				if ((key == SDLK_SPACE) && SDL_GetModState() & KMOD_CTRL) {
+					core->PopupConsole();
+					break;
+				}
+				if ((key == SDLK_f) && SDL_GetModState() & KMOD_CTRL) {
+					// slight delay so one does not immediately re-toggle fullscreen
+					SDL_Delay(200);
+					ToggleFullscreenMode();
+					break;
+				}
+			}
 		case SDL_KEYUP:
 			{
 				SDL_Keycode key = SDL_GetKeyFromScancode(event.key.keysym.scancode);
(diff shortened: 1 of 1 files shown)


bradallred:
Well the console and full screen toggle are completely removed from SDL in my dev branch so I don't have issue merging those without double checking, but the other change looks like it could have unintended side effects. How does SDL1 dispatch the modifier keys? Why are we filtering only shift+ctrl mods? Are scan codes for left and right the same?

kujeger:
Shift and Ctrl are filtered because if they are not, text input fields produce garbage input when they are pressed. Note that they are only filtered from SDL_KEYUP, so they still work for e.g. holding when selecting multiple party members.
The codepath in SDL12Video.cpp just passes the events back to SDLVideo.cpp, but SDL2 cannot -- it handles key input events through SDL_TEXTINPUT, among other things.

KMOD_SHIFT contains both KMOD_LSHIFT and KMOD_RSHIFT (and same with KMOD_CTRL), so left and right are both hit.

bradallred:
ok, sounds like you tested adequately. It would be nice if you updated the comment to something like "filtering keys that will produce SDL_TEXTINPUT events"...

Still uncertain why it is that the mod keys should be filtered here. Do you know where this garbage text actually comes from? This is perhaps the wrong way to deal with it. After all it would prevent any control from ever receiving such an event, and while this might not break anything currently, its far from an ideal solution. Please at least add a FIXME for breaking controls abilities to ever get such events.

kujeger:
I'll update and add a comment. I'll also try to dig a little deeper around this, but I do somewhat suspect that it might be just as well to -- if/when the sdl2 version is on par with the sdl1 version -- remove everything sdl1, and rewrite stuff. But that would come after you merge your other work, in any case.

bradallred:
I doubt removing SDL1 is an option. The reason I'm not allowed to use modern C++ is due to supporting old/obscure platforms that don't support it and also likely don't have SDL 2. I also don't see how "rewriting stuff" (beyond the rewriting I've already done) will help with this particular issue; we don't even yet know the origin of the garbage text.

Now that I actually have looked at this more I also see that you've introduced inconsistent behavior. You are making the SDL2 driver trigger fullscreen/console on keydown vs keyup for SDL1... I also don't see why this is necessary since it looks like the shared code path ought to handle it, and indeed certainly used to when I originally authored it. Either platform inconsistency or something changed in SDL I figure; the kind of thing that would warrant more testing, but since this is soon to be completely removed I don't much care. I wont be surprised if on some platforms using SDL this would now toggle fullscreen on keydown and then undo it on keyup.

I also cant imagine that `SDL_Delay(200);` is actually a good thing. You are blocking everything when you do that. The comment makes it sound like what I said above is true about toggling and untoggling and this delay works around it because the events get dequeued after some time waiting.

Are you using SDL 2.0.4?

kujeger:
Using SDL 2.0.4, yes.

Rewriting things with SDL2 was more of a general thinking, not necessarily this particular issue. You are also correct about the fullscreen thing. I will change that.

lynxlynxlynx:
Btw, I can easily open and close the console.

Thanks for looking into this!

kujeger:
ah, yeah, now I understand about the console -- fixing the double-spacing by also ignoring "space" in SDL_KEYUP/DOWN broke the console, but I thought it had been broken all along.

If keys are handled with SDL_TEXTINPUT (like all the "normal" ones are), they need to be filtered out so they aren't treated by the common SDLVideo.cpp code for KEY_UP/DOWN (which otherwise leads to double keys).

But since some keys are special, and need to be sent on anyway.. This is the reason why the hotkey for fullscreen does not work in SDL2 -- the letter "f" was already filtered.

kujeger:
OK, I think things are better now (see commit msg).

bradallred:
This looks much better; enough to satisfy me. I'll let Lynx have the final say.

bradallred:
BTW, I saw your comment on IRC. We can't drop SDL_TEXTINPUT; several platforms, including iOS and Android, would be unable to input text without it since they dont have physical keyboards. Even if they did, SDL_TEXTINPUT does some nice things that we don't want to have to do ourselves such as translating keypresses into a UTF8 string. This is one of the reasons why we are using SDL in the first place.

kujeger:
Yeah, I thought that was why. It was more of a thougt born of frustration than a real suggestion, hehe.

lynxlynxlynx:
there are still problems:
- cheat keys don't work well; eg ctrl-j to jump around also opens the journal (j) — if a cheat key is used, the press/unpress should be consumed
  - because of this, ctrl-m to debug an actor is practically impossible, since the cursor moves
  - I haven't managed to trigger ctrl-shift + m or n either
- numpad keys with numpad on work as numbers, but also as up/down/... without it, unexpected chars
- console issues are fixed (double spacing, mod keys)

kujeger:
I believe SDL2 behaviour wrt keys and input should now completely match SDL1 behaviour.

Regarding style: I've split up the different if-s for when to break/ignore keys into several to avoid getting a ridiculously long condition. Hope that is OK.

bradallred:
please get rid of those magic constants. People will have no idea what `1073741908` and `1073741927` are supposed to represent.

kujeger:
I've changed them to SDL_Keycode constants (the comment under also explains what/why they are)

lynxlynxlynx:
Thanks, it works much better now!

PR #43, opened 2016-04-27
IWD2::CG: adjusts skill points calculation

MarcelHB:
When creating a new PC, there were more skill points available than there were supposed to be.

This changes the partial formula from:

(class pts + human pts + int bonus pts) * 4

to:

(class pts + human pts) * 2 + int bonus pts * 4

This returns values equal to the original game.


lynxlynxlynx:
Thanks!

PR #44, opened 2016-04-30
IWD2::Skills: prevents scrolling beyond skill items

MarcelHB:
The empty item line at the bottom is gone. In addition, this saves a Python exception.

`TopIndex` is used around the object without checking whether we should really go there by scrolling down.

Behaves a little more like IWD2, where you can move the scrollbar at the bottom parts without any updates.

### DIFF
--- a/gemrb/GUIScripts/iwd2/Skills.py
+++ b/gemrb/GUIScripts/iwd2/Skills.py
@@ -99,8 +99,11 @@ def RedrawSkills():
 def ScrollBarPress():
 	global TopIndex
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Did you find a better way, does it break stuff or in other words — why have you closed your own pull request?

MarcelHB:
Nothing like that, just did some clean up in my own fork and hit the wrong one. Sorry.

lynxlynxlynx:
Thanks!

PR #45, opened 2016-04-30
GUI: makes scroll bar work with repeatable key strokes

MarcelHB:
If you launch IWD2 and go CG in `Skills` or `Abilities` scripts, you see enabling of repeatable keys to make +/- buttons work a little better. Problem: *any* scroll bar hardly drag-scrolls anymore unless the RK-setting is properly reset since this state is managed globally. Which isn't.

This is a patch to what should be refactored: The logic invoking `MouseUp` calls when recognizing held mouse buttons now uses `GEM_MB_ONGOING_ACTION` instead, sharing the bit `GEM_MB_ACTION`. This way, we can leave `rk_flags` handling untouched while having a way to overcome interrupting events. This allows e.g. `ScrollBar` to not reset itself by receiving `GEM_MB_ONGOING_ACTION`.

Did test changes on BG and IWD2.

### DIFF
--- a/gemrb/core/GUI/Button.cpp
+++ b/gemrb/core/GUI/Button.cpp
@@ -524,7 +524,7 @@ void Button::OnMouseUp(unsigned short x, unsigned short y,
 			return;
 	}
 
-	if ((Button&GEM_MB_NORMAL) == GEM_MB_ACTION) {
+	if (Button & GEM_MB_ACTION) {
 		if ((Mod & GEM_MOD_SHIFT) && ButtonOnShiftPress)
 			RunEventHandler( ButtonOnShiftPress );
 		else
--- a/gemrb/core/GUI/EventMgr.cpp
+++ b/gemrb/core/GUI/EventMgr.cpp
@@ -327,7 +327,10 @@ void EventMgr::MouseDown(unsigned short x, unsigned short y, unsigned short Butt
 void EventMgr::MouseUp(unsigned short x, unsigned short y, unsigned short Button,
 	unsigned short Mod)
 {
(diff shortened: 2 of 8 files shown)


lynxlynxlynx:
Besides the extraneus whitespace changes, this looks good to me (untested). I'll wait for @bradallred to chime in and add his (by now) classical note that the whole input system is about to get rewritten.

btw, you can add a &w=1 to github urls to mimic diff -w, which makes reviewing that switch->if commit much easier to read.

MarcelHB:
For the future, what about removal of trailing white-spaces? I usually `git add` them after touching files as my editor auto-removes them, but I'll respect your policy.

lynxlynxlynx:
Any formatting-only changes should be in a separate commit (you can use add/commit with -p to finetune patches), so it is trivial to ignore. Not only does this make actual changes easier to review, but it also makes them easier to revert if needed, reduces chances of merge conflicts and doesn't pollute the git blame log.

bradallred:
I wouldn't want to disappoint @lynxlynxlynx ... so the entire input system is being rewritten! :)

Indeed much of this will conflict with that work, but if it works I have nothing against merging this in the interim.

lynxlynxlynx:
I tried it now. It does fix scrolling stopping if the focus leaves the scrollbar. Did you mean to fix also something else?

I noticed mousewheeling always results in an up motion, but that's a bug in our sdl2 backend.

MarcelHB:
From my point, I only focused on losing the scrollbar. Side-healings welcome.

lynxlynxlynx:
Thanks, I merged it manually.

PR #47, opened 2016-05-01
Encoding: decouples TLK and FS path encoding

MarcelHB:
My setup is as follows: I have a german IWD2 (TLK encoding ISO-8859-1), located on a Linux FS running GemRB, resource listing reveals UTF-8 file paths. The `Sounds` directory consists of 60% of files with names containing umlauts.

When I CGen a new party, the sound files list
1. displays visual garbage around umlaut-chars,
2. the resource manager is unable to find the selected sound file (please review #46 first). No sound then.

In order to overcome 1, it is crucial to replace `TLKEncoding` by `FSPathEncoding` where necessary, since the global encoding makes no sense here. For 2, from the Python interface/IWD2 `CSounds` script, I made everything on the way aware of multi-byteness of the file path to use. Otherwise, it ends up in that beautiful `strnlwrcpy` that disrupts everything. I try not to break anything.

**But**: I don't not know the situation at large, as e.g. `pl_uppercase`/`pl_lowercase` must have been used for files as well. Looking at the name, I assume that some polish version _might_ be broken then, but that's just a guess. This is where I need further information, we could make `FSPathEncoding` configurable I guess.

### DIFF
--- a/gemrb/GUIScripts/iwd2/CSound.py
+++ b/gemrb/GUIScripts/iwd2/CSound.py
@@ -69,7 +69,7 @@ def BackPress():
 	return
 
 def NextPress():
-	GemRB.SetToken("VoiceSet", TextAreaControl.QueryText())
+	GemRB.SetToken("VoiceSet", TextAreaControl.QueryText(1))
 	if SoundWindow:
 		SoundWindow.Unload()
 	GemRB.SetNextScript("CharGen8") #name
@@ -82,10 +82,10 @@ def SelectSound():
 def PlayPress():
 	global SoundIndex
 
-	CharSound = TextAreaControl.QueryText()
+	CharSound = TextAreaControl.QueryText(1)
 	MyChar = GemRB.GetVar ("Slot")
 
-	GemRB.SetPlayerSound (MyChar, CharSound)
+	GemRB.SetPlayerSound (MyChar, CharSound, 1)
 	# play sound as sound slot
 	GemRB.VerbalConstant (MyChar, int(VerbalConstants[SoundIndex]))
 
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -254,6 +254,10 @@ Interface::Interface()
 	TLKEncoding.widechar = false;
 	TLKEncoding.multibyte = false;
 	TLKEncoding.zerospace = false;
+	FSPathEncoding.encoding = "UTF-8";
+	FSPathEncoding.widechar = false;
+	FSPathEncoding.multibyte = true;
+	FSPathEncoding.zerospace = false;
 	MagicBit = HasFeature(GF_MAGICBIT);
 
 	gamedata = new GameData();
--- a/gemrb/core/Interface.h
+++ b/gemrb/core/Interface.h
@@ -397,6 +397,7 @@ private:
 	std::string Encoding;
 public:
 	EncodingStruct TLKEncoding;
+	EncodingStruct FSPathEncoding;
 	Holder<StringMgr> strings;
 	GlobalTimer * timer;
 	Palette *InfoTextPalette;
--- a/gemrb/core/Interface.h
+++ b/gemrb/core/Interface.h
@@ -140,14 +140,6 @@ struct TimeStruct {
 	unsigned int attack_round_size;
 };
 
-struct EncodingStruct
-{
-	std::string encoding;
-	bool widechar;
-	bool multibyte;
-	bool zerospace;
-};
-
 struct SpellDescType {
 	ieResRef resref;
 	ieStrRef value;
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -8344,12 +8344,19 @@ void Actor::SetPortrait(const char* ResRef, int Which)
 	}
 }
 
-void Actor::SetSoundFolder(const char *soundset)
+void Actor::SetSoundFolder(const char *soundset, bool mbStringPath)
 {
 	if (core->HasFeature(GF_SOUNDFOLDERS)) {
 		char filepath[_MAX_PATH];
+		bool useMbPath = mbStringPath && core->FSPathEncoding.multibyte;
 
-		strnlwrcpy(PCStats->SoundFolder, soundset, 32);
+		if(useMbPath) {
+			memset(PCStats->SoundFolder, 0, SOUNDFOLDERSIZE);
+			memcpy(PCStats->SoundFolder, soundset, SOUNDFOLDERSIZE - 1);
+		} else {
+			strnlwrcpy(PCStats->SoundFolder, soundset, 32);
+		}
+
 		PathJoin(filepath, core->GamePath, "sounds", PCStats->SoundFolder, NULL);
 		char file[_MAX_PATH];
 
@@ -8363,7 +8370,12 @@ void Actor::SetSoundFolder(const char *soundset)
 		} else {
 			return;
 		}
-		strnlwrcpy(PCStats->SoundSet, file, 8);
+
+		if(useMbPath) {
+			memcpy(PCStats->SoundSet, file, 8);
+		} else {
+			strnlwrcpy(PCStats->SoundSet, file, 8);
+		}
 	} else {
 		strnlwrcpy(PCStats->SoundSet, soundset, 8);
 		PCStats->SoundFolder[0]=0;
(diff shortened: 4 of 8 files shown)


MarcelHB:
Damn, local Clang build looks fine here, gonna check the next days, hopefully.

lynxlynxlynx:
If you check the log, only the osx build is failing (setlocale).

Comments:
- having setsoudnfolder and querytext take extra params is unwieldy, is there no way to autodetect the proper way to handle a string?
  - could the non-mb path be folded in? Or does this break iwd2 with a, let's say, English tlk?
  - the extra benefit would be, that also the other games would get fixed automatically
- we don't have a function yet to lowercase mb strings?
  - did the original German leave them as-is too in char sound list?
- polish and czech used a different non-utf encoding, so we do some extra translation there, as some chars are remapped (this is where the pl_ functions are from)
  - see gemrb/docs/en/default_ini.txt

bradallred:
you need to include `<clocale>` (not `<locale>`) to use the locale functions as per the C++ standard.

I agree with lynx, seems unnecessarily cumbersome/hacky. The scripts shouldn't need to know anything about encodings or file systems etc.

Also, I suspect there is a better way for dealing with file system encodings then adding yet another configuration for it. This should at the very least be something detectable at runtime instead of something the user must configure, or worse, set at compile time as it seems you have it now.

MarcelHB:
Might be `<locale>` vs. `<clocale>` ...

* Thing about `SetSoundFolder`/`QueryText` is: I expect they are used around scripts that might `QueryText` from some TLK-asset content. And even maybe `SetSoundFolder` by values not originating from reading the file system on demand? I haven't covered 100% of uses yet, so that's the proposal. But I can look at that. Auto-detecting encoding is complete no-go, and since we parsing regular Python values, I don't see easy ways to inject necessary states/flags to circumvent args.
 * I don't see ASCII-only paths breaking. I'm not sure about ISO-8859x as how such paths are used in the wild. I have a feeling of endangering pl/cz as they might have ISO encoding while `FSPathEncoding.encoding` is currently set to UTF-8, that's what the wide use of `strnlwrcpy` makes me think.
 * Yes, at least anything equiavalent to my setup.
* Haven't found it yet, but could check `String` for such a thing. Since std-things mostly to rely on `std::locale` and no one ever included them before, I guess not. :)
  * AFAIR, for all my DE-locale games, the names were uppercased and displayed correctly (Windows, NTFS), but definetely for IWD2.
* Yes, ack'd. And since all the string-processing went through the mentioned `strnlwrcpy`, I expect sound files names (as far as they were subject to non-ASCII chars, but this is something I don't know) to a) have always been broken so far or b.) be broken by now.

lynxlynxlynx:
SetSoundFolder is definitely safe to use directly with whatever the filesystem encoding is, since we generate its list by inspecting the toplevel sound folder. The soundset likely aren't even mentioned in the tlk.

QueryText is less clear. Fine for selectables (lists) like sound and emulated spinners (just numbers), but it's also used for general textedits and editable textareas. The latter are broken right now and can't be tested, but some, like the initial biographies will come straight from the tlk.

I don't know much about unicode, so this may be a silly question: can't QueryText check its result to see if it starts with a byte-order-mark? We currently do have a mask of allowable chars in savegame names, which I think applies to all textedits. Not sure if that works with the CJK locales, but I'm just thinking out loud of what kind of input (encoding) we get, because in the end, we don't want to save text written in two.

Can we treat all editables (or just editable TAs) as multibyte containers? They are few and far in between, so it wouldn't be much of a memory hit to always convert them. There are nice flags to check against.

MarcelHB:
`SetSoundFolder` was given my umlaut `Kämpfer_männlich_1` (`ä` being 2 bytes in UTF-8) string, already returned correctly from the Py context, and made it pass `strnlwrcpy`. This makes the string in console debugging look like `Kpfer_mnlich_1`. In my proposal, I explicitly avoid this last step on the road. Then, and only then, I was able to preserve the path bytes and actually hear sound/make the RM find the file.

I haven't noticed any explicit BOMs in any input (they are not mandatory for UTF-8 or maybe `strlen` ignores these bytes? Haven't tested that.).

I don't know, somebody made GemRB use `std::wstring` for reasons, affecting things around TAs. ;)

lynxlynxlynx:
For SetSoundFolder I meant that it should be fixable without needing an extra param for the gui scripts. Is there anything else but some old FAT partitions that don't use unicode paths by default?

bradallred:
the use of `w_char` is because we need 16 bit characters to properly map to the BAM fonts (for CJK versions specifically): one byte is the cycle and one byte is the frame. It also allows us to use any unicode characters that occupy a maximum of 2 bytes or the CJK TLK encodings which use 16bit chars.

The editable fields/TAs are going display based off of whatever font is set for them individually. If you are using standard fonts then only characters that exist in the standard fonts will display (basically extended ASCII, but depends on locale of the game data). Any character that doesnt have an entry in the BAM will have the little box outline glyph. If you are using TTF fonts (and have iconv) then it should be UTF16, but with the previously mentioned caveat of being incompatible with characters using more than 16bits.

as for FS that don't support unicode... FAT is the only one that comes to mind, but I think some are UTF16 instead of UTF8... anyway, I'm fine with assuming UTF8... lynx is usually more concerned about archaic device compatibility :) but `sys.getfilesystemencoding()` is available from the Python scripts to get an exact answer about what encoding should be used with file names.

`strlen` (and any other plain cstring function) is dumb and completely unaware of encodings so its not skipping anything, the BOM isn't there.

MarcelHB:
Had a view on the german original of IWD2 for Windows today. The game reads and stores single-byte characters, ISO/Win-1252.

I'm not not sure how to continue. We could close this PR to keep it for the archives until someone comes up with a better approach to deal with these issues.

lynxlynxlynx:
Ok, so the problem is still just the filesystem encoding — that's good. :) I don't know if any other translation has the same problem, but I pinged Edheldil about the Czech one.

The pragmatic thing to do would be to autodetect German on startup by just checking for specific known files/folder (we do the same for GameType) and normalize the sound folder names (likely needed also for scripts/) to ascii (ae, ss ..).

lynxlynxlynx:
The Czech one doesn't touch filenames, they're still English.

MarcelHB:
Good to know.

I'm not a big cross-platform fs-encoding compatibility type, but can we assume to have `std::setlocale(LC_ALL, "")` (or some other `LC_`?) in C++ and the mentioned `sys.getfilesystemencoding()` in Python to reliably determine the encoding of file-paths at runtime?

I'm not sure about savegame compat. across platforms as sound file names are taken-as-given. Could we ignore this for now?

bradallred:
assuming `std::setlocale(LC_ALL, "")` is safe enough, it may not do anything, but nothing will break (that isn't already) by using it.

I wouldn't worry about save game compatibility across platforms or localizations (for now anyway). This isn't really a new problem; map notes and other saved tidbits are going to suffer similar problems if you take the saves across platforms/localizations, but they are just annoyances rather then anything game breaking.

If loading a save shouldnt break if the sounds cant be found... if it does that is a separate issue that we can fix.

Good work so far :)

MarcelHB:
Any future plans here? Other approaches?

I went through some voice and sound issues, and all male PC voices for IWD2 (because of German `MÄNNLICH` in their sound file name) are still absent or throw exceptions when selected in CSound.

bradallred:
I cant see a reason not to merge it per se (we certainly want that `allocatedBytes` fix), but it seems like this is a band aid. Is the DirectoryIterator returning the expected strings? that seems like the real place to be addressing this.

MarcelHB:
Unaware of this PR still being pending, I had deleted my fork in the meantime. I guess this PR is then frozen to any further modifications.

lynxlynxlynx:
Can't download the patches, but the code is still viewable here.

Anyway, did a quick test:
- touch sounds/ačč000A.wav
- DirectoryIterator returns the mangled version before anyone else can do anything to it
- dirent has just a normal char array to hold the name, so that's to be expected
A bunch of things would have to be converted to use wchar_t to get that working.

bradallred:
I assume the "mangled" version is expected and is going to be the raw utf8 bytes. I assume GDB can interpret it that way if you ask it.

What I'm saying is: the DirectoryIterator ought to be what has the `EncodingStruct` as a static class member, so that things that read from it can know what to pass to `StringFromEncodedData` et al. This way the problem is solved for everything that displays information from the FS rather then this single point.

lynxlynxlynx:
Btw, I managed to recover the commits after all back then and have them in a branch. I can push it up if you two think there's still something to salvage.

PR #48, opened 2016-05-02
IWD2::Skills: fixes exploits on going back in GC

MarcelHB:
GC in IWD2 allows me to spend plenty of skill points. Simply by going back. Adressing

1. If we return to skills in GC, we get all the points to spend *again*.
2. It subtracts `RacialBonus` on >= 0 in CG, as it adds over and over on continuing.


lynxlynxlynx:
I did something simpler with 0f2d12fbc305, but thanks for the heads up.

lynxlynxlynx:
ehh, the web clipboard doesn't work, I meant: 841940524ecbd5

PR #49, opened 2016-05-02
IWD2: changes main menu background when playing at night

MarcelHB:
In the original game, playing between 6:00pm and 6:59am yields a
different background with a flare animation. Cozy.


lynxlynxlynx:
Nice find, especially locating the animation itself!

Thanks, I've manually merged it.

PR #50, opened 2016-05-05
GUIScript: allows callbacks for window-scoped key presses

MarcelHB:
One thing I'm really missing is changing PCs (in the inventory window) by pressing keys 1-6.

I did some research but did not find a way to have a flexible way to customize key stroke events outside of controls (i.e. do the actual logic in Python). Callbacks are limited to controls and certain events so far.

This PR allows scripts to make use of reading pressed keys on the unspecified window level and to act on them. If you think this is a good idea, we can do something like this to solve the inventory issue: https://github.com/MarcelHB/gemrb/compare/window_keys...MarcelHB:inventory_pc_key?expand=1

### DIFF
--- a/gemrb/core/GUI/EventMgr.cpp
+++ b/gemrb/core/GUI/EventMgr.cpp
@@ -38,7 +38,7 @@ EventMgr::EventMgr(void)
 	// Function bar window (for function keys)
 	function_bar = NULL;
 	// Last window focused for keyboard events
-	last_win_focused = NULL;	
+	last_win_focused = NULL;
 	// Last window focused for mouse events (eg, with a click). Used to determine MouseUp events
 	last_win_mousefocused = NULL;
 	// Last window we were over. Used to determine MouseEnter and MouseLeave events
@@ -363,7 +363,8 @@ void EventMgr::KeyPress(unsigned char Key, unsigned short Mod)
 	Control *ctrl = last_win_focused->GetFocus();
 	if (!ctrl || !ctrl->OnKeyPress( Key, Mod )) {
 		// FIXME: need a better way to determine when to call ResolveKey/SetHotKey
-		if (core->GetGameControl()
+		if (!last_win_focused->OnKeyPress(Key, Mod)
+			&& core->GetGameControl()
 			&& !MButtons // checking for drag actions
 			&& !core->IsPresentingModalWindow()
 			&& !core->InCutSceneMode()
--- a/gemrb/plugins/GUIScript/GUIScript.cpp
+++ b/gemrb/plugins/GUIScript/GUIScript.cpp
@@ -2564,6 +2564,48 @@ static PyObject* GemRB_Control_SetEvent(PyObject * /*self*/, PyObject* args)
 	Py_RETURN_NONE;
 }
 
+PyDoc_STRVAR( GemRB_Window_SetKeyPressEvent__doc,
+"===== Window_SetKeyPressEvent =====\n\
+\n\
+**Prototype:** _GemRB.Window_SetKeyPressEvent (callback)\n\
+\n\
+**Description:** Sets a callback function to handle key press event on window scopes. \n\
+\n\
+**Parameters:**\n\
+  * callback - Python function that accepts (windowIndex, key, mod) arguments and returns\n\
+	* 					 1 indicating succesful key press consumption or 0 otherwise. \n\
+\n\
+**Return value:** N/A\n\
+\n\
+**Example:**\n\
+    Window.SetKeyPressEvent (KeyPressCallback)\n\
+    return\n\
+\n\
+**See also:** [[guiscript:QuitGame]]\n\
+"
+);
+
+static PyObject* GemRB_Window_SetKeyPressEvent(PyObject* /*self*/, PyObject *args) {
+	int windowIndex;
+	PyObject *fn;
+
+	if(!PyArg_ParseTuple(args, "iO", &windowIndex, &fn)) {
+		return AttributeError(GemRB_Window_SetKeyPressEvent__doc);
+	}
+
+	WindowKeyPressHandler handler = NULL;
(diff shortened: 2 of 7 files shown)


gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -2564,6 +2564,48 @@ static PyObject* GemRB_Control_SetEvent(PyObject * /*self*/, PyObject* args)
...
+static PyObject* GemRB_Window_SetKeyPressEvent(PyObject* /*self*/, PyObject *args) {
+	int windowIndex;
+	PyObject *fn;
+
+	if(!PyArg_ParseTuple(args, "iO", &windowIndex, &fn)) {
+		return AttributeError(GemRB_Window_SetKeyPressEvent__doc);
+	}
+
+	WindowKeyPressHandler handler = NULL;
+	if(fn != Py_None && PyCallable_Check(fn)) {
```
lynxlynxlynx: doesn't this make it impossible to turn off a callback without creating a noop function first?

gemrb/core/GUI/EventMgr.cpp
```diff
@@ -363,7 +363,8 @@ void EventMgr::KeyPress(unsigned char Key, unsigned short Mod)
 	Control *ctrl = last_win_focused->GetFocus();
 	if (!ctrl || !ctrl->OnKeyPress( Key, Mod )) {
 		// FIXME: need a better way to determine when to call ResolveKey/SetHotKey
-		if (core->GetGameControl()
+		if (!last_win_focused->OnKeyPress(Key, Mod)
+			&& core->GetGameControl()
```
lynxlynxlynx: Does switching windows still work via hotkeys after this? Eg. inventory to guirec or journal.

I'm just about to leave for a week, so couldn't take but a cursory look.

bradallred:
judging by the kind of problems you are trying to fix perhaps I ought to recruit you to help finish the view hierarchy rewrite I've been working on (which includes event dispatch/hot keys etc). This (I believe from what I gather) is already a solved problem in that branch... as was the scrollbar focus thing you did.

would you be interested in that? It would save me a mighty merge conflict mess :)

gemrb/core/GUI/EventMgr.cpp
MarcelHB: Had this in my mind, it works.

gemrb/plugins/GUIScript/GUIScript.cpp
MarcelHB: It should be roughly the same practice as `Control_SetEvent`. There is one handler per window and if we set it to `NULL` a few lines down - if this branch is not being take - there are about a dozen of checks (or so) in the python helpers that discard any attempt to call the function.

MarcelHB:
Will have a look at Travis later. And probably need a GCC setup.

"Kind of problems"? :) My strategy is or was to attempt to make IWD2 more stable/authentic by simply play-and-work.

Do you have something in mind or even a code branch for what you are doing? Looking at what I have seen so far, the code for GUI/script/input/events looks limited to extend. I guess your work would be rather huge. On the other hand, I still have no idea about all the open issues that need to be solved to reach remaining milestones (let's simply say, making IWD2 finishable?). Just to avoid premature over-engineering considering the time *I* can spend.

bradallred:
I will push up the branch and post some notes on the wiki when I get home and you can decide for yourself if you are interested in assisting. Almost certainly anything related to the GUI/Event dispatch is addressed there. Any other issues are safe to address from master without much conflict.

Since you seem specifically interested in IWD2, that branch at some point needs to have somebody fixup the IWD2 scripts due to API changes. I've been developing on BG2 so nothing else works yet. The BIK videos (I think IWD2 uses these) may also be broken as I didn't have IWD2 handy to test the new video player with.

My email should be on my profile if you want to discuss it; yours does not appear to be.

lynxlynxlynx:
Yay, we'll finally get to see your branch.

IWD2 remnants are mostly rules and reverse-engineering related things though.

MarcelHB:
Glad to see you back.

Wasn't getting into much work here as well. The next thing for me I trying to help bradallred with IWD2 compatbility.

About RE'ing, has any one ever figured out how to maybe exploit cheat codes `jeffkattacks`/`jeffkdefends` for injecting custom script execution (as seen e.g. in BAF files)? Haven't watched deeper yet, but if someone already knows ...

lynxlynxlynx:
What he was talking about should be pretty trivial work, likely largely regexable. It will be needed for all the gametypes anyway.

IWD2-specific bugs are here, with the easier stuff in the other half:
http://www.gemrb.org/wiki/doku.php?id=iwd2_bugs

As for the two cheat codes: that looks trivial, just a spawn of the creature and perhaps script swap. There's no need to inject anything into scripts themselves. If you meant how the original didi it, I doubt it is exploitable. Why would you use it anyway, when all the scripts are available to edit directly?

MarcelHB:
Just for seperation of concerns. Cheating is limited in IWD2 - compared to some other games - like no `CreateCreature`. And if I had some way to execute scripts loaded at command execution time, without touching assets, I'd find this quite handy.

lynxlynxlynx:
That seems to be the only major one missing, but the jeff thing could be used, just make a copy of the wanted cre and rename it as jeffk.cre or jeffk2.cre (supposedly the only thing missing). So not doable without touching assets, especially if you want custom scripts.

The long way is doing it in gemrb (a shorthand cc() is defined and you can use ExecuteString() for everything), saving and loading, but that also runs the risk of finding file format details differences (saves eventually become incompatible), but that's a bag of bugs for another day.

lynxlynxlynx:
Almost forgot to merge this with all the (off-topic) discussion going on — thanks. The forum is a better place for such things, or even irc for smaller stuff.

PR #51, opened 2016-05-06
SDL2: Fix RMB stuck cursor

kujeger:
If not using touch input,
SDL_MOUSEMOTION/SDL_MOUSEBUTTONDOWN/SDL_MOUSEBUTTONUP would never break,
instead moving on to SDL_KEYDOWN -- which would read
event.key.keysym.scancode and erronously break on the scancode event
read when RMB+MouseMovement occurred, never moving on to return the
event to SDLVideoDriver::ProcessEvent . We now explicitly return the
event when event.button.which is not SDL_TOUCH_MOUSEID .

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -733,7 +733,6 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
 			else {
 				// we do not want these events to cascade down to SDL_KEYDOWN, so we return here instead of at default .
 				return SDLVideoDriver::ProcessEvent(event);
-				break;
 			}
 		case SDL_KEYDOWN:
 			{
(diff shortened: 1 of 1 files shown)


gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -725,14 +725,16 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
...
-			}
 		case SDL_MOUSEBUTTONDOWN:
 		case SDL_MOUSEBUTTONUP:
 			if (event.button.which == SDL_TOUCH_MOUSEID) {
 				break;
 			}
+			else {
+				// we do not want these events to cascade down to SDL_KEYDOWN, so we return here instead of at default .
+				return SDLVideoDriver::ProcessEvent(event);
+				break;
```
bradallred: break after a return is going to cause a warning I would think.

kujeger:
Seems not. Only reason I left it there was (percieved) symmetry/clarity -- it can certainly be removed.

bradallred:
You should remove it. I don't know what compiler/flag combo you are building with or what flags our Travis builds use, but I expect there is some compiler out there that will issue an "unreachable code" warning.

kujeger:
Sure. I've removed it.

PR #52, opened 2016-05-07
IWD2: updates some avatar size values

MarcelHB:
In IWD2, if you look closely, you notice lots of (N)PCs not carrying a weapon. Actually they do, but since the 0x5000 to 0x6000 range of avatars lacks size values, everything looks like fist-fighting.

Value range taken from BG2 and tested on humans lurking around Targos harbour area.


bradallred:
Very nice. Thanks!

PR #53, opened 2016-05-15
Inventory pc key

lynxlynxlynx:
Opening Marcel's diff separately, to start the discussion of the python side.


gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -950,4 +951,18 @@ def AbilitiesItemWindow ():
...
+def HookKeyPresses():
+	Window = GUIINV.InventoryWindow
+	Window.SetKeyPressEvent(SwitchPC)
+	return
+
+def SwitchPC(wIdx, key, mod):
+	if key >= 49 and key <= 54:
+		GemRB.GameSelectPCSingle(key - 48)
+		GUICommonWindows.SelectionChanged()
+		GUIINV.UpdateInventoryWindow ()
```
lynxlynxlynx: running GUICommonWindows.RunSelectionChangeHandler () instead would allow for the function to be completely generic. So it can be moved to GUICommonWindows or GUICommon and reused in other modules with a portrait bar.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -950,4 +951,18 @@ def AbilitiesItemWindow ():
 	Window.ShowModal (MODAL_SHADOW_GRAY)
 	return
 
+def HookKeyPresses():
+	Window = GUIINV.InventoryWindow
+	Window.SetKeyPressEvent(SwitchPC)
+	return
```
lynxlynxlynx: This function looks redundant. The binding will always happen in a context where we know the window we want to work with, right? So `InventoryWindow.SetKeyPressEvent (SwitchPC)` is better than going through another layer.

lynxlynxlynx:
As the last thing, the other GUIINV and similar windows need the same call. Luckily besides the journal, most are shared and it won't be that much duplicated work.

gemrb/GUIScripts/InventoryCommon.py
MarcelHB: Good idea.

MarcelHB:
Yes, I will do, at least for BG and IWD2.

gemrb/GUIScripts/InventoryCommon.py
MarcelHB: Ok, I can change that.

MarcelHB:
Damn, I messed it up the branches. I will rewrite the changes as you recommended and do a new PR.

Please close this one.

PR #54, opened 2016-05-15
GUIScripts::IWD2: navigation among PCs inside GUIs

MarcelHB:
Ok, as you suggested in #53, I did some minor modifications to the code. In IWD2, everything works well.

I tried BG1 and managed to get the same work for `GUIINV` and `GUIREC`. In contrast, `GUIMG` and `GUIPR` fail as there is something inside the Python scripts shaking `last_win_focused` to set the options window focused, not the desired center window. If you enter such a GUI on a PC that can actually cast spells, it works until reaching a PC with a grayed window.

I tried some things but this is a little beyond what I'm trying to focus on right now. So I skipped that for now.

### DIFF
--- a/gemrb/GUIScripts/GUICommonWindows.py
+++ b/gemrb/GUIScripts/GUICommonWindows.py
@@ -2074,7 +2074,6 @@ def RealRestPress ():
 def SwitchPCByKey (wIdx, key, mod):
 	if key >= 49 and key <= 54:
 		GemRB.GameSelectPCSingle(key - 48)
-		SelectionChanged ()
 		RunSelectionChangeHandler ()
 		return 1
 	return 0
--- a/gemrb/GUIScripts/GUICommonWindows.py
+++ b/gemrb/GUIScripts/GUICommonWindows.py
@@ -2070,3 +2070,12 @@ def RestPress ():
 def RealRestPress ():
 	GemRB.RestParty(0, 0, 1)
 	return
+
+def SwitchPCByKey (wIdx, key, mod):
+	if key >= 49 and key <= 54:
+		GemRB.GameSelectPCSingle(key - 48)
+		SelectionChanged ()
+		RunSelectionChangeHandler ()
+		return 1
+	return 0
+
--- a/gemrb/GUIScripts/GUIREC.py
+++ b/gemrb/GUIScripts/GUIREC.py
@@ -134,6 +134,8 @@ def OpenRecordsWindow ():
 	GUICommonWindows.SetSelectionChangeHandler (UpdateRecordsWindow)
 	UpdateRecordsWindow ()
 
+	Window.SetKeyPressEvent (GUICommonWindows.SwitchPCByKey)
+
 	OptionsWindow.SetVisible (WINDOW_VISIBLE)
 	Window.SetVisible (WINDOW_VISIBLE)
 	PortraitWindow.SetVisible (WINDOW_VISIBLE)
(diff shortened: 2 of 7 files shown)


lynxlynxlynx:
They're not marked as visible, but grayed (AdjustWindowVisibility) and I guess that means they can't receive focus. Looking at guipr, we're also setting WINDOW_FRONT later, which is odd, but likely an oversight.

gemrb/GUIScripts/GUICommonWindows.py
```diff
@@ -2070,3 +2070,12 @@ def RestPress ():
 def RealRestPress ():
 	GemRB.RestParty(0, 0, 1)
 	return
+
+def SwitchPCByKey (wIdx, key, mod):
+	if key >= 49 and key <= 54:
+		GemRB.GameSelectPCSingle(key - 48)
```
bradallred: shouldn't calling `GameSelectPCSingle` ought to call `SelectionChanged` which in turn ought to call `RunSelectionChangeHandler`

MarcelHB:
I guess so, If you use that multi-selection function, the portraits are highlighted correctly, `GameSelectPCSingle` obviously doesn't.

bradallred:
I think adding `core->SetEventFlag(EF_SELECTION);` to `Game::SelectPCSingle` would be a good thing. I can't imagine wanting to call it and _not_ wanting to run `SelectionChanged`; indeed it looks like every caller is manually doing so (at least on my branch). Even if not, it should be benign to do it.

MarcelHB:
"highlighted correctly" in terms of _nothing happening_. I can change this, I think that's good. I just try to keep the core-ish intrusion as low as possible for the time I may oversee or brake things by doing such modifications.

lynxlynxlynx:
I'm not completely sure what you're talking about either, but Brad is right — there's no reason not to always run the change handler (afaik). This selection function is for single-pc environments (ie. not game window), but that's where the divergence ends. While it is cleaner to run the handler in python directly (all this EF_ stuff is a cludge), it's also not so wieldy.

MarcelHB:
It's just the portrait border change that is not happening unless I call `SelectionChanged` manually, aside from this, the PC is switched flawlessly.

lynxlynxlynx:
I doubt it, if you exit the screen the old selection should persist — in other words, the border is simply the only visual cue, but more is wrong.

bradallred:
really? when I looked at the code earlier it seemed like `core->SetEventFlag(EF_SELECTION);` should cause the selection handler to run...

Anyway I don't _really_ care, I just thought it would be nice for all those double/triple function calls to be a single call (plus it would handle both Python and C++ calls that changed the selection).

lynxlynxlynx:
I'm not sure we disagree.

bradallred:
I didn't think so 😃 That was a response to @MarcelHB 

I interpreted (probably incorrectly) his post to mean that he still had to call `SelectionChanged` manually even after adding `core->SetEventFlag(EF_SELECTION);`

MarcelHB:
Ok guys, the commit right now moves that magic whatever-selection-effect into `Game`. It's working, but I don't know if this fully correct.

lynxlynxlynx:
RunSelectionChangeHandler should be extraneus now too. I'll check the details late tomorrow.

bradallred:
We could/should also clean up the other callers too, but that can be done later.

lynxlynxlynx:
With a too long delay (hoped subviews would be done sooner), this is now in. Also pst and iwd1 scripts use it. I didn't touch pr/mg, since it would require turning their modality off.

Thanks!

PR #55, opened 2016-07-24
CMake improvements

winterheart:
Make all optional dependecies available trough OPTION, which make them
accessible via ccmake or cmake-gui. They also can be redefined via
-DUSE_FOO=ON or -DUSE_FOO=OFF during cmake invocation.
Also rewrited SDL1/SDL2 and OpenGL options to clean look.
Pass -DSDL_BACKEND=SDL or -DSDL_BACKEND=SDL2 for SDL choice, and
-DOPENGL_BACKEND=None, -DOPENGL_BACKEND=OpenGL or -DOPENGL_BACKEND=GLES
for desired configuration.


lynxlynxlynx:
Hey, thanks for the PR. Here are my notes:
- this should really be a series of commits
- you're not handling the migration from the previous values
- disabling optional dependency lookup is already possible through cmake directly: -DCMAKE_DISABLE_FIND_PACKAGE_foo (but the option approach is nicer)

winterheart:
Ok, I'll split commit into series.
What values I left? I thought that they are all covered. USE_SDL2 moved to SDL_BACKEND, USE_OPENGL=(1|2) left unchanged, but there OPENGL_BACKED for user, which handle USE_OPENGL settings. Something about I forgot?

lynxlynxlynx:
I meant that the old flags should continue to work and only be deprecated years later. We don't want to make extra work for existing packagers by breaking their build scripts. So just add a (reverse) translation somewhere in the beginning.

PR #56, opened 2016-07-27
Cmake improvements #2

winterheart:
Replace for PR #55
Make all optional dependecies available trough OPTION, which make them
accessible via ccmake or cmake-gui. They also can be redefined via
-DUSE_FOO=ON or -DUSE_FOO=OFF during cmake invocation.
Also rewrited SDL1/SDL2 and OpenGL options to clean look.
Pass -DSDL_BACKEND=SDL or -DSDL_BACKEND=SDL2 for SDL choice, and
-DOPENGL_BACKEND=None, -DOPENGL_BACKEND=OpenGL or -DOPENGL_BACKEND=GLES
for desired configuration.


CMakeLists.txt
```diff
@@ -190,49 +195,76 @@ ELSE()
...
+	INCLUDE(FindSDL)
+	IF(USE_SDLMIXER)
+		FIND_PACKAGE(SDL_mixer MODULE)
+	ENDIF()
 ENDIF()
 
-# look for sdl2 if sdl1 was not found or a preference was stated
-IF(USE_SDL2 OR NOT SDL_FOUND)
+# FIXME: Compatibility - remove USE_SDL2 in the future
+IF((SDL_BACKEND STREQUAL "SDL2") OR USE_SDL2)
```
lynxlynxlynx: Is there still a default now? We had the flags for sdl, so one can override what is used if both are found on the system. We don't want to require everyone to explicitly state a choice now.

winterheart: Default is still SDL 1.2. It's defined here (line 32):
```SET(SDL_BACKEND "SDL" CACHE STRING "libSDL version (possible values: SDL or SDL2)")```
So, if anyone have both libraries, cmake without options will choose SDL 1.2 as intended earlier. For choosing SDL2 there need to add -DUSE_SDL2 (old behavior) or -DSDL_BACKEND=SDL2 (new one).

lynxlynxlynx: Ok, but what if you have only sdl2 installed? That worked seamlessly before too.

And since you're not unsetting USE_SDL2, wouldn't switching back to sdl1 via the new option now not work properly (even before the latest commit)?

winterheart:
OK, here autodetection mechanism, that uses SDL by default and SDL2 when both backends are found and USE_SDL is active.

lynxlynxlynx:
Thanks! It will look for sdl twice if auto is used, but even if the modules don't shortcircuit that, it shouldn't be a problem.

PR #58, opened 2016-10-19
fix incorrect usage of SDL_GetNumTouchDevices

miartad:
use it as a function not as a function pointer

causes gcc 5.1 warning: error: ordered comparison of pointer with integer zero [-Werror=extra]


lynxlynxlynx:
Thanks, my bad!

PR #59, opened 2017-03-04
Fix link to forum

pwaring:
Old link was redirecting to forums home page instead of specific forum for GemRB.


lynxlynxlynx:
Thanks, at this point it is clear they'll never fix their link schema to be backwards compatible. Don't even have FTP access anymore. :cactus:

PR #60, opened 2017-09-18
IWD2: restores footprint sounds for the party

MarcelHB:
More than a year later, I found some time continuing a little :)

So when I start IWD2, you notice that the party chars steps are silent - we are supposed to hear steps on snow, on stone, on wood planks ... I caught some issues the way down:

1. There is no `walksnd.2da` for IWD2. I just copied the one from BG2 and - for no single clue on animation IDs and creature refs for this game so far (InfExpl tells me little about them) - removed everything but `DEFAULT` entries. This is enough for humans, it seems.
2. For IWD2, walking sounds have a different enumeration (1, 2, ..., n) than previously given (none, a, b, ..., n). I'm not sure about the switch condition but since it's used quite often per sec, I'd refrain from going unhardcoded ways.
3. There's `Map#SrchMap` that takes the Area bitmap, but only for reachability things. Map bits are completely mapped to `Passable` indices, losing everything about floor compositions (11: snow, 1: dirty ways, 2: wood, ...). Surprinsingly, someone already mapped all these things correctly in `pathfind.2da` for each game. So to break nothing, I dup'ed the `SrchMap` to `MaterialMap` for walking sounds.

I have tested BG and IWD2. Comments welcome.


lynxlynxlynx:
Cool and just before a new release. 

1. Some more info on animation IDs: https://gibberlings3.github.io/iesdp/files/ids/iwd2/animate.htm , but you can also look at avatars.2da. You can leave the whole how or  bg2 table in place, since there are bound to be more hits that way. Any extra animation slots not covered for iwd2 can be added later.

3. I don't understand why this change is necessary. The pathfind table already has the sound mapping and the addition of walksnd.2da should be enough. As far as I can tell, it only increases the memory consumption.

GF_IWD2_SCRIPTNAME is used by all iwds, but GF_3ED_RULES isn't, so it's the better choice. Technically this should have a new game flag, but it's fine.

MarcelHB:
`SrchMap` is set like this

```cpp
unsigned short value = sr->GetAt(x,y) & PATH_MAP_AREAMASK;
size_t index = y * Width + x;
SrchMap[index] = Passable[value];
```
And the `Passable` thing is filled by `pathfind.2da`, for IWD2: `8 1 1 1 1 1 1 1 0 1 8 1 0 8 3 1`

So for a snowy position (BMP index 11) at some x,y, this original code

```cpp
int type = GetInternalSearchMap( Pos.x/16, Pos.y/12 )&PATH_MAP_AREAMASK;
memcpy(sound, terrainsounds[i].Sounds[type], sizeof(ieResRef) );
```
will return `type: 1` which maps to `FS_DIRT`, not `FS_SNOW`. So far, I don't see any place where the unmapped values are stored, so they'd be in `MaterialMap` now. This doubles the memory, of course, and another approach could be a lazy mapping for the original bits to the `Passable` index where necessary.

For your other suggestions, I'll have a look.

lynxlynxlynx:
ah, this would explain the wonkiness in other games too, good find! 

We could store the original values and do lookups for the search map, but I think that's a hot path, so more memory sounds better. How would you get a Passable index in your lazy loading idea? The values are not unique.

Additionally, the "if (cnt) {" in the bg2 case looks erroneous or to be more precise, the -1 in the roll does. I can't think of any good reason to skip playing walksounds at a whim.

MarcelHB:
About the sounds: IWD2 contains a few `WAL_` files by the old cycling name pattern but mostly `FS_` files. So I left the entries of BG2 that map to these certain ones, and changed everything else to `SOUND`. Otherwise, the logs is full of unfound `WAL_` files.

MarcelHB:
About my other proposal, lets say we do:
```cpp
SrchMap[index] = sr->GetAt(x,y) & PATH_MAP_AREAMASK
```

Now, everything left in the code expects `SrchMap` to contain `Passable` values, e.g. let's take `Map::GetBlocked`:
```cpp
unsigned int ret = SrchMap[y*Width+x];
if (ret&(PATH_MAP_DOOR_IMPASSABLE|PATH_MAP_ACTOR)) {
  ...
```

Now, the proper change will be
```cpp
unsigned int ret = Passable[SrchMap[y*Width+x]];
if (ret&(PATH_MAP_DOOR_IMPASSABLE|PATH_MAP_ACTOR)) {
  ...
```

This way, we only have the original bitmap and apply the `Passable`-filter for all these former uses.

lynxlynxlynx:
cnt: correct, didn't lookup the filenames before. The WALs could be leftovers, but they don't hurt and the one for slimes is definitely appropriate (not sure if there are any though).

Passable: so what I just described. Bigger memory footprint is better.

MarcelHB:
Great. The failing CI setup is not related to my changes, I guess. ;)

lynxlynxlynx:
It isn't, the appveyor config needs to be improved, since you can't push artifacts on PRs.

lynxlynxlynx:
Boom, fixed. Travis has infra problems, but the mac build was fine before the last commit, so I'm not waiting.

lynxlynxlynx:
Thanks!

PR #61, opened 2017-09-20
ACMReader: fixes underflow on pointer magic

MarcelHB:
Minor nerdy things:

I attempt to put my build setup somewhere else. There I have a GCC v7.2.0 on x86_64 and after setup, IWD2 crashes right at the moment of loading menu music.

The error can be tracked down to this very line: the author once wanted to fill a ptr-wide buffer full of 0xF bytes, shift them aside and add the result to `buff_middle`. This certainly works on 32bit and surprinsingly my VBox Linux, but here `-1ul` evaluates to `0x00000000FFFFFFFF` for this op. The shifted bit pattern will then throw the pointer far away.


bradallred:
Indeed, good find. I was going to make a comment about how we should use `uintptr_t`, but apparently that was introduced in C++11.

PR #62, opened 2017-09-20
Various fixes for Windows and GCC v7

MarcelHB:
I made GemRB compiling in Windows and a brand new GCC v7.2.0 (targeting 64bit). These are some non-CMake related proposals for issues obviously related to any of those components.

See comments below for details. And let's have a look on AppVeyor.


gemrb/includes/plugindef.h
```diff
@@ -135,6 +135,7 @@ GEM_EXPORT_DLL const char* GemRBPlugin_Version();
 #ifdef WIN32
 #include <windows.h>
 
+BOOL APIENTRY DllMain(HANDLE, DWORD,	LPVOID);
```
MarcelHB: See above.

gemrb/core/Core.cpp
```diff
@@ -44,11 +44,13 @@
 #include <crtdbg.h>
 #endif
 
+BOOL WINAPI DllEntryPoint(HINSTANCE, DWORD, LPVOID);
```
MarcelHB: Conflicts with `-fmissing-declarations`, obviously missing on MinGW.

gemrb/core/PluginLoader.cpp
```diff
@@ -98,7 +98,11 @@ static inline voidvoid my_dlsym(void *handle, const char *symbol)
 static void PrintDLError(const char *msg=NULL)
 {
 	char buffer[_MAX_PATH*2];
+#ifdef _MSC_VER
 	_strerror_s(buffer, NULL);
+#else
+	sprintf(buffer, "code %lu", GetLastError());
+#endif
```
MarcelHB: Looks like some MS-exclusive function. Slightly different version available in C11.

gemrb/includes/win32def.h
```diff
@@ -34,7 +34,11 @@
 
 #ifdef WIN32
 # define WIN32_LEAN_AND_MEAN
+
+# ifndef NOMINMAX
 # define NOMINMAX
+# endif
```
MarcelHB: Somewhere defined in `<cstring>` for g++.

gemrb/plugins/BIKPlayer/mem.cpp
```diff
@@ -33,24 +33,33 @@
...
+#elif HAVE_ALIGNED_MALLOC
+	ptr = _aligned_malloc(size, 16);
+#elif HAVE_MEMALIGN
+	ptr = memalign(16, size);
+#else
+	ptr = malloc(size + 16);
 	if(!ptr)
 		return ptr;
-	diff= ((-(long)ptr - 1)&15) + 1;
+	long diff = ((-(size_t)ptr - 1) & 15) + 1;
```
MarcelHB: `long` was the core problem.

Taken a patch from `https://github.com/FFmpeg/FFmpeg/blob/4fb311c804098d78e5ce5f527f9a9c37536d3a08/libavutil/mem.c`, but we still need the last case as a working fallback.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -66,8 +66,10 @@
 #include <algorithm>
 #include <cstdio>
 
+#if !defined(WIN32) && !defined(__MINGW32__)
 // MIPSPro fix for IRIX
 size_t strlcpy(char *, const char *, size_t);
+#endif
```
MarcelHB: On MinGW, it's decorated by some `__declspec(dllimport)`, so this redef. breaks.

lynxlynxlynx:
Just FYI, before I look into this in detail — we're only doing msvc builds on appveyor, not mingw.

gemrb/plugins/BIKPlayer/mem.cpp
```diff
@@ -33,24 +33,33 @@
...
-	ptr = malloc(size+16);
+#if HAVE_POSIX_MEMALIGN
+	if (posix_memalign(&ptr, 16, size))
+		ptr = NULL;
+#elif HAVE_ALIGNED_MALLOC
+	ptr = _aligned_malloc(size, 16);
+#elif HAVE_MEMALIGN
+	ptr = memalign(16, size);
+#else
+	ptr = malloc(size + 16);
```
lynxlynxlynx: do you know on which platforms which of these defines would be relevant? Because we don't test for any yet.

gemrb/plugins/GUIScript/GUIScript.cpp
lynxlynxlynx: I think it would be nicer to just check in cmake and then use a specific define. Look at the main file and cmake/cmake_config.h.in

gemrb/core/PluginLoader.cpp
lynxlynxlynx: This whole thing can be removed. While it did improve debugging, it turned out the final reports were almost as useless. Now it's potentially only adding a dependency on msvcr. But yeah, no way for you to have known.

gemrb/plugins/BIKPlayer/mem.cpp
MarcelHB: As far as I googled: native POSIX systems or MSVC. If they aren't configured at all, I can remove them. Just look a more trustworthy than the one below. ;)

gemrb/plugins/GUIScript/GUIScript.cpp
MarcelHB: Ok.

gemrb/core/PluginLoader.cpp
MarcelHB: :+1:

gemrb/core/PluginLoader.cpp
```diff
@@ -114,7 +104,7 @@ static bool FindFiles( char* path, std::list<char*> &files )
 	strcat( path, "*.dll" );
 	if (( hFile = ( long ) _findfirst( path, &c_file ) ) == -1L) {
 		//If there is no file matching our search
-		PrintDLError("Error looking up dlls: %s");
+    Log(ERROR, "PluginLoader", "Error looking up dlls.");
```
lynxlynxlynx: Misaligned.

And probably now #include <string.h> is not needed any more.

MarcelHB: Oh, yep.

gemrb/plugins/BIKPlayer/mem.cpp
lynxlynxlynx: Maybe more efficient too. Worth adding more checks to cmake imo.

cmake/modules/config.cmake
```diff
@@ -43,5 +43,9 @@ if (WIN32_USE_STDIO OR APPLE)
 	SET (NOCOLOR Yes)
 ENDIF()
 
+IF (IRIX)
+	SET (DECLARE_STRLCPY 1)
+ENDIF()
```
MarcelHB: Like this?

lynxlynxlynx: CHECK_FUNCTION_EXISTS("strlcpy" HAVE_STRLCPY), which we actually already do. So just check against that define.

MarcelHB: This doesn't work for me (`#ifdef HAVE_STRLCPY`). My cmake says none was found but CMake doc mentions some possible false negative cases: https://cmake.org/cmake/help/v3.8/module/CheckFunctionExists.html

lynxlynxlynx: Are you saying you have it and it ends up defined?

MarcelHB: There is `strldup` for me obviously, but Cmake says the contrary.

MarcelHB: Btw. I'd still need to add the entry to `cmake_config.h.in` in case. Or do something different to this check.

lynxlynxlynx: I guess you meant strlcpy. It's a BSD function, so you shouldn't have it — it's odd that you do. Are you sure the error doesn't stem from our definition (can be ifdef-ed the same way). In any case, the check needs to be fixed if it's broken or removed.

MarcelHB: Yes, of course `strlcpy`. I'll check that.

MarcelHB: Mystery solved: It just didn't cover with `core/System/String.h` previous' signature because of `GEM_EXPORT`.

lynxlynxlynx:
Cool, thanks! I'll test it later and combine some of the commits.

lynxlynxlynx:
Merged.

PR #63, opened 2017-09-22
OpenGL: some fixes

MarcelHB:
iwd2s

There are some minor but annoying rendering issues when using OpenGL:

1. Mysterious black pixels in the snow. Blending on area tiles is enabled even if there is no reason to (as of the docs, tiles contain no alpha-values and `BLIT_HALFTRANS` isn't set here), so I set blending to opaque for the area background.
2. Char shadows are opaque and item piles are always outlined. Both these issues are half-related since there is no support on `BLIT_NOSHADOW` and `BLIT_TRANSSHADOW`. I put this handling to the fragment shaders - my experience in GLSL is limited, but they are ES compliant and for now, they work. ;)
3. Ground circles are mostly missing (see the piece of green circle close to the goblin at the center?) and ground reticles lack the outer bound. Something is messed up around scissoring, so for `DrawEllipse`, I disabled it. And `SetPixel` failed trying something SDL-related, so it has turned into some simple GL override.

The chimney smoke also looks weird, but it's nearly the same in the SW renderer so I skipped that for now.

iwd2_after

Tested on IWD2 and BG on Windows.

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20GLVideo.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20GLVideo.cpp
@@ -441,8 +441,32 @@ void GLVideoDriver::drawPolygon(Point* points, unsigned int count, const Color& 
 	glDeleteBuffers(1, &buffer);
 }
 
+void GLVideoDriver::SetPixel(short x, short y, const Color& color, bool clipped) {
+	if (clipped) {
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Awesome, someone cares about the GL driver!

PR #64, opened 2017-09-22
OpenGL: adds missing header update

MarcelHB:
In #63 I missed adding `SDL20GLVideo.h` so the `OPENGL_BACKEND=OpenGL` setup won't compile. CI is probably not building that. :grin:


lynxlynxlynx:
Yeah, we're only testing different platforms and compilers, not build modes.

PR #65, opened 2017-09-23
GUIScripts: the clock in IWD2

MarcelHB:
Contrary to other games, the globe-clock is visible nearly everywhere and its control ID was wrong so there is no natural way to retrieve the time by tooltip (and at all?). 

It actually moves in map overview and inventory - and as in the original, if you click it in the inventory, the puppet stops moving. :+1:

### DIFF
--- a/gemrb/GUIScripts/GUICommonWindows.py
+++ b/gemrb/GUIScripts/GUICommonWindows.py
@@ -256,6 +256,8 @@ def SetupMenuWindowControls (Window, Gears=None, CloseWindowCallback=None):
 		UpdateClock ()
 	else:
 		rb = OptionControl['Rest']
+		if iwd2:
+			UpdateClock ()
 
 	# Rest
 	if Window.HasControl (rb):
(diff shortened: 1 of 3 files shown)


gemrb/GUIScripts/GUICommonWindows.py
```diff
@@ -2042,12 +2044,14 @@ def UpdateClock ():
...
+		if OptionsWindow:
+			if GameCheck.IsIWD2():
+				Clock = OptionsWindow.GetControl (10)
+			elif OptionsWindow.HasControl(9):
+				Clock = OptionsWindow.GetControl (9)
 		elif ActionsWindow and ActionsWindow.HasControl (62):
 			Clock = ActionsWindow.GetControl (62)
-
-		if Clock and Clock.HasAnimation("CGEAR"):
+		if Clock and (Clock.HasAnimation("CGEAR") or GameCheck.IsIWD2()):
```
lynxlynxlynx: Is it safe to assume Clock will always be non-null in iwd2? Eg., do all the window packs have it and at the same control id?

MarcelHB: All the `CGEAR`s I found in InfExp here were indexed by 9. I can double-check for positivity, and then I guess I'd had to find something calling `UpdateClock` without a clock but options. I just assumed this for now as I really couldn't find a counter-example by clicking through all reachable window packs.

MarcelHB: I claim that we are good to go here, for IWD2 ist options bar <=> clock. And for every found `GCOMM` bitmap, it's index 10.

PR #66, opened 2017-09-23
Infravision

MarcelHB:
As you see in the screenshots of #63, infravision is enabled (by the half-orc) even at daytime while it should be enabled by night or inside dungeons only. For at least IWD2, infravision is not red but white.

Moreover, the infravision is far too dark (current implementation in screen 1 vs original in screen 2/3): It's set relative to the light map and global tints. For infravision and my research, actors always glow. For non-IWD2, I've hardcoded the value to some medium red which is very close the original one (screen 2) - somehow, we have a little more global dark blue from somewhere (screen 3 vs screen 4) so I suggest to live with these values for now.

Since there is no documentation in both code and docs, I took the unused `BLIT_GLOW` bit to circumvent global tinting. To me, the whole thing shouldn't take place in the video drivers, but I think that's another issue.

In addition, no game except IWD2 seems to take infravision into account as a racial bonus (not covered here).

infra
infra_bg
infra2
infra3

### DIFF
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -40,6 +40,7 @@
 #include "GUI/GameControl.h"
 #include "System/DataStream.h"
 #include "System/StringBuffer.h"
+#include "Video.h"
 
 #include <algorithm>
 #include <iterator>
@@ -532,7 +533,7 @@ int Game::JoinParty(Actor* actor, int join)
 				CopyResRef(actor->LargePortrait, ptab->QueryField(actor->LargePortrait, "REPLACEMENT"));
 			}
 		}
-		
+
 		if (size) {
 			ieDword id = actor->GetGlobalID();
 			for (size_t i=0;i<size; i++) {
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -2055,6 +2055,21 @@ const Color *Game::GetGlobalTint() const
 	return NULL;
 }
 
+// applies the global tint, if any
+void Game::ApplyGlobalTint(Color &tint, ieDword &flags) const
+{
+	const Color *globalTint = GetGlobalTint();
+	if (globalTint) {
+		if (flags & BLIT_TINTED) {
+			Color::MultiplyTint(tint, globalTint);
+		} else {
+			flags |= BLIT_TINTED;
+			tint = *globalTint;
+			tint.a = 255;
+		}
+	}
+}
+
 bool Game::IsDay()
 {
 	ieDword daynight = core->Time.GetHour(GameTime);
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -707,7 +707,7 @@ bool Game::SelectActor(Actor* actor, bool select, unsigned flags)
 	if (! (flags & SELECT_QUIET)) {
 		core->SetEventFlag(EF_SELECTION);
 	}
-	Infravision(actor);
+	Infravision();
 	return true;
 }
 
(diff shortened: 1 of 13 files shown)


lynxlynxlynx:
This is trickier. BLIT_GLOW was likely used for the bg2 planetar/deva animations (forums currently down, so no link to the comparison).

But more importantly, infravision disabling global tint will conflict with things like timestop and dream areas. The first is easy to check, but I highly doubt the original did it this way — I don't remember any exceptions.

MarcelHB:
Ok, I'll have a look on this and try to find something else if required.

lynxlynxlynx:
btw, I moved the global tint application out of the driver in the subviews branch (27a8d189a1), so you could backport that and the two commits before it, but there'll be merge conflicts. Easier to mess with the tint though.

MarcelHB:
Great, will probably do after another minor issue. :+1:

MarcelHB:
Ok, merge is done correctly so I hope. Infravision updates to follow soon ...

MarcelHB:
Ok, some more things:
* No more `BLIT_GLOW`, we can now do this locally. Timestop works, any overrides (weather, darkness) sound intended (I don't know ... where can fog weather be found?).
* The `infravision` setting seemed to be misunderstood. It only allows other PCs from profiting from one PC's infravision ability when inside the selected group or not - then its reserved for exclusive selection of a single one. Not for disabling it globally if set to 0.
* The previous light map check doesn't match BG2/IWD2 originals (or it's a different one), so I dropped that.

gemrb/core/Game.cpp
```diff
@@ -1995,25 +1995,35 @@ bool Game::TimeStoppedFor(const Actor* target)
 	return true;
 }
 
-//recalculate the party's infravision state
-void Game::Infravision()
+/*
+ * Infravision: on specific actor with infravision or whole party with
+ *  at least one infravision actor and group-wide setting enabled.
+ */
```
lynxlynxlynx: Are you saying that in the original, if you didn't select a pc with infravision, it didn't matter if anyone else had it?

gemrb/core/Game.h
```diff
@@ -476,7 +476,7 @@ class GEM_EXPORT Game : public Scriptable {
 	/** check if the passed actor is a victim of timestop */
 	bool TimeStoppedFor(const Actor* target=NULL);
 	/** updates the infravision info */
-	void Infravision();
+	void Infravision(const Actor *target=NULL);
```
lynxlynxlynx: this compiles? Renamed parameter.

MarcelHB:
And something we could possibly fix as well: I have a savegame from a native BG2 with Jaheira. Here, she has no IV so I guess this racial attribute is not persisted into the savegame but works by global rules - that we don't have for now.

gemrb/core/Game.cpp
MarcelHB: Yes. For BG2 I just verified: Without "Group infravision" being ticked, it's reserved for exlusive selection of IV PCs. When ticked, we only need at least one such a guy in the selection.

lynxlynxlynx: You overwrite it globally when passed an actor though.

lynxlynxlynx:
Fog was never implemented in the originals. Same for snow, I think, but maybe iwds had it.

gemrb/core/Game.h
MarcelHB: To me and my compilers (both GCC), this looks valid won't issue warnings.

gemrb/core/Game.cpp
MarcelHB: To my understanding, this rule then has no effect at all.

lynxlynxlynx: I mean that you change `hasInfra` unconditionally; the setting is currently ignored. So if the ini option is on and you select a char without IV, it will disable it even if another PC has it. Am I reading it wrong?

MarcelHB: If the INI option is on and you select a (single one or set of) char without IV, it will disable it even if another PC has it. Yes.

That's how BG2 actually behaves.

lynxlynxlynx: That just sounds broken, I'm for overriding that behaviour.

MarcelHB: This is the description I picked from my German BG2 `dialog.tlk` - it's the same for BG, IWD2 just calls it "darkvision":

> Aktiviert diese Option, wenn Infravision für alle Charaktere angezeigt werden soll, auch wenn nur einer der ausgewählten Charaktere über Infravision verfügt. Ist diese Option deaktiviert, wird Infravision nur angezeigt, wenn tatsächlich ein Charakter mit Infravision ausgewählt ist.

Which translates to:

> Enable this option to apply infravision for every character even if only just one selected character has infravision. If this option is disaled, infravision is only used if a character is selected that actually has infravision.

To me this sounds intentional.

If not, what kind of override would you prefer?

lynxlynxlynx: Enable this option to apply infravision for every character even if only just one ~selected~ character has infravision.

MarcelHB: Ok, did that. Even with settings disabled, the originals flawed a little when selecting e.g. twi IV PCs by shutting down the light again. This doesn't.

MarcelHB:
Let me examine the work for adding racial attributes for IV before merging. I guess setup of PC actors is a good point do this only once. If the hassle isn't too big, I can do this as well.

lynxlynxlynx:
I'll merge anyway, the racial stuff would be substantial. 

It's not just IV, but also attack bonuses for elves. Many of them from the descriptions weren't actually there (pretty sure about the vs. giants one, unclear about saves vs charm), so it's messy. And technically an effect could change one's race, so it's indeed better to do this in Actor at runtime, rather than CREImporter on load.

I have to fix something in IESDP, since now the iwd2 way isn't visible.

lynxlynxlynx:
iwd2 lists racial abilities in this table:
https://gibberlings3.github.io/iesdp/files/2da/2da_iwd2/racespab.htm

but it's only for castable innates. So we need another column in our races.2da. Depending on the effects we need to apply, a spell name may be enough, otherwise it has to be a table reference like above with racespab.2da.

"TODO: Elves get a racial THAC0 bonus with all swords and bows in BG2 (but not daggers)" is one of the places where it needs to show.

sucks that the forum is still down, I'm pretty sure this was discussed in greater detail somewhere.

MarcelHB:
Ok, then let's do this separately.

lynxlynxlynx:
Cleaning up a bit by manually picking and squashing. Seems all the noGlobalTint stuff is redundant now.

lynxlynxlynx:
Thanks, it's now merged.

lynxlynxlynx:
So, bgee uses racethac.2da to determine who gets infravision and a separate table for thac0:
http://gibberlings3.net/forums/index.php?showtopic=27462&p=237119

Seems I made a list a while ago:
http://gibberlings3.net/forums/index.php?showtopic=19986

Sleep immunity for elves definitely wasn't there, fixpack adds it. So it seems safe to assume that only those two tables matter. Since we eventually want ee compatibility, it's best to use them. And if it turns out we need/want another table, so moddability is better, the races.2da comment from before applies. Maybe it'd be needed already for iwd2, maybe not, it has many tables for racial stuff that we already handle.

MarcelHB:
Ok, I'll examine that.

There is something else - and much simpler I hope - I want to look at before.

lynxlynxlynx:
Sure, no rush. Just something to stick into InitActorTables and then use at the end of RefreshEffects (infravision) and the todo line for the tohit. 

I've checked now and there actually isn't any effect that could change the race, so the infravision table handling could go in CREImporter. But that would be a bit harder, wouldn't make it any cleaner and wouldn't allow for an eventual opcode to be treated properly. So Actor it is.

PR #67, opened 2017-09-24
OpenGL: better text rendering

MarcelHB:
On OpenGL, some texts look close to illegible:
font1

The code did only unset full-transparent pixels when it should unset anything but `0xFF` for keyed palettes. But then we can also save instructions and failing branch predictions for this check completely.

Magic effects still look the same (and some a little dark).


lynxlynxlynx:
Thanks!

PR #68, opened 2017-09-24
IWD2: re-mapping for party voice samples

MarcelHB:
For IWD2, there is not `vcremap.2da` so voices are completely confusing or annoying (like hit sounds for ack'ing commands).

The one from IWD is not comprehensive in terms of the sound palette but it's not mis-mapping. So it's just a copy of that one.


lynxlynxlynx:
Good enough start, thanks!

PR #69, opened 2017-09-24
IWD2: thief visuals

MarcelHB:
In IWD2, unlike BG2, thieves actually have these slim hoodie dresses. This fixes the wrong BAM prefixes. 

`CxxT3` look like bards, but I couldn't identify them solely from [this list](http://gemrb.org/iesdp/files/ids/iwd2/animate.htm) and none of them in this 2DA, so I guess they are still wrong.


lynxlynxlynx:
the main columns are for different armour levels, so the animation used depends on what is equipped (heavy/medium/light/none). So try looking at the combinations and maybe the t3s will become apparent. You'll need to cheat or use feats to get proficiency though.

MarcelHB:
I'm not sure about multi-classes but since thieves are restricted to light armour, the values are fine I guess.

lynxlynxlynx:
They're not restricted at all, this isn't the adnd arbitrariness. :)

MarcelHB:
Ok, but this is the schema of T1 to T3 for all races and m/f, the last one simly doesn't fit medium appearance: the gnome one has fancy trousers, I guess it belongs to a completely different anim. ref., the bard.

avatars

IWD2 doesn't ship anything else, I've scrolled through the whole ANIMchar.

lynxlynxlynx:
Can't tell just from the bams, I'd have to see the originals. The last one has a longer armor and no cowl, so I could easily imagine it being a medium armored char.

It's easy to verify, you don't even need proficiencies.

Looking at the first pregen-ed party, the rogue doesn't change appearance from none->light. Then does for medium (no more cowl) and then again for heavy (big shoulder pads).

MarcelHB:
Wow, yes, this was surprising :) Ok, set the T3 and F4 for everything else. That's what they appear to look for the test group.

PR #70, opened 2017-09-29
OpenGL: bad clipping calculations

MarcelHB:
bg_fisheye

Something I started to notice only in BG/BG2 is this fisheye effect at the viewport fringe. The OpenGL implementation failed to calculate texture coordinate, so we always squeeze the whole texture instead of clipping it appropriately.


lynxlynxlynx:
Thanks!

PR #71, opened 2017-09-30
OpenGL: no more flickering on overlays

MarcelHB:
There is one problem where GUI drawing and OpenGL bite each other: screen areas that are not redrawn inside a frame only redraw the cursor into the back buffer. For the cursor movements, this produces tons of artifacts on the background of overlaying windows or modals. For modals, thanks to double buffering, we have a some nauseaing 15Hz flickering in addition (one grayed background and the previous one without gray tone).

flickering

For OpenGL, for now I propose to create a texture of everything that has been drawn to the background the last time before we create a modal or put a window on top. A suspended window then isn't really redrawn but just the top-most active window(s) on top of the screenshot of the disabled ones.

This isn't perfect yet: e. g. I cannot remove a cursor that sticks on an element the last time it had been drawn correctly and won't ever be drawn again since it's outside of the draw area now. Maybe your `subviews` will address these issues more accurately. But for now, the rendering is smooth again: at max. one bad cursor and no more flickering.

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -3040,6 +3040,8 @@ void Interface::DrawWindows(bool allow_delete)
 {
 	//here comes the REAL drawing of windows
 	static bool modalShield = false;
+	static size_t windowStackShield = 0;
+
 	if (ModalWindow) {
 		if (!modalShield) {
 			// only draw the shield layer once
@@ -3050,8 +3052,11 @@ void Interface::DrawWindows(bool allow_delete)
 				shieldColor.a = 0xff;
 			}
 			video->DrawRect( Region( 0, 0, Width, Height ), shieldColor );
+			video->TakeBackgroundBuffer();
 			RedrawAll(); // wont actually have any effect until the modal window is dismissed.
 			modalShield = true;
+		} else {
+			video->DrawBackgroundBuffer();
 		}
 		ModalWindow->DrawWindow();
 		return;
@@ -3059,6 +3064,15 @@ void Interface::DrawWindows(bool allow_delete)
 	modalShield = false;
 
 	size_t i = topwin.size();
+	bool backgroundRedrawn = false;
+
+	if (windowStackShield != i) {
+		if (i > 1) {
+			video->TakeBackgroundBuffer();
+		}
+		windowStackShield = i;
+	}
+
 	while(i--) {
 		unsigned int t = topwin[i];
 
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
@bradallred any thoughts?

bradallred:
For now I would have a look at what has been done in the subviews branch for SDL 1.2. We no longer draw the cursor on the interface, it has its own buffer as does each window. It might be better to fix this on that branch, but that will be significant work (which needs to be done anyway). I'm not super great at OpenGL so I haven't yet given it much thought or attention. 

I'll be back from vacation on the 5th and can help more then.

gemrb/core/Interface.cpp
```diff
@@ -3077,6 +3091,9 @@ void Interface::DrawWindows(bool allow_delete)
 				}
 			} else if (win->Visible) {
 				win->DrawWindow();
+			} else if (!win->Visible && (t <= topwin.size() - 2) && !backgroundRedrawn) {
```
lynxlynxlynx: where does the -2 come from?

lynxlynxlynx:
That'll take too long, there's no good reason for the usable gemrb not to be improved, especially before a release. It can be refactored later as needed.

gemrb/core/Interface.cpp
MarcelHB: I remember having a top-window in IWD2 that was invisble: the loading screen. Which then ultimately blocked the game screen. I checked again, _that_ doesn't occur with `backgroundRedrawn ` any more as well. The latter one was only to draw that once, for performance, but I'm not sure if this really cancels out other such cases.

MarcelHB:
I think merge-conflict related impacts to `subviews` are fairly low here. Best case: the added video interface methods become superfluous and can be dropped, worst case: still no better approach for the GL renderer.

gemrb/core/Interface.cpp
lynxlynxlynx: Well, if it is still needed, it deserves a comment. Even comparing t to topwin size looks odd considering what it holds.

By using < instead of <= the constant can be lowered to -1, but that still sticks out then.

MarcelHB: Ok, I will look into it.

MarcelHB:
Ok, I couldn't reproduce this observation and dropping this check broke nothing.

lynxlynxlynx:
Thanks. It can be outright undone on the subviews branch if needed.

bradallred:
I just wanted to point out that things have been reengineered for future reference; I'm not suggesting not to merge this or any other improvements. Its hopefully useful information for Marcel to keep in mind as he proceeds. I thought perhaps that might influence design decisions going forward (backporting some interfaces or something).

Keep up the good work.

lynxlynxlynx:
Sure, I just didn't want him to get lost in the rabbit hole.

PR #72, opened 2017-10-03
OpenGL: fix for map tile rendering

MarcelHB:
In #63 , I tried to fix "mysterious black pixels on snow" for OpenGL. This fix was very rough, it breaks BG water rendering.

bg_water

Not that bad, this way I discovered some more general bug as `0` has been treated as invalid palette index for OpenGL. But since this is the default color key index for transparency, by default, for every paletted texture, the pixels of color index `0` are then discarded regardless of intended transparency.


lynxlynxlynx:
Thanks 👍

PR #73, opened 2017-10-04
OpenAL: z-axis for listener

MarcelHB:
Maybe it's a more subjective issue, but let's consider the chanting dudes in this screenshot below:

audio

When centering through them when their samples are playing, inside the inner rectangle you have a very loud sound with very little fade-in, and each red line immediately flips one source (on stereo) from one speaker to the other one. This feels very unbalanced.

On OpenAL, the listener's position is ground (`{ x, y, 0 }`) as all sources are. I propose to put it to the camera (height!), so it has a longer fade-in/out and less amplitude right above. Values `< 100` don't address the impression well, `> 200` makes sounds too low.

Feel free to try it and propose something else, but I consider `0` a bad choice. For distance calculations I think it's Ok to leave it as-is.

Other addressed examples include:
* single miraculously loud foot-steps
* the isolated fireplace sound in the Candlekeep Inn

### DIFF
--- a/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
+++ b/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
@@ -613,7 +613,7 @@ int OpenALAudioDriver::CreateStream(Holder<SoundMgr> newMusic)
 
 void OpenALAudioDriver::UpdateListenerPos(int XPos, int YPos )
 {
-	alListener3f( AL_POSITION, (float) XPos, (float) YPos, 0.0f );
+	alListener3f( AL_POSITION, (float) XPos, (float) YPos, LISTENER_HEIGHT );
 	checkALError("Unable to update listener position.", WARNING);
 }
 
--- a/gemrb/plugins/OpenALAudio/OpenALAudio.h
+++ b/gemrb/plugins/OpenALAudio/OpenALAudio.h
@@ -61,6 +61,8 @@
 #define REFERENCE_DISTANCE 50
 #define ACM_BUFFERSIZE 8192
 
+#define LISTENER_HEIGHT 100.0f
+
 namespace GemRB {
 
 class OpenALSoundHandle : public SoundHandle {


lynxlynxlynx:
very nice! Have you checked how spellcasting is affected? We have a problem for example with magic missile being too loud.

Here's what I wrote on the todo:
> RESEARCH spell effect volume (eg. magic missile hit) is often too loud, especially compared to gui sounds (same volume setting). We do play them differently, but this is audible even in sdlaudio, which has no spatial aspect to sounds (GEM_SND_RELATIVE). The files just have vastly different dbfs/gain (compare tra_22 and hit_07) and it looks like nothing in the acm header could be used to normalize. As far as wavc, there's only one field that's mostly the same, supposedly unused, that could perhaps be the marker. But it's also true that all trap/hit/release sounds are hardcoded in the exe, so perhaps it did something manually.

Ok, so maybe it won't be affected.

MarcelHB:
Yes, spells sound a little odd but they do on any setting.

bradallred:
an improvement for sure. if we want the camera don't we have to angle this since its isometric? not sure how noticeable that would be, but the math should be pretty simple to figure out.

MarcelHB:
I have also thought of this. I will give it a try.

MarcelHB:
To my opinion, this doesn't feel good: this way, the loudest perception of a source is when it just appeard through the bottom of the screen.

bradallred:
ok then, I think we are happy with this then unless @lynxlynxlynx has objections.

lynxlynxlynx:
I don't hear the original choir problem*. I tried the branch anyway. The choir is much too quiet now when they speak on their own. Compare it to when you initiate dialog. Do you also hear the difference?

* my right speaker hardly gets used though, but I know it works http://www.audiocheck.net/audiotests_stereo.php

MarcelHB:
I know the difference: But when you click on them individually, don't the lose any spatial attributes for sound? It then sounds as loud as clicking onto a party member, which to me sounds like equal gain on both speakers, i.e. no relative position at all.

You can maybe try to lower the height a little. Tomorrow, I try to give an example.

MarcelHB:
Ok, I did it now. This is the fly-by I did for these samples upon hearing at least two voices, red line roughly indicates the center of the screen, fountains and some foot steps in the 2nd half of the records. Sound settings are the same for all runs.

audio_fly

This is the original game (bg.mp3):
bg_sound

No spatial effects at all but the fade is Ok.

This is OpenAL with ground zero (gemrb_al_0.mp3):
al_0

You see the higher loudness flipping from right to left in short time quite good on scrolling.

This is z=150 (gemrb_al.mp3):
al_150

Yes, this lower but better positioning than the original one while not being as harsh as z=0.

Try yourself, these are the MP3 records.

[sound_samples.zip](https://github.com/gemrb/gemrb/files/1364660/sound_samples.zip)

MarcelHB:
Maybe 100 is a better choice? Then we have around the same amplitude at the center as the original game while having not fading much longer.

lynxlynxlynx:
Same amplitude definitely sounds like an improvement. I've just tested in the original and there is no audible difference between them chanting on their own and you forcing them.

As for unimplemented stuff, at least configurable, there's eax. I did my testing with that off though.

MarcelHB:
So, no change, the lower value than initially proposed...?

I just think there is a valid - tiny - distinction between a dialog (having actors centered, being right in front of a source) and being a spectator flying over the map.

EAX/EFX surely is another approach but AFAIR the games did a rather poor job working with it. I even have some ideas on it but this is certainly much larger approach.

lynxlynxlynx:
I wasn't testing flying over, but standing near them, viewport affixed.

Retesting in gemrb, I can hear what you meant. Scrolling with the kb or mouse wheel makes the voices pretty jarring and they don't even sound loudest at felt like the origin. But there's no difference in volume if you're standing there and initiate their chanting vs them doing it on their own.

MarcelHB:
So, what's next? I lowered the height but I guess anything else requires some advanced approach under OpenAL. I mean, feel free to close - I have no better idea right now to bring.

lynxlynxlynx:
It's fine, thanks, I just didn't notice the amended commit, since that doesn't trigger notifications.

PR #75, opened 2017-10-11
Shadow animations

MarcelHB:
As you can see in #70 and #73, some PCs/NPCs in Baldur's Gate lack shadows. Outside and for sun, this really looks odd.

For BG, shadow animations for certain animation IDs are stored separately and their attached displaying is hardcoded in the original. Essentially for Sarevok and "everyone else".

I'm not sure if this is this best way to solve this but as far as I see, this only affects some specific BG actor animations.


lynxlynxlynx:
Nice! From the notes on the todo, we only lacked them for the full-fledged animations (pc races), so your id list looks fine. avatars.2da supports id ranges, so if you just copied the segment from there, it's likely a few animations slipped. Please move the file to the shared unhardcoded path.

MarcelHB:
Ok, moved it.

lynxlynxlynx:
I checked and all games have these bams save for pst. Which didn't complain. Good job. :)

PR #76, opened 2017-10-15
OpenAL: simple EFX for reverbing sounds

MarcelHB:
Consider this a little experimental, but it's already working better than expected.

Background: Creative's EAX wrapper tool doesn't work with IWD2 and I personally disliked BG2 settings a little bit here and there so some work went into manual assignment. Also because the ARE files claim to contain a reverb ID but I haven't found anything but 0 there anywhere.

Later IE games seem to have up to 18 predefined environment settings (see `REVERB.IDS`, `REVERB.2DA`, `EAXENVIR.IDS`), referring to the preset of the OpenAL SDK and adding some more adjustments - I could  not do anything meaningful with `DAMPING` and `VOLUME` yet, but the given profiles already do their job.

For BG1, BG2 and IWD2 I went through the areas and assigned a reverb profile number, for the other games (or unmentioned areas - don't have the BG1 addon available right now) there is some best guessing to have some basic reverb.

For the future, we should also equip any call to `Play()` by some more meaningful hint to better identify sounds subject to 3D (such as named in `SNDCHANN.2DA`) and apply some volume mixing.

I mean it's possibly far from an ultimate EFX/EAX experience yet, but neither have the originals been, so this is already quite good ;) e. g. take the initial dungeon in BG2.

Btw. this is likely the last PR for the next months.


MarcelHB:
Will look at the clang issues later.

MarcelHB:
Damn, I tried some things to make both GCC and Clang happy on old C++ syntax without having the need to somehow rewrite the reverb presets as they are taken from the SDK.

But CI clang fails with something nasty, seems to be this one: https://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis

Any better idea?

bradallred:
> Any better idea?

personally I'm all for moving to C++11 😛. In all honesty I don't recall the reason why it is disallowed.

To make it work with 98, you would need to supply a constructor. the `{}` initialization syntax is only allowed at variable declaration time in 98.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
```diff
@@ -220,12 +230,76 @@ bool OpenALAudioDriver::Init(void)
...
 	speech.ambient = false;
 	return true;
 }
 
+bool OpenALAudioDriver::InitEFX(void) {
+	ALCdevice *device = alcGetContextsDevice (alutContext);
+	ALCint auxSends = 0;
+	hasEFX = false;
+
+	/* Would be nice to switch that on and off arbitrarily. */
```
lynxlynxlynx: this setting is already wired in guiopt, so toggling should work.

gemrb/override/shared/REVERB.2DA
```diff
@@ -0,0 +1,22 @@
+2DA V1.0
+-1
+		ENVIRONMENT	VOLUME	DECAY	DAMPING	REVERB_LEVEL
+NO_REVERB	GENERIC		0.0	0.0	0.0	0.0
+SMALL_ROOM	CITY		0.10	1.0	0.40	1.0
+MEDIUM_ROOM	AUDITORIUM	0.15	1.5	0.50	1.0
+LARGE_ROOM	AUDITORIUM	0.15	2.1	0.50	1.0
+OUTSIDE	        CITY		0.0	0.0	0.00	0.0
+DUNGEON	        AUDITORIUM	0.15	2.5	0.50	1.0
```
lynxlynxlynx: these two are offset due mix of whitespace

MarcelHB: Raw extracted that ... I will fix this.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
MarcelHB: Ok, I try to move it then.

lynxlynxlynx:
big work & interesting! I've checked the other games: IWD is also missing them. PST has values in that reverb field (2-5 from a random sampling), eg 4 for most outdoor areas, which matches reverb.2da. So it should be used. Too bad we don't have any RE experts around anymore — with that options switch, it would be trivial to find what code uses it and then more interestingly, what it actually does.

I suggest you store the resolved reverb id in the song header field (like you expected from the pst case), so it doesn't have to be looked up all the time. Think of entering and exiting buildings, where it would keep switching between the master area reverb and something else.

We are conservative when it comes to compilers as not all platform that we work on have access to recent ones. And historically Avenger stuck to MSVC5, as it had a better debugger than later free editions. But that particular problem is gone, so we can slowly remove the hacks (like the loop variable redeclaration bug).

MarcelHB:
So, you like to have the reverb ID set at map parsing time? But we keep these lists then, right?

I just moved to something silly for this clang issue, I'll rebase this history a little when the CI gives green light.

lynxlynxlynx:
Yes, in AREImporter. Then you use that field in core and PST automatically works (with slightly different reverb levels). You can reuse `core->HasFeature(GF_PST_STATE_FLAGS)` to single it out from reassignment.

MarcelHB:
Ok, I hopefully solved all the build issues.

Choosing reverb data has been moved to `AREImporter` process time.

Feel free to test.

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -478,7 +478,15 @@ Map* AREImporter::GetMap(const char *ResRef, bool day_or_night)
...
 		str->ReadDword( map->SongHeader.SongList + i );
 	}
-	str->Seek( RestHeader + 32, GEM_STREAM_START );
+
+	if (core->HasFeature(GF_PST_STATE_FLAGS)) {
+		str->Seek( RestHeader + 32, GEM_STREAM_START );
+		str->ReadDword(&map->SongHeader.reverbID);
+	} else {
+		map->SongHeader.reverbID = EFX_PROFILE_REVERB_INVALID;
+	}
```
lynxlynxlynx: The seeks look odd, am I missing something?
1. The non-pst case also needs to seek forward for the next reads to start at the right location.
2. The rest header is a separate structure, so I don't think it's reading the reverb id either. It looks 48 bytes ahead of the last song resref to me:
https://gibberlings3.github.io/iesdp/file_formats/ie_formats/are_v1.htm#formAREAV1_0_Song_entries

So I suggest you put the original seek after the whole if block and add a new one to get the reverb id to the pst case. Or use the if only to reset the id for non-pst games, after reading the value.

lynxlynxlynx:
Besides that, it looks good. :+1: 

the reverb.2da override is to make things sound better? 2das support comments and I'd like to add one in it to explain why it's needed / what changed wrt the original ones. Eeh, that reminds me: since pst used the same file, we now override it — the only change being the reverb level in the dungeon (1 vs 0.7). If you want to keep pst using the original and override the rest, the 2da needs to get out of shared/ or less cleanly, pst needs its version recopied into our override.

MarcelHB:
I wasn't sure which games do or don't have this file (BG doesn't) and the remaining ones look identical (aside from whitespace formatting) so I guess there was no harm.

Even if PST has some minor difference, since we haven't figured out what to do with the other columns, it's probably less relevant for now. I don't have it on the disk right now.

gemrb/plugins/AREImporter/AREImporter.cpp
MarcelHB: Oh, mistake, yes.

MarcelHB: SongHeader + 0x50 should be correct, rest-parsing was misused.

lynxlynxlynx:
So why not put it into unhardcoded then? It would mean the original get used when present.

MarcelHB:
Just thought "same name as shipped" => `overrides`. I moved it.

MarcelHB:
Meh, I'm not sure how you master releases but AppVeyor gives:

`Looking for include file C:/include/AL/efx.h - not found`

Idea?

lynxlynxlynx:
this table explains the search paths a bit better:
http://www.gemrb.org/wiki/doku.php?id=developers:mods#how_to_start

but basically our override should have only overrides that are really needed — true overrides. In this case it doesn't matter that much since probably nobody touches reverb.2da, but a principle is a principle.

I think for appveyor you have to use check_include_file_cxx — it's not a simple file existence check.

MarcelHB:
Hm, no help.

lynxlynxlynx:
sucks that it's annoying to dig out the actual program it tries. The function can be further modified and maybe setting the include path would do:
https://cmake.org/cmake/help/v3.0/module/CheckIncludeFiles.html

otherwise you can always use find_file instead.

MarcelHB:
No, all of them failed. But `EXISTS` works - at least I could reproduce all failures and successes. :see_no_evil:

MarcelHB:
We can't get an AppVeyor PR build artifact for validation, do we?

lynxlynxlynx:
There aren't any or they expired (at least for the main builds appveyor has a temporary copy). I'll merge this now and you can download the most recent build afterwards.

Thanks again! I'm deep in a prerelease testing run and no major snags so far, so regular users should be able to see your recent work sometime in November. :)

MarcelHB:
Nice. Build tested, sounds fine!

PR #77, opened 2017-11-28
Fix some spelling issues reported by lintian

berenm:
Hello,

Here is a patch for some spelling issues that the lintian tool reported in the debian gemrb package. I guessed you might be interested.

Cheers,


lynxlynxlynx:
Thanks!

PR #80, opened 2018-03-28
CMake: adds some tweaks for MinGW64

MarcelHB:
I noticed that I always had some manual tweaks to get gemrb compiled on most recent MinGW64, like missing definitions for Python, missing libraries on working with SDL2 on Windows and a something I'm not really sure about on the Ogg plugin: since these static callbacks are never used, yet enabled (except on Mac), they break for unused variable errors.

These changes are very narrow but as AppVeyor and Clang seem to work, I prefer the changes to be conservative.


lynxlynxlynx:
Looks fine, thanks!

PR #81, opened 2018-03-29
IWD2: savegame compatibility fix

MarcelHB:
When saving an IWD2 game in GemRB, loading the game into the original game reveals an interesting issue: you cannot enter buildings anymore. Doors open correctly and the cursor shows a trigger, but nothing actually happens.

I diffed the ARE files on triggers and found a significant difference: The "alternative use point" in IESDP claims that there are two words used for x/y coords. Actually, in 9.1 AREs, these 4 bytes are unused, and instead it makes use of two dwords right behind that in the *unknown* region.

Then, I made IWD2 both read and write into these regions in addition. It's a little workaround on the lower two bytes since x/y coords are `short`s internally and I better don't simply touch that now.

Anyway, now we can better cross-play as far as I have advanced, which makes error-spotting a little easier. ;)

### DIFF
--- a/gemrb/plugins/AREImporter/AREImporter.cpp
+++ b/gemrb/plugins/AREImporter/AREImporter.cpp
@@ -552,8 +552,16 @@ Map* AREImporter::GetMap(const char *ResRef, bool day_or_night)
 		str->ReadResRef( Script );
 		str->ReadWord( &PosX);
 		str->ReadWord( &PosY);
-		//maybe we have to store this
-		str->Seek( 36, GEM_CURRENT_POS );
+		/* ARE 9.1: 4B per position after that, but let's just try the lower two ones. */
+		if (16 == map->version) {
+			str->ReadWord(&PosX);
+			str->Seek(2, GEM_CURRENT_POS);
+			str->ReadWord(&PosY);
+			str->Seek(30, GEM_CURRENT_POS);
+		} else {
+			//maybe we have to store this
+			str->Seek( 36, GEM_CURRENT_POS );
+		}
 
 		if (core->HasFeature(GF_INFOPOINT_DIALOGS)) {
 			str->ReadResRef( WavResRef );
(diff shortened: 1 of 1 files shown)


bradallred:
shouldn't this check the version in the ARE header? rather than `core->HasFeature(GF_IWD2_SCRIPTNAME)`

good find. 

as an aside, converting Point to use `int` would be a welcome patch ;) I did a quick change (to the point where it compiles) and yeah, its an unfortunate amount of changes and any importer that uses points would have to be changed. I could post a patch but it would be based off the subviews branch.

bradallred:
I'm assuming in this case it is a difference in version else this problem would appear in other games (maybe it does? we do crash when loading a GemRB save in BG2)

Is the afore mentioned door fix needed now that this is fixed?

BTW, I would suspect the other file loaders using `Point`s to have the same issue for IWD2 at least. Its hard to imagine they used 2 bytes in some files and 4 in others. WEDImporter comes to mind... not sure what else.

re `Point`: of course. that makes sense.

MarcelHB:
From what I see in the code: unlikely, since it [0] does flag magic [1] for this game. But I can test that later this day. I can also try BG2 again.

About WED: Valid point. I'm not too familiar with its data but under this assumption, I guess there would be much more corruption than I have noticed yet.

[0] https://github.com/gemrb/gemrb/blob/master/gemrb/plugins/AREImporter/AREImporter.cpp#L338
[1] https://gibberlings3.github.io/iesdp/file_formats/ie_formats/are_v9.1.htm#formAREAV9_1_Door

MarcelHB:
1. This IWD2 special door handling cannot be removed, otherwise doors won't work at all.
2. I had no difficulties with BG2, bidirectional load and triggers work on any save game (both unpatched and patched).

lynxlynxlynx:
Slow down. :) Please no Point conversion until we sort out subviews — it would only make the merge harder.

This patch doesn't affect the flags, so the door fix is still needed. But yes, both could check versus the version instead.

I'll check the data a bit, it's murkier than it looks.

lynxlynxlynx:
Here's a dump of the points:
http://paste.debian.net/1017766/
You can see that even regions that have the bit set (useAlt) have inconsistent values. 

Looking at ar1000 in nearinfinity it is obvious only travel regions have alternative use point set. The code confirms something was stored there too (and we're naming it differently!):
https://github.com/NearInfinityBrowser/NearInfinity/blob/9bd2d8b18ce968ea5b27ba6b6fe724a2cf94bf09/src/org/infinity/resource/are/ITEPoint.java#L192

... yet none of the travel regions have the alt bit set and only two non-zero current point:
http://paste.debian.net/1017776/

... but actually iwd2 renames (see TRAP_USEPOINT setup) also a bit here, so ignore the mentions above.

So I think everything is ok, except that we use a different name.

MarcelHB:
Ok, so I get that 
* the NI code looks a little more comprehensive in terms of completeness on the data here, and `AREImporter` remains open anyway,
* `OVERRIDE_POINT` possibly sets all travel trigger points to (0,0) or is ignored completey but `ALTERNATE_POINT` seems to come to the rescue -- even there is no bit indicator to say so explicitly, right? Aside from the two outliers.
* for the larger list, anything to respect there? These notably large values hardly look like coordinates. If they aren't indeed, we have unknown data there, which isn't known to NI too, right?

So what's actually left to do: Changing the check to include IWD as well? Treating these 2x4B as coordinates in case of travel points only?

lynxlynxlynx:
In iwd2 the bit is the non-npc one (0x200). It matches the dump above in the sense that all but two travel regions are set up to use the alternate point (in NI speech).

What's left to do? The GF check already includes iwd1, but since we don't know of any saving problems there, it's more likely this would break something. So changing it to a format version check would be better.

MarcelHB:
Modified that.

lynxlynxlynx:
Thanks!

PR #82, opened 2018-04-05
ControlAnimation: palette updates on pause

MarcelHB:
When playing IWD2 and hitting Pause, then changing into the inventory, the paperdoll's palette is never adjusted, so we see this standing frozen:

paperdoll

`UpdateAnimation` will always either skip updating or proceed in frames, which is both not fully what is useful here. I reworked that a little in a way that `SetPaletteGradients` will cause redraw without continuing the animation.

The whole thing has some annoying design, but at least this will stop rainbow colours here.


lynxlynxlynx:
Thanks, looks pretty clean to me. Any thoughts @bradallred?

bradallred:
Yeah, probably nothing to worry about. I dont know enough about it to make any improvement suggestions so I would say merge it.

PR #85, opened 2018-04-22
Win64 fixes for finding files and making directories

nerd-rage:
Tried to build and run on 64-bit Windows, had to change a couple lines that assumed 32-bit. Also strtok was called with a wrong constant.


lynxlynxlynx:
Looks good, thanks! Why is the strtok change needed though, surely nulls are not valid path chars on windows?

nerd-rage:
strtok expects a 0-terminated string as its second parameter. As luck would have it, PathDelimiter wasn't followed by a 0 in my builds, but by 'c', so "Cache2"  was split in to "Ca" and "he2", making a wrong folder with a subfolder.

lynxlynxlynx:
uff, even the man page doesn't mention that. Good catch!

PR #86, opened 2019-01-07
Some more bugfixes

Tomsod:
I took another try at BG, this time finishing the first game.
These are some of the bugs that I uncovered this time.


Tomsod:
Huh.  It seems d6fc8d9 fails one of the builds because of an unsigned/signed comparison.  I've seen this warning before, but I ignored it because that comparison was precisely what I meant to write, and converting one of the arguments to signed would be, IMO, slightly wrong.  Is this crucial?

Tomsod:
Oh, and I also have a list of other bugs (as of v0.8.5) that I noticed but haven't bothered to fix.  Where do I best post it?  Todo wiki page, perhaps?

Tomsod:
Update: I googled a bit, and apparently the mixed sign comparison thing can cause actual errors on 32-bit platforms?  The more you know.  I'll fix this now.

lynxlynxlynx:
Great, thanks! I'll soon be able to review in detail, but first a question - is the dual kit stuff just for the gnomes/illusionists?

Tomsod:
No, gnomish whatever/illusionist characters are multiclass.  Dualing into a kit was possible in BG1, wherein a human PC (like Imoen) could choose to specialize in a school when dualing into a mage.  In BG2 that option was removed, since other classes could have kits and CRE does not support a kit dualed into another kit.  But if we place a limit of one kit per character, CRE can unambiguously define that kit, even when it's for a second class.  The game just needs to be taught to recognize which class is kitted, which is what I did.

lynxlynxlynx:
Actually iwd2 uses the kit field as a bitfield and there's a slight chance the rest did too, since there are actions that can effectively stack kits. But i doubt it. I was asking to see if it was a crazy mod thing, as specialist mages have another layer through their schools - mageschool.2da. It seemed a bit redundant when we were ironing out the class/lu code, but it appears there is a need for it. I'll check if we can simplify the solution for this.

lynxlynxlynx:
The first two fixes are in, but I reworked the first (0a0cb62) — now it's togglable in the table, so there's less code and people can tweak as desired.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -925,7 +925,7 @@ bool Actor::ApplyKit(bool remove, ieDword baseclass, int diff)
 			if (multiclass & msk) {
 				max = GetLevelInClass(i);
 				// don't apply/remove the old kit clab if the kit is disabled
-				if (i == kitclass && !IsDualInactive()) {
+				if (i == kitclass && !IsKitInactive()) {
```
lynxlynxlynx: this inverts the check for cases covered by the previous one.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -1131,7 +1131,7 @@ static void pcf_class (Actor *actor, ieDword /*oldValue*/, ieDword newValue)
 {
 	//Call forced initbuttons in old style systems, and soft initbuttons
 	//in case of iwd2. Maybe we need a custom quickslots flag here.
-	actor->InitButtons(newValue, !iwd2class);
+	actor->InitButtons(actor->GetActiveClass(), !iwd2class);
```
lynxlynxlynx: if i remember testing correctly, this is not how the original behaved. Do you have any data to the contrary?

gemrb/GUIScripts/GUICommon.py
```diff
@@ -408,12 +408,15 @@ def GetActorClassTitle (actor):
...
 					ClassTitle = CommonTables.KitList.GetValue (Dual[1], 2)
-				elif Dual[0] == 2:
+				else:
 					ClassTitle = CommonTables.Classes.GetValue (GetClassRowName(Dual[1], "index"), "CAP_REF")
 				if ClassTitle != "*":
 					ClassTitle = GemRB.GetString (ClassTitle)
 				ClassTitle += " / "
-				ClassTitle += CommonTables.Classes.GetValue (GetClassRowName(Dual[2], "index"), "CAP_REF", GTV_REF)
+				if Dual[0] == 3:
+					ClassTitle += CommonTables.KitList.GetValue (Dual[2], 2, GTV_REF)
```
lynxlynxlynx: Please use named row/column lookups wherever you can. We specifically moved to that to be safe about any table changes and crossgame differences.

gemrb/GUIScripts/GUICommon.py
```diff
@@ -502,35 +505,36 @@ def IsDualClassed(actor, verbose):
...
 					ClassIndex = CommonTables.Classes.FindValue ("ID", i)
 					if ClassIndex == FirstClassIndex:
-						Mask = 1 << i
 						continue
-					DualInfo.append (ClassIndex)
+					SecondClassIndex = ClassIndex
 					break
-				Mask = 1 << i
-			if len(DualInfo) != 3:
+			else:
```
lynxlynxlynx: nice, not a construct one sees often!

gemrb/GUIScripts/LevelUp.py
```diff
@@ -159,20 +159,23 @@ def OpenLevelUpWindow():
...
 		# not dual, must be single
 		if IsDual[0] == 0:
 			Classes = [Class]
-		else: # make sure Classes[1] is a class, not a kit
-			if IsDual[0] == 1: # kit
-				Classes[1] = CommonTables.KitList.GetValue (IsDual[1], 7)
-			else: # class
-				TmpClassName = GUICommon.GetClassRowName (Classes[1], "index")
-				Classes[1] = CommonTables.Classes.GetValue (TmpClassName, "ID")
+		else: # resolve kits to classes (new class goes first)
```
lynxlynxlynx: why the reverse order?

gemrb/core/Scriptable/Actor.cpp
Tomsod: Sorry, I don't comprehend your comment.  The previous what?  What is inverted?

Anyway, this part of code seems to govern updating kit abilities e.g. on level up.  It used to skip inactive duals since their kit must be also inactive, but here I add the aforementioned "rudimentary support for dualing into kits" by checking if it's truly inactive.  Right?

gemrb/core/Scriptable/Actor.cpp
Tomsod: I've just re-tested this in BG2 (a Trilogy build with ToBEx, to be specific) by creating a thief PC and dualing him into mage.  After dual he got spell quickslots instead of thief ability buttons, only regaining the latter after thief class reactivation.  It might possibly be a ToBEx thing, though.

gemrb/GUIScripts/GUICommon.py
Tomsod: Oh, sorry, was not aware.  Does this also apply for C++ code?  I might need to redo a couple of functions if so.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: What I meant was, if we forget about the extra case now handled. IsKitInactive calls IsDualInactive. But the return value is inverted, so the behaviour for other cases changed.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Ok, I'll recheck on vanilla.

gemrb/GUIScripts/GUICommon.py
lynxlynxlynx: It does, but there we try to preload everything. That one call in core getkitindex or whatwasit is an outlier in both senses.

gemrb/GUIScripts/LUCommon.py
```diff
@@ -293,8 +300,11 @@ def SetupHP (pc, Level=None, LevelDiff=None):
 		#we only get the hp bonus if the old class is reactivated
 		if (Levels[0]<=Levels[1]):
 			return
-		Class = GUICommon.GetClassRowName(Dual[2], "index")
-		Class = [CommonTables.Classes.GetValue (Class, "ID")]
+		if Dual[0] == 3:
+			Class = [CommonTables.KitList.GetValue (Dual[2], 7)]
+		else:
+			ClassRow = GUICommon.GetClassRowName(Dual[2], "index")
+			Class = [CommonTables.Classes.GetValue (ClassRow, "ID")]
```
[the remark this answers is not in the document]
Tomsod: Yes, much of 899fb8b could benefit from refactoring, since there's a lot of code like
```
if kit:
    action
else: #class
    similar action but phrased differently
```
But I guess I was feeling hacky when I wrote it.

gemrb/GUIScripts/LevelUp.py
Tomsod: I don't even know!  I was just parroting the line 162 above.

Tomsod:
Re: simplification of 7aeebba (right?)
You're the boss!  Although my fix is not _that_ complex or computationally diffucult either.

Re: b479f08 (weapon headers)
My bad: the header only seems to be reset when you load the game with the weapon in question being actively wielded.  It is preserved on re-equip even w/o my fix.

Re: d6fc8d9 (fatigue)
For instance, Haste tries to modify fatigue directly when it wears off.  In my experience, the fatigued icons only last for a tick and then removed.  Old Actor.cpp:3484 clearly rewrites it entirely.

(I feel like there must be some way to reply to those three comments directly, but I'm not seeing it?  Oh well.)

gemrb/core/Scriptable/Actor.cpp
Tomsod: Nope, `IsKitInactive()` returns `false` if `IsDualInactive()` is `false`, so no inversion here.  Notice the exclamation point:
`if (!IsDualInactive()) return false;`

gemrb/GUIScripts/GUICommon.py
Tomsod: Yes, things like that are what makes Python such a beautiful language.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Yes, but the call to IsKitInactive is negated too.

Tomsod: But the call to `IsDualInactive()` it replaced was also negated!  So there's a triple negation instead of a single negation, which means no change.  Although I now see how this can be confusing.  Should I add a comment somewhere?

gemrb/GUIScripts/GUICommon.py
Tomsod: OK, then I'll add lookups in Python and preloads in core.

lynxlynxlynx:
I should've commented on the commits, it seems the review is not posted there.

Re: b479f08 (weapon headers): ah, ok. Pushed.

Re: d6fc8d9 (fatigue): no, that only sets the base and Modified is recalculated. But thanks for the test case — let's see.

gemrb/GUIScripts/GUICommon.py
lynxlynxlynx: Wait a bit with this dualclass stuff. I feel like it can be redesigned, but need to both read the rest of it thoroughly and think about other options. The aforementioned MC bit would only take care of a part.

cygarniczka:
Cavalier Kit Level Up Button doesn't work. Tested both on PC and mod added Cavalier Isra.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: You're right, sorry. No need for comments, since this is confusing only while looking at the diff.

lynxlynxlynx:
@cygarniczka with this PR or without?

cygarniczka:
PR?

lynxlynxlynx:
pull request... the code we're discussing here.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3463,38 +3463,46 @@ void Actor::UpdateFatigue()
...
-	int FatigueBonus = core->GetConstitutionBonus(STAT_CON_FATIGUE, Modified[IE_CON]);
-	// pst has TNO regeneration stored there
-	if (core->HasFeature(GF_AREA_OVERRIDE)) FatigueBonus = 0;
-	FatigueLevel = (signed)FatigueLevel - FatigueBonus >= 0 ? FatigueLevel - FatigueBonus : 0;
-	FatigueLevel = ClampStat(IE_FATIGUE, FatigueLevel);
-
-	// don't run on init or we automatically make the character supertired
-	if (FatigueLevel != BaseStats[IE_FATIGUE] && TicksLastRested) {
-		int OldLuckMod = LuckMod;
-		NewBase(IE_FATIGUE, FatigueLevel, MOD_ABSOLUTE);
```
Tomsod: Re: modified fatigue being recalculated, I can't find IE_FATIGUE being changed in Actor.cpp anywhere else but here before my fix.  And recall that `UpdateFatigue()` happens in the middle of `RefreshEffects()`, shortly after all the `Modified[]` is reset to `Base[]`.  And, importantly, `NewBase()` actually changes `Modified[]`, too.

lynxlynxlynx: RefreshPCStats is ran after all the effects have been applied and Modified restored (save for any pcf changes — irrelevant for fatigue). NewBase makes sure to conserve the diff between BaseStats and Modified. Anyway, it's better to just test it out properly, which I already have on the todo.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
```diff
@@ -2367,7 +2367,29 @@ int fx_to_hit_modifier (Scriptable* /*Owner*/, Actor* target, Effect* fx)
...
+	case MOD_ADDITIVE:
+	default:
+		target->ToHit.HandleFxBonus(fx->Parameter1, fx->TimingMode==FX_DURATION_INSTANT_PERMANENT);
+		break;
+	case MOD_ABSOLUTE:
+		if (fx->TimingMode == FX_DURATION_INSTANT_PERMANENT) {
+			target->ToHit.SetBase(fx->Parameter1);
+		} else {
+			//FIXME: two such effects probably should not stack, but they do for now
+			target->ToHit.SetFxBonus(fx->Parameter1 - target->ToHit.GetBase(), MOD_ADDITIVE);
```
lynxlynxlynx: This is tricky. The generic bonus is also based on effects, so undoing things properly is harder. The whole system was added for iwd2, while bg2 just used one main stat. According to IESDP, it only worked on the base stat too, but there could have been a mistake.

So in this case we'd ideally be affecting the total. Not sure if it's worth bothering with, since I don't know of any users.

The more actionable problem is that this line doesn't work for iwd2 — it would just set the bonus to param1, since bab starts at 0.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4246,7 +4246,7 @@ bool Actor::CheckSpellDisruption(int damage, int spellLevel)
 	int concentration = GetSkill(IE_CONCENTRATION);
 	int bonus = 0;
 	// combat casting bonus only applies when injured
-	if (HasFeat(FEAT_COMBAT_CASTING) && BaseStats[IE_HITPOINTS] != Modified[IE_HITPOINTS]) {
+	if (HasFeat(FEAT_COMBAT_CASTING) && Modified[IE_MAXHITPOINTS] != Modified[IE_HITPOINTS]) {
```
lynxlynxlynx: probably too strong either way, but we haven't reverse engineered this — it could be it's applied only when an injured hp interval or worse is in effect.

Good catch.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
Tomsod: In the BG2 section IESDP says the opcode works on modified THAC0, but IIRC it turned out not to be true for potions of heroism (which were the reason for this fix).  I'll recheck this.
IDW2 may need special handing anyway, since from what I understand changes to BAB can also affect APR.  Or not?
Also I don't get why `ToHit.GetBase()` would return 0 for IWD2?  `Actor::SetBaseAPRandAB()` seems to set base pretty much at the same time both for 2e and 3e.

Tomsod: Okay, so I tested a bit on my BG2 install and the percentage modifier, at least, does not respect previous changes by the additive modifiers of the same opcode, just as I remembered.  I'm pretty sure by now that it does operate on base THAC0 (but I haven't tested _all_ cases).  The absolute modifier is indeed non-cumulative with itself; furthermore, it's not cumulative with the percentage modifiers either!  (Only the latest applied works.)  The percentage modifier is not cumulative with itself, either.  It seems that the percentage type works more like absolute type behind the scenes, rather than like a bonus.  The most faithful to original way to reproduce it would be, it seems, to create _yet another_ ToHit field that would be populated exclusively by those two modifier types.  Or we could make `base` rewritable without affecting `Owner`.
On the topic of users: after searching with DLTC Editor, it seems that at least some "natural weapon" items do use the absolute modifier of this opcode, e.g. `SLAYERWP.ITM`.  They're by nature unstackable, though.

gemrb/core/Scriptable/Actor.cpp
Tomsod: Again, I've no idea how it's supposed to behave as I don't even have IWD2.  I just saw HP being compared with itself and guessed the intended meaning.

Tomsod:
It should, yes.  So far, a824f55, 699ff59, b479f08 and 7c5b9c8 were merged with some tweaks.

cygarniczka:
I just added some 1PP bugs to the list with paperdolls and avatars. Components Avatar Switching and Thief Galore don't work properly probably because they make BGMain.exe edits.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
lynxlynxlynx: IWD2: I was just going off the fact that base starts at 0, not 20, but didn't check how we store it. 

Likely not worthy to lose more time on this. What some people could miss is the tobex extension, which is anothermod type though. It was introduced to handle stacking better.

lynxlynxlynx:
@cygarniczka this PR is unrelated to your play, so please don't comment here.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7447,6 +7447,12 @@ void Actor::UpdateActorState(ieDword gameTime) {
 		lastattack = 0;
 	}
 
+	// delayed HP adjustment after max HP modification
+	if (checkHP) {
+		checkHP = 0;
+		pcf_hitpoint(this, 0, BaseStats[IE_HITPOINTS]);
+	}
+
```
lynxlynxlynx: this is an ugly place to move the check to, did it not work in any of the more natural candidates dealing with effect application?

Tomsod: I guess it is.  But the whole idea of this hack was to delay HP check for a tick, and anything in `RefreshEffects()` runs way too often for that.  We _could_ move the check into, say, `RefreshHP()`, but we'd have to manually check if a tick has passed.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
```diff
@@ -1782,7 +1782,11 @@ int fx_morale_modifier (Scriptable* /*Owner*/, Actor* target, Effect* fx)
 {
 	if(0) print("fx_morale_modifier(%2d): Mod: %d, Type: %d", fx->Opcode, fx->Parameter1, fx->Parameter2);
 
-	//FIXME: in bg2 this is hacked to set param1=10, param2=1, we might need some flag for this
+	if (core->HasFeature(GF_FIXED_MORALE_OPCODE)) {
+		BASE_SET(IE_MORALE, 10);
```
lynxlynxlynx: This makes it permanent, which I highly doubt was the intention. Would be better to just reset param1&2 (once, so check fx->FirstApply).

gemrb/GUIScripts/LUProfsSelection.py
```diff
@@ -221,8 +221,10 @@ def SetupProfsWindow (pc, type, window, callback, level1=[0,0,0], level2=[1,1,1]
 			# don't give too many points; exact formula unknown, but this works with f/m and f/m/t
 			ProfsPointsLeft += sum(level2)/IsMulti[0]/ProfsRate
 		else:
-			# sum the levels, since the rate is for the combined multiclass
-			ProfsPointsLeft += (sum(level2) - sum(level1))/ProfsRate
+			# average the levels, since the rate is for the combined multiclass
+			# FIXME: should give points when the fasted-profs-rate class (f > c/t > m) levels,
+			# instead of using the average level
+			ProfsPointsLeft += sum(level2)/IsMulti[0]/ProfsRate - sum(level1)/IsMulti[0]/ProfsRate
```
lynxlynxlynx: Please provide exact test cases, so they can be added to the comments or at least the commit description. It's great for when you look at this again in 5 years.

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -601,11 +601,6 @@ Map* AREImporter::GetMap(const char *ResRef, bool day_or_night)
 		ip->StrRef = StrRef; //we need this when saving area
 		ip->SetMap(map);
 		ip->Flags = Flags;
-		// ensure repeating traps are armed, fix for bg2 ar1404 mirror trap to fire
-		// BUT probably not in PST, since that breaks the mausoleum portal in ar0200
-		if (ip->TrapResets() && !core->HasFeature(GF_PST_STATE_FLAGS)) {
-			ip->Trapped = true;
-		}
```
lynxlynxlynx: that's the main point of resetting traps, they reset. However Zelazko put it better, was this the real problem you faced?

> resettable traps can't be disarmed after they reappear

gemrb/core/GameScript/Triggers.cpp
```diff
@@ -2088,7 +2088,16 @@ int GameScript::CheckStat(Scriptable* Sender, Trigger* parameters)
 		return 0;
 	}
 	Actor* actor = ( Actor* ) target;
-	if ( (signed) actor->GetStat( parameters->int1Parameter ) == parameters->int0Parameter) {
+	int stat;
+	// hack for mods that reuse 'empty' stats 176-182, corresponding to scripting states 20-26
+	if (core->HasFeature(GF_FAKE_SCRIPTING_STATES)
+	    && (parameters->int1Parameter >= IE_SCRIPTINGSTATE1 + 20)
+	    && (parameters->int1Parameter <= IE_SCRIPTINGSTATE1 + 26)) {
+		stat = actor->fakeScriptingStates[parameters->int1Parameter - IE_SCRIPTINGSTATE1 - 20];
```
lynxlynxlynx: Indeed a hack! I forgot why Avenger wanted to keep the stat count below 255, but there are pst-only stats that can be used for this instead. The first two of the current list and surely more, but there's a simpler candidate: just refactor  IE_SPLSTATE_ID1 as the TODO notes. A simple thing, since it's not a saved stat and only has a few users.

gemrb/plugins/AREImporter/AREImporter.cpp
Tomsod: Resetting traps are supposed to reset _after triggering_, not after being disarmed!  (Even if they were to reset, it wouldn't be triggered by reload but rather some in-game event.)  After a trap is properly disarmed by a thief, it shouldn't ever reappear on it's own and yes I've tested this.  You _can_ disarm them again, but it's both a chore and an XP farm.

gemrb/core/GameScript/Actions.cpp
```diff
@@ -5243,18 +5243,17 @@ void GameScript::AdvanceTime(Scriptable* /*Sender*/, Action* parameters)
 	core->GetGame()->ResetPartyCommentTimes();
 }
 
-//advance at least one day, then stop at next day/dusk/night/morning
-//oops, not TimeODay is used but Time (this means we got hours)
-//i'm not sure if we should add a whole day either, needs more research
+// advance at the beginning of the specified hour (minus one tick? unsure)
+// the parameter is HOURS (time.ids, 0 to 23)
+// never advance a full day or more (in fact, duplicating this action does nothing)
```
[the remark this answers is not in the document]
Tomsod: I've written my own script for testing, but arrival on the werewolf island in BG1 has many repeated 		`DayNight(MIDNIGHT)` and `DayNight(NOON)` actions to simulate the passage of time.

gemrb/GUIScripts/LUProfsSelection.py
Tomsod: I did describe the logic in the commit, but for exact examples, let's take a thief/mage.  She is supposed to get profs at T4/M3 and T8/M7, since thief is the proficiency-gaining class here.  With my fix she gets profs at T5/M4 and T8/M8, as I used the average level.  As I noted, it's _slightly_ incorrect.  Without my fix, she would not get profs at all when leveling one level at a time due to rounding errors, but if leveled from T1/M1 to T8/M8 all at once, she would gain _three_ profs instead of two.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
Tomsod: I figured it's supposed to ignore timing mode as well, yes.  My evidence is `SPPR108.SPL` (Remove Fear), which has instant/permanent timing for this opcode.  If it respected timing, wouldn't this spell give fixed morale forever?

gemrb/core/GameScript/Triggers.cpp
Tomsod: FWIW, there are only six IWD spell states under 255, and we need seven.  But other than that, I agree that moving spell states would look neater.  We can even dynamically allocate them based on  `SPLSTATE.IDS` length, like how ToBEx does with its 401+ stats.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3158,6 +3161,16 @@ void Actor::RefreshEffects(EffectQueue *fx)
...
 
+	//delayed HP adjustment hack (after max HP modification)
+	//as it's triggered by PCFs from the previous tick, it should probably run before current PCFs
+	if (first && checkHP == 2) {
+		//could not set this in the constructor
+		checkHPTime = core->GetGame()->GameTime;
+	} else if (checkHP && checkHPTime != core->GetGame()->GameTime) {
+		checkHP = 0;
+		pcf_hitpoint(this, 0, BaseStats[IE_HITPOINTS]);
+	}
```
Tomsod: Okay, how about here? Seems more natural, although I had to add a time check as it's reached way too often.

lynxlynxlynx: Better, yes — thanks! It would probably also work by checking against Ticks, since that's saved to disk only for areas  — which don't have hp.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -921,18 +948,73 @@ def SellPressed ():
...
+		#but double-click has another function here
+		OpenItemAmountWindow (STORE_BAG)
+		return
+	if Inventory:
+		return
+	pc = GemRB.GameGetSelectedPCSingle ()
+	RightIndex = GemRB.GetVar ("RightIndex")
+	Slot = GemRB.GetSlotItem (pc, inventory_slots[RightIndex])
+	Item = GemRB.GetItem (Slot['ItemResRef'])
+	if Item['Function'] & 4:
```
lynxlynxlynx: Possible future improvement: add named constants for these bits to GUIDefines.py; it would also make it obvious this is checking for a container.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -9383,13 +9474,43 @@ static PyObject* GemRB_SetPurchasedAmount(PyObject * /*self*/, PyObject* args)
 	Py_RETURN_NONE;
 }
 
+// a bunch of duplicated code moved from GemRB_ChangeStoreItem()
+static int SellBetweenStores(STOItem* si, int action, Store *store)
+{
+	CREItem ci = *reinterpret_cast<CREItem *>(si);
```
lynxlynxlynx: We have a CopySTOItem method to do this — as far as it is possible, since not all the fields are shared.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -1513,35 +1633,38 @@ def SetupItems (pc, Slot, Button, Label, i, type, idx, steal=0):
...
-			if not steal and type == ITEM_PC:
+			if not steal and type != ITEM_STORE:
 				Price = 1
 		else:
 			Name = GemRB.GetString (Item['ItemNameIdentified'])
 			Button.EnableBorder (0, 0)
 
 		GemRB.SetToken ("ITEMNAME", Name)
-		if Inventory:
+		if Inventory or type == ITEM_STORE and steal:
```
lynxlynxlynx: Please use parens in all such checks (mixed and+or), so the evaluation order is explicit.

gemrb/core/GameScript/Triggers.cpp
lynxlynxlynx: You can reuse also the existing 176/177, since they're pst-only.

Tomsod:
I've done the spell states thing.  Looks better, but I was unable to test it properly as I don't have IWD2, so please do test it!

`GetHighestValue()` is a separate func as it'll be useful for ToBEx `SetStat` opcode when it'll be fully implemented (it uses the same mechanic for determining how much space to malloc).

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -9546,21 +9566,21 @@ void Actor::SetOverlay(unsigned int overlay)
...
-	Modified[pos]|=bit;
+	if (spellStates[pos]&bit) return true;
+	spellStates[pos]|=bit;
 	return false;
 }
 
 //returns true if spell state is already set
 bool Actor::HasSpellState(unsigned int spellstate) const
 {
-	if (spellstate >= SS_MAX) return false;
```
lynxlynxlynx: SS_MAX is now an unused macro — please remove it too. No other comments, looks good!

Tomsod: Dang, I knew I forgot something!

Tomsod:
> Do you mean setting stats past the 255 range? I don't if that's worthy of copying, would need to see a user first.

The same BG2 mods that reuse stats 176-182 also use ToBEx stats 401+ for a similar purpose -- that's Sword Coast Stratagems in my case.  I even made my own implementation of that opcode to make the mod work (focusing on 401+ stats) but it was super hacky and yours is better.  I'll probably redo it properly next time I'll be playing the game.

lynxlynxlynx:
Does SCS extend stats.ids for that or does it rely on numbers?

Tomsod:
It definitely extends `stats.ids`.  As I've mentioned before, ToBEx determines how much stats to allocate by scanning that file, so it wouldn't work otherwise.

About the most recently pushed commit: is it possible to combine it with the previous one when merging?  I'm somewhat uneasy about contributing one-line commits.

lynxlynxlynx:
Ah, for detectable spells. Never thought about that, even though several mods use it. We have to check the hardcoded list too and add any missing synonyms:
https://github.com/Gibberlings3/SwordCoastStratagems/blob/master/stratagems/ds/ds.tph#L460

Our table is a true override, so whatever the mod does will be ignored. Meaning we should manually import any static synonyms, but the dynamic additions will cause problems — we'd need them in a separate file or some such. Luckily this is a maintained mod, so we could also improve compatibility there.

Re commits: both are fine standalone, but sure, I can squash them together.

Tomsod:
I think it would be better to start with game-supplied `stats.ids` which will already have modded additions, and _then_ add GemRB-specific indices.  If that's possible, anyway.

If you can squash commits, then that's good, since I was also unsure what to do about other one-line fixes proposed by you.  I'll write them as commits, then.

lynxlynxlynx:
The whole reason why we have the override is to not have to bother with differences between the games. IIRC the originals aren't complete either. They are invisible to gemrb, whether in a biff or in the game override, since our override has the highest precedence in the search path.

Tomsod:
> Did you see the edit before?

Oh, sorry, either GitHub doesn't show edits until you refresh or I'm blind.  Anyway, I couldn't find this file in my install, so that's a no.

gemrb/GUIScripts/GUISTORE.py
Tomsod: Honestly, `and` being of higher precedence than `or` seems super self-evident to me, but I guess I'm not everyone.

gemrb/plugins/GUIScript/GUIScript.cpp
Tomsod: It's even integrated into a constructor!  I completely missed this.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
lynxlynxlynx: Ah, indeed and the immunity effect that would block it is later in the queue. Odd spell, even more if you consider the description. FX_PERMANENT might be more semantically correct, but it doesn't matter at all.

gemrb/plugins/AREImporter/AREImporter.cpp
lynxlynxlynx: Had amazing luck with this today, just because I was using a different ie data tool.

lynxlynxlynx:
I'm swamped at work, so the remaining review/testing will have to wait. 

For the stats.ids stuff, I started a discussion on the forum. Not much focused response, needs a poke, but anyway, the easiest would be to modify the mods to also install a copy as let's say stats2.ids and then we don't have a name collision and can do a merge like we do with triggers, actions and projectiles (gem.*.ids).

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: So, some numbers. The spell just adds 4 fatigue points, while 7 is the normal limit for feeling tired. Zelazko confirmed in the original that using it right after resting doesn't make you sleepy (Xzar with 10 Con and Viconia with 8 Con didn't get tired), so at least this is consistent and just the spell description is misleading.

Looking at the effect, it's of course delayed, but then doesn't stick around. Are you sure your version fixed haste for you? It should affect the base stat to have the desired effect, since it's not setting a good Modified for long.

Tomsod: I'm back!  Yes, on further checks, Haste does not appear to affect fatigue at all, even after my fix.

But I think this might be because of another bug.  Look here:
https://github.com/gemrb/gemrb/blob/308010894dcf6e9fb811d36bfba6a4048564a79b/gemrb/core/EffectQueue.cpp#L209-L214
This array determines how delayed effect timing types transform after the delay expires. Is `FX_DURATION_DELAY_PERMANENT` really supposed to decay into `FX_DURATION_AFTER_EXPIRES`?  If I change it to `FX_DURATION_INSTANT_PERMANENT`, Haste fatigue starts to work (with my fix).

lynxlynxlynx: I came to a similar conclusion — that either this or the opcode needs to change. But before changing something so central as fx_triggered, I'd like to see all other users, so probably need to write a short iesh script. Looks like NI can only look at externall eff files when searching, while dltcep can't do even that.

lynxlynxlynx: Wrote the script and there are many effects with this timing mode in bg1. Most are just to play a sound on spell expiry, but here are the examples of each:

> SPIN103 | 174 0xae play sound (bhaal power Draw upon holy might)
> SPIN117 | 12  0xc damage
> SPIN117 | 139 0x8b display string
> SPIN996 | 151 0x97 ReplaceCreature
> SPPR205 | 146 0x92 Spell:Cast
> SPPR207 | 122 0x7a Item:CreateInventory
> SPWI402 | 124 0x7c DimensionDoor
> SPWI416 | 172 0xac Spell:Remove
> SPWI982 | 68  0x44 UnsummonCreature FX_APPLIED in second case

All but one case of the last opcode are one-off effects and they'd just pollute the effect queue if we kept them around. But your suggestion should still work, just didn't test if there are 2 correct applications. BG2 has more examples of which only ApplyEffect is problematic, since it can return any of the application modes by design. But it does seem like the fatigue opcode itself is fine and doesn't only work on the base stat.

Tomsod: I think the effects won't be kept if they return `FX_NOT_APPLIED`, regardless of timing mode.  So, do I add the `fx_triggered` fix?  Or does it need more testing WRT `ApplyEffect`?

lynxlynxlynx: I want to check the other games first, but so far it does seem like the change is fine. ApplyEffect should be fine too. In bg2 only these four spells use it with this timing mode: SPWI324, SPWI720, SPCL742, SPIN891. All are ok, 3/4 just using it to play an expiry sound.

lynxlynxlynx: The rest looks fine too, not completely certain only about one very complicated pst opcode, but the green light is lit.

For posterity's sake, all the opcodes used with this timing mode:
> bg1 hits:
> SPIN117 | 12  0xc damage
> SPWI982 | 68  0x44 UnsummonCreature
> SPPR207 | 122 0x7a Item:CreateInventory
> SPWI402 | 124 0x7c DimensionDoor
> SPIN117 | 139 0x8b display string
> SPPR205 | 146 0x92 Spell:Cast
> SPIN996 | 151 0x97 ReplaceCreature
> SPWI416 | 172 0xac Spell:Remove
> SPIN103 | 174 0xae play sound
> 
> bg2
> SPIN952 | 0xd / 13 death
> SPWI305 | 0x5d / 93 fatigue PERMANENT
> SPIN916 | 0x8d / 141 VisualSpellHit
> SPIN846 | 0xa8 / 168  RemoveCreature
> SPIN952 | 0xaa / 170 DamageAnimation
> SPWI720 | 0xb1 / 177 ApplyEffect variable!
> SPWI910 | 0xd3 / 211  State:Imprisonment  
> SPWI813 | 0xd5 / 213 Maze 
> SPIN793 | 0xd7 / 215 PlayVisualEffect  variable!
> SPWI705 | 0xe6 / 230 DispelSecondaryTypeOne
> SPWI705 | 0xf0 / 240 Icon:Remove
> SPPR720 | 0x108 / 264 DropWeapon
> SPWI304 | 0x109 / 265 ModifyGlobalVariable
> 
> ApplyEffect bg2: SPWI324, SPWI720, SPCL742, SPIN891
> 
> pst extras:
> 0x1a / 26 | SPWI409 
> 0x6e / 110 | SPWI505 
> 0x8a / 138 | SPWI103 
> 0xbb / 187 | SPWI701 
> 0xbc / 188 | SPWI212 
> 0xbd / 189 | SPWI401
> 0xbf / 191 | SPWI905
> 0xc0 / 192 | SPIN104 
> 0xc1 / 193 | SPWI705 
> 0xc2 / 194 | SPWI409 
> 0xc4 / 196 | SPWI201 
> 0xc9 / 201 | SPWI704 ?
> 0xcf / 207 | SPWI308 
> 
> iwd2 extras:
> 0x42 / 66 | SPIN181
> 0x5d / 93 | SPIN136 fatigue +7
> 0xe8 / 232 | SPWI606
> 0xe9 / 233 | SPWI707
> 0x103 / 259 | SPIN985
> 0x104 / 260 | SPIN985
> 0x109 / 265 | SPPR409
> 0x113 / 275 | SPWI710
> 0x192 / 402 | SPWI713
> 0x19a / 410 | SPIN282
> 0x1a4 / 420 | SPWI606
> 
> iwd1 extras:
> 0x116 / 278 | SPWI511
> 0x4e / 78 | SPPR322

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3463,38 +3463,46 @@ void Actor::UpdateFatigue()
...
-		DisablePortraitIcon(PI_FATIGUE);
+
+	bool updated = false;
+	if (!TicksLastRested) {
+		// just loaded the game; approximate last rest
+		TicksLastRested = game->GameTime - (2*core->Time.hour_size) * (2*GetBase(IE_FATIGUE)+1);
+		updated = true;
+	} else if (LastFatigueCheck) {
+		ieDword FatigueDiff = (game->GameTime - TicksLastRested) / (4*core->Time.hour_size)
+		                    - (LastFatigueCheck - TicksLastRested) / (4*core->Time.hour_size);
```
lynxlynxlynx: How can this ever be non-zero?

Tomsod: Either right after a travel or exactly 4*n hours after a rest.  E.g. 4 hours after resting it evaluates to something like (4:00 / 4:00) - (3:59:59 / 4:00) = 1 - 0 (rounding) = 1.

lynxlynxlynx: Ah, right. :+1:

Tomsod:
Okay, I moved it, and it seems not to crash still.  And if during the map change new ambients are added before old ones are deleted, it shouldn't cause crashes like on quickload, so no regression.

By the way, I still have that list of ~40 noticed-but-unfixed bugs I mentioned before.  Do you need it?

lynxlynxlynx:
Is it in any particular format? Please dump it to a pastebin, so it becomes easier to decide how to proceed.

Tomsod:
There you go: https://pastebin.com/e09302Mn
Thankfully, it's in English, although some of the notes may need to be elaborated.  I planned to rewrite and systematize them on publishing, I think.

lynxlynxlynx:
Thanks, a quick scan:
http://paste.debian.net/1070306/

some were already known and/or noted on the wiki todo, but most of those haven't been moved to this tracker yet. Some are very nice finds. :+1: It'd be great if you reported them one per bug here; most are clear enough as-is.

Tomsod:
Okay, I'll open the issues, soonish.  "Original" means "the original game had this, so it's okay", right?

lynxlynxlynx:
Yep, thanks!

gemrb/GUIScripts/LUProfsSelection.py
lynxlynxlynx: Thanks, I now pushed a fix that takes the fastest rate into account.

Tomsod: Does it work correctly for cleric/thieves?  They're tricky because clerics and thieves have the same prof rate, but different XP level growth.  It seems the original BG2 gives a prof point when the cleric subclass levels, although I'm not sure why -- switching their XP requirements does not change it, so it's not because cleric levels last, at least.

lynxlynxlynx: I didn't try, but from memory I'd guess it uses the cleric one, simply because they come earlier in the table.

lynxlynxlynx:
It's done, thanks again! If you can still elaborate on the short list above, please open a new report with it. I did one just with the "ripe" section, since those are clear: #171.

PR #156, opened 2019-08-04
PluginLoader: fixes compiler errors for GCC and Win

MarcelHB:
Hey,

I just tried to build the most recet `HEAD` on my Windows with GCC v9.1 and encountered the following errors:
```
./gemrb/core/PluginLoader.cpp: In function 'void GemRB::LoadPlugins(char*)':
./gemrb/core/PluginLoader.cpp:89:73: error: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'GemRB::Version_t' {aka 'const char* (*)()'} [-Werror=cast-function-type]
   89 | #define GET_PLUGIN_SYMBOL( handle, name )  GetProcAddress( handle, name )
      |                                                                         ^
./gemrb/core/PluginLoader.cpp:197:40: note: in expansion of macro 'GET_PLUGIN_SYMBOL'
  197 |   Version_t LibVersion = ( Version_t ) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Version" );
      |                                        ^~~~~~~~~~~~~~~~~
./gemrb/core/PluginLoader.cpp:89:73: error: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'GemRB::Description_t' {aka 'const char* (*)()'} [-Werror=cast-function-type]
   89 | #define GET_PLUGIN_SYMBOL( handle, name )  GetProcAddress( handle, name )
      |                                                                         ^
./gemrb/core/PluginLoader.cpp:198:49: note: in expansion of macro 'GET_PLUGIN_SYMBOL'
  198 |   Description_t Description = ( Description_t ) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Description" );
      |                                                 ^~~~~~~~~~~~~~~~~
```

Looks like 979db2def8b2efc7c29916dd04120ef5065f0d5b already attempted to fix two similar issues around that code but wasn't sufficient.

On more recent C++ standards, one seems to be able to use `reinterpret_cast` without objection here, but so far we have to go on with this cast chain.

### DIFF
--- a/gemrb/core/PluginLoader.cpp
+++ b/gemrb/core/PluginLoader.cpp
@@ -194,8 +194,8 @@ void LoadPlugins(char* pluginpath)
 		//printStatus( "OK", LIGHT_GREEN );
 		//using C bindings, so we don't need to jump through extra hoops
 		//with the symbol name
-		Version_t LibVersion = ( Version_t ) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Version" );
-		Description_t Description = ( Description_t ) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Description" );
+		Version_t LibVersion = ( Version_t ) (void*) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Version" );
+		Description_t Description = ( Description_t ) (void*) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_Description" );
 		ID_t ID = ( ID_t ) (void*) GET_PLUGIN_SYMBOL( hMod, "GemRBPlugin_ID" );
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, the travis failure is unrelated, just new clang warnings apparently:

```
874 /home/travis/build/gemrb/gemrb/gemrb/plugins/BIKPlayer/GetBitContext.cpp:33:2: error: 
875 'register' storage class specifier is deprecated and incompatible with
876 C++17 [-Werror,-Wdeprecated-register]
877 register unsigned int tmp;
```

I planned to switch to cxx11 on the subviews branch, since it would help brad, but it just didn't happen yet. Enough time has passed now that we can just do it in master, but it's also the sort of thing that's better done right after a release. I'll open a reminder.

PR #157, opened 2019-08-04
SDL2: enables pasting of text into console

### LINKED ISSUE #154: [enhancement] SDL2: paste into the console
SDL2, but not SDL1, supports clipboards natively - implement pasting into the console on middle-click at cursor location — when the console is open.

https://wiki.libsdl.org/CategoryClipboard

MarcelHB:
Thought I'd pick something from the issue list again.

I did some minmal decoupling in order to let `Interface` be the one that moves the clipboard text to whatever it considers appropriate -- the console only so far.

Tested successfully on Windows and Linux.

Closes #154

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -731,8 +731,29 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
 				break;
 			}
 			else {
-				// we do not want these events to cascade down to SDL_KEYDOWN, so we return here instead of at default .
-				return SDLVideoDriver::ProcessEvent(event);
+				/**
+				 * As being SDL2-only, try to query the clipboard state to
+				 * paste when middle clicking the mouse.
+				 */
+				if (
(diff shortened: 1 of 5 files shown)


gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -731,8 +731,29 @@ int SDL20VideoDriver::ProcessEvent(const SDL_Event & event)
...
+				 */
+				if (
+					   event.button.button == SDL_BUTTON_MIDDLE
+					&& event.type == SDL_MOUSEBUTTONDOWN
+					&& SDL_HasClipboardText()
+				) {
+					const char *pasteValue = SDL_GetClipboardText();
+
+					if (pasteValue != NULL) {
+						String *pasteValueString = StringFromCString(pasteValue);
```
MarcelHB: Oh wait, this does not make me happy with umlaute and stuff alike.

MarcelHB: Well, we should live with this for now: Typing non-ASCII values into the console already produces garbage anyway, so ...

lynxlynxlynx:
LGTM, thanks!

PR #160, opened 2019-08-07
Actor animation: shadow palette

### LINKED ISSUE #123: [bug] Sarevok BG1 animation has red shadow
**Describe the bug**
The problem is known and noticeable in Gorion's death cutscene and even in mods that use Sarevok BG1 animation **0x6404** e.g. Sarevok ([Wiederherstellung](http://www.baldursgate-refugium.de/mods/sarevok-wiederherstellung/)) Restore mod.

<details>

Sarevok
Sarevok RED

</details>

MarcelHB:
Please check if this is what we were talking about in #123, and why I was hestitating to reuse `DrawActorSprite` for this purpose: It always uses a ragdoll palette for drawing - even if we pass different animations to it, so I prefer not to touch that so much and use a different path.

Closes #123

### DIFF
--- a/gemrb/core/CharAnimations.cpp
+++ b/gemrb/core/CharAnimations.cpp
@@ -479,6 +479,10 @@ Palette* CharAnimations::GetPartPalette(int part)
 	return palette[type];
 }
 
+Palette* CharAnimations::GetShadowPalette() const {
+	return shadowPalette;
+}
+
 static inline int compare_avatars(const void *a, const void *b)
 {
 	return (int) (((const AvatarStruct *)a)->AnimID - ((const AvatarStruct *)b)->AnimID);
@@ -653,6 +657,7 @@ CharAnimations::CharAnimations(unsigned int AnimID, ieDword ArmourLevel)
 		modifiedPalette[i] = NULL;
 		palette[i] = NULL;
 	}
+	shadowPalette = NULL;
 	nextStanceID = 0;
 	StanceID = 0;
 	autoSwitchOnEnd = false;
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
I don't like all the code duplication. How about you introduce an optional parameter to DrawActorSprite to toggle between the palettes?

MarcelHB:
Yes, this is what I thought about the "hack factor" in the linked issue. I see these approaches (just sketched, not tested):

1. "Use the anim's palette"-flag: given some flag `useAnimPalette`, we make https://github.com/gemrb/gemrb/blob/e346a86c339ac7ee343f8cb167e59d9b53a8ea2e/gemrb/core/Scriptable/Actor.cpp#L8084
into

```cpp
... useAnimPalette ? anim->GetFrame(0)->GetPalette() : ca->GetPartPalette(partnum) ...
```

2. "It's a shadow"-flag: given some flag `isShadowAnimation`, we do like 1., but additionally for

https://github.com/gemrb/gemrb/blob/e346a86c339ac7ee343f8cb167e59d9b53a8ea2e/gemrb/core/Scriptable/Actor.cpp#L8051

also - as I'm not sure if we really need multipart support? -

```cpp
int PartCount = isShadowAnimation ? 1 : ca->GetTotalPartCount();
```

3. Refactoring against code duplication (as additional control flags tend to make the state space explode), let's say by dedup'ing that code block (and maybe the flags above) into it's own method (and the required palette as parameter):

https://github.com/gemrb/gemrb/blob/e346a86c339ac7ee343f8cb167e59d9b53a8ea2e/gemrb/core/Scriptable/Actor.cpp#L8071-L8084

Here again the question, I'm not sure if shadow drawing needs that loop for multipart, and z-ordering? If not, unless flagging around that with thinks like in 1. or 2., this is probably wasting soom cycles.

lynxlynxlynx:
I don't think the part counter is much of an overhead, so I don't see a problem with a few extra calls to make things cleaner. The assert will be gone in releases binaries too. So I propose approach 1.

MarcelHB:
Okay, I have reworked it like this then.

lynxlynxlynx:
Excellent, thanks!

PR #162, opened 2019-08-10
Talking circles

### LINKED ISSUE #159: [feature] NPCs should have white circles when vocalizing
**Description**
I don't know all the circumstances under which this is supposed to happen, but generally speaking, when an NPC talks into the wild without a prior interaction, its circle should turn white and oscillating for the time of the sound file playing. 

**Additional context**
Here, you see what I mean (BG, candlekeep, singing priests) in the original game:
people_talking

MarcelHB:
This covers many cases when the original games indicate the sound emitting actor by a pulsating white circle, essentially most of the time when there is a verbal constant requested but also existence sounds. I have bound it to the time of the sound file (presumably) playing, without looking for cases of early interruption -- if you know some, I can try to implement them.

The animation was no exact rocket science but it's the numbers that really make it look like in BG1.

Closes #159

### DIFF
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -522,6 +522,11 @@ void DisplayStringCore(Scriptable* const Sender, int Strref, int flags)
 		unsigned int len = 0;
 		core->GetAudioDrv()->Play(Sound, channel, 0, 0, speech, &len);
 		ieDword counter = ( AI_UPDATE_TIME * len ) / 1000;
+
+		if (Sender->Type == ST_ACTOR && len > 0) {
+			reinterpret_cast<Actor*>(Sender)->SetAnimatedTalking(len);
+		}
+
 		if ((counter != 0) && (flags &DS_WAIT) )
 			Sender->SetWait( counter );
 	}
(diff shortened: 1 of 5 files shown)


gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -522,6 +522,11 @@ void DisplayStringCore(Scriptable* const Sender, int Strref, int flags)
 		unsigned int len = 0;
 		core->GetAudioDrv()->Play(Sound, channel, 0, 0, speech, &len);
 		ieDword counter = ( AI_UPDATE_TIME * len ) / 1000;
+
+		if (len > 0) {
+			reinterpret_cast<Actor*>(Sender)->SetAnimatedTalking(len);
```
lynxlynxlynx: There's no guarantee that the Scriptable is of Actor type, so please add a check first: Sender->Type == ST_ACTOR

lynxlynxlynx:
Overall the code looks good, thanks! The original may have oscillated also the color a bit, but neither I'm sure nor should it matter.

gemrb/core/GameScript/GSUtils.cpp
MarcelHB: Oh yeah, I was fooled by that line above which is not formally a barrier, I guess: https://github.com/gemrb/gemrb/blob/23ae06a166caa3e5723fc82fc119955b31260f8b/gemrb/core/GameScript/GSUtils.cpp#L461

MarcelHB: Fixed that.

Issue #155, opened 2019-07-29
[enhancement] Set the window manager icons


lynxlynxlynx:
FLOSS platforms support WM icon(s), stuff visible in the window titlebar, program lists, window switcher lists ... now the default X11 logo is used instead.

SDL1 has [SDL_WM_SetIcon](https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlwmseticon.html), but that has a somewhat limited format.
SDl2 only has [SDL_SetWindowIcon](https://wiki.libsdl.org/SDL_SetWindowIcon), which can do arbitrary sizes also on Windows. So we can ignore SDL1 ... Both cover only part of the use cases, while the rest is handled by the .desktop file for anyone that uses it.

https://github.com/gemrb/gemrb/blob/master/artwork/ has the logos, the "rb" square one being most appropriate. Nowadays it should be enough to install the SVG into /usr/share/icons/hicolor/scalable/apps ([as suggested by XDG icon theme spec](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html)). ~We currently install only the PNG to ICON_DIR. The gemrb.desktop file tells the WM the name to look for, so remember to rename it to gemrb.svg on install.~

For SDL, a fixed size surface is needed, so I suggest loading a 128x128px version.

EDIT: as of #166 windows gets an icon through cmake. Not sure if any other cases needs to be handled for that platform.

EDIT: GIMP can generate the C code, yay. :) https://blog.gibson.sh/2015/04/13/how-to-integrate-your-sdl2-window-icon-or-any-image-into-your-executable/

> I managed to get icons to work in sdl2 and linux by calling SDL_SetWindowIcon right after SDL_CreateWindow, before SDL_CreateRenderer.

lynxlynxlynx:
Actually we already install the svg and it works fine for the prebuilt binaries ... I think, I noticed now that the icon got cached for me and I can't wipe it clean. @fenuks can you please try downloading one of the linux builds and running it, just to see if you get the proper taskbar icon?

https://sourceforge.net/projects/gemrb/files/Buildbot%20Binaries/Linux/

fenuks:
So, icon does render correctly for me using the latest build at the moment (gemrb-2020-06-305b0cf79.tar.bz2).

![obraz](https://user-images.githubusercontent.com/1249829/86147451-f0010880-bae8-11ea-82b3-760eb1078520.png)

lynxlynxlynx:
[...]

PR #166, opened 2019-08-18
Window icon for Windows

MarcelHB:
As requested by #155, I thought this is solvable for Windows without any additional SDL code and catches both an executable and window icon in one move. This is a 256x256 icon file, that is supposed to be compatible to Windows XP and newer (tested on Win 7).
windows_icon


MarcelHB:
I had wanted to check AppVeyor for an artifcat, but it does not produce any.

lynxlynxlynx:
It's not configured for PRs. Getting it to work was nasty and it still looks horrible; seems cmake and windows don't get along well THEN you have all the other packaging dependency problems ...

Thanks, I'll just merge it and you can check that build too.

MarcelHB:
Ok, good to know.

Interestingly, the SDL window does not copy the console's window icon in that build, but then it is using SDL v1.2 whereas I have tested SDL v2.0, so ...

PR #170, opened 2019-08-19
Map: statically adjusting some animation flags for PST

### LINKED ISSUE #163: [bug] Blending issues in PST, AR1202
**Describe the bug**
Enter AR1202 to notice a blending issue of that altar-thing, and some red-ish whatever artifcat that I cannot identify (Ignus' remainings after defeat?).

GemRB version: `HEAD` of right now.

**To Reproduce**
1. Enter AR1202 of PST.
2. Look around

**Expected behavior**
See here:

pst_blending_good

**Screenshots**
See here:
pst_blending_bad

MarcelHB:
That's what we've synthesized in #163, it fixes every major map animation annoyance for PST. There is probably details left, like this one that I cannot explain yet, but you probably don't even notice unless playing side by side:

ar0202_lights

Closes #163

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -3899,7 +3899,7 @@ AreaAnimation::AreaAnimation()
 	palette=NULL;
 	covers=NULL;
 	appearance = sequence = frame = transparency = height = 0;
-	Flags = startFrameRange = skipcycle = startchance = 0;
+	Flags = originalFlags = startFrameRange = skipcycle = startchance = 0;
 	unknown48 = 0;
 	Name[0] = 0;
 	BAM[0] = 0;
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -3912,6 +3912,7 @@ AreaAnimation::AreaAnimation(AreaAnimation *src)
 	sequence = src->sequence;
 	animation = NULL;
 	Flags = src->Flags;
+	originalFlags = src->originalFlags;
 	Pos.x = src->Pos.x;
 	Pos.y = src->Pos.y;
 	appearance = src->appearance;
--- a/gemrb/plugins/AREImporter/AREImporter.cpp
+++ b/gemrb/plugins/AREImporter/AREImporter.cpp
@@ -1190,6 +1190,8 @@ Map* AREImporter::GetMap(const char *ResRef, bool day_or_night)
 		}
 	}
 
+	int pst = core->HasFeature( GF_AUTOMAP_INI );
+
 	core->LoadProgress(90);
 	Log(DEBUG, "AREImporter", "Loading animations");
 	str->Seek( AnimOffset, GEM_STREAM_START );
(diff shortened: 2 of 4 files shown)


MarcelHB:
Ok, I will look at it later.

But mentioning savegames, I notices that for all my savegames that were written by GemRB, PST crashes already. Can you confirm?

lynxlynxlynx:
It's likely, PST is extra strict. Last I checked loading a mortuary save worked, but more complicated saves (states) didn't. It's a similar hit and miss with other games, but I don't look at it often (iediff is super handy for this though). But we don't like to make it worse on purpose. :)

MarcelHB:
Ok, are there scenarios when these flags get modified programmatically? A simple way may be to backup the flag set on load and use it for writing the area file, but if this is the case, there must be something more sophisticated I suppose.

lynxlynxlynx:
We could store the original, but I think it's better to just reverse it before save. Carrying an extra member for something so simple seems like overkill.

MarcelHB:
But it is not revertible using these bit operations, is it? I mean, we cannot reconstruct any obscure bit pattern that way.

MarcelHB:
I moved the whole thing into AREImporter.

gemrb/core/Map.cpp
```diff
@@ -3912,6 +3912,7 @@ AreaAnimation::AreaAnimation(AreaAnimation *src)
 	sequence = src->sequence;
 	animation = NULL;
 	Flags = src->Flags;
+	originalFlags = src->originalFlags;
```
lynxlynxlynx: Please also add it to the default constructor.

lynxlynxlynx:
Thanks!

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -1209,6 +1211,7 @@ Map* AREImporter::GetMap(const char *ResRef, bool day_or_night)
 			str->ReadWord( &anim->sequence );
 			str->ReadWord( &anim->frame );
 			str->ReadDword( &anim->Flags );
+			anim->originalFlags = anim->Flags;
```
lynxlynxlynx: This change broke the purple laser "doors" in the pocket plane. Scripts eventually disable them, but now saving stores the old value, so on load they're back again.

Changing this to only affect pst would help, but would leave the bug in pst. I guess the proper way to fix it at save time is to:
1. Get the changed bits between adjusted originalFlags and current Flags
2. Store originalFlags with those bits applied on top

MarcelHB: Your steps at the save time only work when not dealing with PST - since there will be bit diffs that we caused by `AdjustPSTFlags`. If the other games do not suffer from different bit settings, and presumably PST never changes them ingame, wouldn't it be easier just to switch between saving `originalFlags` (PST) and `Flags` (the rest)

lynxlynxlynx: Unfortunately pst also modifies them, just not with AmbientActivate. But yeah, for non-pst we can just save the current ones.

MarcelHB: In PST, is it the portals in the maze? This was one example of on/off animations that I did not know how to trigger when quickly checking the maps.

Other than that, we need to know the bits that are touched in that case to properly write them back at the PSTish position when saving - for compatibility.

lynxlynxlynx: The gears in ar0502 are one example. It's just A_ANI_ACTIVE, so we can do it like I proposed.

PR #181, opened 2019-08-25
CharAnimations: palette update

MarcelHB:
This fixes:
- PST: Grace, Ebb and some guards having overall bad animation colors.
- BG2: Umber hulks and fire giants having bad attack animation colors.
- IWD: Arundel and some colorfully misplaced pixels.

Closes #168

By the way, the PR template is not taking effect yet!

PR #184, opened 2019-08-27
CharAnimations: fixes missing palette resref on stance change

MarcelHB:
In BG, I had some random encounter with Gnolls or Flinds that make the game fatally exit over `Palette is supposed to be named, but got no name!`. Fixes behaviour introduced by #181.

Should have been that way at first place I guess, but I did not correctly check the cases of the destructor of `CharAnimations`.

## Checklist


lynxlynxlynx:
Thanks, I also missed it, obviously.

PR #188, opened 2019-08-28
Map: prevents crash if map note has encoding issues

### LINKED ISSUE #185: [bug] Crash in PST on loading map notes from savegame on Linux
**Describe the bug**
I just copied PST (GoG version) from my Windows machine to Linux, together with the existing savegames from the native game

GemRB version: Current HEAD

**To Reproduce**
1. On Linux, start PST.
2. Load a savegame created by the original game

**Expected behavior**
The game loads correctly.

**Screenshots**
The game crashes with the given stacktrace:

```
(gdb) bt
#0  __strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:101
#1  0x00007ffff7ef7359 in GemRB::Interface::UpdateString (this=0x55555556aed0, strref=4294967295, text=text@entry=0x0) at ./gemrb/gemrb/core/Interface.cpp:2101
#2  0x00007ffff7f17a58 in GemRB::MapNote::MapNote (color=<optimized out>, text=<optimized out>, this=0x7fffffffca60) at ./gemrb/gemrb/core/Map.h:199
#3  GemRB::MapNote::MapNote (this=0x7fffffffca60, text=<optimized out>, color=<optimized out>) at ./gemrb/gemrb/core/Map.h:193
#4  0x00007ffff7f1598f in GemRB::Map::AddMapNote (this=this@entry=0x555555c75760, point=..., color=<optimized out>, text=text@entry=0x55555727f330) at ./gemrb/gemrb/core/Map.cpp:3173
#5  0x00007ffff53c622e in GemRB::AREImporter::GetMap (this=this@entry=0x555555763960, ResRef=ResRef@entry=0x555555db1000 "ar0500", day_or_night=<optimized out>)
    at ./gemrb/gemrb/plugins/AREImporter/AREImporter.cpp:1357
#6  0x00007ffff7ec87b0 in GemRB::Game::LoadMap (this=0x555555cafed0, ResRef=0x555555db1000 "ar0500", loadscreen=<optimized out>) at ./gemrb/gemrb/core/Game.cpp:970
#7  0x00007ffff7ec89d4 in GemRB::Game::GetMap (this=this@entry=0x555555cafed0, areaname=areaname@entry=0x555555db1000 "ar0500", change=change@entry=true) at ./gemrb/gemrb/core/Game.cpp:757
#8  0x00007ffff7eb260c in GemRB::GameControl::ChangeMap (this=this@entry=0x555555c76d80, pc=0x555555db0dc0, forced=forced@entry=true) at ./gemrb/gemrb/core/GUI/GameControl.cpp:2507
#9  0x00007ffff7f002ed in GemRB::Interface::HandleFlags (this=0x55555556aed0) at ./gemrb/gemrb/core/Interface.cpp:642
#10 0x00007ffff7f00658 in GemRB::Interface::Main (this=0x55555556aed0) at ./gemrb/gemrb/core/Interface.cpp:993
#11 0x00005555555551e2 in main (argc=<optimized out>, argv=0x7fffffffdf48) at ./gemrb/gemrb/GemRB.cpp:99
(gdb) f 1
#1  0x00007ffff7ef7359 in GemRB::Interface::UpdateString (this=0x55555556aed0, strref=4294967295, text=text@entry=0x0) at ./gemrb/gemrb/core/Interface.cpp:2101
2101		bool changed = strcmp(text, current) != 0;
(gdb) p text
$1 = 0x0
(gdb) p current
$2 = 0x55555727f9a0 ""
(gdb) 
```
It does not affect all savegames, i.e. not the ones created from a new game by GemRB, some others work, some don't.

MarcelHB:
I'm not fully sure what you meant by "fallback" exactly - I inserted a question mark for that case - since I did not really see where this bubbles through at the end of PST since it's actually loaded into the game but just not updated in the string table.

Closes #185 

## Checklist

### DIFF
--- a/gemrb/core/Map.h
+++ b/gemrb/core/Map.h
@@ -196,8 +196,13 @@ public:
 		if (text) {
 			//update custom strref
 			char* mbstring = MBCStringFromString(*text);
-			strref = core->UpdateString( strref, mbstring);
-			free(mbstring);
+			if (mbstring) {
+				strref = core->UpdateString( strref, mbstring);
+				free(mbstring);
+			} else {
+				strref = core->UpdateString(strref, "?");
+				Log(WARNING, "Map", "Failed to update string from map note, possibly an enconding issue.");
+			}
 		}
 	}
 	MapNote(const ieStrRef ref, ieWord color)


lynxlynxlynx:
LGTM!

gemrb/core/Map.h
```diff
@@ -196,8 +196,13 @@ class MapNote {
...
 			//update custom strref
 			char* mbstring = MBCStringFromString(*text);
-			strref = core->UpdateString( strref, mbstring);
-			free(mbstring);
+			if (mbstring) {
+				strref = core->UpdateString( strref, mbstring);
+				free(mbstring);
+			} else {
+				strref = core->UpdateString(strref, "?");
+				Log(WARNING, "Map", "Failed to update string from map note, possibly an enconding issue.");
```
Mingun: I think it will be good log also `strref`, because this log only indicates problem _somewhere_, but is useless when you want fix this problem

lynxlynxlynx: This can only happen with custom map notes, so it will all be in the TLK override. With a brand new strref, so I'm not sure it'd be that useful.

MarcelHB: It wasn't even a custom map note, but one from the string table containing a `ß` (ISO 8859-1 `0xDF`). At some point in `String.cpp`, it is converted into a wide char string by this https://github.com/gemrb/gemrb/blob/89b2bfa1cf8ec8dd3b543bfbb9c79fbd6c96c60b/gemrb/core/System/String.cpp#L87 Trying to get this through `wcstombs` fails on my Linux.

lynxlynxlynx: For "why", I think you nailed it. Unless the original disabled it, we shouldn't patch over it.

As for the underlying issue, it still reminds me of the Czech and Polish versions, which also used a different encoding (not sure if even standard), which we then remap. See here: https://github.com/gemrb/gemrb/blob/master/gemrb/docs/en/default_ini.txt

PR #241, opened 2019-09-17
fix reversed inventory item pickup sounds

CrevsDaak:
Fixes inventory item pickup sounds being reversed in BG1&2. PsT and IwD work fine so I didn't touch them. I just switched the columns with a dirty sed script, but everything works fine (tested in both games).


lynxlynxlynx:
Nice, thanks! It's indeed correct to rename all values; renaming columns wouldn't work in this case, since here the lookup is index based.

PR #255, opened 2019-09-29
Add _DJ* Werebear animations to avatars.2da

caillean7:
As requested [here](https://www.gibberlings3.net/forums/topic/30655-werebear-animations-on-gemrb/?tab=comments#comment-272919)


lynxlynxlynx:
Thanks!

Issue #263, opened 2019-10-01
[bug] Flying creatures are interactable


caillean7:
**Description**
Split from #258:

Birds and bats should not be valid targets for dialogue. On GemRB they are, they get a tooltip and the talk cursor drawn, which also shouldn't happen. Plus, they should not be affected by AoE spells.

**GemRB version**
0.85-git

**To Reproduce**
- go to areas like AR2500 (Underdark exit) or AR0300 (Athkatlan docks)
- find yourself a seagull, bird or bat and try talking to it
- notice the talk cursor
- your character should walk up to the creature to initiate dialogue, resulting in something like "Bat / Bird - has nothing to say to you"

**Expected behavior**
Make these creatures non-interactable like on the original

- [x] make bats and birds not valid for dialogue: Actor::ValidTarget(GA_TALK) returning false
- [ ] omit them from drawing the cursor
- [ ] don't show tooltips ~~(?)~~ yep, no tooltips in IE

- [x]  make them unaffected by AoE spells

lynxlynxlynx:
For the cursor you can look at GameControl::OnMouseOver near the first mention of IE_CURSOR_TALK.

If that's not enough also for the tooltip, check GameControl::DisplayTooltip.

caillean7:
Thanks for the support :) I'll look into the cursor thing tonight, but at first glance this one looks a little more complicated.

caillean7:
Another matter that Marcel mentioned in #258: those creatures should not be affected by AoE spells, but they currently are on GemRB.

caillean7:
Oi, the no cursor change works, and it also takes care of the tooltip, take a look:
https://pastebin.com/4GxvFYw1

lynxlynxlynx:
Please use the gdup function, so it's clearer what you changed.

caillean7:
https://github.com/caillean7/gemrb/commit/66ba039eaca40221e881dc6e9d60e5f302ad644d

lynxlynxlynx:
They're completely non-interactive, so one check is enough. You can move it before the "if (lastActor) {" block. Also, please use GetStat instead of directly resolving Modified.

caillean7:
`if (lastActor->GetStat(IE_DONOTJUMP) == DNJ_BIRD)`

With this I'm getting segfault crashes, I don't know anymore, sorry :(

lynxlynxlynx:
[...]

PR #265, opened 2019-10-01
No cursor change for birds and bats

caillean7:
Fixes the two remaining points from #263 (talk cursor and tooltips).


lynxlynxlynx:
Thanks!

PR #316, opened 2019-10-08
Subviews: fixing guideath position and display

### LINKED ISSUE #285: [subviews] Place death message dialogs at the bottom.
In all of the games, when the main/last PC died, the message asking for reloading is always centered, not at the bottom.

More like an aesthetic bug:

death_screen

MarcelHB:
Now they are at the bottom, and BG has the action bars visually disabled, too. 

Closes #285


bradallred:
thanks!

Issue #332, opened 2019-10-11
[bug] BG2: Immediate invisibility when hiding in shadows


caillean7:
Split from #171:

**Description**
(Original-except it shouldn't happen on failure) hiding in shadows makes you invisible immediately (before the per-round roll)

Hiding is part of so called modal actions. You can look at the modal struct in Actor.cpp to see how they work in general. They're triggered through GUIScript.cpp, as the buttons get pressed, and Actor::TryToHide has the main rolling for deciding success. I'd start there to see why failure is not handled properly.

ifyGecko:
I should have a copy of iwd1&2 any day now, I can work on this when they arrive.

lynxlynxlynx:
Cool, thanks. Comment here if you get into trouble. IWD1 will be the better choice for debugging this, since the nomenclature will be more familiar.

ifyGecko:
This issue is due to light and is not consistent, sometimes the variable of "lightness" in Actor::TryToHide is assigned a value of 0 right outside "Winter's Cradle Tavern" in iwd. When the issue does not present itself "lightness" gets a value of 100.

Variables when the bug arises:
lightness: 0
light_diff: 37675114
Chance: 39182171

I have a hackish work around by setting "lightness" to 100 if it is 0 before calculating "light_diff" but this is not satisfactory. I will continue by looking into why GetLightLevel sometimes returns 0 instead of 100.

lynxlynxlynx:
Sounds like a different issue, though it makes sense to ensure light_diff doesn't underflow like that.

Returning 0 is valid, that's from data directly. It looks like ar1000lm.bmp (the lightmap) has high enough values though, so just step through GetLightLevel to see which part of the calculation/data is suspect.

ifyGecko:
In the if(tint) block c.r-tint->r, c.g-tint->g, and c.b-tint->b each are returning 0 so the calculation for lightness returns 0. I have poked around a small bit to try and see how a lot of the code involved is working but I am lost as to where to look for r-tint, g-tint, and b-tint.

ifyGecko:
The tint returned in GetLightLevel is FogTint, rgb=ffffff, and the color returned from GetPixel is also ffffff so the c.r - tint->r calculation and so on will be zero which is why GetLightLevel returns zero.

I am not sure as to why things are calculated the way they are regarding lightness when a tint is applied, could someone give some clarity on the if(tint) block in GetLightLevel?

MarcelHB:
[...]

PR #355, opened 2019-10-28
Temporary fix for issue #332

ifyGecko:
## Description
Removed Fog from GetGlobalTint and GetWeather to prevent miscalculated lightness in GetLightLevel. FogTint variable was also removed due to compilation error "un-used variable" with FreeBSD clang version 6.0.1.

## Checklist

### DIFF
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -2105,7 +2105,7 @@ void Game::Infravision()
 static const Color DreamTint={0xf0,0xe0,0xd0,0x10};    //light brown scale
 static const Color NightTint={0x80,0x80,0xe0,0x40};    //dark, bluish
 static const Color DuskTint={0xe0,0x80,0x80,0x40};     //dark, reddish
-static const Color FogTint={0xff,0xff,0xff,0x40};      //whitish
+//TODO: add FogTint when a solution that does not hinder lightness calculation is implemented
 static const Color DarkTint={0x80,0x80,0xe0,0x10};     //slightly dark bluish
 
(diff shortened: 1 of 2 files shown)


gemrb/core/Game.cpp
```diff
@@ -2130,9 +2130,7 @@ const Color *Game::GetGlobalTint() const
 		if (WeatherBits&WB_RAIN) {
 			return &DarkTint;
 		}
-		if (WeatherBits&WB_FOG) {
-			return &FogTint;
-		}
+		//TODO: add FogTint when it is fixed to not hinder lightness calculation
```
lynxlynxlynx: One todo is enough if they're all equal. You can remove this one and the one before it. As we found out, tinting can't be used to get a fog effect, at least not the way we do it now. But it would probably look wonky either way: http://i.imgur.com/yBxaLO5.jpg

Something with light dithering would perhaps work, but that's for another time.

gemrb/core/Map.cpp
```diff
@@ -3759,9 +3759,7 @@ int Map::GetWeather()
 	if (Snow>=core->Roll(1,100,0) ) {
 		return WB_SNOW;
 	}
-	if (Fog>=core->Roll(1,100,0) ) {
-		return WB_FOG;
-	}
+	//TODO: add FOG when FogTint is fixed to not hinder lightness calculation
```
lynxlynxlynx:
```suggestion
	// TODO: handle WB_FOG the same way when we start drawing it
```

lynxlynxlynx:
Thanks!

Issue #324, opened 2019-10-11
[bug] BG2:  Known spells reverse their order on save-load


caillean7:
Split from #171

**Description**

> known spells reverse their order on save-load

I'm pretty sure I can remember that from a couple of years ago, but can't reproduce anymore. Might be already taken care of?

lynxlynxlynx:
Trivial to reproduce with a mage. Remember, it's the known spells, not the memorized ones.

ifyGecko:
The known spell list reverses on save. Continually re-loading the same save does not reverse the order.

lynxlynxlynx:
Yeah, most likely just the iteration when saving has to be reversed.

Tomsod:
Continuously re-loading can't give different results either way, because you're repeating the same deterministic operation.  It would probably be more prudent to test (1) saving in IE and loading in GemRB, and (2) saving in GemRB and loading in IE.  If (1) reverses spells, fix loading; if (2) does, fix saving.  That way the fix will be compatible with IE.

caillean7:
I just tried real quick:

> (1) saving in IE and loading in GemRB

Load order is fine in GemRB (ascending alphabetically)

> (2) saving in GemRB and loading in IE

A save from GemRB reloaded in GemRB has the order reversed, while loading the same in IE has the right load order.

Tomsod:
Tried the same, and I think I get it.  Indeed, loading a save in (original) BG2 sorts the spells in alphabetical order!  I never noticed that before.  It respects localization, as well (my spells were sorted according to their Russian names), so the order must be computed on the fly.

Meanwhile, in GemRB learning a new spell adds it to the end of the list, and the list is also flipped on each save/load.  If you're testing with an IE save, you won't notice, but I played through BG1 in GemRB and my spellbook is a mess.

Also, saving the game in IE and then loading it in GemRB preserves the alphabetical ordering.  That means it's not merely a display thing and IE actually overwrites the spell order in the save.  Also, it probably means that GemRB reverses spells on save, not load (otherwise it would load spells in Z-A order).  But I think we should just drop the whole reversal issue and simulate IE behavior instead.

caillean7:
It certainly looks as if IE does the sorting on both saving and loading.

lynxlynxlynx:
[...]

PR #357, opened 2019-10-28
Fixed issue #324

ifyGecko:
## Description
The spells must be saved in reverse order to load them in the same order as seen in the spell-book. Inverted the loop from a 0 start index to a (count-1) start index. The use of unsigned integers can not be used for a (count-1) starting index. Since count could equal 0 this can causes the sign bit from the subtraction to start the loop at a VERY high number resulting in a crash from an out of bounds index reference.

If you want to keep  the prior use of unsigned integers an "if(!count) continue;" statement will need to be placed before the start of the inner most loop to prevent the loop on that iteration to execute when count==0.

## Checklist

### DIFF
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -2105,7 +2105,6 @@ void Game::Infravision()
 static const Color DreamTint={0xf0,0xe0,0xd0,0x10};    //light brown scale
 static const Color NightTint={0x80,0x80,0xe0,0x40};    //dark, bluish
 static const Color DuskTint={0xe0,0x80,0x80,0x40};     //dark, reddish
-static const Color FogTint={0xff,0xff,0xff,0x40};      //whitish
 static const Color DarkTint={0x80,0x80,0xe0,0x10};     //slightly dark bluish
 
 const Color *Game::GetGlobalTint() const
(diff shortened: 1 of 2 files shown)


ifyGecko:
I noticed the commits from issue #332 also showed up in this pull request. I am new to using git from a terminal and github outside of online code storage. Does anyone know how I can prevent this from happening in the future?

lynxlynxlynx:
Please work in branches or force reset after master changes, otherwise you bring in all those irrelevant commits and we'll have to continue squash merging.

Both GetSpellLevelCount and GetKnownSpellsCount return unsigned, so this change needlessly changes the types. Is it not enough for k to be signed? You can cast count in the inline assignment.

MarcelHB:
@ifyGecko Look at this flow here: https://gist.github.com/Chaser324/ce0505fbed06b947d962

ifyGecko:
@lynxlynxlynx Since the type change wouldn't impact calculation I was trying to make it uniform for readability, in hindsight readability is very subjective. I will avoid that in the future and only change what is necessary. How do you want the inner most loop initialized, int k = int(count)-1?

@MarcelHB THANK YOU! I have been needing that. I was under the impression typical flow was suppose to be fork, branch, commit, push, merge to fork/master then pull.

lynxlynxlynx:
`git reset --hard somecommithash` would **delete** all history from now up to the specified commit. That commit would be the new HEAD of the branch you're on. So if you do it to some commit before your changes, you can easily updated back to current upstream.

If you have to touch up on existing work, that's not so useful and branches save the day.

PR #358, opened 2019-10-29
fixed issue #324

ifyGecko:
## Description
Saving reverses the order of known spells in the PC's spell-book, saving the spells in reverse order preserves the original order on next load. I Inverted the loop to start indexing from (count-1) to fix the issue.

## Checklist

### DIFF
--- a/gemrb/plugins/CREImporter/CREImporter.cpp
+++ b/gemrb/plugins/CREImporter/CREImporter.cpp
@@ -2952,7 +2952,7 @@ int CREImporter::PutKnownSpells( DataStream *stream, Actor *actor)
 		unsigned int level = actor->spellbook.GetSpellLevelCount(i);
 		for (unsigned int j=0;j<level;j++) {
 			unsigned int count = actor->spellbook.GetKnownSpellsCount(i, j);
-			for (unsigned int k=0;k<count;k++) {
+			for (int k=count-1;0<=k;k--) {
 				CREKnownSpell *ck = actor->spellbook.GetKnownSpell(i, j, k);
 				assert(ck);
 				stream->WriteResRef(ck->SpellResRef);


gemrb/plugins/CREImporter/CREImporter.cpp
```diff
@@ -2952,7 +2952,7 @@ int CREImporter::PutKnownSpells( DataStream *stream, Actor *actor)
 		unsigned int level = actor->spellbook.GetSpellLevelCount(i);
 		for (unsigned int j=0;j<level;j++) {
 			unsigned int count = actor->spellbook.GetKnownSpellsCount(i, j);
-			for (unsigned int k=0;k<count;k++) {
```
lynxlynxlynx: Are you sure this is ok? Like you warned before it looks like it could underflow for a count of 0.
If I remember my c++ type promotion rules correctly, mixing signed and unsigned will promote all operands to unsigned, so rhs would be UINT_MAX which would then get mangled again into a signed int. Ugly even if it works (did you try?).

Please also use the same order for all expressions, so it matches the convention: k>=0.

ifyGecko: Your memory would be correct. It could if you wanted to keep everything uniform with unsigned int as it originally would. I set k to int so regardless of what count is promoted to the calculation will be the same since GetKnownSpellCount can not return an unsigned int so large that it's MSB can be interpreted as a sign bit during promotion.

ifyGecko: Just for clarity I have a tested this on one of my game saves and it works.

From a language theory perspective you are correct, I think at least at first glance, but all in all it boils down to bits on the hardware. If count == 0, it will be the same bit representation in signed and unsigned form. A 0-1 situation will assign k to -1 since it is of type int. If we kept it as unsigned int, as i warned, then it would be represented as a huge number and crash the program.

lynxlynxlynx:
Thanks!

Issue #356, opened 2019-10-28
[bug] Can't create a Paladin in IWD


MarcelHB:
**Describe the bug**
When creating a PC as a paladin in IWD, you cannot finish the wizard.

GemRB version: HEAD.

**To Reproduce**
1. Create a human paladin.
2. Go through all the steps
3. Try to finish after giving a name

**Expected behavior**
PC to be accepted

**Logs**
```python
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "./GUIScripts/iwd\CharGen.py", line 364, in AcceptPress
[Python/ERROR]:     GUICommon.ResolveClassAbilities (MyChar, ClassName)
[Python/ERROR]:   File "./GUIScripts\GUICommon.py", line 189, in ResolveClassAbilities
[Python/ERROR]:     AddClassAbilities (pc, ABTable, Levels[0], Levels[0])
[Python/ERROR]:   File "./GUIScripts\GUICommon.py", line 228, in AddClassAbilities
[Python/ERROR]:     Spellbook.LearnSpell (pc, ab[3:], IE_SPELL_TYPE_INNATE, 0, 1, LS_MEMO)
[Python/ERROR]:   File "./GUIScripts\Spellbook.py", line 683, in LearnSpell
[Python/ERROR]:     raise RuntimeError, "Failed learning spell: %s !" %(spellref)
[Python/ERROR]: RuntimeError: Failed learning spell: SPIN152 !
```

MarcelHB:
It's probably a typo in one of the 2DA files (clabpala). I cannot find SPIN152 but SPIN125.

lynxlynxlynx:
Yeah, though I'm surprise they get that.

lynxlynxlynx:
Ah, 125 is only named Neutralize poison, while it is in fact their disease resolver. And 152 again only doesn't exist in non-how iwd (smite evil). What does your description say they should get besides remove disease, lay on hands, detect evil and protection from evil?

MarcelHB:
When I'm back at the computer with an optical drive, I can start the original with a Paladin a look for the proper spells.

Other than that I'd say, for non-HoW it should be the same as BG1, and that's less than what the 2DA currently says.

caillean7:
Starting spells for the paladin should be Lay On Hands (SPIN109), Protection from Evil (SPIN121), Cure Disease (SPIN125), Smite Evil (SPIN152) 5x Detect Evil (SPIN120).

[...]

PR #360, opened 2019-10-29
Cleanup of invalid spells in IWD clab files

MarcelHB:
See #356.

## Description
Non-HoW IWD only knows innate spells until SPIN125 so these values have nothing to do here. I have manually tested if upleveling between level 10 and 14 gives alternate innate spells - but there is nothing.


lynxlynxlynx:
Thanks!

Issue #267, opened 2019-10-02
[subviews] Action bar not disabled while in inventory or other screens


bradallred:
It doesn't break anything (that I have noticed), but since GC is disabled you cant do anything, and the bar buttons will be for the PC that is actually selected in the game as opposed to the PC you have selected for inventory or other purposes.

We should probably hide this window while in this faux selection mode.

PR #363, opened 2019-10-31
subviews: Really disabling action bar buttons

### LINKED ISSUE #283: [subviews] Main action controls look and behave weird when disabled
Also in many games (BG, BG2, IWD), some PCs don't have the features enabled yet that render some action bar items disabled. This has some issues:

1. The color of disabled items is not as in the original - Ok, it's also like that in master, you can ignore that.
2. You can click anyway, sometimes you end up in an empty bar (item features), or the items turn active for no good reason.

BG par exemple:

bg_action_controls

MarcelHB:
This is more like patchwork: I guess somewhere something is changing control flags of the button while it is already its own disabled state but this does not matter for `Control::IsDisabled` when starting the action, so let's reinforce that even if the state's already there.

closes #283

### DIFF
--- a/gemrb/core/GUI/Button.cpp
+++ b/gemrb/core/GUI/Button.cpp
@@ -317,12 +317,12 @@ void Button::SetState(unsigned char state)
 		return;
 	}
 
+	// FIXME: we should properly consolidate IE_GUI_BUTTON_DISABLED with the view Disabled flag
+	SetDisabled(state == IE_GUI_BUTTON_DISABLED);
+
 	if (State == state) {
 		return; // avoid redraw
 	}
-
-	// FIXME: we should properly consolidate IE_GUI_BUTTON_DISABLED with the view Disabled flag
-	SetDisabled(state == IE_GUI_BUTTON_DISABLED);
 
 	if (State != state) {
 		MarkDirty();


bradallred:
I'll have to poke into the history to see if there is a reason things are in the order they are in.

However, I think just disabling the window would be a better solution anyway. Disabled windows are drawn differently and disable all controls in the window, not just buttons (I know we don't have any other controls currently).

I haven't looked in the python code to verify, but this makes me think we are doing something silly like iterating all the buttons and disabling them one by one. seems like we could delete that code and replace it with a single line to disable the window instead. Is this accurate?

bradallred:
oh, I guess the comment `return; // avoid redraw` tells me the order is likely just a minor efficiency thing... still would be nice to know if we can make things even better by removing some python code somewhere :)

lynxlynxlynx:
What do you mean by disabling the window? The actionbar is only on a window that always stays interactive when present, iow there's no use case where disabling it whole and still drawing it would be sensible.

bradallred:
> What do you mean by disabling the window? The actionbar is only on a window that always stays interactive when present, iow there's no use case where disabling it whole and still drawing it would be sensible.

then hide it instead of disabling it. doesn't matter to me.

bradallred:
I think IWD2 combines the action window with some other, so maybe hiding or disabling the entire window isn't an option anyway.

lynxlynxlynx:
I'm still confused, since this issue is about the individual buttons. Dead pcs aren't selectable and we already hide everything if you happen to have a single pc selected die on you.

bradallred:
sorry, I was dragging #267 into this too

bradallred:
Sorry, I've been away. This looks good and I checked history and the side effects for `SetDisabled` and see no issues with this.

PR #380, opened 2019-11-11
cmake: Haiku install paths

korli:
## Description

add Haiku install layout for cmake.

## Checklist

- [x] I checked and it doesn't seem necessary to extend the documentation


lynxlynxlynx:
Thanks, but is this really needed? You only change 3 paths from the FHS layout that could be specified when invoking cmake:
https://github.com/haikuports/haikuports/blob/7e5f711456dd0d016b0e518ee4d63b23a86369eb/games-engines/gemrb/gemrb-0.8.5.recipe#L71
`cmake -DLAYOUT=fhs -DDATA_DIR=${CMAKE_INSTALL_PREFIX}/data/gemrb` and so on. The defaulting in the first chunk would be changed to fhs and then there's one less thing to specify.

korli:
@extrowerk what do you think?

CMakeLists.txt
```diff
@@ -111,6 +113,24 @@ elseif (${LAYOUT} MATCHES "fhs")
...
+		if ( ${CMAKE_INSTALL_PREFIX} STREQUAL "/usr" )
+			SET( SYSCONF_DIR /etc/gemrb )
+		else ()
+			SET( SYSCONF_DIR ${CMAKE_INSTALL_PREFIX}/settings/gemrb )
+		endif ()
+	ENDIF( NOT SYSCONF_DIR )
+	SET_PATH( DOC_DIR ${CMAKE_INSTALL_PREFIX}/documentation/gemrb )
+	SET_PATH( ICON_DIR ${CMAKE_INSTALL_PREFIX}/share/pixmaps )
+	SET_PATH( SVG_DIR ${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps )
+	SET_PATH( MENU_DIR ${CMAKE_INSTALL_PREFIX}/share/applications )
```
extrowerk: Menu entries, svg icons and maybe the pixmaps have no use for us.

lynxlynxlynx:
I don't think /dev/null would work, but you can try it, of course. What would for sure is just ifdefing the relevant install commands with `if (NOT HAIKU)`. They're further down the file, just search for the paths you want to exclude.

korli:
@lynxlynxlynx any chance to use GNUInstallDirs instead?

lynxlynxlynx:
What would be the benefit?

korli:
For example, it will define CMAKE_INSTALL_BINDIR based on command line arguments, and default to ${CMAKE_INSTALL_PREFIX}/bin. These CMAKE_INSTALL_* are then standard names.

lynxlynxlynx:
Sure, but you can already do that, so what's the benefit *over the current situation*?

korli:
Our tool provides already a variable $cmakeDirArgs that provides what GNUInstallDirs requires. Now for gemrb we have to add specifically DATADIR, MANDIR, etc.. (and before a layout).
https://github.com/haikuports/haikuports/commit/db6c225250d239135f232fbb07b66a9b806d3571#diff-5531098333ba98b87c4ea9bb7e44f5b7R74
I suppose this isn't really an issue specific to Haiku, therefore I asked "any chance to use GNUInstallDirs instead?"

lynxlynxlynx:
It would bloat the file quite a bit, since we'd have to leave in the old stuff for compatibility. But I'll add the haiku exception at the end, so it doesn't install the stuff you don't need and therefore don't need to maintain a patch.

lynxlynxlynx:
Your patch is in; it required only a small change to apply. Thanks!

If anything else crops up, do let us know.

PR #382, opened 2019-11-11
AREImporter: saving anim flags in a better way

MarcelHB:
See #163, #170.

Please check if we had no misunderstanding here.


lynxlynxlynx:
I was thinking of a more general solution, but this is fine as well. Thanks!

PR #421, opened 2019-11-24
subviews: fix for properly clearing render targets in some cases

### LINKED ISSUE #410: [subviews] Cursor sprite collecting garbage
This is a problem I just discovered on Linux - I'm pretty sure it wasn't there some time ago. Anyway, it's specific to subviews, my Linux computer and SDL2.

After moving the cursor around a bit, it starts to look like this.

cursor

I try to bisect it later.

MarcelHB:
I removed your other change since there is no reason IMO to keep that switch - and it did not help here.

closes #410

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -128,7 +128,7 @@ int SDL20VideoDriver::UpdateRenderTarget(const Color* color)
 
 	if (screenClip.Dimensions() == screenSize)
 	{
-		// SDL does _not_ seem to like having a clip rect of the entire renderer size
+		// Some SDL backends complain on having a clip rect of the entire renderer size
 		// I'm not sure if it is an SDL bug; possibly its just 0 based so it is out of bounds?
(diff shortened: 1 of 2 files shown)


gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -126,14 +126,7 @@ int SDL20VideoDriver::UpdateRenderTarget(const Color* color)
 		return ret;
 	}
 
-	if (screenClip.Dimensions() == screenSize)
```
bradallred: we actually need this to silence complaints from the Metal backend. It too might just be an SDL bug for all I know, but debug builds of the renderer will assert without this under the current SDL 2.0.10 at least.

MarcelHB: Argh, ok

MarcelHB: I rephrase the comment in a way non-metal readers understand. :wink:

MarcelHB:
Restored your fix.

bradallred:
thanks!

PR #527, opened 2019-12-16
Feet2Pixels Correction

ifyGecko:
## Description
Adjusted calculation so the angle used starts with 0 degrees being on the positive x-axis instead of positive y-axis.

## Checklist


lynxlynxlynx:
Thanks!

Issue #161, opened 2019-08-08
define the C++ coding style


lynxlynxlynx:
Previous discussion here: #158 

Basically Tom's [original proposal](http://titanpad.com/LHPyb9r96s) has been lost and we have to start from scratch (or someone could ask him, if he has a copy, of course). The goal is to reduce ambiguity, eventually inconsistency and enable the use of patch formatters to integrate their checking with CI (plus potentially local git hooks).

http://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting

Check if there is a style-deducer for some format, something that would read the code and assert a style format. Otherwise it will probably have to be trial and error until a format with minimum style violations is found.

QGIS has an overly complicated script that could be trimmed down and reused (runs astyle as a pre-commit hook)
https://github.com/qgis/QGIS/blob/master/scripts/prepare-commit.sh

gemrb/docs/en/CodingStyle.txt has just our comprehensive "#include" ordering rules. The core style ideas are noted in CONTRIBUTING.

notes: uppercase the literal suffix 0.0f -> 0.0F (avoids confusion with l, 1 and L)

another option: https://github.com/sk-/git-lint

OBS example we could use at least for PRs: https://github.com/obsproject/obs-studio/blob/master/.github/workflows/clang-format.yml

lynxlynxlynx:
Same goes for python; now that we're part of github actions beta, this pep8 checker needs to be evaluated: https://github.com/marketplace/pep-8-speaks

Who knows how our current style differs and if the checker doesn't choke on our c++ modules (eg. snake-food does).

lynxlynxlynx:
Github actions also has cpp-lint, clang-tidy and pyton-lint support.

lynxlynxlynx:
As an interim partial solution we now enforce a few of our whitespace requirements for python in pull requests:
https://github.com/gemrb/gemrb/blob/master/.github/workflows/style.yml

Ideally we'd do all the checking in a pre-receive hook, so pushes would verbosely fail if non-compliant. And no setup would be required on the user side, since that runs on the server. Unfortunately github does not support this hook, so only local ones like pre-commit are game if we want to maintain clean history. That means extra deps for developers, but we can make them suggested only and also rely on pull request checkers for non-team changes.

- write the hook (succeeds if the checkers are not present)
- install the pre-commit hook via a cmake target included in the main build target (once/idempotent)
- write the docs, warmly suggest installing the chosen checkers

kmfrick:
[...]

PR #528, opened 2019-12-16
replaced VOODOO_SHOUT_RANGE constant with Feet2Pixels function call

ifyGecko:
## Description
Replaced VOODOO_SHOUT_RANGE with a call to Feet2Pixels. Also adjusted Map::Shout to accept a flag signifying if the call was from GameScript::GlobalShout or not to still allow for the unlimited shout distance of GameScript::GlobalShout.

## Checklist

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -1398,7 +1398,6 @@ void Map::Shout(Actor* actor, int shoutID, bool global)
 		if (!global) {
 			/* Audible range was confirmed to be 3x visual range in EEs, accounting for isometric scaling 
 			and more than 1x visual range in others in EE, the '48' (3*16) default can be set by the 'Audible Range'
-			game option in baldur.lua
 
 			This is a bit tricky, it has been show to not be very consistent. The game used a double value of visual 
 			range in several places, so we will use '3 * visual_range / 2' */
@@ -4182,4 +4181,3 @@ void Map::SetupReverbInfo() {
 }
 
 }
-
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -1397,7 +1397,8 @@ void Map::Shout(Actor* actor, int shoutID, bool global)
 
 		if (!global) {
 			/* Audible range was confirmed to be 3x visual range in EEs, accounting for isometric scaling 
-			and more than 1x visual range in others in EE, the '48' (3*16) default can be set by the 'Audible Range'
+			and more than 1x visual range in others;
+			in EE, the '48' (3*16) default can be set by the 'Audible Range' option in baldur.lua
 
 			This is a bit tricky, it has been show to not be very consistent. The game used a double value of visual 
 			range in several places, so we will use '3 * visual_range / 2' */
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -1406,7 +1406,7 @@ void Map::Shout(Actor* actor, int shoutID, bool global)
 			const Point& B = listener->Pos;
 			double angle = atan2(B.y - A.y, B.x - A.x);
 			int distance = (3 * actor->GetStat(IE_VISUALRANGE)) / 2;
-			if (Distance(A, B)>Feet2Pixels(distance, angle)) {
+			if (Distance(A, B) > Feet2Pixels(distance, angle)) {
 				continue;
 			}
 		}
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -1396,10 +1396,17 @@ void Map::Shout(Actor* actor, int shoutID, bool global)
 		}
 
 		if (!global) {
-		        const Point& A = actor->Pos;
+			/* Audible range was confirmed to be 3x visual range in EEs, accounting for isometric scaling 
+			and more than 1x visual range in others in EE, the '48' (3*16) default can be set by the 'Audible Range'
+			game option in baldur.lua
+
+			This is a bit tricky, it has been show to not be very consistent. The game used a double value of visual 
+			range in several places, so we will use '3 * visual_range / 2' */
+			const Point& A = actor->Pos;
 			const Point& B = listener->Pos;
 			double angle = atan2(B.y - A.y, B.x - A.x);
-			if (Distance(A, B)>Feet2Pixels(30, angle)) {
+			int distance = (3 * actor->GetStat(IE_VISUALRANGE)) / 2;
+			if (Distance(A, B)>Feet2Pixels(distance, angle)) {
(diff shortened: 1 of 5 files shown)


gemrb/core/Map.cpp
```diff
@@ -1395,8 +1395,11 @@ void Map::Shout(Actor* actor, int shoutID, unsigned int radius)
 			continue;
 		}
 
-		if (radius) {
-			if (Distance(actor->Pos, listener->Pos)>radius) {
+		if (!global) {
+		        const Point& A = actor->Pos;
```
lynxlynxlynx: Mixed indentation.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4339,7 +4339,8 @@ void Actor::PlayExistenceSounds()
 	int xpos, ypos;
 	audio->GetListenerPos(xpos, ypos);
 	Point listener(xpos, ypos);
-	if (nextComment && !Immobile() && Distance(Pos, listener) <= VOODOO_SHOUT_RANGE) {
+	double angle = atan2(listener.y - Pos.y, listener.x - Pos.x);
+	  if (nextComment && !Immobile() && Distance(Pos, listener) <= Feet2Pixels(30, angle)) {
```
lynxlynxlynx: Same distance problem and misindentation.

gemrb/core/Map.cpp
ifyGecko: line number?

gemrb/includes/voodooconst.h
```diff
@@ -60,13 +60,6 @@ static const int VOODOO_FINDTRAP_RANGE = 10;
 // The distance of operating a trigger, container, dialog buffer etc.
 static unsigned int MAX_OPERATING_DISTANCE IGNORE_UNUSED = 40; //a search square is 16x12 (diagonal of 20), so that's about two
 
-// used for the shout action, supposedly "slightly larger than the default visual radius of NPCs"
-// while it looks too big, it is needed this big in at least pst (help())
-// bgee testing shows a radius of 72 feet
-// In the EE Help() has the range of "visualrange" * 3 * (16 * (1 or 3/4 in the vertical)) == 3x visual range, converted via Feet2pixels
-// the '48' (3*16) default can be set by the 'Audible Range' game option in baldur.lua
```
lynxlynxlynx: Please rather edit and move this documentation to the function, so it's clear there was research done and that we shouldn't be changing it without a very good reason. Something like:
```
// audible range was confirmed to be 3x visual range in EEs, accounting for isometric scaling and more than 1x visual rangein others
// in EE, the '48' (3*16) default can be set by the 'Audible Range' game option in baldur.lua
```

The tricky thing is that this is clearly not consistent. The game used a double value of visual range in several places, so our value in the code should actually be 3 * stat / 2.

gemrb/core/Map.cpp
lynxlynxlynx: 1399. No spaces should be used for indentation anywhere.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4339,7 +4339,9 @@ void Actor::PlayExistenceSounds()
 	int xpos, ypos;
 	audio->GetListenerPos(xpos, ypos);
 	Point listener(xpos, ypos);
-	if (nextComment && !Immobile() && Distance(Pos, listener) <= VOODOO_SHOUT_RANGE) {
+	double angle = atan2(listener.y - Pos.y, listener.x - Pos.x);
+	int distance = (3 * this->GetStat(IE_VISUALRANGE)) / 2; // refer to Map::Shout regarding audible range
```
lynxlynxlynx:
```suggestion
	int distance = (3 * GetStat(IE_VISUALRANGE)) / 2; // refer to Map::Shout regarding audible range
```

gemrb/core/Map.cpp
```diff
@@ -1395,8 +1395,18 @@ void Map::Shout(Actor* actor, int shoutID, unsigned int radius)
 			continue;
 		}
 
-		if (radius) {
-			if (Distance(actor->Pos, listener->Pos)>radius) {
+		if (!global) {
+			/* Audible range was confirmed to be 3x visual range in EEs, accounting for isometric scaling 
+			and more than 1x visual range in others in EE, the '48' (3*16) default can be set by the 'Audible Range'
```
lynxlynxlynx:
```suggestion
			and more than 1x visual range in others;
			in EE, the '48' (3*16) default can be set by the 'Audible Range' option in baldur.lua
```

gemrb/core/Map.cpp
```diff
@@ -1395,8 +1395,18 @@ void Map::Shout(Actor* actor, int shoutID, unsigned int radius)
...
+			game option in baldur.lua
```
lynxlynxlynx:
```suggestion
```
(gh doesn't support suggestions on multiple lines)

gemrb/core/Map.cpp
```diff
@@ -1395,8 +1395,18 @@ void Map::Shout(Actor* actor, int shoutID, unsigned int radius)
...
+
+			This is a bit tricky, it has been show to not be very consistent. The game used a double value of visual 
+			range in several places, so we will use '3 * visual_range / 2' */
+			const Point& A = actor->Pos;
+			const Point& B = listener->Pos;
+			double angle = atan2(B.y - A.y, B.x - A.x);
+			int distance = (3 * actor->GetStat(IE_VISUALRANGE)) / 2;
+			if (Distance(A, B)>Feet2Pixels(distance, angle)) {
```
lynxlynxlynx:
```suggestion
			if (Distance(A, B) > Feet2Pixels(distance, angle)) {
```

ifyGecko: I completely agree with that suggestion. That is how I would have written it but I was emulating what was there before. As far as I understand it this project has loosely governed coding standards and primarily runs on mirroring the style of already implemented code. Would you say its better to just keep adhering to the styles already used and correct things during review of a pull request or use personal judgement for the minor aesthetics like this?

lynxlynxlynx: We are never going to have the chance to fix the style in one wholesome swoop, since it would just mess up the history too much. So the rule of thumb is, when you're modifying a line already, touch it up stylistically as well. The rule in the PR template is not meant to be read on a line-by-line basis. You can see from the nearby code that we like spacing around our operators too.

At one point we'll define it and enforce it on PRs, but that's a TODO noted in #161.

ifyGecko: Thanks for the clarity!

lynxlynxlynx:
Thanks!

PR #544, opened 2019-12-20
Fix casting sound for magic schools in IWD1 & 2; BG2

### LINKED ISSUE #543: [bug] iwd uses wrong casting sounds
Continuing the investigation post #518.

Better mapping:
https://github.com/gemrb/gemrb/compare/master...caillean7:master

What are the open issues after that?

I noticed that the numbers are just odd in there. Are the even one of the other gender, something else?

caillean7:
Spell schools use the correct wav files in IWD1 with and without HoW installed. Tested. Fixes #543 

EDIT: Add modified IWD2 table, too, and correct Abjuration / Enchantment mix-up in BG2.


lynxlynxlynx:
Please check if iwd2 has the same problem.

caillean7:
Had that planned for tomorrow, but just checked, and yes, they're off in IWD2, too. I'll add IWD2s table tomorrow then.

lynxlynxlynx:
While you're at it, might as well recheck if everything is fine in bg1 and bg2 too. The initial iwd numbers were copied from bg2 ...

caillean7:
BG1 is wrong, too, but the table has the correct numbers. Something else must be off with this one.

lynxlynxlynx:
Please give an example, it used different files.

caillean7:
Try Armor - cast on the original it plays CAS_PM3*. GemRB plays CAS_PM1*. It's a conjuration spell, and the table lists '03'. Could be the zero, maybe...

lynxlynxlynx:
D-oh, it's the same problem as yesterday, just for bg1. Is 9 the highest number to be used?

caillean7:
8 is the highest for BG1. I'm checking BG2 now, it doesn't play any casting sounds at all :/

lynxlynxlynx:
There, bgs are back up.

caillean7:
Nothing changed with that commit - BG1 casting sounds are wrong and BG2 has none...

caillean7:
Eh, strike BG2, it's the installation - it works on the other one.

lynxlynxlynx:
bg1 plays CAS_PM3M.wav when casting armor by Edwin here.

caillean7:
*cough* Feel free to bite my head off. There was a cgtable in the override. Perhaps I managed to get it there yesterday when dealing with IWD, it's the same prefix. Sorry :sweat: 

I think in BG2 Abjuration uses the Enchantment sound, but I'll check that again. Two or three times.

caillean7:
Okay, got it now. BG2 really had Abjuration and Enchanting mixed up, all others are fine.

IWD2 is changed and tested, too.

lynxlynxlynx:
Thanks!

caillean7:
Sorry again about that mess with the override. I have thought about a routine to avoid such things in the future.

lynxlynxlynx:
That's why it's easiest to run from the build dir — any modifications stay in the source tree and are obvious (git diff, git status, git commit -p ...).

caillean7:
And on the installation side backing up the override and restoring a fresh copy before going about testing stuff might be a good idea, too.

Issue #554, opened 2019-12-22
[bug] PST: After defeating Ravel, the cutscene that plays after leaving the area breaks down completely


chilvence:
[000000217-GRB Exhausting Extrication.zip](https://github.com/gemrb/gemrb/files/3992083/000000217-GRB.Exhausting.Extrication.zip)

This save is made from after the prolonged conversation with Ravel and defeating her in combat, if you walk to the north you should get the option to escape the maze

The first problem here is that none of the audio plays in the cutscene that is initiated, which is a bit annoying but not serious, to my mind something about PlaySoundNotRanged is not working right

The second problem is that Ravel is meant to actually start this cutscene on the floor pretending to play dead, but in GemRB she is just standing: https://www.youtube.com/watch?v=nQ89yJw5JVU

The last problem is that after Ravel and the Transcendent one have finished rattling their bone boxes, they start to have a scripted magic fight, which eventually crashes the game:

```
[ResourceManager]: Found 'bluetint.spl' in 'GemRB Unhardcoded data'.

Thread 1 "gemrb" received signal SIGSEGV, Segmentation fault.
GemRB::SDLVideoDriver::SetFadeColor (this=0x5555555ad4d0, r=32, g=32, b=255) at /home/chiv/code/gemrb/gemrb/plugins/SDLVideo/SDLVideo.cpp:1362
1362		long val = SDL_MapRGBA( extra->format, fadeColor.r, fadeColor.g, fadeColor.b, fadeColor.a );
```
I think it is Elysium's Tears that causes this. 

I am pretty sure this is more to do with the fact that the video stuff is in the process of being overhauled, and I was able to progress past this point before, and of course problems like this should vanish once bugs in the magic system are squashed, but all the same, I cannot go any further at this point in time.

There is a lot going on in this sequence and it is a choke point that can't be avoided, but I still think it is quite cool you don't get ruined by an unavoidable point of failure until two thirds of the way through the game, I mean Beamdog said they had a hard time getting this game to work even though they had a copy of the source code and access to the living breathing Chris Avellone, so...yeah.  I mean, I could try and kidnap the guy, if it helps? Just trying to think outside the box :)

lynxlynxlynx:
Your save works for me, I successfully ended up in the Outlands. I'll assume that particular crash is SDL2 related like your previous one.

0610cut3 is the cutscene the dialog starts. The sounds are just a bit silent, since they don't use the sender position. Fixed.

[...]

PR #555, opened 2019-12-22
SDLVideo: prevents a crash on screen fading under SDL2

MarcelHB:
This is patchwork against the crash in #554. There is some more code in SDL2 that once seemed to address fade colors, but @bradallred had put (parts of) it into comments. So I guess it makes more sense to see this working in subviews than trying to recover that in master.

Just prevent people from seeing the crash in PST right now.

## Checklist


lynxlynxlynx:
Thanks!

Issue #434, opened 2019-11-28
[bug] Save-before-exit dialog appearing when in CharGen


MarcelHB:
**Describe the bug**
When you try to hard-exit the game while in CharGen, the dialog pops up that asks you if you want to save before exit. You can actually click Save, and Python errors appear, putting some of the games into a buggy state.

GemRB version: HEAD of r/n.

**To Reproduce**
1. Go into CharGen of whatever game, in IWD2 you need to go into the custom group setup.
2. Click "X" or whatever button on your window title bar to close the game.
3. Click "Save" in the dialog.

**Expected behavior**
_Save_ should be disabled, makes no sense anyway.

**Screenshots**
IWD (and some other):
```py
[GUIScript/ERROR]: Runtime Error:
[Python/ERROR]: .gemrb/GUIScripts\GUISAVE.py:97: RuntimeWarning: tp_compare didn't return -1 or -2 for exception  for j in range(min(6, MAX_PARTY_SIZE)):
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "./gemrb/GUIScripts\GUIOPT.py",line 619, in SaveGamePress
[Python/ERROR]:     GUISAVE.OpenSaveWindow ()
[Python/ERROR]:   File "./gemrb/GUIScripts\GUISAVE.py", line 97, in OpenSaveWindow
[Python/ERROR]:     for j in range(min(6, MAX_PARTY_SIZE)):
[Python/ERROR]: RuntimeError: Cannot find window index #0 (unloaded?)
```

PST:
```py
[GUIScript/ERROR]: Runtime Error:
[GUIScript/ERROR]: Syntax Error:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "./gemrb/GUIScripts/pst\GUIOPT.py", line 441, in SaveGame
[Python/ERROR]:     GUISAVE.OpenSaveWindow ()
[Python/ERROR]:   File "./gemrb/GUIScripts\GUISAVE.py", line 49, in OpenSaveWindow
[Python/ERROR]:     if GUICommon.CloseOtherWindow (OpenSaveWindow):
[Python/ERROR]:   File "./gemrb/GUIScripts\GUICommon.py", line 44, in CloseOtherWindow
[Python/ERROR]:     OtherWindowFn ()
[Python/ERROR]:   File "./gemrb/GUIScripts/pst\GUIOPT.py", line 59, in OpenOptionsWindow
[...]

PR #557, opened 2019-12-22
subviews: prevent saving in exit dialog when there is no game

MarcelHB:
See #434, but this time for the subviews branch.


bradallred:
perfect. thank you.

Issue #577, opened 2020-01-04
[subviews] SDL1.2 won't compile on GCC


MarcelHB:
Eh, I tried to make SDL1.2 with your latest changes running, and failed:
```cpp
./gemrb/plugins/SDLVideo/SDLSurfaceDrawing.h: In function 'void GemRB::DrawHLineSurface(SDL_Surface*, GemRB::Point, short int, const GemRB::Region&, const GemRB::Color&) [with bool BLENDED = true]':
./mingw64/include/SDL/SDL_stdinc.h:282:25: error: this statement may fall through [-Werror=implicit-fallthrough=]
  282 |         case 0: do {    *_p++ = _val;  \
      |                         ^
./gemrb/plugins/SDLVideo/SDLSurfaceDrawing.h:185:3: note: in expansion of macro 'SDL_memset4'
  185 |   SDL_memset4(px, c, numPx);
```

Turns out that `SDL_memset4` consists of a macro that is incompatible to the current fallthrough settings in GCC v9.2.

Is that SDL-memset crucial to use here? If so, we probably need to scale down the fallthrough strictness for that plugin.

bradallred:
its not crucial per se, but memset4 is 4x faster than memset :)

this is a function that runs a lot now with how the filled polygons work, which are also what make the sprite masks so its definitely one that you should feel a difference if your system actually provides a _real_ memset4. some just hit a fallback which is what you appear to be seeing, if gcc/linux doesn't have a memset 4 we can ifdef it for them.

bradallred:
I just realized SDL 1.2 has a different implementation of `SDL_memset4` than SDL 2 so no platform benefits from a real memset4. We should probably copy the SDL2 implementation into our code.

MarcelHB:
Well, for my Linux computer, `SDL_memset4` is about five times slower for large memory than whatever `memset` I get linked without specifying any optimizations. Same goes for the SDL2 version.

MarcelHB:
If you really want to write 4 bytes, the SDL version is 15% better than the most naive way of writing this in a loop. If you unroll your loop for grouping 4 values/16 bytes per iteration, we already twice as good as SDL again.

And what looks even more promising since we're C++11 now: `std::fill`

MarcelHB:
So, we should either switch between that Maccish `memset_pattern4` and a loop like this (instead of copying that SDL stuff), or test if we can use STL containers and algorithms instead.

bradallred:
[...]

PR #580, opened 2020-01-05
subviews: makes SDL v1.2 code compile on GCC again

MarcelHB:
Interestingly, `std::fill` is about as fast as that `SDL_memset4` on Linux, but only half as good on an old Windows 7 with Lynnfield CPU.

Anyway, that's more like the canonical way to have it compiling without errors (#577).

### DIFF
--- a/gemrb/plugins/SDLVideo/SDLSurfaceDrawing.h
+++ b/gemrb/plugins/SDLVideo/SDLSurfaceDrawing.h
@@ -182,7 +182,7 @@ void DrawHLineSurface(SDL_Surface* dst, Point p, short x2, const Region& clip, c
 		Uint32 c = SDL_MapRGBA(dst->format, color.r, color.g, color.b, color.a);
 
 		int numPx = std::min(x2 - p.x, dst->w - p.x);
-		SDL_memset4(px, c, numPx);
+		std::fill(px, px + numPx, c);
 	}
 }
 


bradallred:
maybe measure the 2 and see what the difference is?

MarcelHB:
None for Windows, slightly worse on Linux.

bradallred:
thanks for checking

bradallred:
tho, to be clear you did measure using release builds only right? debug should never be used to measure performance.

MarcelHB:
I tried with and without `-O2` (never `-g`) but that did not change the difference between either version.

bradallred:
why not -O3? thats the only worthwhile optimization setting for games :)

MarcelHB:
I was damaged by some project that had weird Heisenbugs on `-O3`, so ...

But to answer your question. With `-O3` everything is roughly on par, even slightly better than the old code on Linux, but `_n` still makes no difference.

bradallred:
sounds like you need to be introduced to UndefinedBehavior sanitizer ;)

bugs that come out at higher optimization levels are almost always a result of invoking undefined behavior when they aren't a compiler bug.

anyway, sounds like we are good to use what is already merged. thanks.

lynxlynxlynx:
-Os is sometimes better, but I agree on the dangers of -O3.

PR #586, opened 2020-01-06
BG1 avatars.2da: Additional entry for ghouls

caillean7:
It misses a line for 0x7700 (Ghouls), so the game falls back to the ettercap animation. This adds that line, tested locally.


lynxlynxlynx:
Thanks!

Issue #592, opened 2020-01-07
[subviews] Compiler issue in BIKPlayer


MarcelHB:
When trying to compile BIKPlayer with GCC v9.2 on Windows r/n, I get this error:
```cpp
In file included from ./gemrb_subviews/gemrb/plugins/BIKPlayer/rational.cpp:31:
./gemrb_subviews/gemrb/plugins/BIKPlayer/rational.cpp: In function 'int av_reduce(int&, int&, int64_t, int64_t, int64_t)':
./gemrb_subviews/gemrb/plugins/BIKPlayer/common.h:47:25: error: comparison of integer expressions of different signedness: 'uint64_t' {aka 'long long unsigned int'} and 'int64_t' {aka 'long long int'} [-Werror=sign-compare]
   47 | #define FFMIN(a,b) ((a) > (b) ? (b) : (a))
      |                     ~~~~^~~~~
./gemrb_subviews/gemrb/plugins/BIKPlayer/rational.cpp:62:27: note: in expansion of macro 'FFMIN'
   62 |             if(a1.den) x= FFMIN(x, (max - a0.den) / a1.den);
      |                           ^~~~~
./gemrb_subviews/gemrb/plugins/BIKPlayer/rational.cpp:64:43: error: comparison of integer expressions of different signedness: 'long long unsigned int' and 'int64_t' {aka 'long long int'} [-Werror=sign-compare]
   64 |             if (den*(2*x*a1.den + a0.den) > num*a1.den) {
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
```

It looks as if you can trivially fix this by changing the type of `x` from `uint64_t` to `int64_t`, but I'm not sure if there is some numerical reason for this in the first place - or how and why that worked before.

bradallred:
we should use `std::min<type>` (and delete `FFMIN`) for whatever type is on the left (type of x?).

lynxlynxlynx:
I changed x to signed — video playback hasn't suffered. Didn't look into its use though, just a quick break from work.

bradallred:
of course that works too. I still like to lean toward API consistency and using <algorithms> when we can :shrug:

lynxlynxlynx:
Doesn't std::min also require the arguments to be of the same type? If the <type> version works, doesn't it just hide the underlying problem by casting implicitly?

bradallred:
[...]

PR #601, opened 2020-01-10
subviews: fix for BIKPlayer breaking builds

MarcelHB:
See #592; no obvious video glitches yet.


bradallred:
Would you mind changing the other uses of `FFMIN` and deleting the macro as well?

MarcelHB:
I don't. I just hestiated to do too much inside of some alien code. :wink: Wait a minute.

bradallred:
beautiful. Thank you.

PR #650, opened 2020-04-03
Allow importing character files with non-ASCII characters

fenuks:
## Description
I've fixed import issue with importing characters from files containing non-ASCII letters. More context and discussion is contained within issue report.

iconv is used for encoding, and since it isn't hard dependency, it's optional (without iconv no convention between encoding will be made, resulting in the same error).

`<langinfo.h>` is being used for getting user system encoding. As I can see, it might be not available on Windows, so that might require change. Unfortunately, I don't have Windows machine, so I would be grateful if someone could test that.

Polish version of Icewind Dale that I have is using `WINDOWS-1250` for encoding dialogue, so I added ini file to configure that. This information is being used during string convention, because GUI correctly displays Polish letters only in this encoding.

I don't know if any adjustments to documentation are needed, probably it could be mentioned that in case of Polish ID: HOW, one must change encoding in gemrb.cfg.

import screen

## Checklist


fenuks:
Oh, and I forgot to mention. Possibly, generated `config.h` file has an error. Even though my system is equipped with `iconv` and it is detected, `#define HAVE_ICONV 1` is not added. There is only commented out `/* #undef HAVE_ICONV */`. I had to add that line manually to test my changes with iconv enabled.

gemrb/core/Interface.cpp
```diff
@@ -250,6 +251,7 @@ Interface::Interface()
 	SpecialSpellsCount = -1;
 	SpecialSpells = NULL;
 	Encoding = "default";
+	SystemEncoding = nl_langinfo(CODESET);
```
lynxlynxlynx, line 254: You're right that this isn't available on windows. How about using localeconv instead?

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,48 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+	    return clone(string);
+	}
+
+	/* buf = (char*)realloc(buf, newlen+1); */
+	size_t used = out_len - out_len_left;
+	buf = (char*)realloc(buf, used + 1);
+	buf[used] = '\0';
+	return buf;
+#else
+	return clone(string)
```
lynxlynxlynx, line 146:
```suggestion
	return clone(string);
```

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,48 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+	char* buf = (char*) malloc(out_len);
+	char* buf_out = buf;
+	#if __FreeBSD__
+		size_t ret = iconv(cd, (const char **)&in, &in_len, &buf_out, &out_len_left);
+	#else
+		size_t ret = iconv(cd, &in, &in_len, &buf_out, &out_len_left);
+	#endif
+
+	if (ret == (size_t)-1) {
+	    perror("iconv");
```
lynxlynxlynx, line 135: We have a Log function for this sort of thing, please change all these perrors to that and make them more descriptive.

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,48 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+	    free(buf);
+	    return clone(string);
+	}
+
+	/* buf = (char*)realloc(buf, newlen+1); */
```
lynxlynxlynx, line 140:
```suggestion

```

gemrb/core/System/String.cpp
```diff
@@ -230,6 +280,14 @@ int strnlen(const char* string, int maxlen)
 }
 #endif // ! HAVE_STRNLEN
 
+char* clone(const char * str) {
+	size_t len = strlen(str);
+	char * buf = (char*) malloc(len + 1);
+	strcpy(buf, str);
+	buf[len] = '\0';
+	return buf;
+}
```
lynxlynxlynx, line 289: Is this really needed? We use strndup.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -6689,17 +6693,20 @@ static PyObject* GemRB_TextArea_ListResources(PyObject * /*self*/, PyObject* arg
...
 				if (pos == String::npos || (type == DIRECTORY_CHR_SOUNDS && pos-- == 0)) {
+					free(str);
 					delete string;
 					continue;
 				}
 				string->resize(pos);
 			}
-			StringToUpper(*string);
 			strings.push_back(*string);
+			free(str);
```
lynxlynxlynx, line 6709: You can just move these two frees to right after `string` is created, since they're not used later any more.

lynxlynxlynx:
You're right about iconv and I see the bug, fixing ...

gemrb/core/Interface.cpp
fenuks: I think localeconv don't have encoding information, it seems to have only information related to currency symbols, digit separators, and such. From what I quickly read, it will be required to use Windows-specific function, but I will investigate that more thoroughly later.

lynxlynxlynx: maybe this would do, storing it before we override it:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-current-locale?view=vs-2019

fenuks: I've been reading a bit on the subject, and I'm afraid that I won't be able to add Windows support, mainly because I don't have Windows installed on my computer, so I cannot find solution with trial and error. As I understand, one can easily fetch code point, and iconv supports many of these. Windows also has native functions to convert from and to wide strings, which potentially could be used. https://docs.microsoft.com/en-us/windows/win32/intl/code-pages

My apologies, but could someone having Windows computer and expertise help me to finish this merge request?

gemrb/unhardcoded/how/polish.ini
```diff
@@ -0,0 +1,2 @@
+[encoding]
+TLKEncoding = WINDOWS-1250
```
lynxlynxlynx, line 2: please also copy it to ../iwd/

gemrb/core/Interface.cpp
lynxlynxlynx: I suggest we just if-def it to UTF-16 in that case. I'm not sure about the defaults on windows and there's some problem, but it appears it is the 16 bit variant.

Also, is it really about the system encoding? You can have different encodings per partition on linux.

gemrb/core/System/String.cpp
```diff
@@ -32,6 +32,11 @@
 #endif
 #endif
 
+#if HAVE_ICONV
+#include <iconv.h>
+#include <errno.h>
```
bradallred, line 37: this is a c++ project, please use `<cerrno>`

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,47 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+
+
+	char * in = (char *) string;
+	size_t in_len = strlen(string);
+	// twice as much bytes should be safe for most encodings
+	size_t out_len = in_len * 2 + 1;
+	size_t out_len_left = out_len;
+	char* buf = (char*) malloc(out_len);
+	char* buf_out = buf;
+	#if __FreeBSD__
```
bradallred, line 128: we should use `#if ICONV_ACCEPTS_NONCONST_INPUT` instead

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,47 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+	    free(buf);
+	    return strdup(string);
+	}
+
+	size_t used = out_len - out_len_left;
+	buf = (char*)realloc(buf, used + 1);
+	buf[used] = '\0';
+	return buf;
+#else
+	return strdup(string);
```
bradallred, line 145: we should add a warning log message about no iconv I think.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -2007,7 +2007,11 @@ static PyObject* GemRB_Control_QueryText(PyObject * /*self*/, PyObject* args)
 	if (!ctrl) {
 		return NULL;
 	}
-	char* cStr = MBCStringFromString(ctrl->QueryText());
+
+	String wstring = ctrl->QueryText();
+	std::string nstring(wstring.begin(), wstring.end());
+	char * cStr = (char*) ConvertCharEncoding(nstring.c_str(),
```
bradallred, line 2013: why are we doing a const cast? just change the return type of the function please.

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,47 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
 	return dbString;
 }
 
+// returns new string converted to given encoding
+// in case iconv is not available, or there is encoding error, copy of original
+// string is returned
+const char* ConvertCharEncoding(IGNORE_UNUSED const char* string,
+        IGNORE_UNUSED const char* from, IGNORE_UNUSED const char* to) {
+#if HAVE_ICONV
```
bradallred, line 113: I think it would be cleaner to do this:

```
#if HAVE_ICONV
const char* ConvertCharEncoding(const char* string, const char* from, const char* to)
{
...
}
#else 
const char* ConvertCharEncoding(const char*,  const char*, const char*)
{
...
}
#endif
```

that also fixes the unused parameter warnings without allowing them for the iconv case

gemrb/core/System/String.cpp
lynxlynxlynx: We already do, he'll just have to rebase to current HEAD.

gemrb/core/Interface.cpp
fenuks: Alright, we can do that. I will update review.

> Also, is it really about the system encoding? You can have different encodings per partition on linux.

Well, usually yes. As I can read Gnome's Glib assumes that filenames are encoded in UTF-8, and Qt assumes they are in user encoding. Tools like `ls` also seem to depend on user-locale, `LANG=c ls` doesn't show correctly filenames with non-ASCII characters on my system. I suppose assuming that filenames are encoded in user-locale is reasonable thing to do.

lynxlynxlynx:
Thanks for taking care of most of the feedback. You should have rebased though, now all the new commits from master show up here. You can make a new branch, reset this one to master and cherrypick your work back. A force push then should clean up the presentation on github.

Looking back at @MarcelHB's similar problem with sound sets in #47, apparently:
- std::setlocale from <clocale> can also be used to look up the current locale
- utf-8 was used a default on windows as well

fenuks:
> Thanks for taking care of most of the feedback. You should have rebased though, now all the new commits from master show up here. You can make a new branch, reset this one to master and cherrypick your work back. A force push then should clean up the presentation on github.

Should be fixed now.

> Looking back at @MarcelHB's similar problem with sound sets in #47, apparently:
> 
>     * std::setlocale from  can also be used to look up the current locale

Yes, in general `setlocale` returns name of locale if it was correctly set, it was my intention at the beginning to get encoding information from that string, but it would require parsing something like `pl_PL.UTF-8`. Not practical.

lynxlynxlynx:
Not sure why it doesn't link on mac, but it looks like it didn't construct the right linker command.

lynxlynxlynx:
@MarcelHB can you give this a try to verify it works on windows?

Travis is currently in maintenance, so the logs aren't visible, but @bradallred please check if this doesn't build for you either.

gemrb/core/Interface.cpp
```diff
@@ -250,6 +254,11 @@ Interface::Interface()
 	SpecialSpellsCount = -1;
 	SpecialSpells = NULL;
 	Encoding = "default";
+	#ifndef WIN32
+	SystemEncoding = nl_langinfo(CODESET);
+	#else
+	SystemEncoding = "UTF-16";
```
MarcelHB, line 260: You can't do this in C++11 anymore. I suggest to use `std::string` that and feed its `.c_str()` where required.

fenuks: You meant `SystemEncoding` not being `const`? Right, how about adding `const` modifier, instead of using `std::string`?

MarcelHB: Sure, if it suits.

gemrb/core/System/String.cpp
```diff
@@ -100,6 +105,49 @@ static String* StringFromEncodedData(const ieByte* string, const EncodingStruct&
...
+	size_t in_len = strlen(string);
+	// twice as much bytes should be safe for most encodings
+	size_t out_len = in_len * 2 + 1;
+	size_t out_len_left = out_len;
+	char* buf = (char*) malloc(out_len);
+	char* buf_out = buf;
+	#if ICONV_ACCEPTS_NONCONST_INPUT
+		size_t ret = iconv(cd, &in, &in_len, &buf_out, &out_len_left);
+	#else
+		size_t ret = iconv(cd, (const char **)&in, &in_len, &buf_out, &out_len_left);
```
MarcelHB, line 130: Compiler error:
```gcc
./gemrb/gemrb/core/System/String.cpp:130:26: error: invalid conversion from 'const char**' to 'char**' [-fpermissive]
  130 |   size_t ret = iconv(cd, (const char **)&in, &in_len, &buf_out, &out_len_left);
      |                          ^~~~~~~~~~~~~~~~~~
      |                          |
      |                          const char**
In file included from ./gemrb/gemrb/core/System/String.cpp:36:
./mingw64/include/iconv.h:82:43: note:   initializing argument 2 of 'size_t libiconv(libiconv_t, char**, size_t*, char**, size_t*)'
   82 | extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
```

lynxlynxlynx: That's odd, how come FindICONV.cmake didn't define ICONV_ACCEPTS_NONCONST_INPUT for you?

fenuks: Apparently, Windows version of `iconv` expects non-const input, and cmake didn't detect that. What your cmake log says, can you see iconv-related warnings about cmake not being able to determine `iconv` signature?

MarcelHB: I'll check that again, it looks as if I have some artifacts of the attempt to reproduce that SDL1-iconv-AppVeyor in place.

MarcelHB: Clean CMake fixed _that_.

Now my linker complains about them with `lib` prepended to these functions, like:
```gcc
./mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: E:/Devel/Sources/Cpp/gemrb/gemrb/core/System/String.cpp:128: undefined reference to `libiconv'
```
I'll try to find out what this is about.

MarcelHB:
I'm not sure if this the way to go, but something like this is still missing:

```diff
diff --git a/gemrb/core/CMakeLists.txt b/gemrb/core/CMakeLists.txt
index 3255f77a3..5df267ea2 100644
--- a/gemrb/core/CMakeLists.txt
+++ b/gemrb/core/CMakeLists.txt
@@ -150,6 +150,9 @@ if (STATIC_LINK)
 else (STATIC_LINK)
 	ADD_LIBRARY(gemrb_core SHARED ${gemrb_core_LIB_SRCS})
 	TARGET_LINK_LIBRARIES(gemrb_core ${CMAKE_DL_LIBS} ${COREFOUNDATION_LIBRARY})
+	if (HAVE_ICONV)
+		TARGET_LINK_LIBRARIES(gemrb_core ${ICONV_LIBRARY})
+	endif()
 	IF(WIN32)
 	  INSTALL(TARGETS gemrb_core RUNTIME DESTINATION ${LIB_DIR})
 	ELSE(WIN32)
```

MarcelHB:
In CharGen of IWD1, I now encounter those error (German):
```
[ResourceManager]: Searching for 'kämpfer_männlich_4/hemf_09'...
[ResourceManager]: Found 'kämpfer_männlich_4/hemf_09.wav' in 'Sounds'.
[String/ERROR]: iconv failed to convert string Kämpfer_männlich_1 from ISO-8859-1 to UTF-16 with err
or: Arg list too long
```

fenuks:
> In CharGen of IWD1, I now encounter those error (German):
> 
> ```
> [ResourceManager]: Searching for 'kämpfer_männlich_4/hemf_09'...
> [ResourceManager]: Found 'kämpfer_männlich_4/hemf_09.wav' in 'Sounds'.
> [String/ERROR]: iconv failed to convert string Kämpfer_männlich_1 from ISO-8859-1 to UTF-16 with err
> or: Arg list too long
> ```

Could you please try again with this patch applied?

```diff
diff --git a/gemrb/core/System/String.cpp b/gemrb/core/System/String.cpp
index 46a3854e18..13ebaf9fb4 100644
--- a/gemrb/core/System/String.cpp
+++ b/gemrb/core/System/String.cpp
@@ -120,7 +120,7 @@ char* ConvertCharEncoding(const char* string, const char* from, const char* to)
 	char * in = (char *) string;
 	size_t in_len = strlen(string);
 	// twice as much bytes should be safe for most encodings
-	size_t out_len = in_len * 2 + 1;
+	size_t out_len = (in_len + 1) * 2;
 	size_t out_len_left = out_len;
 	char* buf = (char*) malloc(out_len);
 	char* buf_out = buf;
```

MarcelHB:
I can check this later tomorrow. Nonetheless, this looks mandatory to fix.

gemrb/core/System/String.cpp
bradallred: I don't get this `#if` a `char**` (or any pointer for that matter) can always be passed to a function that takes the `const` variant. What am I missing?

lynxlynxlynx: The const variant isn't guaranteed to be there. There's also ICONV_ACCEPTS_CONST_INPUT if we wanted to invert the check.

bradallred: I know the const variant isn't guaranteed to exist. However, `oldchar` is already a `char*` so it could be passed to either a `const` or non `const` function. I don't understand the need for this still :)

lynxlynxlynx: Neither is guaranteed to exist, not just the one.

bradallred: I'm still lost. Those parameters should be good to pass to either variety... it shouldn't matter which function exists.

lynxlynxlynx: Ahh, I see what you meant now!

MarcelHB:
With your patch, I seem to end up in an error that converts into the other direction, obviously:
```
[String/ERROR]: iconv failed to convert string K▒mpfer_m▒nnlich_5 from UTF-16 to ISO-8859-1 with error: Illegal byte sequence
```

gemrb/core/System/String.cpp
fenuks: It was adaptation of code used in TTFImporter. @bradallred makes a valid point, I will change that.

fenuks:
> With your patch, I seem to end up in an error that converts into the other direction, obviously:

This is called to convert filename to TLK encoding. I *think* problem might lie in GemRB not using wide variant of `_findfirst` in [its dirent polyfill](https://github.com/gemrb/gemrb/blob/491b32ee028812f8e9235ed1cbe8a7a4959c39ff/gemrb/core/System/VFS.cpp#L100). https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/findfirst-functions?view=vs-2019

MarcelHB:
Urgh, yes. The whole thing looks broken for this purpose. I'll try to do something about it.

MarcelHB:
Well, I'd prepare a PR for Windows myself first. Reason is that GemRB ist *not* defining `UNICODE` anywhere, and the code isn't fully prepared to simply allow that. This is great for supporting Windows before XP, but probably won't bring us anywhere here.

MarcelHB:
Ok, I got something and I remember why I hate the whole thing so much. If you work with Win32 API like its 2000 or later, you do this exclusively over UTF-16 strings. We did not so far, and other platform just take multibyte strings.

I experimentally tested one approach: let Windows-VFS/FS/PluginLoader encode everything for WinAPI. This works but the assumption is that everything that hits these things is multibyte encoded -- but this is probably breaking some IE distributions. We can maybe take locales into account, but this competes with iconv in some ways.

Other approach: _Everything_ that goes VFS & co must be encoded _beforehand_ when running Windows. Asset file names, config values, sound files, save games, ... This is the approach of this PR but  we better have some thorough testing if this is the case for everything. In order not to break signatures, we'd blindly convert the wide-char `char*` inputs as `wchar_t*` in the Windows code sections, with the inverse direction handled as well.

Not sure what is to prefer right now.

MarcelHB:
@lynxlynxlynx Opinion? I just expect this PR to hang unless this is resolved in one way or another.

lynxlynxlynx:
This is beyond me, but the way you describe it, the first approach sounds a bit less extreme.

MarcelHB:
I agree. Then I'll test again how iconv is helping here.

bradallred:
I can't help but feel that at some point we will just need to step up and make all internal strings utf16 and convert things at the TLK/VFS boundaries. I say utf16 instead of utf8 or anything else because the BAM fonts require 2 bytes. we probably also need to have some way of detecting/setting the encoding for the filesystem so we actually can properly convert paths.

of course this does nothing for the additional complications of what possible encoding the 2da/ini/config files are in. We should probably convert all those to UTF16 too. The joys of things developed prior to unicode I guess. Definitely a lot more work to do it right than to keep hacking at it lol.

lynxlynxlynx:
At least that last part is easy: all data is in ascii. We're lucky they used strrefs everywhere. :) The only thing that comes to mind (besides our own cfg with paths) as a potential exception are the keymap.ini files, but even those I doubt have different binding names or keys.

MarcelHB:
@fenuks May I ask you to rebase this branch on master? I don't expect any trouble with non-Windows platforms, but there will be for Windows. I hope the solution is as easy as removing the `"UTF-16"` hardcoding.

fenuks:
Rebased on top of the latest master.

MarcelHB:
Thanks, I will test that later today.

MarcelHB:
Looks good (setting the locale string to what my system is) but still has some missing link: Querying `setlocale` on Windows returns strings that are incompatible to iconv. You can also query the code page identifier and map it to iconv. It's just a [non-trivial amount](https://docs.microsoft.com/de-de/windows/win32/intl/code-page-identifiers) of code pages out there, I found this one to look like what we are needing: https://github.com/win-iconv/win-iconv/blob/master/win_iconv.c#L158

The file is claimed to be public domain, and from [what I get, we can import it then](https://www.gnu.org/licenses/gpl-faq.en.html#CombinePublicDomainWithGPL)? I just don't have a better idea right now.

lynxlynxlynx:
Yeah, we can include it, though I would trim it down to the list we actually need.

MarcelHB:
@fenuks I propose to work on a copy of your branch and do what is missing to get Windows running. You can then cherry-pick or merge or just apply the patch into yours. I should find the time to do this no later than weekend.

fenuks:
> I propose to work on a copy of your branch and do what is missing to get Windows running. You can then cherry-pick or merge or just apply the patch into yours. I should find the time to do this no later than weekend.

Fine with me. Unfortunately I don't usually have enough time during week to do something substantial on GemRB, but on weekend I will be able to spend more time on this, and some other improvements targeting Icewind Dale.

MarcelHB:
@fenuks  I was testing around with my [Windows patch](https://github.com/fenuks/gemrb/compare/icw_import...MarcelHB:icw_import_win.diff) (works well!) when I noticed something on Linux: TLK is ISO-8859-1, system encoding is UTF-8. On IWD2, I can see the sound files having correctly shown umlauts but when clicking, GemRB can't find the resource file -- to be sure, I tested that in your branch as well. So can you reproduce this?

fenuks:
With this review patch set (and also with yours applied atop of them) I can play character sounds with non-ASCII letters. I tested this on both Icewind Dale games, works in case of sequel as well, although in my case I need to add polish.ini to specify that encoding is in fact `WINDOWS-1250` (prior to that change I get numerous errors opening voice selection screen). Are you certain that your game encoding is `ISO-8859-1`? These both encoding are similar, but not the same.

During work on this review, I had problem with non-ASCII strings that were uppercased. Perhaps it is somehow related to your issue. Thing is that on case-sensitive filesystems GemRB tries to do the following: find exact match, if it doesn't exist convert casing of searched and system resource, and compare them for match. Convention doesn't work for non-ASCII letters, therefore there will be no match. As consequence, for example, if your sounds folder contains non-ASCII letter and in config you didn't specify it with cased exactly as on filesystem, then you will get error. Could you paste logs from console?

lynxlynxlynx:
Please then also just remove the existing polish.ini additions and add it to gemrb/unhardcoded/shared/ instead.

Maybe both Polish and German need to have the uppercase chars manually mapped like ./gemrb/unhardcoded/shared/czech.ini?

MarcelHB:
Here is a log line, `ä` is the thing in question.
```
[ResourceManager]: Searching for 'k�mpfer_m�nnlich_3/08'...
[ResourceManager/WARNING]: Couldn't find 'k�mpfer_m�nnlich_3/08'... Tried k�mpfer_m�nnlich_3/08.wav k�mpfer_m�nnlich_3/08.ogg k�mpfer_m�nnlich_3/08.acm k�mpfer_m�nnlich_3/08.wav
```
I must go through the string table do check if it is really something different to ISO-8859-1 as this is the default 8bit encoding that also includes German. For the umlaute it's no difference to Windows-1252 either -- that's what I have on Windows 7.

MarcelHB:
Case-theory is the best guess then, I think, my path is `Kämpfer_männlich_3` etc. I'll try to get what the czech table is doing and try the same for German.

lynxlynxlynx:
Check out gemrb/docs/en/default_ini.txt.

MarcelHB:
Cool, that made it. I've added an INI file to the patch file above. So please reapply before pushing. :slightly_smiling_face:

fenuks:
> Please then also just remove the existing polish.ini additions and add it to gemrb/unhardcoded/shared/ instead.

I added it just now to test ICW2 since MarcelHB reported problems with this particular game, since without it I got numerous errors. However, I focus on testing IWD1, but I guess it I could add `polish.ini` for IWD2 in this review anyhow.

fenuks:
> Cool, that made it. I've added an INI file to the patch file above. So please reapply before pushing. slightly_smiling_face

Thank you very much for your help with Windows support! Applied it and pushed to this branch.

lynxlynxlynx:
> > Please then also just remove the existing polish.ini additions and add it to gemrb/unhardcoded/shared/ instead.
> 
> I added it just now to test ICW2 since MarcelHB reported problems with this particular game, since without it I got numerous errors. However, I focus on testing IWD1, but I guess it I could add `polish.ini` for IWD2 in this review anyhow.

You need it for iwd, so I highly doubt it will be any different for iwd2. The move is just a way to reduce the amount of files. In that path it will be used for all games.

gemrb/core/System/String.cpp
fenuks: We could, but I'm afraid that would only obfuscate log, as this function can be called many times.

fenuks:
@bradallred @lynxlynxlynx This changeset should now be more or less complete after Windows support was added, could you please take another look on it?

lynxlynxlynx:
You didn't include all of Marcel's work. And his new files are missing the license headers (same everywhere).

MarcelHB:
What's the most recent one to use? Found some very old ones around the code so I was not sure.

lynxlynxlynx:
https://github.com/gemrb/gemrb/blob/master/gemrb/core/Game.cpp looks fine and I don't remember seeing much diversity, save for some address errors that bots notified us about.

fenuks:
> You didn't include all of Marcel's work. And his new files are missing the license headers (same everywhere).

I have applied attached patch, and everything that came with it.

Will add headers in a moment.

lynxlynxlynx:
You didn't include all the files, eg. it created gemrb/unhardcoded/shared/german.ini, which you haven't committed.

gemrb/core/CodepageToIconv.cpp
```diff
@@ -1,3 +1,21 @@
+/* GemRB - Infinity Engine Emulator
+ * Copyright (C) 2004 The GemRB Project
```
lynxlynxlynx, line 2: Like discussed, make them 2020.

fenuks: Fixed.

lynxlynxlynx:
Looks fine to me now, thanks. To make it perfect, you can fix also brad's comment for that iconv ifdef.

fenuks:
> You didn't include all the files, eg. it created gemrb/unhardcoded/shared/german.ini, which you haven't committed.

If I didn't break anything rewriting that commit, it should now contain all changes from updated Marcel patch.

fenuks:
> Looks fine to me now, thanks. To make it perfect, you can fix also brad's comment for that iconv ifdef.

Thanks for reminding me of that. Changed.

lynxlynxlynx:
Reread it all and have one suggestion for you and one question for @MarcelHB. Please add a shortcircuit to ConvertCharEncoding in case (from == to), since this gets called even when conversion isn't needed. Eg. I won't be needing it, but I do have iconv installed.

Marcel: you need to set Encoding=german for the ini file to take effect, right? If so, we need to fix those config comments, since they're saying it's only needed for polish. But this can be done separately from this PR.

MarcelHB:
I see `default_ini.txt` is still untouched. But then let's maybe clean this up altogether then because `Except of polish BG1, leave disabled.` (GemRB cfg sample) sounds confusing w. r. t. the available config options.

fenuks:
@MarcelHB
Which IE games need tweaking Encoding in German versions?

In case of Polish, `default_init.txt` states that only BG1 requires that, but in fact, I had to specify encoding for IDW1 and IWD2 (have to check if BG2 should also be added to list).

lynxlynxlynx:
All of them, like Polish. So the docs to change are: cfg samples, manpage, default_ini.txt. Just remove the whole warning sentence. The website is fine and I already added german there.

Thanks for the shortcut. It could be cleaner in other languages (I forgot the technical term, sort of self-modifying code), but I guess we could add a static variable if it ever becomes a bottleneck.

fenuks:
Hopefully, I grepped through all occurrences that required change.

lynxlynxlynx:
Wohoo, a little saga complete. Thanks everyone, it was a nice team effort. :)

fenuks:
Thank you and Brad for patient guidance, and Marcel for Windows bits. Glad to finally see it in master, as I had to rebase my other changes against this branch, if I wanted to check something that touched importable characters. XD

chilvence:
Sorry to bump after everything is already over

I was updating my repo to gemrb/master, and found that in order for this patch to work for me on Visual Studio 2019, I have to add `#include <locale.h>`  to gemrb.cpp, or the [setlocale line](https://github.com/gemrb/gemrb/blob/4f1da4803a975cb8a2243c01923dde3556b2fc06/gemrb/GemRB.cpp#L59) doesn't work 

I don't know why MSVC has to be different. I don't have any different language versions of the games any more to test, but it looks like that was all that was needed? 

Can anyone throw me a dialog.tlk file to check it out?

bradallred:
go ahead and submit a PR, but change that include to `<clocale>`.

Issue #651, opened 2020-04-06
The counter on a stack of expendable items can go down to 0 and then wraps through to 65535.


kindaro:
**Describe the bug**

Suppose I have a stack of 3 bandages in a quick item slot. I can use them 3 times and have a stack of 0 bandages. The same works with charcoal charms. If I use the stack again, I shall have what looks like a stack of 6553 items, but I suspect the last digit simply did not fit. Essentially this means infinite stacks.

GemRB version: 0.8.6

**To Reproduce**
1. Get a stack of usable items.
2. Use them up via the circular menu, one by one.
3. Observe a stack of 0 items.
4. Use the stack once again.
5. Observe a very large stack.

**Expected behavior**
Stack disappears when the counter reaches 0.

**Screenshots**
![stack](https://user-images.githubusercontent.com/2599460/78548473-8767ff80-7809-11ea-8995-f7f30c305b3c.png)

lukas-valenta:
This looks like a good starting issue. Can I give it a shot?

lynxlynxlynx:
Thanks for the report and thanks for volunteering! What a furtuitiuous coincidence, two new users and bam. :)

Kindaro, does it matter if the game is paused or not?

Lukas: I'd start by looking at gemrb/GUIScripts/pst/FloatMenuWindow.py, which handles this window's logic. Check where it calls Gemrb.UseItem or maybe links into InventoryCommon.py or GUICommon.py.

kindaro:
@lynxlynxlynx  No, it does not matter whether the game is paused.

lynxlynxlynx:
Hmm, this will likely then be a problem in core, not in the script.

I specifically disabled aura pollution (how other games limit you to one use per round) in 78593b74. Perhaps the hack is not good enough. Lukas, I suggest you look at other uses of spellCount in GameControl and where we actually consume the item on use.

lukas-valenta:
@lynxlynxlynx yes, i think it's in the core. i did some preliminary exploration yesterday, but i was blocked by #652 before trying an older tag. i will look more closely today

lukas-valenta:
I found out that replacement item Z_GEN2B.itm is not found and therefore the original item is not deleted. As this is my first foray into this code, I'd appreciate some testing as I don't know what else I broke. Expending items normally seems to be working now.

lynxlynxlynx:
Where is this Z_GEN2B coming from? Items generally aren't replaced on consumption.

[...]

PR #653, opened 2020-04-07
fix expended items not deleting

lukas-valenta:
## Description
My first attempt to contribute. I've found out that depleted items are not deleted because the replacement item "Z_GEN2B.itm" is not present. as @lynxlynxlynx pointed out, it's actually a sound as PST does not use replacement items.

Issue #651 

## Checklist

### DIFF
--- a/gemrb/core/Inventory.cpp
+++ b/gemrb/core/Inventory.cpp
@@ -1420,6 +1420,8 @@ void Inventory::SetSlotItemRes(const ieResRef ItemResRef, int SlotID, int Charge
 			SetSlotItem( TmpItem, SlotID );
 		} else {
 			delete TmpItem;
+			//if the item isn't creatable, we still destroy the old item
+			KillSlot( SlotID );
 		}
 	} else {
 		//if the item isn't creatable, we still destroy the old item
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, let's keep the discussion in the bug for now. Also checked and the build failure is not related to your change, but the recent fix for iconv to get built more often. On it.

gemrb/core/Inventory.cpp
```diff
@@ -1468,7 +1468,8 @@ void Inventory::BreakItemSlot(ieDword slot)
 	const Item *itm = GetItemPointer(slot, Slot);
 	if (!itm) return;
 	//if it is the magic weapon slot, don't break it, just remove it, because it couldn't be removed
-	if(slot ==(unsigned int) SLOT_MAGIC) {
+	//or for pst, just remove it as there is no breaking (the replacement item is a sound)
+	if(slot ==(unsigned int) SLOT_MAGIC || core->HasFeature(GF_HAS_PICK_SOUND)) {
```
lynxlynxlynx, line 1472: This is the correct fix, nice job. :)
Since you're modifying the line, please also fix the inherited stylistic issues:
```suggestion
	if (slot == (unsigned int) SLOT_MAGIC || core->HasFeature(GF_HAS_PICK_SOUND)) {
```

lukas-valenta:
Fixed it ;)

lynxlynxlynx:
Thanks!

Issue #655, opened 2020-04-08
Biographies are confused.


kindaro:
## Description of the bug:

Biographies of characters are confused. For example, the biography of the Nameless One starts with:

> I am Vhailor. I am a Mercykiller. I am justice.

— Which is clearly wrong.

Biography of Morte does not get displayed at all — instead, the game hangs.

## GemRB version:

Current master, commit `4b2632ab6`.

## Steps to reproduce the behavior:

1. Start a new life or resume an existing one.
1. Open the Statistics screen.
3. Click _«Information»_.
4. Click _«Biography»_.

## Expected behavior:
A biography that aligns with what we know about a character from other sources.

lynxlynxlynx:
Yeah, the order is out of whack. Morte ends up with the default value in the table (I think he has IE_SPECIFIC of a human), that's why python gets stuck:
```
[Python/ERROR]: Traceback (most recent call last):                                                                                                  
[Python/ERROR]:   File "../gemrb/GUIScripts/pst/GUIREC.py", line 832, in OpenBiographyWindow                                                        
[Python/ERROR]:     BioText = int (BioTable.GetValue (BioTable.GetRowName (Specific+1), 'BIO'))                                                     
[Python/ERROR]: ValueError: invalid literal for int() with base 10: '*' 
```
Table: gemrb/unhardcoded/pst/bios.2da

So clearly it is being read wrong. We should be checking with some other stat or the name instead.

PR #661, opened 2020-04-09
Fix biography selection.

kindaro:
Fixes issue #655.


kindaro:
This may be the shortest patch I shall ever propose. Why was that `+1` there in the first place?

lynxlynxlynx:
You can look at git blame output. Does this work for all the npcs? Eg. also Ignus, Vhailor.

kindaro:
I do not have saves with all of them, only The Nameless One, Morte and Dak'kon, but all three are now correct.

lynxlynxlynx:
Yep, works for everyone, thanks. :)

Issue #667, opened 2020-04-11
IWD2: Mage spell books are read/saved incorrectly 


MarcelHB:
#### Bug description
Speaking of savegames, there is some open issue in IWD2. The way we write down spellbooks into savegames makes the original confuse.

GemRB version: master

#### Steps to reproduce
1. Take a mage in IWD2, observe mage spells/spell tabs.
2. Save the game.
3. Open in GemRB, have a look, and save again.
4. Open in native IWD2, look at the mage's spellbook again.

It's _not_ completely wrong, IWD2 shows the correct spells in the action bar!

#### Expected behavior
What IWD2 has set up and persists into the save files.
![iwd2_spellbook_mage](https://user-images.githubusercontent.com/238558/79041578-33865d80-7bf1-11ea-9bc6-2f2adb00ba80.jpg)

#### Screenshots
What GemRB reads (mages afaik don't have domain spells, so the tab should be hidden):
![iwd2_spellbook_mage_gemrb](https://user-images.githubusercontent.com/238558/79041601-57e23a00-7bf1-11ea-9aee-052580640894.jpg)

What GemRB writes and what IWD2 then sees:

![iwd2_mage_spellbook_post_gemrb](https://user-images.githubusercontent.com/238558/79041613-6a5c7380-7bf1-11ea-9a30-46bbec280e22.jpg)

lynxlynxlynx:
Saw the domain problem before, but never got annoyed enough to check why we're granting these slots (which is likely the reason it ends up in the save). Or maybe they're misdetected at someplace, since they do contain all sorts of spells, not just divine.

MarcelHB:
Looks as if we screw up CRE bytes 0x3B4 to 0x3B9.

lynxlynxlynx:
```
0x03b4 | 2 (word) | AVClass value (duplicate of class, used for object matching)
0x03b6 | 2 (word) | ClassMsk bitfield value (duplicate of class, used for object matching)
0x03b8 | 2 (word) | Unknown
```
Are you talking just about gemrb saves in the original or also gem in gem?

MarcelHB:
I only checked IWD2 -> GemRB for now:

was (by IWD2): `0B 00 00 04 00 00`
turns into (by GemRB): `0C 00 0C 00 0C 00`

Replacing these makes the save game look better.

lynxlynxlynx:
Ah, ok. The first two are a bit confusing, since there are conflicting ids available (also with 20 suffix iirc).

PR #681, opened 2020-04-13
IWD2: compatibility improvements to save games

MarcelHB:
## Description
See two issues described in #667.

Bit of rationale:
- I stick to the class mask that is calculated by `GetClassMask()` instead of reading it in, as it looks equivalent.
- Looking at this domain spells hack, I think it only makes sense to do so if there are actually spells in the counter. Otherwise, mages get that spellbook tab enabled without anything in it.

## Checklist


lynxlynxlynx:
LGTM, thanks!

Issue #169, opened 2019-08-19
Remove remaining MSVC6 hacks


lynxlynxlynx:
That sad piece of software plagued us for a long time, but now nobody is using it any more, even Avenger.

All the hacks that are specific to v6 can go! Just search the code, they're easy to find.

~This includes many separate "int i" definitions that are redundant and can be inlined into their respective loops. You can find most of the lines to check with: `git grep -E "^\s*(unsigned)?\s*int.*\<i\>"`. Other types didn't pop up in a cursory inspection.~

lynxlynxlynx:
Looks like it all worked, at least as far as building is concerned on appveyor. Closing, finally. :)

PR #683, opened 2020-04-13
Remove an old Win32 workaround that conflicts with cmath in MSVC2019

chilvence:
I am guessing it is ok to remove this completely as GemRB requires a much more recent C++ standard anyway.

### DIFF
--- a/gemrb/includes/win32def.h
+++ b/gemrb/includes/win32def.h
@@ -59,12 +59,6 @@
 
 #include "System/VFS.h"
 
-#ifdef _MSC_VER
-# ifndef round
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
One step closer to closing #169 as well, thanks!

chilvence:
Aha, I will keep that issue in mind if I spot anything else while snooping around.

I will probably stick to building with MSVC 2019 from now on, if you need someone to troubleshoot the windows builds. My only grudge against using visual studio historically was really the fact that you had to pay through the nose for it if you didn't want to pirate it, neither option I can stomach much. 

Even the commodore 64 came with its own dev kit, where along the line did computers cease being tools and start becoming consumer appliances... grr... anyway, that's just my personal axe to grind :)

PR #693, opened 2020-04-18
Windows build tweaks

chilvence:
I was having a bit of a headache trying to relearn to use git from a different approach via MSVC and github desktop. In the end I gave up and decided to just use the command line.

Only two conservative changes here, turns out only this and the one I PR'd previously were actually needed to stop either compilers aborting.

I am still figuring out what to do about the tweaks to cmake, something really strange is happening on my end with the vcpkg integration and I haven't worked out what it is yet. 

The only other thing I need to update after that is .gitignore, because MSVC puts a bunch of its own clutter in the source directory, Including a CmakeSettings.json that defaults to useless settings that I can't work out how to pre-emptively override....

### DIFF
--- a/gemrb/plugins/GUIScript/GUIScript.h
+++ b/gemrb/plugins/GUIScript/GUIScript.h
@@ -23,6 +23,11 @@
 
 // NOTE: Python.h has to be included first.
 
+// This definition needs to be here to keep mingw happy with standard python.org packages
+#if _MINGW32_
+#define hypot _hypot
+#endif
+
 #if defined(WIN32) && defined(_DEBUG)
 #undef _DEBUG
 #include <Python.h>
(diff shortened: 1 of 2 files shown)


gemrb/plugins/GUIScript/GUIScript.h
```diff
@@ -23,6 +23,11 @@
 
 // NOTE: Python.h has to be included first.
 
+// This definition needs to be here to keep mingw happy with standard python.org packages
+#if _MINGW32_
```
lynxlynxlynx, line 27:
```suggestion
#if __MINGW32__
```

chilvence: Whoops...

I'll fix that and try again.

Sometimes I do wonder if I should be allowed in the general vicinity of a computer

PR #697, opened 2020-04-20
Msvc tweaks: updated gitignore & most common compiler warning silenced

chilvence:
This is the pull request I should have started with. I am still trying to do basic housekeeping, but I can't seem to manage to do anything without github desktop tripping over visual studio or vice versa. At least github desktop doesn't create a load of its own junk though. This should cover ignoring all the local stuff that MSVC creates.

I'll split that mingw tweak to another branch for my own sanity. I thought of another scenario which I hadn't thought of before as well, so I need to double check it is ok. Never imagined one line in one file would end up being so annoying...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,8 @@ xcuserdata
 /cmake_uninstall.cmake
 /CMakeCache.txt
 /build
+
+#MSVC generates these
+/.vs
+/CMakeSettings.json
+/out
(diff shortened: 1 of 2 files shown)


bradallred:
I don't have a problem with adding that cmake file to the ignore list, but why is it being generated there instead of in the build directory?

chilvence:
It gets generated by the MSVC cmake integration as soon as you check out from git from the cmakelists in the top level directory (https://github.com/gemrb/gemrb/blob/master/CMakeLists.txt). This happens before you even decide to build or what parameters you want to pass it.

lynxlynxlynx:
Thanks!

Issue #684, opened 2020-04-13
evaluate c++11  <random>


lynxlynxlynx:
C++11 brought `<random>`, which also includes a bunch of statistical power. In particular, perhaps we can now deprecate our internal gemrb/core/RNG/ copy of SMFT, which stands for SIMD-oriented Fast Mersenne Twister (code v1.4.1 from http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/SFMT/index.html).

The candidate for replacement is:
https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

Originally when we switched from dumb random, we did a bunch chi tests to ensure a better result, but these two look like exactly the same algorithm, so I don't think so much caution is needed.
https://www.gibberlings3.net/forums/topic/26213-random-number-issues/?do=findComment&comment=220514

lynxlynxlynx:
We can still simplify things (get rid of our own CDF and ranged rand) by making them a wrapper for `std::uniform_int_distribution`. It's also worth verifying if the extra sanity checks need to stay.

It's all in:
https://github.com/gemrb/gemrb/blob/master/gemrb/core/RNG.cpp

PR #699, opened 2020-04-21
std::mt19937

MarcelHB:
## Description
See #684. You don't have a nice RNG test suite at hand, do you?

Anyway, I followed the comments closely and did not change anything related to the logic, just some tiny cleanup, added headers, kept the name neutral to the underlying RNG.

I don't know what needs to be done for XCode files exactly. BG1 attribute dice game feels like it's supposed to feel. :slightly_smiling_face: 

## Checklist

- [x I extended the documentation, if necessary


gemrb/core/RNG.cpp
```diff
@@ -0,0 +1,185 @@
+/* GemRB - Infinity Engine Emulator
+ * Copyright (C) 2004 The GemRB Project
```
lynxlynxlynx, line 2:
```suggestion
 * Copyright (C) 2020 The GemRB Project

```

gemrb/core/RNG.cpp
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// For complete fairness and equal timing, this should be a roll, but let's skip it anyway
+	if(min == max)
+		return max;
+
+	// Someone wants rand() % -foo, so we compute -rand(0, +foo)
+	// This is the only time where min > max is (sort of) legal.
+	// Not handling this will cause the application to crash.
+	if(min == 0 && max < 0) {
```
lynxlynxlynx, line 87: I know this is from the old file, but please clean up the spaces while you're at it. They're missing after the ifs.

gemrb/core/RNG.cpp
```diff
@@ -0,0 +1,185 @@
...
+	// If the division has a remainder, the result is floored automatically.
+	const uint64_t buckets = std::mt19937_64::max() / diameter;
+
+	// Compute the last valid random number. All numbers beyond have to be
+	// ignored.
+	// If there was no remainder in the previous step, limit is equal to
+	// UINT64_MAX.
+	const uint64_t limit = diameter * buckets;
+
+	uint64_t rand;
```
lynxlynxlynx, line 162: Please move it after the big comment, so it doesn't stand out so much.

gemrb/core/GameScript/GameScript.cpp
```diff
@@ -2061,7 +2061,7 @@ bool GameScript::Update(bool *continuing, bool *done)
 	bool continueExecution = false;
 	if (continuing) continueExecution = *continuing;
 
-	RandomNumValue=RNG_SFMT::getInstance()->rand();
+	RandomNumValue=RAND_DEFAULT_MIN_MAX();
```
lynxlynxlynx, line 2064: Surely we can find a nicer name. I know macros don't support overloading by default and let's not add hacks, but how about `RAND_FULL()`, `RAND_ALL()` or something along those lines? 

And you can add spaces around the equality sign at the same time. :)

gemrb/core/RNG.h
```diff
@@ -0,0 +1,46 @@
+/* GemRB - Infinity Engine Emulator
+ * Copyright (C) 2004 The GemRB Project
```
lynxlynxlynx, line 2: Same here.

lynxlynxlynx:
Thanks and on a general note, does `<random>` not have anything that would obviate our need to manually implement the cdf or a bound `rand()`?

gemrb/core/RNG.h
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ *
+ */
+
+#ifndef RNG_H
+#define RNG_H
+
+#include <random>
```
lynxlynxlynx, line 24: bots are complaining about this:
> error: use of undeclared       identifier 'INT_MAX'

so I guess limits is missing.

gemrb/core/RNG.cpp
MarcelHB: `Game.cpp` :wink: I'll touch them both then.

gemrb/core/GameScript/GameScript.cpp
MarcelHB: Was also not sure about that, I'll take the `ALL` then.

MarcelHB:
I guess the one who implemented the previous version truly spend some statistical thinking in there. From what I see we can maybe look at `std::uniform_int_distribution` but then I'm not statistics guy enough to judge that. So this is more like the get-rid-of-the-lib approach.

lynxlynxlynx:
If we can trust the header with the engine, I'm sure we can trust it with that too. But no problem in doing the change in two steps to get some more eyes and dwell time.

bradallred:
the one issue i see is you are using `std::mt19937_64` to generate 32 bit numbers. If you want 32 bit random numbers use the `std::mt19937` version.

MarcelHB:
When you look at the previous code, you find things like `sfmt_genrand_uint64` and a special note to take care of `UINT64_MAX`. That's why 64bit version.

I have a very vague memory of sampling from a much higher resolution than what you actually need for security-relevant contexts, so to me, this looks good without further research.

bradallred:
fair enough :)

lynxlynxlynx:
Will give it a spin first.

lynxlynxlynx:
Generated 5*10^6d100 from both and basic stats were fine. Couldn't get the chitest in localc to do anything, but the F test as a silly proxy looked fine too.

PR #698, opened 2020-04-21
Feature: Extract GUIScript doc-strings through Python interpreter.

kindaro:
## Description

Solves https://github.com/gemrb/gemrb.github.io/issues/9.

There is a build target added to CMake that builds an executable `run_script` with the obvious function. It initializes as little of the game as was needed for the goal at hand.

#### Directions for further study:

* How many of the GUIScript functions this interpreter is initialized enough to actually run?
* A small nuisance is that the file `include.py` may not be found, resulting in a non-critical error. Possibly we may want to make the importing of this file optional.
* Recent Pandoc — [at least version 2.6][changelog] — is required to convert from the DokuWiki format.
   + I am not sure why it escapes apostrophes, but it does not affect the HTML output.
* I am going to squash and rebase when this branch is ready to be merged.

[changelog]: https://github.com/jgm/pandoc/blob/master/changelog.md#pandoc-26-2019-01-30

## Checklist

- [ ] Commit messages are descriptive and explain the rationale for changes
- [ ] I used the same coding style as the surrounding code 
- [ ] I have tested the proposed changes
- [ ] I extended the documentation, if necessary

PR #701, opened 2020-04-22
Add some editor and formatter configurations.

kindaro:
* Add `.dir-locals.el` — sets up some Emacs variables to make sure C++
  and Python code is consistently indented with tabs.

* Add `.clang-format` — makes the `clang-format` command indent with
  tabs.

* Add `.pep8` — makes the `autopep8` command ignore indentation.

Unfortunately, it is not possible to automatically indent Python code
with tabs, since that is contradictory to the widely accepted coding
standard.

## Description

It has been mentioned to me elsewhere _(https://github.com/gemrb/gemrb/pull/698#issuecomment-617372408)_ that my coding style is not suitable for the project. Since _«coding style»_ is such an immaterial term, I would like to have a hard standard to compare against, and preferably one that does all the work, too.

Use the facilities offered by this commit by running either `clang-format` or `autopep8` over your C++ and Python code, respectively. The configuration files should be read by these tools automatically.

Please review the formatting offered by the tools and let me know if something needs adjustment. Even though the range of possible adjustments is not unlimited, it is quite wide, especially for the `clang-format` tool.

This may solve the long-standing #161 issue, or at least move things that way.

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,137 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortLambdasOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: MultiLine
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
The machine readable standard needs to be decided and developed in #161. Until then, the current style guidelines hold. Partial solutions for tools we don't know we'll be even using are not very helpful.

Btw, is there a way to put this junk into a folder instead, so it doesn't pollute the toplevel dir?

The clang file is long — are you sure it just enforces indentation?

kindaro:
> The machine readable standard needs to be decided and developed in #161. Until then, the current style guidelines hold. Partial solutions for tools we don't know we'll be even using are not very helpful.

It is up to you to decide and develop a standard in the future. I am offering something that can be used now — certainly more helpful than a non-existent perfect standard. Moreover, these tools are each an industry standard, you are unlikely to find something better. So, I am not sure I can perceive your judgement as well-founded. Is there something specific that you dislike about the style offered by the present configurations?

> Btw, is there a way to put this junk into a folder instead, so it doesn't pollute the toplevel dir?

No easy way. Tools such as these only search upwards in the directory tree. That being said, the files are dotted, so they are invisible on Unix by default, and the cluttering problem is thus largely eliminated.

> The clang file is long — are you sure it just enforces indentation?

It is the standard configuration file, minimally adjusted to select the preference of tabs over spaces. So, it sets everything else to default. That is to say, it enforces consistent style throughout.

lynxlynxlynx:
Ok, so then we can't merge this, since anyone using it would easily get unrelated line changes creeping into their diffs. We've had this problem before, so we can't enforce it until a minimally damaging style is defined. Which is the point of the bug.

Thanks anyway!

kindaro:
I see. I was thinking of using this setup to validate new code, but I can see that most work is done on existing files.

It can be advised to run the code base through the formatting tools once. There are some egregious style features currently present in the code. For instance, despite [the axioms of style][axioms], there are very many operators without surrounding whitespace, and even some with asymmetric whitespace. There are many, many other unnecessary idiosyncrasies throughout the code base. It is hard to find a rule that is not violated. There is even a situation where sorting `#include` directives alphabetically breaks the build.

[axioms]: https://github.com/gemrb/gemrb/blob/master/CONTRIBUTING.md#axioms-of-style

At the same time, even though the patch is rather lengthy, it breaks only 2 places in the C++ code, and after fixing them the engine seems to run. So it might not be very expensive to transition.

lynxlynxlynx:
We don't like to pollute the history with needless changes (mistakes of course happen), since we need to inspect it often and mass style fixes for the sake of it really mess with git praise. So the sync will be a slow and ongoing process. First step after we have a defined format is to force it for PRs, the second to make a pre-commit hook.

PR #706, opened 2020-04-25
Retarget the gemrb core dll directory to the same as the excutable fo…

chilvence:
This is my suggestion to solve the problem that windows builds currently place the gemrb core library where the exe can't find it.

I find it a tiny bit more preferable to manually or automatically copying this dll to the game directory, even if the end result is basically the same. 

My brain has essentially melted down trying to reconcile the fact that MinGW is a kind of Unix on windows but still needs to know where its DLL files are, but I have tested this and I think it works on MinGW as well (@MarcelHB, I would be grateful if you would confirm or deny this)

I have been reluctant to make a pull request for this because I am completely green when it comes to working with cmake, but I guess if this were to have unintended consequences, it would be easy enough to revert in any case.


MarcelHB:
Works for me :+1:

lynxlynxlynx:
Thanks! I'll check the newbuilt binary, forgot we don't save them for PRs.

PR #725, opened 2020-05-06
IWD: Don't open racial enemy selection window in chargen for druids

fenuks:
## Description
As @lynxlynxlynx pointed out, druids shouldn't be able to select racial enemy.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/iwd/CharGen.py
+++ b/gemrb/GUIScripts/iwd/CharGen.py
@@ -1524,17 +1524,21 @@ def SkillsPress():
 	LUSkillsSelection.SkillsNullify (MyChar)
 
 	if SkillsState == 0:
-		GemRB.SetVar ("HatedRace", 0)
-		if IsThief!="*":
-			SkillsSelect()
-		elif DruidSpell!="*":
+		if DruidSpell != "*":
 			Skill = GemRB.LoadTable("SKILLRNG").GetValue(str(Level), "STEALTH")
 			GemRB.SetPlayerStat (MyChar, IE_STEALTH, Skill)
-			RacialEnemySelect()
-		elif IsBard!="*":
+		elif IsBard != "*":
 			Skill = GemRB.LoadTable(IsBard).GetValue(str(Level), "PICK_POCKETS")
 			GemRB.SetPlayerStat (MyChar, IE_PICKPOCKET, Skill)
-			SkillsState = 1
+
+		GemRB.SetVar ("HatedRace", 0)
+		GemRB.SetVar ("RacialEnemy", 0)
+		GemRB.SetVar ("RacialEnemyIndex", 0)
+		HateRace = CommonTables.ClassSkills.GetValue(ClassName, "HATERACE")
+		if HateRace != "*":
+			RacialEnemySelect()
+		elif IsThief != "*":
+			SkillsSelect()
 		else:
 			SkillsState = 1
 


gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -1535,7 +1535,10 @@ def SkillsPress():
 		elif DruidSpell!="*":
 			Skill = GemRB.LoadTable("SKILLRNG").GetValue(str(Level), "STEALTH")
 			GemRB.SetPlayerStat (MyChar, IE_STEALTH, Skill)
-			RacialEnemySelect()
+			if ClassName == "RANGER":
+				RacialEnemySelect()
+			else:
+			    SkillsSelect()
```
lynxlynxlynx, line 1541: Come on, look at your diffs. A better way to check who should get it is by checking if the HATERACE column of clsskills.2da (CommonTables.ClassSkills) has an entry.

bradallred: in addition to what you said, this probably shouldn't be under a check for `DruidSpell` either. might technically cover all built in classes, but I assume mods can add this to anything they want.

fenuks: I always look at diff before committing, but as it turns out, diff-so-fancy that I use to make it more readable, shows tabs as four spaces, that's why I couldn't see it. Changed shiftwidth to 8 spaces, so it will be now readable in editor. If I may express my opinion, then let me say that python code style and formatting is rather peculiar. It would be easier to maintain consistent style if code formatter was being used.

> A better way to check who should get it is by checking if the HATERACE column of clsskills.2da (CommonTables.ClassSkills) has an entry.

I see. Is there more comprehensive description of game files than IESDP? It has some useful knowledge, but it's not describing, for example, `clsskills.2da` and its fields.

lynxlynxlynx: Nothing in gemrb/unhardcoded/ is on iesdp, since they are all our files. There is gemrb/docs/en/Tables/, but we'll eventually move them inside the table files themselves, since we otherwise don't remember to update the docs ever.

fenuks: > in addition to what you said, this probably shouldn't be under a check for DruidSpell either. might technically cover all built in classes, but I assume mods can add this to anything they want.

That's good idea I guess, but will complicate checks as race hate opens later skills select, so we will have to ensure it is not opened twice.

lynxlynxlynx: I wouldn't bother with this at all. We rely on only rangers having this table in core as well. It's the druid that doesn't get any special skills, so you can just replace the `DruidSpell!="*"` check.

fenuks:
```diff
diff --git a/gemrb/GUIScripts/iwd/CharGen.py b/gemrb/GUIScripts/iwd/CharGen.py
index 9962d799e1..fe6bf4c543 100644
--- a/gemrb/GUIScripts/iwd/CharGen.py
+++ b/gemrb/GUIScripts/iwd/CharGen.py
@@ -1526,6 +1526,7 @@ def SkillsPress():
 	MageSpell = CommonTables.ClassSkills.GetValue (ClassName, "MAGESPELL")
 	IsBard = CommonTables.ClassSkills.GetValue (ClassName, "BARDSKILL")
 	IsThief = CommonTables.ClassSkills.GetValue (ClassName, "THIEFSKILL")
+	HateRace = CommonTables.ClassSkills.GetValue(ClassName, "HATERACE")
 	LUSkillsSelection.SkillsNullify (MyChar)
 
 	if SkillsState == 0:
@@ -1535,10 +1536,6 @@ def SkillsPress():
 		elif DruidSpell!="*":
 			Skill = GemRB.LoadTable("SKILLRNG").GetValue(str(Level), "STEALTH")
 			GemRB.SetPlayerStat (MyChar, IE_STEALTH, Skill)
-			if ClassName == "RANGER":
-				RacialEnemySelect()
-			else:
-			    SkillsSelect()
 		elif IsBard!="*":
 			Skill = GemRB.LoadTable(IsBard).GetValue(str(Level), "PICK_POCKETS")
 			GemRB.SetPlayerStat (MyChar, IE_PICKPOCKET, Skill)
@@ -1546,6 +1543,12 @@ def SkillsPress():
 		else:
 			SkillsState = 1
 
+		if HateRace != "*":
+			RacialEnemySelect()
+		elif IsThief!="*" or DruidSpell!="*":
+			SkillsSelect()
+
+
 	if SkillsState == 1:
 		ProficienciesSelect()
 ```

Something like this seems to work, but give me more time to test it further.

Comment (lynxlynxlynx) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: Is there some other reason druids need a call to SkillsSelect?

Comment (fenuks) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: Other reason? I don't think I'm on par with you on this, it's just needed distribute skill points. `RacialEnemySelect()` implies calling `SkillsSelect()` afterwards. If you remove `or DruidSpell!="*", then simply skills window doesn't open anymore for druids.

Comment (lynxlynxlynx) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: Yeah, but what skills do they have that they need the window? None that I'm aware ...

Comment (fenuks) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: Oh, I mistook skills for proficiencies. So, to rephrase myself, proficiencies window doesn't open. Perhaps `SkillSelect()` sets `SkillsState = 1` or something like that. I'll check that in the evening.

Comment (lynxlynxlynx) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: Just look at the whole if-block from the diff — if nothing matches SkillsState will be set to 1 and proficiencies immediately opened.

**Comment by lynxlynxlynx** [comment]
Until we get #161 sorted out, I at least now tried enforcing good whitespace in python. Please try merging master in your branch and pushing — I want to test if the plain spaces will be detected.

File: gemrb/GUIScripts/iwd/CharGen.py (hunk 1 above)
Comment (fenuks) [reply gemrb/GUIScripts/iwd/CharGen.py:1541]: I've seen that, that's why I thought it might be solution, but it could be that `SkillsSet()` manipulates global state, and sets something druid might need. After quick check, it appears that setting SkillsState is sufficient.

**Comment by fenuks** [comment]
Coding style would be welcome improvement. I see it is planned before new release, so fingers crossed.

Rebased this branch against master as requested… and it builds.

**Comment by lynxlynxlynx** [comment]
Thanks and yes, I apparently didn't set it up correctly yet.

**Comment by fenuks** [comment]
Command is `pycodestyle --select=E1,E201,E202,E203,E225 --diff gemrb/GUIScripts`, and --diff expects unified diff on stdin. `git diff -u master | pycodestyle2 --select=E1,E201,E202,E203,E225 --diff` works for me locally, and I get:

```
./gemrb/GUIScripts/iwd/CharGen.py:1535:18: E225 missing whitespace around operator
./gemrb/GUIScripts/iwd/CharGen.py:1536:4: E117 over-indented
./gemrb/GUIScripts/iwd/CharGen.py:1539:5: E117 over-indented
./gemrb/GUIScripts/iwd/CharGen.py:1541:4: E101 indentation contains mixed spaces and tabs
./gemrb/GUIScripts/iwd/CharGen.py:1542:14: E225 missing whitespace around operator
./gemrb/GUIScripts/iwd/CharGen.py:1543:4: E117 over-indented
```

**Comment by lynxlynxlynx** [comment]
Yeah, the challenge is to get correct invocation without needing to write our own gh action.

**Comment by fenuks** [comment]
Could you please try to specify args as `--select=E1,E201,E202,E203,E225 --diff < <(git diff -u master)`?

**Comment by lynxlynxlynx** [comment]
I was just about to try with a "run" setup, since that avoids the hassle with branches. Let's see.

**Comment by lynxlynxlynx** [comment]
Unlike travis, didn't find any commit range vars, so in the end I limited it to master again. Will try a fake PR, no reason to continue here.

**Comment by lynxlynxlynx** [comment]
Phew, finally nailed that down.

**Comment by fenuks** [comment]
Glad to hear you got machinery working. I will update review the other day to test it in practise. :)

File: gemrb/GUIScripts/iwd/CharGen.py [hunk 2]
```diff
@@ -1524,17 +1524,21 @@ def SkillsPress():
 	LUSkillsSelection.SkillsNullify (MyChar)
 
 	if SkillsState == 0:
-		GemRB.SetVar ("HatedRace", 0)
-		if IsThief!="*":
-			SkillsSelect()
-		elif DruidSpell!="*":
+		if DruidSpell!="*":
```
lynxlynxlynx, line 1527: Aren't you still giving druids stealth this way? 

Besides the reported spacing issues also keep in mind the space before an opening paren.

gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -1524,17 +1524,21 @@ def SkillsPress():
...
 			GemRB.SetPlayerStat (MyChar, IE_STEALTH, Skill)
-			RacialEnemySelect()
 		elif IsBard!="*":
 			Skill = GemRB.LoadTable(IsBard).GetValue(str(Level), "PICK_POCKETS")
 			GemRB.SetPlayerStat (MyChar, IE_PICKPOCKET, Skill)
-			SkillsState = 1
+
+		GemRB.SetVar ("HatedRace", 0)
+		GemRB.SetVar ("RacialEnemy", 0)
+		GemRB.SetVar ("RacialEnemyIndex", 0)
```
lynxlynxlynx, line 1536: Are these two used later as well? If so, sounds like a refactoring waiting to happen. Either way we should just save and reset the stat directly in RacialEnemySelect and kids.

lynxlynxlynx:
Does this two-line change really not cut it?
```
		GemRB.SetVar ("HatedRace", 0)
		HateRace = CommonTables.ClassSkills.GetValue(ClassName, "HATERACE") ######
		if IsThief!="*":
			SkillsSelect()
		elif HateRace != "*": ########
			Skill = GemRB.LoadTable("SKILLRNG").GetValue(str(Level), "STEALTH")
			GemRB.SetPlayerStat (MyChar, IE_STEALTH, Skill)
			RacialEnemySelect()
		elif IsBard!="*":
```

gemrb/GUIScripts/iwd/CharGen.py
fenuks: They are used in to display racial enemy in character description screen. Without resetting these, selecting ranger, choosing racial enemy, going back, and say, changing class to druid will print later race enemy.

> Either way we should just save and reset the stat directly in RacialEnemySelect and kids.

It has to be reset before it, because it's not guaranteed that `RacialEnemySelect` will be called.

lynxlynxlynx: With direct storage, the description code can check against the stat as well. But a refactoring is best done as a separate PR.

gemrb/GUIScripts/iwd/CharGen.py
fenuks: Yes, it's the same condition as it was previously. Thief if was moved down below. If mod adds RacialEnemy for thief, RacialEnemySelect will be triggered, which in turn will trigger later SkillsSelect.

fenuks: > Besides the reported spacing issues also keep in mind the space before an opening paren.

Should all method/function calls have space before opening parenthesis? Some do, some not.

fenuks:
> Does this two-line change really not cut it?

By quick look, it could, but my version was meant to be in line with Brad's comment. If you think that would be better, I can test it and amend my commit.

gemrb/GUIScripts/iwd/CharGen.py
lynxlynxlynx: Yes, it's one of the oldest and wierdest style decisions in the project.

I forgot this turned into an enhancement request as well, not just about fixing the initial bug.

fenuks:
I also noticed minor disagreement between original and GemRB. In original, skill points are first distributed, then player is prompted for racial enemy. In GemRB reverse is the case. Changing that would increase diff considerably, though.

lynxlynxlynx:
Like I said, one thing at a time. :)

lynxlynxlynx:
Ok, so only the linter still needs to be appeased and the rest belongs to new PRs/reports.

fenuks:
To clarify things: would you like me to go with version you proposed, or with the current one, which is less straightforward, but potentially more flexible/mod friendly?

lynxlynxlynx:
It's fine, just fix the linter errors. The druid stealth fix or any refactoring belongs into new PRs.

lynxlynxlynx:
Thanks and sorry for the delay — I didn't notice the last update.

fenuks:
That's no problem, thanks. I'm not always the swiftest to reply/include your suggestions.

lynxlynxlynx:
Btw, I have the druid stealth fixed, but noticed the setting doesn't survive chargen, so even rangers get a 0 base value in game. Still investigating that (setting works).

fenuks:
That's interesting. I'll try to push forward this`QueryText` thing.

PR #733, opened 2020-05-17
Animation: prevents first frame of new loop to remain for too long

### LINKED ISSUE #732: Background animations have lags
#### Bug description

There is something pedantic about map animations that grinds my gears a bit: they are not running smoothly. A good example is Thalantyr's home of BG1 (see below), but I think I can also notice that in another example in PST.

The lightning (and the crystal) are spinning around non-stop, but they shortly stop when the lightning is close to the ᛚ rune, close to Thalantyr.

GemRB version: master

#### Steps to reproduce
1. Start BG1.
2. Go to AR3202.
3. See the crystal and lightning.

#### Expected behavior
The animation should go without any noticable seam between a loop. Can't take a LICEcap GIF on the original game since it stops on losing the window focus.

#### Screenshots

store

MarcelHB:
fixes #732

So when an animation loop overflows, `starttime` was reset to 0. From what I see, `starttime` is used to determine the timestamp of the last update of the frame counter. But when it is set to 0, this code https://github.com/gemrb/gemrb/blob/4831717a78b1e97cf1823138824b7494619143ed/gemrb/core/Animation.cpp#L133-L139
in the next frame request will make it look like "just now". So we have a whole cycle of pos 0 in addition as this line won't do anything then:

https://github.com/gemrb/gemrb/blob/4831717a78b1e97cf1823138824b7494619143ed/gemrb/core/Animation.cpp#L158

## Checklist

- [ ] The proposed change builds also on our build bots (check after submission) :shrug:

### DIFF
--- a/gemrb/core/Animation.cpp
+++ b/gemrb/core/Animation.cpp
@@ -173,7 +173,6 @@ Sprite2D* Animation::NextFrame(void)
 			pos = 0;
 			endReached = true;
 		}
-		starttime = 0;
 	}
 	return ret;
 }


MarcelHB:
Now:
store_fix

lynxlynxlynx:
Thanks!

PR #738, opened 2020-05-19
WIP: Buffered file reading (BIFImporter)

MarcelHB:
Here is something I'd like to discuss: On my older computer with an HDD, I find loading some bigger savegames painfully slow. So I started profiling around loading the game:

profiling

As you see, lot of time is locked by `ReadDword` and `ReadWord` over BIF files. Looking at the `FileStream` stuff, it does all these bare metal 2/4-bytes read requests -- that's not so much of a performance win.

I then tried to put this above of a simple page-like buffered reader. This cuts down `GAMImporter::LoadGame()` by roughly **50%** of time for me (about 1.6s vs 3.2s).

The code is prototype for now, so RFC, more candidates ...

### DIFF
--- a/gemrb/core/System/BufferedFileInputStream.h
+++ b/gemrb/core/System/BufferedFileInputStream.h
@@ -1,0 +1,148 @@
+#ifndef BUFFERED_FILE_INPUT_STREAM_H
+#define BUFFERED_FILE_INPUT_STREAM_H
+
+#include <algorithm>
+#include <array>
+#include <fstream>
+
+#include "System/DataStream.h"
+#include "System/VFS.h"
+
+#define DEFAULT_READ_BUFFER_SIZE 4096
+
+namespace GemRB {
+
+template<size_t BUFFER_SIZE = DEFAULT_READ_BUFFER_SIZE>
+class BufferedFileInputStream : public DataStream {
+  public:
+    BufferedFileInputStream(const std::string& fileName);
+    ~BufferedFileInputStream() override {};
+
+    bool isOk() const { return ifstream.is_open(); }
+    DataStream* Clone() override {
+      return new BufferedFileInputStream{originalfile};
+    }
+    int Read(void *dest, unsigned int len) override;
+    int Seek(int pos, int startPos) override;
+    int Write(const void*, unsigned int) override {
+      return GEM_ERROR;
+    };
+  private:
+    std::array<char, BUFFER_SIZE> readBuffer;
+    uint64_t bufferPos;
+    size_t bufferLength;
+    bool bufferLoaded;
+
+    std::ifstream ifstream;
+
+    void readToBuffer();
+};
+
+template<size_t BS>
+BufferedFileInputStream<BS>::BufferedFileInputStream(const std::string& fileName)
+  : DataStream(),
+    bufferPos(0),
+    bufferLength(0),
+    bufferLoaded(false)
+{
+  ifstream.open(fileName, std::ios_base::in | std::ios_base::binary);
+
+  if (isOk()) {
+    ifstream.seekg(0, std::ios_base::end);
+    this->size = ifstream.tellg();
+    ifstream.seekg(0, std::ios_base::beg);
+
+	  strlcpy(originalfile, fileName.c_str(), _MAX_PATH);
+  }
+}
+
+template<size_t BS>
+int BufferedFileInputStream<BS>::Read(void *dest, unsigned int length) {
+  if (!isOk()) {
+    return GEM_ERROR;
+  }
+
+  if (Pos + length > size) {
+    return GEM_ERROR;
+  }
+
+  if (length == 0) {
+    return 0;
+  }
+
+  if (!bufferLoaded || Pos < bufferPos || Pos >= bufferPos + bufferLength) {
+    readToBuffer();
+  }
+
+  size_t bytesToRead = length;
+  size_t copyOffset = 0, readOffset = Pos - bufferPos;
+  do {
+    size_t chunkSize = std::min(bytesToRead, bufferLength - readOffset);
+
+    std::copy(
+        std::begin(readBuffer) + readOffset,
+        std::begin(readBuffer) + readOffset + chunkSize,
+        reinterpret_cast<char*>(dest) + copyOffset
+    );
+
+    bytesToRead -= chunkSize;
+    copyOffset += chunkSize;
+    readOffset = 0;
+    this->Pos += chunkSize;
+
+    if (bytesToRead > 0) {
+      readToBuffer();
+    }
+  } while(bytesToRead > 0);
+
+  if (copyOffset != length) {
+    return GEM_ERROR;
+  }
+
+  if (Encrypted) {
+    ReadDecrypted(dest, copyOffset);
+  }
+
+  return copyOffset;
+}
+
+template<size_t BS>
+void BufferedFileInputStream<BS>::readToBuffer() {
+  ifstream.seekg(Pos, std::ios_base::beg);
+  ifstream.read(readBuffer.data(), BS);
+  this->bufferPos = Pos;
+  this->bufferLength = ifstream.gcount();
+  this->bufferLoaded = true;
+}
+
+template<size_t BS>
+int BufferedFileInputStream<BS>::Seek(int newPosition, int seekMode) {
+  if (!isOk()) {
+    return GEM_ERROR;
+  }
+
+  switch (seekMode) {
+    case GEM_STREAM_END:
+      this->Pos = size - newPosition;
+      break;
+    case GEM_CURRENT_POS:
+      this->Pos += newPosition;
+      break;
+    case GEM_STREAM_START:
+      this->Pos = newPosition;
+      break;
+    default:
+      return GEM_ERROR;
+  }
+
+  if (Pos >= size) {
+    Log(ERROR, "BufferedFileInputStream", "Invalid seek position %ld (limit: %ld)", Pos, size);
+    return GEM_ERROR;
+  }
+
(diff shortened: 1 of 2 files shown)


bradallred:
I assume you are exaggerating the "bare metal" claim? The Kernel should be handling buffering at the device level. The speed increase you see is likely due to fewer system calls/copying between buffers rather than lack of buffering, but it's not important.

Now about the code. It seems silly to not just change the existing `FileStream`; we don't need 2 of the exact same tools. This seems like it ought to be a solid improvement everywhere.

I'm not sure a template is the best approach, however. If you wanted a variable buffer size I would presume the optimal choice is some multiple of the device block size which won't be known at compile time. It's probably not worth any concern either, so why not just `constexpr` it and move it all into `FileStream`?

If Windows only had `mmap` that is by far the fastest way to read a file that I know of and would avoid all the read/write calls :)

All this makes me wonder why the C++ `fstream` wasn't used from the get go. Too much effort for too little gained to change it now, but a bit puzzling.

lynxlynxlynx:
Would it make sense to use our MemoryStream to store the buffer instead? Seems like it cut some duplication.

bradallred:
`MemoryStream` by itself would work fantastically if we could use `mmap`... I don't know if there is a Windows equivalent and if so no idea how much effort it would be to use it.

MarcelHB:
> I assume you are exaggerating the "bare metal" claim? The Kernel should be handling buffering at the device level. The speed increase you see is likely due to fewer system calls/copying between buffers rather than lack of buffering, but it's not important.

Maybe, I'm not sure about the details on Windows. Anyway I think I can hear less drive sounds while loading. :wink: 

> Now about the code. It seems silly to not just change the existing `FileStream`; we don't need 2 of the exact same tools. This seems like it ought to be a solid improvement everywhere.

Generally yes, it was just easier to work with for testing.

> I'm not sure a template is the best approach, however. If you wanted a variable buffer size I would presume the optimal choice is some multiple of the device block size which won't be known at compile time. It's probably not worth any concern either, so why not just `constexpr` it and move it all into `FileStream`?

I think 4 kB is somewhere around the 80% coverage of use cases globally -- not sure if alignment gets win in addition there -- I did not see any improvement with larger buffers, but then maybe IE has some formats around that maybe appreciate some tuning? So yes, probably no big need for templatization.

> If Windows only had `mmap` that is by far the fastest way to read a file that I know of and would avoid all the read/write calls :)

There is [something](https://docs.microsoft.com/en-us/windows/win32/memory/file-mapping) but I have never used it myself so far.

> Would it make sense to use our MemoryStream to store the buffer instead? Seems like it cut some duplication.

I also thought of that but then we still need something that moves data in there, any maybe not strictly bound to FS I/O.

bradallred:
> we still need something that moves data in there

No, you don't; that is what mmap (and the Windows equivalent you linked) do. After `mmap`ping  you will have a `char*` or `void*` that you can pass along to `MemoryStream` and read/write to it just like regular memory.

MarcelHB:
Sure, but I guess you want the nmap or Win-alike thing not to be placed in MemoryStream? This probably requires platform-dependent handling of paths, page permissions, flags.

bradallred:
> Sure, but I guess you want the nmap or Win-alike thing not to be placed in MemoryStream? This probably requires platform-dependent handling of paths, page permissions, flags.

I have no idea what to do for Windows, but on everything else it is just 2-4 lines of code:
```
int fd = open("file", O_RDONLY);
struct stat sb;
fstat(fd, &sb);
auto memblock = mmap(NULL, sb.st_size, PROT_WRITE, MAP_PRIVATE, fd, 0);
```

You could just write a factory function in `MemoryStream` that takes a file path and returns a `MemoryStream` object from that.

I thought we already had VFS stuff to deal with paths. If we have a `FILE*` you can get a file descriptor from that instead. I have no idea what "flags", or "page permission" stuff you think you would need.

MarcelHB:
From my experience it's probably just 50% more things to do specify when working with Win API. Just like you don't simply [execve](https://linux.die.net/man/2/execve), you [CreateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa).

bradallred:
It's actually [about the same](https://courses.washington.edu/hypertxt/cwb/cl/windows-mmap.c).

MarcelHB:
Cool, then it's the same estimation. So, corollary:

- We agree on tuning the FS reading for now, writing is not so much of focus but may profit as a side effect.
- We make use of (Win)mmap somewhere (I say _somewhere_ because e. g. BIFImporter currently may lead to some chain of `FileStream` -> `SlicedStream` -> `MemoryStream`) -- is there any supported platform that needs fallback?
- Do we do this generally for every file (type) to read? Is there a valid reason not to do this under some circumstances?

bradallred:
> We agree on tuning the FS reading for now, writing is not so much of focus but may profit as a side effect

that's probably what we are limited to do since `mmap` cannot grow or shrink easily. One of the reasons why mmap doesn't largely replace a lot of read/write code in practice. The other main reason being portability to Windows.

> is there any supported platform that needs fallback?

I think our "supported platforms" are essentially what we (sometimes vaguely) know how to support (Windows, Mac/iOS, Android, Linux/Unix) after that its up to other contributors to supply patches. `mmap` is part of POSIX and is available on everything I listed other than Windows.

> Do we do this generally for every file (type) to read? Is there a valid reason not to do this under some circumstances?

Technically `read()` can be faster sometimes, but it's not worth figuring out. It largely depends on the platform and underlying implementation anyway. The rule of thumb (I say from experience from a purely POSIX perspective) is if your files are small enough you can just map the entire file into memory at once `mmap` should be faster. This is doubly true if you are doing random access.

I guess the question is how large are BIF files generally? If they are large we may not want to map the entire file to memory at once. I'm not actually sure off the top of my head if there is a POSIX flag for controlling that anyway so just don't worry about it I guess.

It should be fine to do this with any file, but BIF seems like a logical place to start. You can measure before applying it to other places if there is doubt.

MarcelHB:
> I guess the question is how large are BIF files generally? If they are large we may not want to map the entire file to memory at once. I'm not actually sure off the top of my head if there is a POSIX flag for controlling that anyway so just don't worry about it I guess.

That's my concern as well: The original BIFs that I found are up to 150MB of size, sometimes with just a few, even just one large movie file. You probably don't want to have too many large files mapped at the same time. For 32bit environments this sounds potentially critical.

Maybe it makes sense to restrict that by file size, or to certain parts, like the entry list of a BIF.

bradallred:
> > I guess the question is how large are BIF files generally? If they are large we may not want to map the entire file to memory at once. I'm not actually sure off the top of my head if there is a POSIX flag for controlling that anyway so just don't worry about it I guess.
> 
> That's my concern as well: The original BIFs that I found are up to 150MB of size, sometimes with just a few, even just one large movie file. You probably don't want to have too many large files mapped at the same time. For 32bit environments this sounds potentially critical.
> 
> Maybe it makes sense to restrict that by file size, or to certain parts, like the entry list of a BIF.

No, just let the OS deal with paging the memory. If you don't forcibly pass flags to immediately map the entire file it's not a problem. Forget I mentioned mapping entire files as an option. Our real savings will come from elimination of syscall overhead and copying between kernel and user space buffers. No need to special case anything.

MarcelHB:
Do I get this schema then? 

1. have a known file type with lots of small items to read from, e. g. a BIF file
2. do something like `FileStream::to_mmaped_stream` by the hopefully determinable offset and size, e. g. offset + num of entries * 16 for the BIF file entry list.
3. instead of tons of `fread` or `ReadFile`, the `::Read` maps that to memory accesses, e. g. by using the `MemoryStream`.
4. when done, unmap

This what sounds like would eliminate most of that without keeping anything for too long.

lynxlynxlynx:
Concerning file size cut-offs, MemoryStream uses a quadruple of what is proposed in this PR. PST, at least in some distributions, has plenty of tiny BIFFs.

bradallred:
It seems to me the easiest thing to do would be:

1. create a subclass to `MemoryStream` that takes a path in the constructor and does the `mmap` there and does the `munmap` in the destructor
2. in `BIFImporter` use this new class instead of `FileStream`.

MarcelHB:
Ok, I'll play around on both platforms and see then.

bradallred:
`mmap`ing  should perform at least as well as the 4kb cache. The OS will map the file one page (4kb on x86) at a time so the buffer size is the same more or less and you also eliminate all the many many `read` syscalls in exchange for only the upfront cost of the `mmap`. You also get cheap random access (no idea if we read non sequentially for BIF files) for free. Should be a big win for large files with lots of tiny `read()`s.

bradallred:
BTW, File IO buffering can be controlled via `setvbuf` so that's another enhancement that could be made. You can also get the block size with `stats.st_blksize`.

MarcelHB:
Thanks, I never noticed that anywhere so far :thinking: Is there a simple `std::istream` equivalent on C++? Java has `BufferedInputStream` that just wraps any other stream and you no more care as it adheres to the `InputStream` interface. `std::basic_filebuf` isn't something I understood by three minutes of reading.

bradallred:
the C++ `fstream` and associated `streambuf ` is the closest thing to an equivalent and it works fine (tho definitely one of the ugliest things in the stdlib), but as I said earlier at this point switching to it is more effort than it is worth and `setvbuf` and/or the `mmap` solution would be easier and just as fast or even perhaps faster.

MarcelHB:
Nice, Windows version works and is unseen speed (600ms).

MarcelHB:
I close this an will open a new PR for the other approach to review the code more thoroughly.

PR #743, opened 2020-05-21
Include <clocale> to make MSVC happy with PR #650

chilvence:
I was not able to build gemrb.cpp on VS2019 without including 'locale.h'

As per Brad's suggestion, using 'clocale' works fine

Without this, the set_locale function chokes the build as 'undefined'

I don't see any reason why it wouldn't work, but if someone throws me a Polish or German dialog(f).tlk set for any game I can verify it works for me.

### DIFF
--- a/gemrb/GemRB.cpp
+++ b/gemrb/GemRB.cpp
@@ -21,6 +21,7 @@
 // GemRB.cpp : Defines the entry point for the application.
 
 #include "win32def.h" // logging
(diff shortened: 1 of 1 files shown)


bradallred:
For a quick explanation on the difference between the `.h` and `c` variants of C files in C++ see [this page](http://www.cplusplus.com/reference/clibrary/). I've had to deal with more than one subtle bug due to the wrong header being included.

PR #745, opened 2020-05-21
BIFImporter: Memory mapped file reading

MarcelHB:
## Description
As discussed in #738, now the approach with mmap and its Windows counter part.

Reduces loading time up to 80%, can't even read some of the loading screen texts any more. Please try out!

## Checklist

- [ ] The proposed change builds also on our build bots (check after submission) :shrug:

### DIFF
--- a/gemrb/core/System/MemoryStream.h
+++ b/gemrb/core/System/MemoryStream.h
@@ -29,16 +29,16 @@ namespace GemRB {
 
 class GEM_EXPORT MemoryStream : public DataStream
 {
-private:
+protected:
 	char *data;
 public:
-	MemoryStream(char *name, void* data, unsigned long size);
+	MemoryStream(const char *name, void* data, unsigned long size);
 	~MemoryStream();
-	DataStream* Clone();
+	DataStream* Clone() override;
 
-	int Read(void* dest, unsigned int length);
-	int Write(const void* src, unsigned int length);
-	int Seek(int pos, int startpos);
+	int Read(void* dest, unsigned int length) override;
+	int Write(const void* src, unsigned int length) override;
+	int Seek(int pos, int startpos) override;
 };
 
 }
--- a/gemrb/core/System/VFS.h
+++ b/gemrb/core/System/VFS.h
@@ -126,6 +126,11 @@ GEM_EXPORT char* CopyHomePath(char* outPath, ieWord maxLen);
 // default directory housing GUIScripts/Override/Unhardcoded
 GEM_EXPORT char* CopyGemDataPath(char* outPath, ieWord maxLen);
 
+void* readonly_mmap(void *fd);
+#ifdef WIN32
+void munmap(void *start, size_t);
+#endif
+
 class GEM_EXPORT DirectoryIterator {
 public:
 	typedef Predicate<const char*> FileFilterPredicate;
--- a/android/GEMRB_Android.mk
+++ b/android/GEMRB_Android.mk
@@ -158,6 +158,7 @@ LOCAL_SRC_FILES :=  main/gemrb/plugins/SAVImporter/SAVImporter.cpp \
 		    main/gemrb/core/System/String.cpp \
 		    main/gemrb/core/System/Logging.cpp \
 		    main/gemrb/core/System/FileStream.cpp \
+		    main/gemrb/core/System/MappedFileMemoryStream.cpp \
 		    main/gemrb/core/System/MemoryStream.cpp \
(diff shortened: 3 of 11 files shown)


bradallred:
This looks good. You should use `static_cast` instead of `reinterpret_cast` tho.

One question about its use in `FileCache`. If `FileCache` is an object that stays around for the duration of the program it will always be mapped. I don't know how big a `FileCache` is, but it could mean lots of RAM use if it is always around.

For files we open then read and close its no problem.

MarcelHB:
I guess `FileCache` is mildly misleading as it does not maintain a cache. It's more the `CacheCompressedStream` inside that does re-open BIF files from one of the Cache directories after they have been inflated.

bradallred:
cool. no worries about that then.

bradallred:
Builds and works fine on my Mac. The performance difference is only about 15% improvement, but I have an NVME drive so probably not a surprise there.

Good work. I'll let @lynxlynxlynx look it over and poke it all he wants, but I see no reason not to merge it.

gemrb/core/System/MemoryStream.h
```diff
@@ -29,16 +29,16 @@ namespace GemRB {
...
-	~MemoryStream();
-	DataStream* Clone();
+	MemoryStream(const char *name, void* data, unsigned long size);
+	virtual ~MemoryStream();
+	virtual DataStream* Clone();
 
-	int Read(void* dest, unsigned int length);
-	int Write(const void* src, unsigned int length);
-	int Seek(int pos, int startpos);
+	virtual int Read(void* dest, unsigned int length);
```
bradallred, line 39: all these are already virtual, but you should mark them as `override` instead.

gemrb/core/System/VFS.h
```diff
@@ -126,6 +126,11 @@ GEM_EXPORT char* CopyHomePath(char* outPath, ieWord maxLen);
 // default directory housing GUIScripts/Override/Unhardcoded
 GEM_EXPORT char* CopyGemDataPath(char* outPath, ieWord maxLen);
 
+void* readonly_mmap(void *fd);
+#ifdef WIN32
+void munmap(void *start, size_t);
+#endif
```
lynxlynxlynx, line 132: Does this work on mingw too or does it cause a redefinition error? Wondering if we need to test munmap presence in cmake.

MarcelHB: I'm not really sure how much `-G "MinGW Makefiles"` inside of an MSYS2 shell is the answer here, but I got no errors when doing `mingw32-make` . And from googling, it looks as if you still need a wrapper anyway.

gemrb/core/System/MemoryStream.h
MarcelHB: I'll have a look later today.

gemrb/core/System/VFS.h
lynxlynxlynx: Ok, we'll deal with it later if someone manages to break it.

gemrb/core/System/MemoryStream.h
MarcelHB: Ok, I learned that there is no need to manually re-virtual down the chain.

bradallred: Yup. All the `virtual` keyword does is tell the compiler to create a vtable for the function. Once a vtable exists it can't not exist. Do mark them as `override` tho.

MarcelHB: I fixed that. I just had the concept in mind that C++ may allow to restrict vtable entries. e. g. virt. A > B > C would lead to invoke only B even if we have C under a polymorph A when not doing virt. A > virt. B > C.

lynxlynxlynx:
Are you done? I want to run some comparison tests here too, but wouldn't like to repeat them if the code is still in flux.

MarcelHB:
Please, I'm waiting.

lynxlynxlynx:
Here are my results, tested cold and hot runs on a hdd and ssd. Memory use in the main menu and ingame is pretty much unaffected, but this was expected, since I didn't measure it during loading.

Load progress 10-100:
Cold start I tested only once and had a 60 % decrease on the hdd, 20 % on the ssd.
Hot start (5 consecutive runs after the cold one) detected no change on the hdd, 19 % on the ssd.

Small sample sizes, thus unreliable, but still  clearly beneficial. My ssd loads games slower than the disky barracuda, regardless of this PR — odd.

lynxlynxlynx:
Thanks!

PR #747, opened 2020-05-24
Substitution for VCPKG's dll placement script

chilvence:
I am suggesting this patch only as temporary fix to the only gripe I have with the almost serendipitious "VCPKG" package manager for Visual Studio. 

To be completely fair, It is actually a really nice package manager that works in general like apt or pacman etc, and works really well up until the point that it.... well.. doesn't

Unfortunately, GemRB seems to be outside of the use cases they may have tested. It makes a grand and well considered effort to place dependency dll files where they should be found, but it unfortunately fails to satisfy GemRB because of the fact that its plugins are in a different directory to the executable file. 

This is a really basic workaround that copies expected dll file names to the target directory so that the game can at least run. 

In the back of my mind, I was trying to consider how to make a script that tries to find all dependencies and deploy them as necessary, but then I realised I was only trying to plug a leak, and I'm probably not qualified for that anyway.

Please pick fault with this PR if you find any. I've made enough variations on the theme that I am probably blind to mistakes by now. I would be mortified if I messed up anyone's build environment though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,14 @@ OPTION(USE_PNG "Enable LibPNG support" ON)
 OPTION(USE_VORBIS "Enable Vorbis support" ON)
 OPTION(USE_ICONV "Enable Iconv support" ON)
 
+#VCPKG dll deployment is circumvented because it doesn't currently work for gemrb
+IF(WIN32 AND _VCPKG_INSTALLED_DIR)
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks! I guess this was the last blocker to cleaning up the instructions? :)

chilvence:
Yep, I have already taken an axe to the MSVC guide I wrote, the only other thing I was going to look at was the config generation, but that is only for the bonus round

PR #759, opened 2020-05-25
BG: disables mission pack button in menu when not installed

### LINKED ISSUE #758: BG: Ghost list of save games accessible from main menu
#### Bug description
I think I just clicked in too much haste and landed in some ghost area. For my BG, below the "Load Game" button, there is blank space. But actually you end up in an empty list of save games.

GemRB version: master

#### Steps to reproduce
1. Go to BG1 main menu.
2. Click the solo play button.
3. Click below the load button.

#### Expected behavior
Nothing should happen.

#### Screenshots
Demo:
<details>
  <summary></summary>

store

</details>

MarcelHB:
## Description
fixes #758 

## Checklist

### DIFF
--- a/gemrb/GUIScripts/GameCheck.py
+++ b/gemrb/GUIScripts/GameCheck.py
@@ -20,7 +20,7 @@
 # GameCheck.py - functions to check GameType
 
 import GemRB
-from ie_restype import RES_WMP, RES_ARE
+from ie_restype import RES_WMP, RES_ARE, RES_2DA
 
 MAX_PARTY_SIZE = GemRB.GetVar ("MaxPartySize")
 
@@ -65,3 +65,6 @@ def HasBGT ():
 
 def HasTutu ():
 	return GemRB.HasResource ("fw0125", RES_ARE)
+
+def HasTOTSC ():
+	return GemRB.HasResource ("toscst", RES_2DA)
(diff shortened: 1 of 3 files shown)


bradallred:
I think I have a preference to deleting unwanted controls because it keeps noise out of your way when debugging issues with the view hierarchy. I think it is odd to keep things around that we don't want.

MarcelHB:
The HTML-history of mine wants to have two distinct buttons with `display: none` being toggled, but then this kind of recycling can be found across the scripts, so :shrug:

lynxlynxlynx:
Deleting would require moving more code into the if block, but that's about it.

bradallred:
it's not a dealbreaker.

why is it called `MoviesButton` tho?

lynxlynxlynx:
That's its label on the first screen, nothing special.

bradallred:
Sure, but is there a reason? it seems like it has nothing to do with "movies" and makes the code much less clear.

lynxlynxlynx:
Yes, the second menu is not actually in a separate window, just everything gets relabelled and the callbacks changed.

MarcelHB:
I can offer `PartTimeMoviesPartTimeMissionPackButton` without reorganizing the menu as a whole.

chilvence:
I had a 1999 flashback. I think the button is actually just for 'movies' and then they hijack it for the mission pack when you install TOTSC.

lynxlynxlynx:
It is for movies, but on on the main screen — this one is one level deeper, once you choose singleplayer.

lynxlynxlynx:
So, since it's the same window, deleting would actually hurt.

Issue #767, opened 2020-05-30
Subviews & MSVC : CopyColorRange & ColorPal templates


chilvence:
This is the final error I have when using MSVC to compile.

It affects 5 files: CharAnimations.cpp, ScriptedAnimation.cpp, WorldMap.cpp GUIScript.cpp and Palette.cpp

```
B:\msvc\src\chiv\gemrb\core\Palette.cpp(122): error C2664: 'void GemRB::Palette::CopyColorRange(const GemRB::Color *,const GemRB::Color *,uint8_t)': cannot convert argument 1 from 'std::_Array_const_iterator<_Ty,16>' to 'const GemRB::Color *'
          with
          [
              _Ty=GemRB::Color
          ]
  ..\..\..\gemrb\core\Palette.cpp(122): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
  ..\..\..\gemrb\core\Palette.cpp(52): note: see declaration of 'GemRB::Palette::CopyColorRange'
```

full output: https://pastebin.com/PnABG6tL

I don't know if this has a good clue, but this is the in editor feedback:
![pallette cpp 122](https://user-images.githubusercontent.com/1923867/83320125-481fc300-a23c-11ea-9262-3ed7475740c6.png)

I will keep trying to make it work, but this is well beyond my really limited understanding of C++. If anyone has any hints, I'm totally stumped. 

It looks like it cant figure out the ColorPal template properly though?

MarcelHB:
What about `pal16.data()` or `&pal16[0]`?

chilvence:
I'm kicking myself for not trying that, if I change every instance of pal16.begin() to &pal16[0] etc, then it builds and I can get in game. 

Unfortunately I think that is not the end of that, since I have to explicitly set the size of the palette range to copy, I think the point of using begin() and end() here was to avoid having to do that?

MarcelHB:
If you have contiguous memory, you can provide both `begin/end()` as well as `&[0]/&[0+n]`. Reason is that the underlying `std::copy` is likely taking anything templated that adheres to some arithmetic operators and dereferencing (duck typing). Both pointers and usual iterators, well, know how to do basic `*`, `->` and `++`.

If it's non-contiguous or something different, then `begin/end()` abstract the underlying stuff away, e.g. linked lists, maps, ...

[...]

PR #769, opened 2020-06-03
Some tweaks to palette handling for MSVC

chilvence:
Related to the discussion in #767 , this allows compilation of some recent palette handling changes that work for gcc/clang but not for 'cl.exe'

This is the last blocking issue for the building this dev branch on Windows 

Tested on Torment and Icewind Dale. I would have tested on BG1/2 as well, but I stupidly forgot that I only have a disc version when I was tidying up files last, and I haven't had a CD/DVD drive for at least 10 years


bradallred:
Thanks; your willingness to help out is greatly appreciated.

chilvence:
No problem! I'll probably be rebuilding and testing on this branch at least once a week from now on , so I can keep you posted if there happens to be anything else Visual Studio complains about

Issue #797, opened 2020-06-07
[subviews] PST Crash when this zombie becomes hostile


chilvence:
Actually this happens with any zombie, but it might be quicker to check this save rather than start a new game (although holding down '1' quickly tears through the opening torrent of dialogue)

[000000261-new game.zip](https://github.com/gemrb/gemrb/files/4741995/000000261-new.game.zip)

When the zombie becomes hostile for any reason, the game almost immediately crashes

I did a bisect and found it works in 246b08e , but not in 67d1ed8

The line at fault appears to be this one: https://github.com/gemrb/gemrb/blob/2e885015d9b8ce7305607bd15d5a9068777e6946/gemrb/core/Scriptable/Actor.cpp#L7814

(it might well happen with more actors, but I didn't manage to get that far after loading my new game save before bumping into this)

chilvence:
This seemed to be enough to avoid the crash : https://github.com/gemrb/gemrb/blob/2e885015d9b8ce7305607bd15d5a9068777e6946/gemrb/core/Scriptable/Actor.cpp#L7814

chilvence:
I think #175 is the main issue related to the PST stances? Correct me if I am wrong. I haven't done a lot of digging to see how much work is involved since I've always pretty much accepted the way it is

The function that searches for the correct animation for PST seems pretty isolated though, and it was pretty straightforward to triage one case, the invisible 'running' zombies: https://github.com/chilvence/gemrb/commit/d43f8cabaf83fd6cafabff9291f320ea995821e1

I can probably rework that function a bit more to make safe assumptions based on the actual data, because the file naming conventions are so strictly organised. I realise that it is far from the desired proper solution, but it seems like an easy temporary compromise, and one that wouldn't spill over to the other games. A few extra safety checks seems better than having problems like this.

lynxlynxlynx:
The patch looks fine, please just put the condition on the same line (&&), so there's less nesting. It's needed regardless of other work (is a regression from master).

Let's discuss other stuff elsewhere. I haven't looked at stances much, but I'd wager there's more work to be done on transitions and making it more robust (CharAnimations is very cludgy).

PR #798, opened 2020-06-07
Actor.cpp: avoid trying to use null 'anims'

chilvence:
Fixes crash from #797

As mentioned in that issue, it seems to just be a symptom of the incomplete PST animation code


bradallred:
thanks!

Issue #822, opened 2020-06-15
[subviews] Unable to visit PST AR0600


chilvence:
Currently when you try to load AR0600, the game crashes abruptly 

```
[AREImporter/DEBUG]: Loading wallgroups
terminate called after throwing an instance of 'std::length_error'
  what():  vector::_M_default_append
```
I couldn't get a backtrace out of gdb, but Visual Studio pointed at Polygon.cpp, the last call was just after Polygon.cpp line 40

Going through the commit history, it doesn't crash before [this commit](https://github.com/gemrb/gemrb/commit/80078c548c448d5518ce78f3da66c9b1183e5a7d)

I am not sure about what the particular line there is for, but just reverting that change means the game can load this area again.

Perhaps there is something funky about the data for this area?

bradallred:
just send a PR with that reverted tho I'm surprised it wasn't encountered before now.

chilvence:
Are you sure? I am not claiming to understand what's happening here, I was just trying blind luck 

(but if that was all that needed to change, that's cool)

bradallred:
yeah, that is the line that allocates the vector. we don't use `push_back` to add lines, we expect them to all be preallocated as we will need them all. I don't know why I thought we didn't need `BBox.h` number of lines, that's just silly.

MarcelHB:
Hold on:

```cpp
if (y_top < 0) y_top = 0;
if (y_bot >= BBox.h) y_bot = BBox.h - 1;
// ...
for (int y = y_top; y < y_bot; ++y) {
```

Given that `BBox.h >= 2`, this loop will then underrun the vector?

bradallred:
if it was an underrun then increasing the size shouldn't have fixed it, but we should check. I still think logically `BBox.h` should equal the vector size tho.

bradallred:
however, this line `if (y_bot >= BBox.h) y_bot = BBox.h - 1;` makes me think the allocation to `BBox.h - 1` is correct. now that i'm home I loaded into the debugger and found we have a 0 height polygon so that's the actual issue since a poly of height 0 is going to lead to an allocation of -1 (aka the size of the known universe) obviously that throws the unhandled exception.

the fix then should be to bail out for a polygon of height (or width) of 0.

PR #823, opened 2020-06-16
Subviews Polygon.cpp reversion for #822

chilvence:
This reverts commit 80078c548c448d5518ce78f3da66c9b1183e5a7d.

Related to issue discussion: #822

I can exhaustively test the game areas if need be, but other than whether or not they load, I am not sure what to be testing or looking for?

### DIFF
--- a/gemrb/core/Polygon.cpp
+++ b/gemrb/core/Polygon.cpp
@@ -42,7 +42,7 @@ Gem_Polygon::Gem_Polygon(const Point* points, unsigned int cnt, Region *bbox)
 
 void Gem_Polygon::Rasterize()
 {
-	rasterData.resize(BBox.h - 1);
+	rasterData.resize(BBox.h);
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Can you determine which polygon / infopoint this is about? I don't see anyone in ar0600 with a height of 0.

bradallred:
it is the wall polygon at index 286

lynxlynxlynx:
Ah, in the wed then. We should just skip it in the loader.

chilvence:
Does that mean it was the data at fault? Is this reversion still needed or does it just need adding to?

bradallred:
no, we shouldn't revert this. the problem wasn't a case of not enough vector space like initially theorized.

lynxlynxlynx:
Yeah, the data is at fault, so we need to fix this in TISImporter or WEDImporter, wherever we read it in.

chilvence:
No problem, glad the culprit has been pinned down all the same

Issue #824, opened 2020-06-16
PST: Critical Hit/Miss lines are mixed up with rare select lines


chilvence:
Another character sound bug I neglected to log

When the characters make a critical miss, the 'rare select 4' line is played instead of the critical miss voice line. When they score a critical hit, the 'rare select 3' line is played.

I was attempting to fix this, but I haven't been able to work out where these sounds are actually defined in the original game, they don't appear to be referenced in the CRE files like the selection/bored/insult etc sounds are.

lynxlynxlynx:
We just shouldn't be playing those VBs in pst and criticals have their own entries in strings.2da. But this is also related to #453, since these could fit nicely into VBs (pst has so many unused slots).

chilvence:
Do you mean we could repurpose the unused voice slots in the CRE files?

lynxlynxlynx:
Perhaps yes; there's mayve enough room to dynamically populate them from a table on startup.

chilvence:
I'd be happy to compile the 2da table, I'm guessing something like columns with character names and rows with voice strrefs would do the job, right?

lynxlynxlynx:
A rough count makes it look like there might be enough space, enough unused slots of the 100 in CRE. But this is a discussion for the other bug (or a new one). First step would be to use the lists from there and compile full ones for all the NPCs, then check their CREs if they have enough space and where — hopefully all share enough holes. Only then can we start thinking of a table.

chilvence:
Oops, I didn't realise I already mentioned this over there anyway, must have forgot to take it off my own list.

lynxlynxlynx:
For this bug of using nonexisting VBs, you could indeed remap both slots to one unused one. Then the other bug can work on the full fix.

PR #826, opened 2020-06-16
PST: Patch for Critical hit/miss annoyance #824

chilvence:
This fixes the annoyance of critical hits/misses playing the rare select 3/4 sounds #824

That side issue can then be closed, while the larger issue is pending.

Indexes 90/91 were simply picked as entries that seem have no existing usages in the cre data, no other particular reason, they could be subject to change as needed

### DIFF
--- a/gemrb/unhardcoded/pst/vcremap.2da
+++ b/gemrb/unhardcoded/pst/vcremap.2da
@@ -7,3 +7,5 @@ SELR3     65            37
 SELR4     66            38
 RESP1     59            58      no second hostile response
 RESP2     60            58      no third hostile response
+CRITMISS  65            90      remap from pst rare select 3
+CRITHIT   66            91      remap from pst rare select 4


lynxlynxlynx:
Why waste two slots? And if there's a hole of size 1 in the list, that would be a better fit. Definitely pick a smaller hole, wherever it may be — this one breaks up the largest contiguous chunk available.

All this to avoid needing to re-move it later.

chilvence:
There are lots of invalid entries that play a Deionerra sound, "I will wait for you in death's halls", I just picked two slots that have none of that, right after the last existing definition in NI (NORDOM_PORTAL)

I was also assuming that there would need to eventually be two slots, since the sounds do actually exist

lynxlynxlynx:
fair enough.

chilvence:
I can probably write a proper description of the slot usages, all the same, since they are currently only roughly defined in NI or still based on BG assumptions.

In any case, the #824 issue is the real target

PR #827, opened 2020-06-17
PST: TNO Avatar Reassignment is using the wrong table row

### LINKED ISSUE #522: [bug] PST:  TNO must change avatar depending on equipped weapon type
PST has a complete set of character animations for each weapon type he can use instead of being a multi part sprite like in BG/IWD . Currently Gemrb seems to only use the fists anim set no matter which weapon equipped

These are the animation sets I could find, with the first four letters being the name of the action in the pst convention:

****NOAB - resdata: 47 - Axe
****NOCB - resdata: 48 - Club
****NODB - resdata: 49 - Dagger
****NOFB - resdata: 50 - Fists
****NOHB - resdata:51 - Hammer 

Also there are two disguises in the game that should probably override that:

****NOMB - resdata:93 - Dustman disguise, which is given by equippable robes
****NOZB - resdata:82 -  Zombie disguise, which is placed on you by an npc.

I can't see anything notable in the itm files that relates to these avatars so I guess they were just hardcoded changes. Even the DUSTROBE.itm file has no specific avatar change action that I can find.

None of the other characters can actually use any other weapons unless they are the same type they started with. Vhailor might be different, but I can't remember offhand. I actually remember on my old Pentium 233 it taking a good second or two to change weapons in the inventory screen, so it must have been preloading the animations to stop the main game window from imploding.

Of course this is probably better dealt with after subviews, just reporting it.

chilvence:
BioTable.GetRowName (Specific+1) was returning VHAILOR

Deleting a single character is enough to close #522

Whoops.

### DIFF
--- a/gemrb/GUIScripts/GUICommonWindows.py
+++ b/gemrb/GUIScripts/GUICommonWindows.py
@@ -1290,7 +1290,7 @@ def UpdateAnimation ():
 
 	BioTable = GemRB.LoadTable ("BIOS")
 	Specific = GemRB.GetPlayerStat (pc, IE_SPECIFIC)
-	AvatarName = BioTable.GetRowName (Specific+1)
+	AvatarName = BioTable.GetRowName (Specific)
 	AnimTable = GemRB.LoadTable ("ANIMS")
 	if animid=="":
 		animid="*"


lynxlynxlynx:
Please verify this is correct for all pcs.

chilvence:
I'm not sure really what to test for, TNO is the only one that has different avatars, and the item restriction rules in the game are so strict none of the characters can even equip items that would potentially change their appearance. Basically if it isn't an Axe, Vhailor isn't interested

I can create some new entries in the table, but there is no art? I am happy to do more work here, but I need a push in the right direction because I'm confused. Or did you notice something else broken by this?

lynxlynxlynx:
This function is called for everyone, but I see now there's nothing else in the table, so that's fine.

chilvence:
If you had any improvement ideas on this though, I'm all ears - I started off sure that this used to work at some point, then I convinced myself it had been so long I started to think I could be imagining it, then I drew up a rough plan to implement it again, then I stumbled upon the existing feature while looking at something else.... so, almost disappointed it was too easy

Issue #825, opened 2020-06-16
 META : Overview list for full PST playability


chilvence:
Copying the idea from Lynx, I thought a meta issue for keeping track of PST bugs would be handy. 

I find the tracker really amazing and helpful, but it has been hard for me to keep so many different issues to test straight in my head

This is basically a combination of my remaining offline (ie post-it) notes, combined with issues that are already tracked (I've summarised some of the issue names best I could for readability), condensed into one giant list

_"e pluribus unum"_

Gameplay:
- [ ] #176 Verify level up behaviour
- [ ] Multi classing, class levels for TNO are not handled correctly
- [ ] #438 You can respawn if killed in the final fight
- [ ] #521 When you respawn, you should have regenerated more than 1 HP
- [ ] #524 Morte loses his taunt ability
- [ ] #598 Actors that use EscapeArea() should try to walk 'off screen' then timeout
- [ ] #599 You don't get locked in the catacombs because of unpredictable scripting

Magic Spells: _(hiss)_
- [ ] #458, #637 Figure out spell casting/round timings
- [ ] #640 Errant stacking of some spell bonuses
- [ ] #641 Errant permanency of some temporary effects
- [ ] #455 Multiple actor targeting needed for some missile spells
- [ ] #639 Spell behind/in front of actor animations don't work
- [ ] #454 Some spells 'should' make spectators stand around as if dumbfounded
- [ ] #not tracked?: High level spells, figure out what is left to implement

Interface: 
- [ ] #293 Fix Text rendering in options window
- [ ] #307 Reform party window has placeholder graphics
- [ ] #445 Dialog window 'click to close' text should not always appear
- [ ] #450 Anti RSI: The original game has 'shift+click' as a run command
- [ ] #451 Original interface stays visible in cutscenes
- [ ] Thieving skills can't be assigned in UI
- [ ] No visual feedback for active thief skills
- [ ] Items are not always stacked automatically on pickup
- [x] Item stack split isn't working [fixed in my WIP branch]
- [ ] Options menu doesn't pause the game
- [ ] Dying shouldn't deselect you, since you just carry on living regardless...
- [ ] Some animated portrait states are chosen incorrectly
- [ ] Clicking on quick items does not engage target mode until after the floatwindow is closed

_[note: the untracked ones here I prefer to send PR's rather than bug reports]_

Graphical Rendering:
- [ ] #167 Area animation clipping by wall groups
- [ ] #177 Floating text support for combat messages etc
[...]

PR #834, opened 2020-06-22
pst/guirec: increase maxhp before current hp

chilvence:
This fixes one small item from #825

When you level up, your total hitpoint limit is increased, but you are shortchanged being given the new HP because they are updated in the wrong order.

Only 60 more to go!


lynxlynxlynx:
Thanks!

Issue #830, opened 2020-06-18
PST travel regression: unable to move around the hive 


chilvence:
It took me starting a new game to notice this, but after escaping the mortuary and clearing the first mini dungeon, I was stuck in the northeast quarter.

Worldmap travel works after you reach AR0500, but before that happens those worldmap trigger points have to default to moving to the nearest area. 

This exit should take you to the north entrance of the Smouldering Corpse bar area for example:

 [000000270-GRB Close to the edge.zip](https://github.com/gemrb/gemrb/files/4799557/000000270-GRB.Close.to.the.edge.zip)

I think I have traced it back to https://github.com/gemrb/gemrb/commit/e4e6ce19a37ee544230cf7ac36a88f4ef8d73ce2
The previous https://github.com/gemrb/gemrb/commit/e1c9774b6f76c98f042b0cfa16d293209b0bef37 allows you to exit.

Commenting out [this single line](https://github.com/gemrb/gemrb/blob/f2c6946425574ac8931e275237ea144754f4a76c/gemrb/core/GUI/GameControl.cpp#L2139) is enough to restore the original behaviour, but it also breaks the ability to worldmap travel later on:

[000000271-GRB Don't push me cause I'm.zip](https://github.com/gemrb/gemrb/files/4799798/000000271-GRB.Don.t.push.me.cause.I.m.zip)

I think the issue is trickier than it seemed at first glance, because until you get access to the worldmap in AR0500, all exits should behave like regular travel triggers.

I will continue trying to solve this, just adding a tracker issue for the details

edit: I haven't had a chance to thoroughly test this yet, but I think it might be the simple solution:

https://github.com/chilvence/gemrb/commit/d6148c5363d5a2ec08a0bf967e448cc530c31db4

The issue is, there wasn't anything technically wrong with the original behaviour before the pst world mapping was added, it's just that it needed to change globally when you hit the lower ward (everything was navigable before, it just took longer). So both ways are correct, just at different stages of the game.

I'm not sure yet what happens in Carceri/Baator, which I think are the only other areas with any kind of 'worldmap' ish navigation, although they only have one icon on the map so I suspect they were written as one 'area'

chilvence:
I'm not totally sure about that - for what its worth, the current worldmap checking works fine, except that it doesn't have a fall back behaviour for the large part of the game before you get to use it.

(the world map area entrances are sometimes strange, but that's another issue)

lynxlynxlynx:
[...]

PR #843, opened 2020-06-27
PST Worldmapping: Fall back to regular area travel if worldmap is not allowed

chilvence:
This is for #830

From the way I read it, the only thing going wrong here is that the "ShouldDoWorldMapping" check should return false if you haven't reached AR0500, because before that point, that's actually exactly the case.

What I think is happening is that the check is returning a false positive, i.e. it is saying "yes you can use the worldmap here", and then somewhere else is ignoring that and not allowing access to the worldmap, because it knows that is not true. 

So as a consequence, before you 'unlock' the worldmap, nothing actually happens when you click the map edge travel regions.

When you force the check to be false before AR0500, then all of the screen edges just behave like normal exits.  Then after reaching ar0500 they automatically adopt the new behavior. I don't even have do do anything specific, just "mta("ar0500") is enough, and that exactly matches the bottleneck from the original game that you have to pass through. (its kind of tricky to test though, cos if you warp into the wrong place the cutscene won't play out)

Areas like Baator, Carceri, Curst etc are unaffected by this, because none of those have exits that would trigger the worldmap in any case. They are just like 'room to room' transitions.

I have a really chronic lack of experience modding IE games, so I am never 100% sure what I should be looking at, but all of the edge transition exits in the hive do actually seem to have proper entrance points that work and are fine to use.

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -1937,6 +1937,12 @@ static std::map<std::string, std::vector<std::string>> pstWMapExits {
 // has to be a plain travel region and on the whitelist
 bool GameControl::ShouldTriggerWorldMap(const Actor *pc) const
 {
+
+	//if we havent reached the key area, we need return false to allow regular travel instead
+	Scriptable* Sender = (Scriptable*)pc;
+	bool keyAreaVisited = core->HasFeature(GF_TEAM_MOVEMENT) && CheckVariable(Sender, "AR0500_Visited", "GLOBAL") == 1;
+	if (!keyAreaVisited) return false;
+
 	if (!core->HasFeature(GF_TEAM_MOVEMENT)) return false;
 
 	bool teamMoved = (pc->GetInternalFlag() & IF_USEEXIT) && overInfoPoint && overInfoPoint->Type == ST_TRAVEL;


gemrb/core/GUI/GameControl.cpp
```diff
@@ -1937,6 +1937,12 @@ static std::map<std::string, std::vector<std::string>> pstWMapExits {
 // has to be a plain travel region and on the whitelist
 bool GameControl::ShouldTriggerWorldMap(const Actor *pc) const
 {
+
+	//if we havent reached the key area, we need return false to allow regular travel instead
+	Scriptable* Sender = (Scriptable*)pc;
+	bool keyAreaVisited = core->HasFeature(GF_TEAM_MOVEMENT) && CheckVariable(Sender, "AR0500_Visited", "GLOBAL") == 1;
+	if (!keyAreaVisited) return false;
+
 	if (!core->HasFeature(GF_TEAM_MOVEMENT)) return false;
```
MarcelHB, line 1946: Purely from the code perspective: Can you drop the 2nd check for `GF_TEAM_MOVEMENT` (always true then), or just move it up and skip yours?

bradallred: I was about to point this out myself. let's fix that.

lynxlynxlynx: No need for Sender either, but let me take a broader look first.

chilvence: Oh yeah, I didn't even notice that! Whoops...

I also wasn't sure why I needed a sender, it just seemed like most instances of CheckVariable also include that. Sometimes it's the 'map' rather than an actor. I didn't ask any questions...

It does work in game though, so this must at least be 15% right :)

Issue #764, opened 2020-05-27
Encoding problems in AI script window


fenuks:
#### Bug description
Current script text area in character AI screen is mangling non-ASCII characters. Problem is related to my work on encoding. Prior to that change, no description was shown due to silent encoding error.

Issue is that QueryText returns now string that is encoded with user system encoding (which is needed if string is filename, like in importable characters list), and here it causes problem (text is [queried here](https://github.com/gemrb/gemrb/blob/22802a62f6465316f5fcb7afa62949305dec3eec/gemrb/GUIScripts/GUIRECCommon.py#L447) (it is in `UTF-8` in my case), and is immediately set in description text area (it expects string encoded in game encoding, `WINDOWS-1250` in my case).

I think situations where text from control is later used to reach some resource on user system that is sensitive to encoding (like filenames) is way rarer, so I think that it will be best to return string in game encoding from QueryText [here](https://github.com/gemrb/gemrb/blob/22802a62f6465316f5fcb7afa62949305dec3eec/gemrb/plugins/GUIScript/GUIScript.cpp#L1999), and explicitly convert between encodings where character is imported [here](https://github.com/gemrb/gemrb/blob/22802a62f6465316f5fcb7afa62949305dec3eec/gemrb/GUIScripts/iwd/CharGen.py#L2708) (and possibly in few other places yet to be discovered).

I've already checked this solution, and it works fine.

GemRB version: 

#### Steps to reproduce
1. Open character screen
2. Click modification -> script
3. See encoding errors

#### Screenshots
![encoding problems](https://user-images.githubusercontent.com/1249829/83075753-e3117500-a063-11ea-9b38-1cf94bde0d30.png)

lynxlynxlynx:
Yeah, adding an optional parameter to QueryText  to enable fs-mode sounds fine. The default then remains the same and only a handful of calls in python need to be updated.

fenuks:
Alright, I will address that in future MR.

PR #853, opened 2020-06-30
Optionally use system encoding in QueryText

fenuks:
## Description
This solves #764. I've added optional boolean parameter to QueryText(), my only concern is whether it should be positional argument, or keyword one. Initially, I wanted to use keyword one, but that requires changing method signature, and I run into issues [here](https://github.com/gemrb/gemrb/blob/5b0cf79e5d388f718585b96586535c1fe5970fb2/gemrb/plugins/GUIScript/GUIScript.cpp#L15801), so I would have add another macro that would cast it to PyCFunction, as it is described [here](https://docs.python.org/2/extending/extending.html#keyword-parameters-for-extension-functions), I suppose. Current `METHOD` macro expects PyCFunction, not PyCFunctionWithKeywords. Do you think positional optional argument is fine, or it should be keyword one, which would require a little more work?

## Checklist

### DIFF
--- a/gemrb/plugins/GUIScript/GUIScript.cpp
+++ b/gemrb/plugins/GUIScript/GUIScript.cpp
@@ -1972,9 +1972,9 @@ static PyObject* GemRB_Window_HasControl(PyObject * /*self*/, PyObject* args)
 PyDoc_STRVAR( GemRB_Control_QueryText__doc,
 "===== Control_QueryText =====\n\
 \n\
-**Prototype:** GemRB.QueryText (WindowIndex, ControlIndex)\n\
+**Prototype:** GemRB.QueryText (WindowIndex, ControlIndex[, UseSystemEncoding])\n\
 \n\
-**Metaclass Prototype:** QueryText ()\n\
+**Metaclass Prototype:** QueryText (UseSystemEncoding=False)\n\
 \n\
 **Description:** Returns the Text of a TextEdit/TextArea/Label control. \n\
 In case of a TextArea, it will return the selected row, not the entire \n\
@@ -1982,6 +1982,7 @@ textarea.\n\
 \n\
 **Parameters:**\n\
   * WindowIndex, ControlIndex - the control's reference\n\
+  * UseSystemEncoding - whether returned text should use OS encoding, False by default\n\
 \n\
 **Return value:** string, may be empty\n\
 \n\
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
It's fine, thanks, we don't use keyword parameters at all.

fenuks:
Alright, I will update documentation then, and test it properly.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -1972,16 +1972,17 @@ static PyObject* GemRB_Window_HasControl(PyObject * /*self*/, PyObject* args)
 PyDoc_STRVAR( GemRB_Control_QueryText__doc,
 "===== Control_QueryText =====\n\
 \n\
-**Prototype:** GemRB.QueryText (WindowIndex, ControlIndex)\n\
+**Prototype:** GemRB.QueryText (WindowIndex, ControlIndex, UseSystemEncoding)\n\
```
lynxlynxlynx, line 1975:
```suggestion
**Prototype:** GemRB.QueryText (WindowIndex, ControlIndex[, UseSystemEncoding])\n\
```

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -1972,16 +1972,17 @@ static PyObject* GemRB_Window_HasControl(PyObject * /*self*/, PyObject* args)
...
 \n\
-**Metaclass Prototype:** QueryText ()\n\
+**Metaclass Prototype:** QueryText (UseSystemEncoding=False)\n\
```
lynxlynxlynx, line 1977: We currently use the above syntax, but this is fine, I guess.

fenuks: If you wish, I can change it, but yes, I think it should be OK.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -1992,15 +1993,19 @@ The above example retrieves the character's name typed into the TextEdit control
...
 **See also:** [[guiscript:Control_SetText]], [[guiscript:SetToken]], [[guiscript:accessing_gui_controls]]"
 );
 
 static PyObject* GemRB_Control_QueryText(PyObject * /*self*/, PyObject* args)
 {
-	int wi, ci;
+	int wi, ci, useSystemEncoding = 0;
 
-	if (!PyArg_ParseTuple( args, "ii", &wi, &ci )) {
+	if (!PyArg_ParseTuple( args, "ii|i", &wi, &ci, &useSystemEncoding )) {
```
bradallred, line 2008: you can use `p` as a format specifier for booleans.

fenuks: I was initially trying to do so, but it was introduced only in python 3.3, I believe.

fenuks: Yes, here is the changelog entry. https://hg.python.org/cpython/file/05274ab06182/Misc/NEWS

lynxlynxlynx:
Do I understand correctly from the checkbox that testing is complete and this is ready for merging?

fenuks:
Yes, I've checked all chargen GUIs, plus all main GUIs in-game. I remember that BG version that @MarcelHB has had encoding problems with sounds, but that was in resource manager, so this change shouldn't cause regression here.

I wanted to ask, automated testing in this kind of program is not always easy, but it would be still useful to have some kind of testing framework integrated in project in case it's possible write to unit/integration test without much effort. What do you think?

lynxlynxlynx:
We have bits and pieces here and there, but yeah, most subsystems are not testable without a running game.

PR #857, opened 2020-07-06
Rpi support

bholland:
## Description

This will bring in the SDL2 and GLES support for the raspberry pi 4 specifically. I will also provide instructions on how to compile SDL2 to use the GLES support and how to configure gemrb to work with retropi. 

## Checklist


bradallred:
you might be interested in ensuring that this merges and works with the subviews branch else it will likely get broken in the near future.

CMakeLists.txt
```diff
@@ -276,69 +289,120 @@ IF((SDL_BACKEND STREQUAL "SDL2") OR USE_SDL2)
...
-
-IF(SDL_FOUND)
-	MESSAGE(STATUS "Looking for SDL: found")
-	# unify SDL variables, so we don't have to differentiate later
-	MESSAGE(STATUS "Looking for libSDL: found")
-ELSE()
-	MESSAGE(WARNING "Looking for SDL: not found!")
-	MESSAGE(FATAL_ERROR "Please get SDL from www.libsdl.org")
+IF(IS_RPI)
+IF(IS_RPI)
```
lynxlynxlynx, line 309: One level isn't needed.

CMakeLists.txt
```diff
@@ -276,69 +289,120 @@ IF((SDL_BACKEND STREQUAL "SDL2") OR USE_SDL2)
...
+  IF(SDL_MIXER_FOUND)
+    MESSAGE(STATUS "Looking for SDL_mixer: found")
+  ELSE()
+    MESSAGE(WARNING "Looking for SDL_mixer: not found!")
+    MESSAGE(WARNING "If you want to build the SDL_mixer plugin, install SDL_mixer first.")
+    MESSAGE(WARNING "Make sure you use a version compatible with the chosen SDL version.")
+  ENDIF()
+
+  SET(OPENGL_BACKEND GLES)
+  SET(SDL_BACKEND SDL2)
```
lynxlynxlynx, line 329: these two are already set above.

lynxlynxlynx:
Uff, this is quite messy. Is it meant for compiling or crosscompiling? We already check for the bcm header file, so you can use the RPI cmake var. Also, I'm sure it can be done without so much code duplication. You also want openal anyway, no need to bother with sdl_mixer.

Choosing the experimental GL backend is also not that advisable. Besides being incomplete, the branch Brad mentions only supports GLEW afaik. So without putting in some dedicated work, GLES support may vanish after the release next month.

CMakeLists.txt
bholland: Good catch, sorry about that.

CMakeLists.txt
bholland: > Uff, this is quite messy. Is it meant for compiling or crosscompiling? We already check for the bcm header file, so you can use the RPI cmake var. Also, I'm sure it can be done without so much code duplication. You also want openal anyway, no need to bother with sdl_mixer.
> 
> Choosing the experimental GL backend is also not that advisable. Besides being incomplete, the branch Brad mentions only supports GLEW afaik. So without putting in some dedicated work, GLES support may vanish after the release next month.

I did my best with the CMake. I don't have a lot of experience with it and someone else can likely do it better but this does work. I don't think this is about cross-compiling. I only compiled this on the pi itself. Getting library links working without the pi itself and specifically Raspbian was very hard.  I am sure it is possible but that would take someone with a lot more experience than I have. 

I didn't see a check for the BCM variables but that also can be Android which I specifically did not want. The Pi is special because it uses GLES 1.2 in hardware. It doesn't emulate and when you get into emulations, your framerate goes to <1 FPS. Using Raspbian, you must link to the /opt/vc library to use the hardware rendering regardless of whatever other GLES or OpenGL libraries you happen to have installed on your pi. 

In addition, the SDL2 comes with Jessy and Buster both to not link properly and have to be compiled from source and installed. I have documentation going up on Gibberling now. This isn't using the experimental backend AFAIK. It is using the hardware rendering engine supported by Buster. SLD2 is the preferred layer on top of that from what I gathered on the RPi forums. I am happy to stand corrected but this is the only way I got BG2 to get 30FPS at 1600x900. I get 5FPS at 1920x1080. This is up from 4FPS and .5FPS, respectively, using the provided libraries and GemRb configuration. 

I would highly recommend not totally dumping GLES if for no other reason than it is the only real way to get decent framerates on the pi. I am happy to work on that when I have some free time or even add a disclaimer to use a specific version for the pi.

bholland: BTW, I don't actually know how to disable or make OpenAL preferred. I have both on my system. I run this using the RetroPie system and I believe that they default to OpenAL. I don't think this breaks with the sdl_mixer support. I am not even sure I use it and I don't know how to test that I do.

lynxlynxlynx:
How about this instead? You can remove the url parameter to get a raw patch to apply:
http://sprunge.us/w7Dmq3?diff

No need to worry about android and sdl_mixer (misread the diff).

bholland:
> How about this instead? You can remove the url parameter to get a raw patch to apply:
> http://sprunge.us/w7Dmq3?diff
> 
> No need to worry about android and sdl_mixer (misread the diff).

I think that will work. I can't apply the diff though. Can you send me the CMakeList.txt file and I will verify that your changes work? I will update the file on my push with your changes.

MarcelHB:
Simply try: `git reset master -- CMakeLists.txt`, then apply the change and do a new commit.

lynxlynxlynx:
The diff was made against master, yes, not 0.8.6. I think you're getting conflicts because of the extra vcpkg stuff now inside.

lynxlynxlynx:
Have you managed to try it yet?

bholland:
I have it on my to do list tonight. I will let you know but the change
looks like it should work.

lynxlynxlynx:
Closing; superseded by #860

Issue #873, opened 2020-07-18
[BG2] Characters pass through the decor


FrElvire:
#### Bug description
While playing with new pathfinding (It's way better than older one !! well done 👍 ) on areas where space is a little constrained, I see two of my characters going through the decor (at least it seems).

GemRB version: master

#### Steps to reproduce
1. load included save
2. do the same as in video

#### Expected behavior
All characters take stairs

#### Video
[path_shortcut-2020-07-18_12.13.25.zip](https://github.com/gemrb/gemrb/files/4941477/path_shortcut-2020-07-18_12.13.25.zip)

#### Save
[000000087-through decor bug.zip](https://github.com/gemrb/gemrb/files/4941478/000000087-through.decor.bug.zip)

FrElvire:
I can't reproduce it with old pathfinding code

kmfrick:
It's an oversight in `BlockSearchMap`, made apparent by the presence of bumping with the new pathfinder: blocking the searchmap overrides `PATH_MAP_IMPASSABLE`, if present. I'll address this in #874 .

Issue #871, opened 2020-07-17
Undefined behavior in `Map::RandomWalk`


bradallred:
#### Bug description
we have a casting overflow in `Map::RandomWalk`. `double dx = 3 * dxRand[I];` can return a negative value which we then implicitly cast though addition with a `short` when passing to `GetBlockedInRadius`. the problem is that `dx` is a `double` type, not an integer so this is not going to work consistently across architectures/compilers.

#### Steps to reproduce
1. build with UBsan
2. load presumably any save game
#### Expected behavior
UBsan shouldn't barf up errors

bradallred:
I'm also not sure why we are using `double` instead of `float` since the table is full of `float`s

lynxlynxlynx:
I'll leave the design bits to @kmfrick, but for me master doesn't even build with ubsan. It doesn't like the core->console calls in SDLVideo.cpp:

> /usr/bin/ld: CMakeFiles/SDLVideo.dir/SDLVideo.cpp.o:(.data.rel+0x158): undefined reference to `typeinfo for GemRB::Console'

Not sure what's going on, since the includes and linker flags are there.

bradallred:
sounds like maybe something is being built without RTTI enabled.

bradallred:
I'm not having the same issue, but perhaps [this stack overflow post](https://stackoverflow.com/questions/37358573/why-does-fsanitize-undefined-cause-undefined-reference-to-typeinfo) has something helpful.

lynxlynxlynx:
Ah, that did it, thanks. I got stuck in another thread, more about virtuals and hacking around it with -fno-rtti.

PR #874, opened 2020-07-18
Pathfinder bughunt

kmfrick:
## Description

Fixes a couple bugs with the pathfinder and tweaks `Map::FindPath` to provide straighter paths and work faster.
Addresses #871 and #873.
## Checklist

- [x] I extended the documentation, if necessary - no documentation necessary

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -2907,24 +2907,21 @@ void Map::BlockSearchMap(const Point &Pos, unsigned int size, unsigned int value
 				unsigned int ppypj = ppy+j;
 				unsigned int ppxmi = ppx-i;
 				unsigned int ppymj = ppy-j;
-				if ((ppxpi<Width) && (ppypj<Height)) {
-					unsigned int pos = ppypj*Width+ppxpi;
-					SrchMap[pos] = (SrchMap[pos]&PATH_MAP_NOTACTOR) | value;
+				unsigned int pos = ppypj * Width + ppxpi;
+				if (ppxpi < Width && ppypj < Height && SrchMap[pos] != PATH_MAP_IMPASSABLE) {
+					SrchMap[pos] = (SrchMap[pos] & PATH_MAP_NOTACTOR) | value;
 				}
-
-				if ((ppxpi<Width) && (ppymj<Height)) {
-					unsigned int pos = (ppymj)*Width+ppxpi;
-					SrchMap[pos] = (SrchMap[pos]&PATH_MAP_NOTACTOR) | value;
+				pos = ppymj * Width + ppxpi;
+				if (ppxpi < Width && ppymj < Height && SrchMap[pos] != PATH_MAP_IMPASSABLE) {
+					SrchMap[pos] = (SrchMap[pos] & PATH_MAP_NOTACTOR) | value;
 				}
-
-				if ((ppxmi<Width) && (ppypj<Height)) {
-					unsigned int pos = (ppypj)*Width+ppxmi;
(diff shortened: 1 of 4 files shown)


gemrb/core/Map.cpp
```diff
@@ -2053,7 +2053,7 @@ bool Map::IsVisibleLOS(const Point &s, const Point &d, const Actor *caller) cons
 bool Map::IsWalkableTo(const Point &s, const Point &d, bool actorsAreBlocking, const Actor *caller) const
 {
 	unsigned ret = GetBlockedInLine(s, d, true, caller);
-	return ret & (PATH_MAP_PASSABLE | (actorsAreBlocking ? 0 : PATH_MAP_ACTOR));
+	return ret & (PATH_MAP_PASSABLE | PATH_MAP_ACTOR | (actorsAreBlocking ? 0 : PATH_MAP_ACTOR));
```
lynxlynxlynx: actorsAreBlocking is now ignored, so if that's fine, the param and last boolean block can be removed.

kmfrick: It was a typo! Fixed in 3f8a1b5

gemrb/core/PathFinder.cpp
```diff
@@ -133,7 +133,17 @@ bool Map::TargetUnreachable(const Point &s, const Point &d, unsigned int size, b
...
-	return FindPath(s, d, size, 0, flags) == nullptr;
+	PathNode *path = FindPath(s, d, size, 0, flags);
+	bool targetUnreachable = path == nullptr;
+	if (!targetUnreachable) {
+		PathNode *thisNode = path;
+		while (thisNode) {
+			PathNode *nextNode = thisNode->Next;
+			delete(thisNode);
+			thisNode = nextNode;
+		}
```
lynxlynxlynx, line 144: This would fit nicely in a destructor. Reading up on c++11 POD changes, it would even still keep PathNode POD, since it would satisfy "standard layout" — if I understand correctly.

lynxlynxlynx: I guess it would have to be selective though, since most of the time we don't want to wipe everything that's linked. :)

gemrb/core/Map.cpp
```diff
@@ -2907,23 +2907,23 @@ void Map::BlockSearchMap(const Point &Pos, unsigned int size, unsigned int value
 				unsigned int ppypj = ppy+j;
 				unsigned int ppxmi = ppx-i;
 				unsigned int ppymj = ppy-j;
-				if ((ppxpi<Width) && (ppypj<Height)) {
-					unsigned int pos = ppypj*Width+ppxpi;
+				unsigned int pos = ppypj*Width+ppxpi;
```
lynxlynxlynx, line 2910: Please use the occasion to also fix the whitespace of the lines you edit. This our main way to harmonize style without polluting history.

gemrb/core/PathFinder.cpp
lynxlynxlynx: I guess again, if the idea is feasible at all, it would better fit in a standalone method and then there's no fallout.

kmfrick: We already have `Movable::ClearPath`, which is where I copied this from. There are no other uses of PathNode than for Movables iirc and even then I'd rather just get a move on and replace PathNode with an `std::list`

lynxlynxlynx: ok

gemrb/core/Map.cpp
lynxlynxlynx: next time please only do **the lines you edit**, otherwise it's no different.

lynxlynxlynx:
Thanks!

PR #906, opened 2020-08-17
Movable: Minor bugfixes

kmfrick:
## Description
- Makes `Movable::DoStep` not abandon path when stopping because of a door: actors should try and re-path, if possible.
- Makes `Movable::WalkTo` only be rate-limited while moving, since the purpose of rate-limiting is avoiding actors being stuck on a moving target.

## Checklist

- [x] I extended the documentation, if necessary N/A


lynxlynxlynx:
Thanks!

PR #915, opened 2020-08-25
VCPkg integration update

nerd-rage:
Adjustments for VCPkg integration so GemRB sets up without configuring anything but SDL2/OpenGL frontend choices when its prerequisites are installed.


lynxlynxlynx:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gemrb/gemrb/pullRequest?prid=6066881&bid=19868835)

lynxlynxlynx:
Thanks ... it opens up several questions, but I'll add them to the commits.

gemrb/plugins/GUIScript/GUIScript.h
```diff
@@ -23,13 +23,7 @@
 
 // NOTE: Python.h has to be included first.
 
-#if defined(WIN32) && defined(_DEBUG)
-#undef _DEBUG
 #include <Python.h>
-#define _DEBUG
-#else
-#include <Python.h>
-#endif
```
lynxlynxlynx, line 32: Probably needed for older python, but it doesn't matter any more. 👍🏾

gemrb/plugins/SDLVideo/Matrix.cpp
```diff
@@ -1,3 +1,4 @@
+#define _USE_MATH_DEFINES
```
lynxlynxlynx, line 1: Isn't this relevant only for MSVC? I think it's better to include win32def.h or better yet, split out the M_PI stuff in a separate header and include that.

gemrb/vcpkg_deps.cmake
```diff
@@ -29,6 +29,8 @@ ENDIF()
 # libcharset is just a relic of the current iconv build
 LIST(APPEND DLL_SET
 	${VCPKG_DATAROOT}/bin/SDL${DLL_SDL_VER}.dll # Cmake doesn't actually find the sdl debug libs
+	${DLL_DIR}SDL2_mixer.dll
```
lynxlynxlynx: This should use ${DLL_SDL_VER}, since there's a mixer also for sdl1.

gemrb/vcpkg_deps.cmake
```diff
@@ -11,68 +11,64 @@ MESSAGE(STATUS "")
...
-	${DLL_DIR}libiconv.dll )
-
-FOREACH(ENTRY IN LISTS DLL_SET)
-
-	IF(NOT EXISTS ${ENTRY})
-		LIST(REMOVE_ITEM DLL_SET ${ENTRY})
-		CONTINUE()
+LIST(APPEND DLL_SET_DEBUG
+	brotlicommon.dll
+	brotlidec.dll
```
lynxlynxlynx, line 21: Why is brotli needed?

gemrb/vcpkg_deps.cmake
```diff
@@ -11,68 +11,64 @@ MESSAGE(STATUS "")
...
+	glew32d.dll
+	libcharset.dll
+	libiconv.dll
+	libpng16d.dll
+	ogg.dll
+	OpenAL32.dll
+	python27_d.dll
+	SDL.dll
+	SDL2d.dll
+	SDL2_mixer.dll
```
lynxlynxlynx, line 33: Missing SDL_mixer.dll x2

gemrb/vcpkg_deps.cmake
```diff
@@ -11,68 +11,64 @@ MESSAGE(STATUS "")
...
+	vorbis.dll
+	vorbisfile.dll
+	zlibd1.dll )
+
```
lynxlynxlynx, line 37: Are you sure this mixing of debug and normal libraries produced working binaries? I think it didn't work for @chilvence, hence the code.

gemrb/vcpkg_deps.cmake
```diff
@@ -11,68 +11,64 @@ MESSAGE(STATUS "")
...
-# after checking dependencies with dumpbin, turns out it is looking for a vorbisfile.dll
-# this may be because the .lib file differs in name from the .dll file
-IF(EXISTS ${DLL_DIR}vorbisfile-1.dll )
-
-	ADD_CUSTOM_COMMAND(TARGET gemrb POST_BUILD
-		COMMAND ${CMAKE_COMMAND} -E copy_if_different
-		${DLL_DIR}vorbisfile-1.dll
-		${CMAKE_BINARY_DIR}/gemrb/vorbisfile.dll)
-
-	INSTALL(FILES ${DLL_DIR}vorbisfile-1.dll DESTINATION ${BIN_DIR} RENAME vorbisfile.dll)
```
lynxlynxlynx, line 71: Is this workaround not needed any more? If it was fixed upstream in vcpkg, great, but let's make sure the plugin still works.

gemrb/plugins/GUIScript/GUIScript.h
nerd-rage: Sadly—as many headers do when they see MSVC—Python.h links to the export lib with a pragma, and it causes a mismatch with Python's debug DLL when fooled this way.

CMakeLists.txt
```diff
@@ -268,7 +268,15 @@ IF((SDL_BACKEND STREQUAL "SDL") AND NOT USE_SDL2)
 ENDIF()
 
 IF((SDL_BACKEND STREQUAL "SDL2") OR USE_SDL2)
-	INCLUDE(FindSDL2)
+	IF(WIN32 AND _VCPKG_INSTALLED_DIR)
+		find_package(SDL2 CONFIG REQUIRED)
+		GET_TARGET_PROPERTY(SDL2_INCLUDE_DIR SDL2::SDL2 INTERFACE_INCLUDE_DIRECTORIES)
```
lynxlynxlynx, line 273: Which version of cmake does this syntax require? We currently set the minimum to 3.1, but I'm not sure that's enough (no reason not to bump it though).

Also, can't this just replace FindSDL2? SDL2 now ships its own and any platform hacks can be done here.

gemrb/plugins/SDLVideo/Matrix.cpp
nerd-rage: A valid point, although it's been around for so long I doubt that any other library uses it differently. Those constants should be replaced in C++20, and ideally everyone should use a retroport of the new header.

CMakeLists.txt
```diff
@@ -342,7 +350,15 @@ IF((NOT OPENGL_BACKEND STREQUAL "None") OR USE_OPENGL)
...
-		INCLUDE(FindGLEW)
+		IF(WIN32 AND _VCPKG_INSTALLED_DIR)
+			find_package(GLEW REQUIRED)
+			GET_TARGET_PROPERTY(GLEW_INCLUDE_DIR GLEW::GLEW INTERFACE_INCLUDE_DIRECTORIES)
+			GET_TARGET_PROPERTY(GLEW_LIBRARY_DEBUG GLEW::GLEW IMPORTED_IMPLIB_DEBUG)
+			GET_TARGET_PROPERTY(GLEW_LIBRARY_RELEASE GLEW::GLEW IMPORTED_IMPLIB_RELEASE)
+			SET(GLEW_LIBRARIES debug ${GLEW_LIBRARY_DEBUG} optimized ${GLEW_LIBRARY_RELEASE})
+		ELSE()
+			INCLUDE(FindGLEW)
+		ENDIF()
```
lynxlynxlynx, line 361: Similarly here, except that cmake itself ships FindGLEW.

gemrb/vcpkg_deps.cmake
nerd-rage: I haven't found SDL mixer in VCPkg, is it a part of some other package?

gemrb/vcpkg_deps.cmake
nerd-rage: Freetype's dependencies.

gemrb/vcpkg_deps.cmake
nerd-rage: Those are debug ones. Not all debug libraries have the “d” suffix. And some have a different one (like Python with its “_d”. That's why I've spit the lists.

gemrb/vcpkg_deps.cmake
nerd-rage: Yes, I've checked.

CMakeLists.txt
nerd-rage: I've tried to avoid intrusion into non VCPkg paths, but yes, the CMake modules should probably be looked into.

CMakeLists.txt
nerd-rage: IConv as well?

CMakeLists.txt
lynxlynxlynx: To me it looks like we can just remove it and make the new block the default.

CMakeLists.txt
lynxlynxlynx: Would require a cmake bump first. Not sure when it was added, but it's not in 3.1.:
https://cmake.org/cmake/help/v3.11/module/FindIconv.html

gemrb/plugins/GUIScript/GUIScript.h
lynxlynxlynx: So does this still work with both release and debug builds?

nerd-rage: Yes, but now the debug build uses the debug Python library.

lynxlynxlynx: Ok, but this is code for everywhere, so I wonder if it still builds with MSYS2. Looking at their packages, it seems they build only the release version. @MarcelHB, can you check if this PR compiles?

CMakeLists.txt
nerd-rage: Well, I wasn't able to test on other platforms, so I dared not.

lynxlynxlynx: We have buildbots for that, but also, the cmake versopm question remains open. AppVeyor ships a very recent version, so the results are not conclusive.

gemrb/plugins/GUIScript/GUIScript.h
nerd-rage: I've tried to check it myself, but MSYS2 and MinGW toolchains seem to always build the release version, which works.

CMakeLists.txt
nerd-rage: Could it be its own PR? After a more thorough investigation.

gemrb/plugins/GUIScript/GUIScript.h
MarcelHB: Can confirm, builds and starts.

lynxlynxlynx: Great, thanks!

CMakeLists.txt
lynxlynxlynx: The cleanup, sure, but it's still important to know if we need to bump the minimal cmake version.

nerd-rage: Oh, all the the syntax is pre-3.0 as far as I understand.

lynxlynxlynx:
Thanks! Please do investigate that little cleanup too.

CMakeLists.txt
nerd-rage: The use of IConv is very limited and arguably not needed on Windows with its own built-in codepage API. Many Unix descendants have it inside their C runtime too. Seems like a weak candidate to bump CMake version up for.

lynxlynxlynx: Agreed, especially since it looks like I nailed the version, which is from 2018.

Issue #901, opened 2020-08-09
 Refactor Actor::CalculateSpeed further


lynxlynxlynx:
Small refactoring todo, dumping notes here while working on something else:
- set IE_ENCUMBRANCE from CalculateWeight, not CalculateSpeed (Owner holds the actor to use)
- guiscript: remove calls or if the above somehow wasn't enough, call inventory.CalculateWeight() instead of CalculateSpeed
- repurpose (move the whole Map::UpdateScripts() block dealing with speed here)

PR #917, opened 2020-08-26
Refactor Actor::CalculateSpeed (WIP)

anewusername:
## Description

I took an initial pass at #901 .

At first glance most of the GUIScript.cpp calls (namely to `Inventory::RemoveItem` and `Inventory::AddSlotItem`) don't recalculate the weight, so there's not much simplification there. I will try to take a look at their other calling points to check if they could do so, unless someone recommends against it.

## Checklist

### DIFF
--- a/gemrb/core/Inventory.cpp
+++ b/gemrb/core/Inventory.cpp
@@ -160,7 +160,7 @@ void Inventory::AddItem(CREItem *item)
 	CalculateWeight();
 }
 
-void Inventory::CalculateWeight() const
+void Inventory::CalculateWeight()
 {
 	Weight = 0;
 	for (size_t i = 0; i < Slots.size(); i++) {
@@ -1500,8 +1500,6 @@ void Inventory::dump(StringBuffer& buffer) const
 	}
 
 	buffer.appendFormatted("Equipped: %d       EquippedHeader: %d\n", Equipped, EquippedHeader);
-	Changed = true;
-	CalculateWeight();
 	buffer.appendFormatted( "Total weight: %d\n", Weight );
 }
 
--- a/gemrb/core/Inventory.cpp
+++ b/gemrb/core/Inventory.cpp
@@ -94,7 +94,6 @@ Inventory::Inventory()
 {
 	Owner = NULL;
 	InventoryType = INVENTORY_HEAP;
-	Changed = false;
 	Weight = 0;
 	Equipped = IW_NO_EQUIPPED;
 	EquippedHeader = 0;
@@ -139,7 +138,6 @@ void Inventory::CopyFrom(const Actor *source)
 	Equipped = source->inventory.GetEquipped();
 	EquippedHeader = source->inventory.GetEquippedHeader();
 
-	Changed = true;
 	CalculateWeight();
 }
 
@@ -151,7 +149,7 @@ CREItem *Inventory::GetItem(unsigned int slot)
 	}
 	CREItem *item = Slots[slot];
 	Slots.erase(Slots.begin()+slot);
-	Changed = true;
+	CalculateWeight();
 	return item;
 }
 
@@ -159,14 +157,11 @@ void Inventory::AddItem(CREItem *item)
 {
 	if (!item) return; //invalid items get no slot
 	Slots.push_back(item);
-	Changed = true;
+	CalculateWeight();
 }
 
 void Inventory::CalculateWeight() const
 {
-	if (!Changed) {
-		return;
-	}
 	Weight = 0;
 	for (size_t i = 0; i < Slots.size(); i++) {
 		CREItem *slot = Slots[i];
@@ -195,7 +190,10 @@ void Inventory::CalculateWeight() const
 			Weight += slot->Weight * ((slot->Usages[0] && slot->MaxStackAmount) ? slot->Usages[0] : 1);
 		}
 	}
-	Changed = false;
+
+	if (Owner) {
+		Owner->SetBase(IE_ENCUMBRANCE, Weight);
+	}
 }
 
 void Inventory::AddSlotEffects(ieDword index)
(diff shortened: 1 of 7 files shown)


gemrb/core/Inventory.cpp
```diff
@@ -196,6 +196,8 @@ void Inventory::CalculateWeight() const
 		}
 	}
 	Changed = false;
+
+	Owner->SetStat(IE_ENCUMBRANCE, Weight, false);
```
lynxlynxlynx, line 200: This needs to be changed to SetBase, so we set the base stat. SetStat sets the modified one, which is wiped each game tick. This is probably why you couldn't just delete the calls from GUIScript (which is the way to go).

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -5439,15 +5435,42 @@ int Actor::GetEncumbranceFactor(bool feedback) const
...
 	}
-
-	int encumbranceFactor = GetEncumbranceFactor(feedback);
-	return speed / encumbranceFactor;
+	if (movementRate) {
+		speed = 1500 / movementRate;
+	} else {
+		speed = 0;
+	}
+	if (core->HasFeature(GF_RESDATA_INI)) {
```
lynxlynxlynx, line 5451: This block is completely separate even in Map. Since the function does two unrelated things at once, the proper thing is two add two more private functions and have CalculateSpeed basically be:
```
if (core->HasFeature(GF_RESDATA_INI)) {
  CalculateSpeedFromIni(feedback);
} else {
  CalculateSpeedFromRate(feedback);
}
```

lynxlynxlynx:
Thanks, a great start! Left some comments, so it can be improved.

gemrb/core/Inventory.cpp
anewusername: I added a commit changing it SetBase, but I'm not sure if it's necessary.
I think I was misunderstanding when CalculateWeight should be called (immediately before `GetWeight()`, gated by `Changed`, rather than on inventory update).
I also removed a few calls to CalculateWeight within other inventory functions to keep them consistent, since I couldn't find anywhere that the weight was accessed without first being recalculated.

lynxlynxlynx: It feels a bit wrong — shouldn't be up to the inventory to keep track of its own state? Division of responsibilities and all that. Now it works only because the recalc is called each tick, so any changes are quickly picked up and there's at most a tick's delay.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -5436,22 +5433,57 @@ int Actor::GetEncumbranceFactor(bool feedback) const
...
 
-// NOTE: for ini-based speed users this will only update their encumbrance, speed will be 0
 int Actor::CalculateSpeed(bool feedback)
 {
-	int speed = GetStat(IE_MOVEMENTRATE);
+	if (core->HasFeature(GF_RESDATA_INI)) {
+		CalculateSpeedFromINI(feedback);
+	} else {
+		CalculateSpeedFromRate(feedback);
+	};
```
lynxlynxlynx, line 5442:
```suggestion
	}
```

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -5436,22 +5433,57 @@ int Actor::GetEncumbranceFactor(bool feedback) const
...
+{
 	int encumbranceFactor = GetEncumbranceFactor(feedback);
-	SetStat(IE_ENCUMBRANCE, encumbrance, false);
-
-	return speed / encumbranceFactor;
+	ieDword animid = BaseStats[IE_ANIMATION_ID];
+	if (core->HasFeature(GF_ONE_BYTE_ANIMID)) {
+		animid = animid & 0xff;
+	}
+	if (animid < (ieDword)CharAnimations::GetAvatarsCount()) {
```
lynxlynxlynx, line 5470: Since we're cleaning up, let's just make this an assert, since it should always be true. Will reduce the nesting by one level as well.

gemrb/core/Inventory.cpp
lynxlynxlynx: SetBase is needed even more than I thought too — if the inventory didn't change, nothing will set the stat that tick, so it will equal the base ... which is unset, so 0.

anewusername: I forgot to add, maybe the right approach would be to call CalculateWeight from GetWeight?
That would ensure you don't get a stale weight, while encapsulating things better. Maybe `Weight` should also be private in that case.

lynxlynxlynx: Weight should definitely remain private — for anyone that needs it, there's a getter already. CalculateWeight is a bit annoying, since it also does some flag cleanup, but it also rechecks everything (we don't update the weight atomically), which saves us having to deal with both the removed and new item if a swap occurred. I don't think it's worth touching it.

But I do think it's better to keep GetWeight const and make the effort to maintain an always correct Weight. `Changed` is there only for the weight handling, so it's easy to see where the updates should happen. Not sure why these two are marked mutable either.

anewusername: I guess I see two paths for a clean interface:
- Use `Weight` to privately cache the value, invalidate using `Changed`, interface using `GetWeight`
- Remove `Changed`, replace Changed=True with calls to `CalculateWeight`. Less performant, but I doubt it'll make a difference.

anewusername:
I refactored with the objective of keeping the weight/encumbrance updated on each slot update, and removed `Changed`.

I think it's reasonable to make `CalculateWeight` private and not const, since at this point it shouldn't need to be used externally.

lynxlynxlynx:
Looks good, thanks. Tried to break it, but didn't succeed, so even better. :)

Issue #902, opened 2020-08-10
Add middle-mouse viewport scrolling


lynxlynxlynx:
A simple feature from later EEs.
https://us.v-cdn.net/5019558/uploads/editor/pf/ndxlo8290ngx.gif

PR #919, opened 2020-08-27
Add middle-mouse viewport scrolling #902

Balrog994:
## Description
I was having a couple of free hours and was taking a look at the issue tracker and found a very simple enhancement I thought I could help with :)

Just tried to implement #902 in the current branch without touching anything critical.
I saw it was under development on the subviews branch but maybe the code can be used as a temporary implementation.

I tested the implementation and seems working but some more testing may be useful.

There is a *breaking change in the code for text pasting*. I had to require the user to keep CTRL pressed for pasting because it was interfearing with the middle mouse button handling in the event system.

## Checklist

### DIFF
--- a/gemrb/core/GUI/EventMgr.h
+++ b/gemrb/core/GUI/EventMgr.h
@@ -63,6 +63,7 @@ class Window;
 
 // Mouse buttons
 #define GEM_MB_ACTION           1
+#define GEM_MB_PAN	            2
 #define GEM_MB_MENU             4
 #define GEM_MB_SCRLUP           8
 #define GEM_MB_SCRLDOWN         16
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
Thanks, yet unfortunately this is mostly wasted effort — as you have noted, we've already implemented the feature for the next release and it was possible to do it more cleanly with the new code. Middle-click paste has been a linux staple since before I can remember as well.

I do like the idea of mouse capture, so you don't have to go back and forth than much, but even that should be done against the subviews branch.

Balrog994:
No problem, I was just having fun taking a look at the actual codebase and changing somthing around 😄 
Feel free to delete this pull request!

Issue #542, opened 2019-12-19
[bug] PST AR0605, game sporadically chokes on 0605ffg.bcs


chilvence:
A script execution error running Grace's 'introduction' script crashes the game on random occasions. The effect is not easy to reproduce, but sometimes occurs immediately after loading this save game file :

[000000199-GRB Pretentious Presupposition.zip](https://github.com/gemrb/gemrb/files/3986069/000000199-GRB.Pretentious.Presupposition.zip)

GemRB version: any

1. Go to AR0605
2. Wait
3. If Waiting doesn't work, try to save and reload the game. The nature of this bug makes it very hard to actively cause it to happen.

chilvence:
I thought I could escape this bug by emptying the aforementioned script, but I just had a crash with a different script in the same area:

```
[ResourceManager]: Found '0605pt3b.bcs' in 'chitin.key'.
[GameScript/ERROR]: Aborting response execution due to object deletion.
                                                                          This should not happen and we need to fix it.
```

So hopefully those two scripts have something obviously in common to go on. Otherwise my only other idea is to fill the script parser with print statements until something makes itself obvious...

bradallred:
it is 100% reproducible just by loading the game if you build with asan 😁 

```
==31569==ERROR: AddressSanitizer: heap-use-after-free on address 0x60400004a9e8 at pc 0x0001014bd575 bp 0x7ffeefbfc670 sp 0x7ffeefbfc668
READ of size 8 at 0x60400004a9e8 thread T0
2019-12-19 19:20:19.287681-0700 atos[31580:19766455] examining /Users/USER/Library/Developer/Xcode/DerivedData/GemRB-bmltmhdkakvalidorcnktnkenaub/Build/Products/Debug/GemRB.app/Contents/MacOS/GemRB [31569]
    #0 0x1014bd574 in std::__1::vector<GemRB::Action*, std::__1::allocator<GemRB::Action*> >::size() const vector:657
    #1 0x1014bd66e in GemRB::Response::Execute(GemRB::Scriptable*) GameScript.cpp:2370
    #2 0x1014bcb85 in GemRB::ResponseSet::Execute(GemRB::Scriptable*) GameScript.cpp:2340
    #3 0x1014bc5a8 in GemRB::GameScript::Update(bool*, bool*) GameScript.cpp:2079
    #4 0x101362425 in GemRB::Scriptable::ExecuteScript(int) Scriptable.cpp:424
[...]

PR #926, opened 2020-09-01
Guard against scripts committing suicide before they're done running

anewusername:
## Description
Fix for #542 .
I added the same logic to both `SetScript` variants, even though I'm not sure the 2-arg variant is currently callable from within a script.

I think this eliminates the need for the `Canary` stuff; I can remove that in another commit or submit a separate PR if desirable.

## Checklist

### DIFF
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -2422,14 +2422,6 @@ int Response::Execute(Scriptable* Sender)
 {
 	int ret = 0; // continue or not
 	for (size_t i = 0; i < actions.size(); i++) {
-		if (!CheckCanary()) {
-			// FIXME: hack to prevent crashing when a script deletes itself.
-			// this object has been deleted and this should not be considered a fix (it may cause unforseen problems too).
-			Log(ERROR, "GameScript", "Aborting response execution due to object deletion.\n \
-									  This should not happen and we need to fix it.");
-			ret = 0;
-			break;
-		}
 		Action* aC = actions[i];
 		switch (actionflags[aC->actionID] & AF_MASK) {
 			case AF_IMMEDIATE:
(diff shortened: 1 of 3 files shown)


gemrb/core/GameScript/GameScript.cpp
```diff
@@ -2067,6 +2069,9 @@ static Condition* ReadCondition(DataStream* stream)
  * if you pass non-NULL parameters, continuing is set to whether we Continue()ed
  * (should start false and be passed to next script's Update),
  * and done is set to whether we processed a block without Continue()
+ *
+ * NOTE: After calling, callers should deallocate this object if it if `dead==true` is set.
```
lynxlynxlynx, line 2073:
```suggestion
 * NOTE: After calling, callers should deallocate this object if `dead == true`.
```

gemrb/core/GameScript/GameScript.cpp
```diff
@@ -1845,6 +1845,8 @@ GameScript::GameScript(const ieResRef ResRef, Scriptable* MySelf,
 	strnlwrcpy( Name, ResRef, 8 );
 
 	script = CacheScript( Name, AIScript);
+	dead = false;
+	running = false;
```
lynxlynxlynx, line 1849: C++11 supports initing in the declaration, which is both cleaner and a Sonar warning.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -204,7 +208,11 @@ void Scriptable::SetScript(int index, GameScript* script)
 		Log(ERROR, "Scriptable", "Invalid script index!");
 		return;
 	}
-	delete Scripts[index];
+	if (Scripts[index] && Scripts[index]->running) {
+		Scripts[index]->dead = true;
+	} else {
+		delete Scripts[index];
+	}
```
lynxlynxlynx, line 215: I've removed this function yesterday, since it's unused, just haven't pushed, but I think git will be smart enough to not care.

lynxlynxlynx:
You can remove the CheckCanary block already here. Thanks!

lynxlynxlynx:
Thank you, very nice!

PR #933, opened 2020-09-02
add OpenGL shaders versioning for GLSL compatibility

### LINKED ISSUE #932: GLSL Shader error [GLES3 - Raspberry Pi4]
#### Description
(Originally reported [here](https://retropie.org.uk/forum/topic/26864/gemrb-support/22), by a RetroPie user. My initial diagnostic is not 100% correct there)

Trying to start GemRB on a Pi4, built with OpenGLES support and SDL2, leads to the following error during shader initialization:
```
...
Core]: Initializing GUI Script Engine...
[SDL 2 GL Driver]: Creating display
[SDL 2 GL Driver]: Creating Main Surface: w=1024 h=768 fmt=SDL_PIXELFORMAT_RGBA8888
[SDL 2 GL Driver/FATAL]: Can't build shader program: 0:1(1): error: syntax error, unexpected NEW_IDENTIFIER
...
```
 The reason I think it fails is because the shaders syntax is specific to the GLSL version corresponding to OpenGL(ES) 2.0 (GLSL 1.x), but are incorrect when the GLSL version resulting from the GL context initialization is higher (in this case is GLSL ES 3.10). 
 When the GL context is initialized here: https://github.com/gemrb/gemrb/blob/ac595aae4d4f0b735635d2c60e1af8d8aa64e75a/gemrb/plugins/SDLVideo/SDL20GLVideo.cpp#L50-L56

 the 2.0 version (ES or not) is requested, but the version requested is just a minimum guarantee, the resulting GL context can be different than OpenGL 2.0 (see https://wiki.libsdl.org/SDL_GLattr#OpenGL).
 
 In this particular case, adding a few debug lines to show the GL(SL) versions show:

* building with `-DOPENGL_BACKEND=GLES`, the GL context is created with GLES 3.1 and GLSL ES 3.10, GemRB fails with the aforementioned error:

```
  [Core]: Initializing GUI Script Engine...
  [SDL 2 GL Driver]: Creating display
  [SDL 2 GL Driver]: Creating Main Surface: w=1024 h=768 fmt=SDL_PIXELFORMAT_RGBA8888
  [SDL 2 GL Driver]: OpenGL version: OpenGL ES 3.1 Mesa 20.2.0-devel (git-e0518800a1), renderer: V3D 4.2, vendor: Broadcom
  [SDL 2 GL Driver]:   GLSL version: OpenGL ES GLSL ES 3.10
  [SDL 2 GL Driver/FATAL]: Can't build shader program: 0:1(1): error: syntax error, unexpected NEW_IDENTIFIER
```
My suspicion is that GLSL ES 3.10 doesn't like the `precision` line (deprecated in newer versions ?). 

* building with `-DOPENGL_BACKEND=OpenGL`, the GL context is created with OpenGL 2.1 and GLSL 1.20 and GemRB starts ok:

```
 [SDL 2 GL Driver]: OpenGL version: 2.1 Mesa 20.2.0-devel (git-e0518800a1), renderer: V3D 4.2s, vendor: Broadcom
 [SDL 2 GL Driver]:   GLSL version: 1.20
```

**GemRB version**: 0.8.7, but happens on `master` or previous versions also.

#### Reproducing
Test on a system supporting GLES 3.x (for instance on a Raspberry Pi 4, running the latest Raspbian/Raspberry Pi OS).

* build `0.8.7` or `master` with
```
-DCMAKE_BUILD_TYPE=Release -DFREETYPE_INCLUDE_DIRS=/usr/include/freetype2/ -DSDL_BACKEND=SDL2   -DCMAKE_INSTALL_PREFIX=/opt/retropie/ports/gemrb -DUSE_SDLMIXER=OFF -DOPENGL_BACKEND=GLES -DDISABLE_WERROR=1
```
* install
* try to run a game

#### Fix (?)

If the GLSL version is added in the shader code, the shaders are compiled and loaded fine, regardless of the OpenGL backend used.
I've tested with [this patch](https://github.com/cmitu/gemrb/commit/df493a7da7251eb5ea33b62a65580d09b440dd74) and it seems ok.

cmitu:
The GL context initialization in 'GLVideoDriver' requests OpenGL (ES) 2.0 from SDL2,
but SDL2 doesn't guarantee the context returned will have the same version [1].

To avoid any compatibility issues with a superior GLSL version, add the GLSL version to the shader code,
depending on the GL context requested (non-ES vs. ES).
 * when expecting an OpenGL 2.0 context, set version to 110
 * when expecting an OpenGL ES 2.0 context, set version to 100

[1] - https://wiki.libsdl.org/SDL_GLattr#OpenGL

- Fixes #932.
- Minor: 
 * replaced a few error messages (_couldnt .._ replaced with the more ubiquotus _Unable to_)
 * added a couple of log messages about the initialized GL context.

====

### DIFF
--- a/gemrb/plugins/SDLVideo/GLSLProgram.cpp
+++ b/gemrb/plugins/SDLVideo/GLSLProgram.cpp
@@ -102,10 +102,16 @@ GLSLProgram* GLSLProgram::CreateFromFiles(std::string vertexSourceFileName, std:
 	return GLSLProgram::Create(vertexContent, fragmentContent);
 }
 
-GLuint GLSLProgram::buildShader(GLenum type, std::string source)
+GLuint GLSLProgram::buildShader(GLenum type, const std::string source)
 {
+	std::string shader_source = source;
+#ifdef USE_GL
+	shader_source.insert(0, "#version 110\n");
+#else
+	shader_source.insert(0, "#version 100\n");
+#endif
     GLuint id = glCreateShader(type);
-	const char* src = source.c_str();
+	const char* src = shader_source.c_str();
 	glShaderSource(id, 1, &src, 0);
     glCompileShader(id);
     GLint result = GL_FALSE;
--- a/gemrb/plugins/SDLVideo/SDL20GLVideo.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20GLVideo.cpp
@@ -66,23 +66,25 @@ int GLVideoDriver::CreateDisplay(int w, int h, int bpp, bool fs, const char* tit
 	window = SDL_CreateWindow(title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, width, height, winFlags);
 	if (window == NULL) 
 	{
-		Log(ERROR, "SDL 2 GL Driver", "couldnt create window:%s", SDL_GetError());
+		Log(ERROR, "SDL 2 GL Driver", "Unable to create window:%s", SDL_GetError());
 		return GEM_ERROR;
 	}
 
 	context = SDL_GL_CreateContext(window);
 	if (context == NULL) 
 	{
-		Log(ERROR, "SDL 2 GL Driver", "couldnt create GL context:%s", SDL_GetError());
+		Log(ERROR, "SDL 2 GL Driver", "Unable to create GL context:%s", SDL_GetError());
 		return GEM_ERROR;
 	}
 	SDL_GL_MakeCurrent(window, context);
+	Log(MESSAGE, "SDL 2 GL Driver", "OpenGL version: %s, renderer: %s, vendor: %s", glGetString(GL_VERSION), glGetString(GL_RENDERER), glGetString(GL_VENDOR));
+	Log(MESSAGE, "SDL 2 GL Driver", "  GLSL version: %s", glGetString(GL_SHADING_LANGUAGE_VERSION));
 
 	renderer = SDL_CreateRenderer(window, -1, 0);
 
 	if (renderer == NULL) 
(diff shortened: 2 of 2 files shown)


lynxlynxlynx:
Thanks!

bradallred:
I have reverted this in the subviews branch because it breaks OpenGL for me

lynxlynxlynx:
Can you be more specific?

MarcelHB:
I remember trouble that Mac is not backwards compatible when it comes to OpenGL in certain context setups. As a PoC, you can maybe try `#version 150`.

bradallred:
after additional scrutiny, it works fine as is. The issue is `#ifdef USE_GL` isn't set. I'm supposing then, that we set `USE_GL` for OpenGL and otherwise assume OpenGLES2?

MarcelHB:
In a second thought, when ending up in this place we are already know that we will build OGL shaders, thus `USE_GL` sounds like a bad name to use here anyway.

bradallred:
yes, it also looks like we already do `OPENGL_BACKEND=GLES` so we really ought to just compare against that.

lynxlynxlynx:
OPENGL_BACKEND is newer and the rest of the non-cmake code was never cleaned up for it. So yeah, go forth and then the forcing you just added to the xcode project can go away as well. You're also setting OPENGL_BACKEND wrong. Same goes for manually adding -DHAVE_ICONV=1 — it shouldn't be needed any more.

bradallred:
> You're also setting OPENGL_BACKEND wrong

meaning?

lynxlynxlynx:
You're setting it to 1 or 2, not the strings.

PR #936, opened 2020-09-08
OpenGL Backend: updates for better GLES integration

cmitu:
Modifications in the commit try to improve OpenGL ES integration during the build/compilation step and add versioning for the GLSL shaders for better compatibility.

### Build system (cmake) modifications:
 * simplified building for the Raspberry Pi platform. RPI models supports either GLES2 or (some of them) OpenGL 2.1.
   A Raspberry Pi (Raspbian/Raspberry Pi OS no) can use:
    - the Broadcom proprietary drivers/GLES libraries, located in `/opt/vc` (Models from 0 to 3, support only GLES 2).
    - the opensource Mesa VC4/VC5 driver, which is similar to any x86/x86_64 platform using an opensource driver (Models 3/4 - supporting GLES2/GLES3 and OpenGL 2.1).
  The default driver/GLES for all models < 4 is the legacy Broadcom drivers/libraries and the MESA driver for 4.

  The build system prefers the Broadcom GLES libraries when a RPI system is detected, but this can be overriden by adding `-DDISABLE_VIDEOCORE=1` - and is necessary for the Pi4.

 **NOTE**: similar to how the RPI is set-up, support could be added for other SBC platforms (which are GLES exclusive) - by providing hints for library location and names (i.e. some Odroid platforms use `libMali.so` for GLES library).

  * added a GLES library/header detection module, similar to `FindOpenGL`, in order to link to the correct libraries, added also explicit linkage to the proper API library

  * when enabling `OPENGL_BACKEND`, add `USE_OPENGL_BACKEND` and a separate pre-processor define, based on the GL API used (USE_OPENGL_API/USE_GLES_API)

### SDLVideo plugin modifications:
  * incorporate the shader GLSL versioning from df493a7da7251eb5ea33b62a65580d09b440dd74:
    ** when expecting an OpenGL 2.0 context, set version to 110
    ** when expecting an OpenGL ES 2.0 context, set version to 100
  * remove GLEW headers dependency, use SDL2's headers for OpenGL/GLES.
  * request a GL context that's at least OpenGL2, just to prevent any older systems that might run OpenGL 1.4 (very unlikely, but happened a few times in RetroPie).
  * minor - add a function to determine the runtime SDL2 version. 

 **NOTES/Caveats**:
  - with the changes added, GLES is correctly detected and compiled in on the Pi4, but at runtime, the new ` SDLTextureV.glsl` shader cannot be compiled and and GemRB stop. Looks like the GLSL ES versions do not support some of the built-ins used in the shader. Error:
```
[SDL20Video/DEBUG]: Renderer: opengles2
[SDL 2 GL Driver]: OpenGL version: OpenGL ES 3.1 Mesa 20.2.0-devel (git-e0518800a1), renderer: V3D 4.2, vendor: Broadcom
[SDL 2 GL Driver]:   GLSL version: OpenGL ES GLSL ES 3.10
[SDL 2 GL Driver/FATAL]: Can't build shader program: 0:7(15): error: `gl_ModelViewProjectionMatrix' undeclared
0:7(46): error: `gl_Vertex' undeclared
0:7(15): error: operands to arithmetic operators must be numeric
0:8(11): error: `gl_Color' undeclared
0:9(19): error: `gl_MultiTexCoord0' undeclared
0:9(14): error: cannot construct `vec2' from a non-numeric data type
```
  - I haven't tested outside of Linux, so the elimination of GLEW might affect the Windows build (?)
  - Raspberry Pi related:
     * the location of the Broadcom libraries might change in the future (`/opt/vc`), their new beta version of Raspberry Pi OS already changed that, but right now that version is 64bit only (`aarch64`). If that's the case, perhaps a `-DRPI=true` might be necessary.
     * the OpenGL MESA driver/library that enables OpenGL can be used on the RPI3 models, but needs explicit configuration (modifying the bootloader configuration `config.txt` and enabling the `v3d-kms-rpi3` overlay). It's not a common configuration, but it can happen. In this case, the build should use `-DDISABLE_VIDEOCORE=1`, just like the Pi4 model.


bradallred:
@MarcelHB can you give some feedback here? Specifically about GLEW, but also in general.

CMakeLists.txt
```diff
@@ -297,7 +299,7 @@ IF(USE_SDL2)
 	MESSAGE(WARNING "USE_SDL2 has been deprecated, use SDL_BACKEND instead: Auto, SDL, SDL2.")
 ENDIF()
 IF(USE_OPENGL)
-	MESSAGE(WARNING "USE_OPENGL has been dropped, use OPENGL_BACKEND instead: None, OpenGL")
+	MESSAGE(WARNING "USE_OPENGL has been dropped, use OPENGL_BACKEND instead: None, OpenGL, GLES")
```
lynxlynxlynx, line 302: Please also change line 44.

CMakeLists.txt
```diff
@@ -319,44 +321,25 @@ ELSE()
...
 	ENDIF()
 
-	IF(NOT APPLE)
-		INCLUDE(FindGLEW)
-		IF(GLEW_FOUND)
-			MESSAGE(STATUS "Looking for Glew: found")
-		ELSE()
-			MESSAGE(SEND_ERROR "Looking for Glew: not found!")
-			MESSAGE(FATAL_ERROR "Please install the Glew library and headers first!")
+	IF(OPENGL_BACKEND STREQUAL "GLES")
```
lynxlynxlynx, line 336: Nit: this could be an ELSEIF conjoined to the previous block to save one comparison in one case.

CMakeLists.txt
cmitu: Should be fine now.

gemrb/plugins/SDLVideo/CMakeLists.txt
```diff
@@ -1,13 +1,21 @@
...
-			TARGET_LINK_LIBRARIES( SDLVideo ${SDL_LIBRARY} ${OPENGL_LIBRARY} ${CMAKE_THREAD_LIBS_INIT} ${COCOA_LIBRARY_PATH})
+		target_compile_definitions( SDLVideo PRIVATE USE_OPENGL_BACKEND)
+		target_compile_definitions( SDLVideo PRIVATE USE_$<UPPER_CASE:${OPENGL_BACKEND}_API>)
+		TARGET_LINK_LIBRARIES( SDLVideo ${SDL_LIBRARY} ${CMAKE_THREAD_LIBS_INIT} ${COCOA_LIBRARY_PATH})
+
+		IF(OPENGL_BACKEND STREQUAL "OpenGL")
+			IF(APPLE)
+				TARGET_LINK_LIBRARIES( SDLVideo ${OPENGL_LIBRARIES})
+			ELSE()
+				TARGET_LINK_LIBRARIES( SDLVideo ${OPENGL_LIBRARIES})
```
lynxlynxlynx, line 15: This looks redundant now. Previously the only difference was in glew, but now you removed it ...

CMakeLists.txt
cmitu: Done.

gemrb/plugins/SDLVideo/CMakeLists.txt
cmitu: Indeed - fixed.

gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -58,29 +58,26 @@ int SDL20VideoDriver::CreateDriverDisplay(const Size& s, int bpp, const char* ti
...
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
-	*/
 #endif
 
-#if SDL_VERSION_ATLEAST(2, 0, 10)
-	SDL_SetHint(SDL_HINT_RENDER_BATCHING, "1");
-#endif
+	if ( sdl2_version >= 2010) {
```
lynxlynxlynx, line 71: Shouldn't this be sdl2_runtime_version? But neither appear set anywhere.

lynxlynxlynx (changes requested):

gemrb/plugins/SDLVideo/SDL20Video.cpp
lynxlynxlynx: FindOpenGLES2 is missing as well.

cmitu: Yes, this is an error. I added it, but didn't initialize it (and the variable is wrong). I'll fix it.

cmitu: Ok, fixed.

cmake/modules/FindOpenGLES2.cmake
```diff
@@ -0,0 +1,35 @@
...
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(OpenGLES2
+            REQUIRED_VARS OPENGLES2_gl_LIBRARY OPENGLES2_INCLUDE_DIR)
+
+
+if(OPENGLES2_FOUND)
+    set(OPENGLES2_LIBRARIES ${OPENGLES2_gl_LIBRARY})
+    set(OPENGLES2_INCLUDE_DIRS ${OPENGLES2_INCLUDE_DIR})
```
lynxlynxlynx, line 32: OPENGLES2_INCLUDE_DIRS is never used. Is that really ok?

cmitu: I think so, we include `SDL_opengles2.h` for any `gl` related functions. 
Same situation is for `OPENGL_INCLUDE_DIR`, which is never used, though it's filled in by _FindOpenGL.cmake_ (though it probably wasn't needed because GLEW was support for it).

CMakeLists.txt
```diff
@@ -319,44 +321,23 @@ ELSE()
...
+		ENDIF()
+	ELSEIF(OPENGL_BACKEND STREQUAL "GLES")
+		INCLUDE(FindOpenGLES2)
+		IF(NOT OPENGLES2_FOUND)
+			MESSAGE(FATAL_ERROR "OpenGLES2 library not found!")
 		ENDIF()
 	ENDIF()
+
 ELSE()
 	SET(OPENGL_BACKEND 0)
```
lynxlynxlynx, line 342: This can now go as well. :)

INSTALL
```diff
@@ -19,7 +19,7 @@ Optional dependencies:
   - FreeType (for the TTF fonts plugin)
   - ICONV (for the TTF fonts plugin and only needed when using a non unicode compatible TLK)
   - (lib)VLC 2.0+ (for movies in some mac distributions of the games)
-  - OpenGL drivers and GLEW
+  - OpenGL/OpenGLES drivers
```
lynxlynxlynx, line 23: Please also include your note somewhere, so other builders will have an easier time:
> The build system prefers the Broadcom GLES libraries when a RPI system is detected, but this can be overriden by adding -DDISABLE_VIDEOCORE=1 - and is necessary for the Pi4.

CMakeLists.txt
cmitu: Should be ok now.

INSTALL
cmitu: Yes, that would be useful - I'll include some build instructions.

lynxlynxlynx:
Ok, that's all I can read from the diff, besides some whitespace issues. But those can be tackled at the end. I'll try it on linux in the following days, thanks!

gemrb/plugins/SDLVideo/OpenGLEnv.h
```diff
@@ -21,10 +21,16 @@
...
 
-#if __APPLE__
+#define GL_GLEXT_PROTOTYPES 1
+
+#ifdef __APPLE__
 	#include <OpenGL/gl3.h>
 #else
-	#include <GL/glew.h>
+#ifdef USE_OPENGL_API
+	#include <SDL_opengl.h>
```
MarcelHB, line 30: That does not work for everything beyond OpenGL 1. While the header declares everything, there is nothing to link the OpenGL 2.0+ stuff against (like ` undefined reference to __imp_glCompileShader`). And for everything SDL2 may need internally, it does the procedure lookup [manually](https://github.com/spurious/SDL-mirror/blob/579d3b8ef46f995832a10561f2de743051874f36/src/render/opengl/SDL_shaders_gl.c#L473). So GLEW again makes that easy.

cmitu: This does work for OpenGL2 - as reported in #932 - though it might not work for OpenGL >= 3.
 SDL2 loads the GL implementation dynamically, if the GL library has the necessary support, it will not be a problem.

Regarding the error you get - is this from a clean compile ? I think I've seen this error during testing after removing GLEW, but with previous object files containing GLEW references

MarcelHB: I am not sure where in #932 you can imply this? Anyway, I really did a fresh build env to double-check. No change. Look at this link chain: ` -lSDL2 -lopengl32 `. None of them supports OpenGL2.0+ here by the required symbols. GLEW _does_ and links or sets up dispatches to their GL proc in some way when calling `glewInit()` at runtime. 

Thinking about it: you need to have all the `glX` stuff assigned by their runtime lookup results, and better only once. Therefore, you either do this yourself (using `SDL_GL_GetProcAddress `, or whatever your OS asks you), or you have it globally assigned by some compilation unit that has reserved space to be linked. That is what GLEW or epoxy are doing for the whole OpenGL2.0+ stuff, by eager or lazy loading. Apple prebakes this up to OGL3, so there is no need for GLEW either. ES seems to follow that but I cannot verify.

Look at this [minimal example](https://gist.github.com/koute/7391344) that follows your idea, it exposes the same compilation error for me (on Windows MSYS2), while it some other platform seems to work. MacOS probably?

cmitu: > None of them supports OpenGL2.0+ here by the required symbols. GLEW _does_ and links or sets up dispatches to their GL proc in some way when calling `glewInit()` at runtime.

You're correct. Maybe because Windows' `opengl32.lib/dll` is just implementing OpenGL 1.1 ?
So GLEW is still needed for Windows (at least) - I'll add it back.

> Look at this minimal example that follows your idea, it exposes the same compilation error for me (on Windows MSYS2), while it some other platform seems to work. MacOS probably?

I've tested on Linux only. The example linked works fine there (at least it compiles/links without errors).

cmitu:
Updated to address the pending changes after review
 - added back GLEW when OpenGL is chosen as OPENGL_BACKEND on non-macOS platforms.
 - added note about building on a Raspberry Pi to the `INSTALL` document.

bradallred:
I don't understand why we now have `USE_OPENGL_API` and `USE_GLES_API` instead of just using the value of `OPENGL_BACKEND`

cmitu:
> I don't understand why we now have `USE_OPENGL_API` and `USE_GLES_API` instead of just using the value of `OPENGL_BACKEND`

String comparison doesn't work in pre-processor commands  
```
#if ABC == ZYX
```
always yields true, so it cannot be used as a conditional. This is why I opted to explicitly add the 2 defines.

bradallred:
that's a good reason alright. I probably shouldn't be surprised I guess. An error or warning would be useful tho. Thanks.

lynxlynxlynx:
So what are the cases when glew is needed on linux?

MarcelHB:
Not sure, actually. `libGL.so` by Mesa exposes what we need.

cmitu:
@lynxlynxlynx I think right now, for Linux, the SDL2 headers provide support for OpenglGL2 functions and - as @MarcelHB mentioned - any link time requirements are satisfied by MESA.
If you'll want to use something from OpenGL3 and raise the minimum required GL version, then SDL2 headers will not be enough.

lynxlynxlynx:
Yeah, let's keep it optional then. A separate commit is fine.

gemrb/plugins/SDLVideo/CMakeLists.txt
```diff
@@ -1,13 +1,17 @@
-INCLUDE_DIRECTORIES( ${SDL_INCLUDE_DIR} ${GLEW_INCLUDE_DIR} )
+INCLUDE_DIRECTORIES( ${SDL_INCLUDE_DIR} $<$<BOOL:${WIN32}>:${GLEW_INCLUDE_DIR}>)
```
lynxlynxlynx, line 1:
```suggestion
INCLUDE_DIRECTORIES(${SDL_INCLUDE_DIR} $<$<BOOL:${WIN32}>:${GLEW_INCLUDE_DIR}>)
```

gemrb/plugins/SDLVideo/CMakeLists.txt
```diff
@@ -1,13 +1,17 @@
...
+	IF(NOT OPENGL_BACKEND STREQUAL "None")
 		ADD_GEMRB_PLUGIN( SDLVideo ${COMMON_FILES} SDL20Video.cpp GLSLProgram.cpp)
-		IF(APPLE)
-			TARGET_LINK_LIBRARIES( SDLVideo ${SDL_LIBRARY} ${OPENGL_LIBRARY} ${CMAKE_THREAD_LIBS_INIT} ${COCOA_LIBRARY_PATH})
+		target_compile_definitions( SDLVideo PRIVATE USE_OPENGL_BACKEND)
+		target_compile_definitions( SDLVideo PRIVATE USE_$<UPPER_CASE:${OPENGL_BACKEND}_API>)
+		TARGET_LINK_LIBRARIES( SDLVideo ${SDL_LIBRARY} ${CMAKE_THREAD_LIBS_INIT} ${COCOA_LIBRARY_PATH})
+
+		IF(OPENGL_BACKEND STREQUAL "OpenGL")
+			TARGET_LINK_LIBRARIES( SDLVideo ${OPENGL_LIBRARIES} $<$<BOOL:${WIN32}>:${GLEW_LIBRARIES}>)
```
lynxlynxlynx, line 12:
```suggestion
		target_compile_definitions(SDLVideo PRIVATE USE_OPENGL_BACKEND)
		target_compile_definitions(SDLVideo PRIVATE USE_$<UPPER_CASE:${OPENGL_BACKEND}_API>)
		TARGET_LINK_LIBRARIES(SDLVideo ${SDL_LIBRARY} ${CMAKE_THREAD_LIBS_INIT} ${COCOA_LIBRARY_PATH})

		IF(OPENGL_BACKEND STREQUAL "OpenGL")
			TARGET_LINK_LIBRARIES(SDLVideo ${OPENGL_LIBRARIES} $<$<BOOL:${WIN32}>:${GLEW_LIBRARIES}>)
```

gemrb/plugins/SDLVideo/CMakeLists.txt
```diff
@@ -1,13 +1,17 @@
...
 		ELSE()
-			TARGET_LINK_LIBRARIES( SDLVideo ${SDL_LIBRARY} ${OPENGL_LIBRARY} ${GLEW_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
+			TARGET_LINK_LIBRARIES( SDLVideo ${OPENGLES2_LIBRARIES})
```
lynxlynxlynx, line 14:
```suggestion
			TARGET_LINK_LIBRARIES(SDLVideo ${OPENGLES2_LIBRARIES})
```

gemrb/plugins/SDLVideo/OpenGLEnv.h
```diff
@@ -21,10 +21,20 @@
...
 #else
+#ifdef USE_OPENGL_API
+#ifdef _WIN32
 	#include <GL/glew.h>
+#else
+	#include <SDL_opengl.h>
+#endif
+#else // USE_GLES_API
+	#include <SDL_opengles2.h>
+#endif
```
lynxlynxlynx, line 37:
```suggestion
	#ifdef USE_OPENGL_API
		#ifdef _WIN32
			#include <GL/glew.h>
		#else
			#include <SDL_opengl.h>
		#endif
	#else // USE_GLES_API
		#include <SDL_opengles2.h>
	#endif
```

lynxlynxlynx:
Thanks, left the final few style nitpicks as suggestions.

cmitu:
@lynxlynxlynx OK, I eliminated the stray spaces and added some clarifying indents to the header file. 
Nitpick away, it's better to get things right in the first place than patch it up later.

lynxlynxlynx:
Thanks! Any reason for it still being marked as a draft? Looks ready to be merged to me. 👍🏾

cmitu:
The idea was to get some feedback for it and maybe wait for the shader changes so that using GLES would be feasible. 
I removed the draft status - you can decide it if you want to merge it now or wait for the shader changes.

lynxlynxlynx (approved):

lynxlynxlynx:
Ah, right. Since this looks fine already and fixes a few problems, let's go ahead so it doesn't get stale. I'll split the shader stuff into a fresh issue, so we don't have to continue in the closed one for master.

gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -58,29 +62,26 @@ int SDL20VideoDriver::CreateDriverDisplay(const Size& s, int bpp, const char* ti
...
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-	   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
-	*/
 #endif
 
-#if SDL_VERSION_ATLEAST(2, 0, 10)
-	SDL_SetHint(SDL_HINT_RENDER_BATCHING, "1");
-#endif
+	if (sdl2_runtime_version >= SDL_VERSIONNUM(2,0,10)) {
```
lynxlynxlynx, line 75: @cmitu  This broke the coverity workflow — it also needs to check the build time SDL version, not just the runtime one, since otherwise SDL_HINT_RENDER_BATCHING may not be defined. That's the case on github, since SDL is at 2.0.8.

cmitu: Should be fixable. I see Ubuntu 18.04 LTS still uses SDL 2.0.8, so it's still a viable target - https://github.com/gemrb/gemrb/pull/939.

PR #947, opened 2020-09-28
Window.rc : the artwork directory is now one more level above relatively

chilvence:
Now that the platforms directory has been created, this fixes a tiny issue of MSVC not being able to find the program icon while building, as the artwork directory is now two levels above instead of one.


lynxlynxlynx:
Thanks!

PR #948, opened 2020-09-28
PST clskills.2da fixes and skills.2da draft

chilvence:
This is the prerequisite first step of implementing the common skill scripts in PST

The first commit seems to be a simple oversight or typo

The second commit adds the skills.2da table that will be needed later with the correct label strrefs (there is no description text needed/used since the level up screen is much simplified)

I think the skill id's are correct, or at least, they match up with the entries for all the other games. This part I am not 100% sure how they are used or whether they would be any reason for them to be different for PST.

## Checklist

### DIFF
--- a/gemrb/unhardcoded/pst/clskills.2da
+++ b/gemrb/unhardcoded/pst/clskills.2da
@@ -1,6 +1,6 @@
 2DA V1.0
 *
-                    HATERACE   CLERICSPELL MAGESPELL  STARTXP  BARD  THIEF    LAYHANDS  TURNLEVEL  BOOKTYPE  SAVEBONUS  SPONCAST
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Looks fine, thanks. I don't see any code relying on the old names, so it's probably safe to alter.

chilvence:
Ok, I will undraft it

There is more to come, but I am just sifting through and tidying individual parts piecemeal

lynxlynxlynx:
Please make skills.2da  align properly, so it will be more readable. If you used tabs there, that could explain it.

chilvence:
No problem, although I'm not sure what the 'proper' method is for aligning 2da files, is it literally just spaces?

chilvence:
Is it better now? I changed the formatting, but I am really not sure what editor to use any more, what I see in notepad++ doesn't ever quite match up to what I see on github

lynxlynxlynx:
Yeah, that will do. Thanks!

PR #950, opened 2020-09-30
MSVC build tweaks

chilvence:
Various tweaks for building with MSVC - these were just some other small things I had to change to get back to normal

1: Renables deployment of dll's to the build directory for quick testing. I think the original purpose of this probably didn't appear clear, so it went away with the cleanup.

2: Disables a new warning related to gamescript.h, hundreds of "c4138: /* found outside of comment" messages. Not sure what suddenly caused them to start appearing.

3: When 'installing' Dll deployment was not happening when RelWithDebInfo option is used. I tend to use this one, it uses the same release library set

4: Prefer copying the system python version if it exists, otherwise use the vcpkg version. I hope this change works fine with appveyor, am I correct in thinking that making a pull request will test that/show what happens in the log?

For what it's worth, I did manage to finally get mine building with the vcpkg version of python, it was all down to an annoying Visual studio quirk that was saving some of my old variables even after purging the cmake cache. The only downside is that it doesn't work from the build dir unless you manually set the pythonhome variable to point the the Lib folder, because it doesn't actually set itself to work system wide. However, it does work fine if you hit 'install', since the change to copying the Lib folder over as well. 

I'm not 100% sure what to suggest. I am also not quite certain what happens with appveyor if you don't use the vcpkg version of python. So I am just putting this here for feedback.

## Checklist

### DIFF
--- a/gemrb/vcpkg_deps.cmake
+++ b/gemrb/vcpkg_deps.cmake
@@ -84,7 +84,19 @@ INSTALL(FILES ${DLL_PATHS_DEBUG} CONFIGURATIONS Debug DESTINATION ${BIN_DIR})
 INSTALL(FILES ${DLL_PATHS_RELEASE} CONFIGURATIONS Release RelWithDebInfo DESTINATION ${BIN_DIR})
 
 # copy over python core modules, so the buildbot binaries work without python installed
-INSTALL(DIRECTORY ${VCPKG_DATAROOT}/share/python2/Lib DESTINATION ${BIN_DIR})
+
+#this copies the modules bundled with the standard python installer if found
+#otherwise, this uses the vcpkg bundled python modules if detected
+#site.py is searched for by name because the vcpkg uninstall process doesn't properly purge the folder
+#so if you switch between them for any reason, it will try to copy the wrong folder
+
+GET_FILENAME_COMPONENT(PYTHON_PARENT_DIR ${PYTHON_INCLUDE_DIR} DIRECTORY)
+
+IF(EXISTS ${PYTHON_PARENT_DIR}/Lib/site.py )
+	INSTALL(DIRECTORY ${PYTHON_PARENT_DIR}/Lib DESTINATION ${BIN_DIR})
+ELSEIF(EXISTS ${VCPKG_DATAROOT}/share/python2/Lib/site.py)
+	INSTALL(DIRECTORY ${VCPKG_DATAROOT}/share/python2/Lib DESTINATION ${BIN_DIR})
+ENDIF()
 
 MESSAGE(STATUS "Dependency DLL's will be copied to the build and install directory")
 MESSAGE(STATUS "Disable option VCPKG_AUTO_DEPLOY to skip this")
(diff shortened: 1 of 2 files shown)


gemrb/vcpkg_deps.cmake
```diff
@@ -66,6 +66,19 @@ FOREACH(ENTRY IN LISTS DLL_SET_RELEASE)
...
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different
+		${DLL_PATHS_DEBUG}
+		${CMAKE_BINARY_DIR}/gemrb/)
+ELSE()
+	ADD_CUSTOM_COMMAND(TARGET gemrb POST_BUILD
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different
+		${DLL_PATHS_RELEASE}
+		${CMAKE_BINARY_DIR}/gemrb/)
+ENDIF()
+
```
lynxlynxlynx, line 81: I think it would be cleaner with a bit less duplication. My suggestion:
- split the if into determining which path to use
- then unconditionally add the custom command with the new var

lynxlynxlynx:
Looks fine, thanks! The appveyor build was also successful. @nerd-rage any comments from your side?

chilvence:
Unless there's any more feedback to this, I'll unlock it. I am back home on my windows machine so I can always amend it.

lynxlynxlynx:
You could improve it with my last code comment, but other than that, it's ready to merge.

chilvence:
Awk, I forgot to test that out, sorry, hold on one moment...

chilvence:
Well, I thought this was finished, but now I have an inexplicable new problem, the debug build refuses to build because python27_d.lib does not get installed with the python installer.

The crazy thing is, this has never caused a problem before, I've built and run the debug build dozens of times while testing before, it usually just falls back to the release library.

I think this is happening because I installed vcpkg python for testing, then removed it again for more testing. Reinstalling makes the problem go away, but I have no idea how to return to how I was building previously...

lynxlynxlynx:
Did you try in a clean build dir?

chilvence:
Yes, even a brand new profile. But I think it's probably something more to do with the way Visual Studio itself works. It's done stuff like this to me before.

I had a really nice time last week with away with my PI, managed to make loads of progress without any of this nonsense...

I'm sure I can figure it out with some googling though.

lynxlynxlynx:
Well, before the last change, we didn't even try to use python27_d. Now we do the proper thing, which also enabled some header cleanup.

See here how to install it properly, since it's an extra option:
https://stackoverflow.com/a/45407558/1494504

chilvence:
Well, probably the best solution then is to require the vcpkg version of python regardless, since that is the only way to get the debug lib. Neither the python 2 or 3 installers that ship with visual studio contain debug libs, so the only workaround would be to go back to the way things were, which isn't an option. Mildly infuriating that it doesn't just work, but it looks like many other people have been slapped by this problem looking around the web. 

Best I can do is find someone at Microsoft to have a rant at (whether I'll get anywhere...)

lynxlynxlynx:
Ok, I'll just document it in the guide then and merge this. Thanks!

chilvence:
I'll be sticking with VS anyway, so I'll revisit this if I find a better solution

chilvence:
Sent a bug report to VS team:

https://developercommunity.visualstudio.com/content/problem/1215312/cannot-link-to-python-debug-lib-because-it-is-not.html

Now, the waiting game :)

PR #956, opened 2020-10-05
SDLAudio caching

Northfear:
## Description

Added simple Mix_Chunk caching for SDLAudio. Pretty much the same as one in OpenAL plugin, but with fixed bug, that allowed removing cached sounds just after adding them.
Judging by the docs, removing the chunks should be done with Mix_FreeChunk, but it resulted in a massive memleak for me, so I've tried doing everything manually.

### DIFF
--- a/gemrb/plugins/SDLAudio/SDLAudio.cpp
+++ b/gemrb/plugins/SDLAudio/SDLAudio.cpp
@@ -120,14 +120,14 @@ bool SDLAudio::evictBuffer()
 	// Room for optimization: this is O(n^2) in the number of buffers
 	// at the tail that are used. It can be O(n) if LRUCache supports it.
 	unsigned int n = 0;
-	void* p;
-	const char* k;
+	void *p;
+	const char *k;
 	bool res;
 
 	SDL_LockAudio();
 
 	while ((res = buffercache.getLRU(n, k, p)) == true && buffercache.GetCount() >= BUFFER_CACHE_SIZE) {
-		CacheEntry* e = (CacheEntry*)p;
+		CacheEntry *e = (CacheEntry*)p;
 		bool chunkPlaying = false;
 		int numChannels = Mix_AllocateChannels(-1);
 
@@ -142,8 +142,8 @@ bool SDLAudio::evictBuffer()
 			++n;
 		} else {		
 			//Mix_FreeChunk(e->chunk) fails to free anything here
-			SDL_free(e->chunk->abuf);
-			SDL_free(e->chunk);
+			free(e->chunk->abuf);
+			free(e->chunk);
 			delete e;
 			buffercache.Remove(k);
 		}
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Nice, thanks! I'll try to finish the clang-format work first, so these can be a nice test.

Can you elaborate on the openal bug, so we can fix that as well?

I see there's some duplication between the plugins, but it looks just annoying enough to not be worth it to split the shared code out.

Northfear:
"Bug" is a bit of a strong word for it, but in rare cases `evictBuffer();` that's called after `buffercache.SetAt(ResRef, (void*)e);` was deleting newly added chunk.
Shouldn't really matter with cache size of 100. I've only seen this behavior during testing with really small cache sizes and busy audio channels.

bradallred:
is this the bug where if you click really fast to skip the intro videos you sometimes don't get the music on the start screen?

Northfear:
I don't think that music is cached, so probably no.
Once again it's not a "bug" really. More like a different approach. One could only see it when cache is full and all of the cached files are in use (which wouldn't allow playing of the newly added file anyway. The only exception could be speech probably, since it has reserved channel).

bradallred:
yeah, it was a race condition that should be fixed now

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -117,52 +113,93 @@ void SDLAudio::music_callback(void *udata, unsigned short *stream, int len) {
...
-	if (!ResRef) {
-		if (flags & GEM_SND_SPEECH) {
-			Mix_HaltChannel(0);
-		}
-		return Holder<SoundHandle>();
+	if (!ResRef[0]) {
+		return chunk;
+	}
+
+	if(buffercache.Lookup(ResRef, p)) {
```
lynxlynxlynx, line 183:
```suggestion
	if (buffercache.Lookup(ResRef, p)) {
```
The style report will take a while, so let's just polish this manually.

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -177,11 +214,56 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
...
+	e = new CacheEntry;
+	e->chunk = chunk;
+	e->Length = time_length;
+
+	if (buffercache.GetCount() >= BUFFER_CACHE_SIZE) {
+		evictBuffer();
+	}
+
+	buffercache.SetAt(ResRef, (void*)e);
+	//print("LoadSound: added %s to cache. Cache size now %d", ResRef, buffercache.GetCount());
```
lynxlynxlynx, line 233:
```suggestion

```

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -177,11 +214,56 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
...
+	(void)flags;
+
+	if (!ResRef) {
+		if (flags & GEM_SND_SPEECH) {
+			Mix_HaltChannel(0);
+		}
 		return Holder<SoundHandle>();
 	}
+
+	chunk = loadSound( ResRef, time_length );
```
lynxlynxlynx, line 258:
```suggestion
	chunk = loadSound(ResRef, time_length);
```

lynxlynxlynx:
Thanks!

PR #959, opened 2020-10-06
Fix loading and display of frames for widescreen resolutions

bernds:
## Description
A widescreen fix for the subviews branch. This makes the frames work in my 960x600 resolution, and I've tested with 1024x768 as well.

## Checklist


bradallred:
thank you!

Issue #629, opened 2020-02-12
[bug] Mouse right click won't move a single character


burner1024:
**Describe the bug**
In classic BG2, right click works for directing path for a group as well as a single character. In GemRB, it works only for a group. For a character, right click doesn't register.

GemRB version: 0.8.6-git

**To Reproduce**
1. Select a single character.
2. Right click to walk somewhere.

**Expected behavior**
The character should move to the designated spot.

bradallred:
heh, I would be of the opinion that the group behavior is the unwanted bug :)

Why do we need right click to do what left click already does? Sounds like unnecessary code complexity for no gain to me. This is especially true if this behavior really does only apply to BG2 as per the tag.

burner1024:
I think you got it all backwards. If anything, why do we need left click to do what _right_ click already does? :)

![war4](https://user-images.githubusercontent.com/1279852/75087950-277b0980-5579-11ea-8686-e2809637f3a5.jpg)


Anyway,
1. Sorry, I don't understand what do you mean by "group behaviour", why it's unwanted and a bug on top. I was talking about single character behaviour.
2. Somewhere in the code, it reads:
   ```
   if mouse.right_click()
     n = count_selected_characters()
     if n > 1:
       work()
      else:
       dont_work()
   ```
    That's what seems like unnecessary code complication to me. (Maybe digging git history would shed some light on why it's done that way.)
3. On the broader scope of right vs left, the difference is semantical. Right click only moves. You can't accidentally talk to someone, or attack someone, or pass through a door, or open a guarded container with it. It's _safe_.
   Left click is what you use when you want to _interact_: attack, travel, etc. If you miss, you may be screwed.
  I think many games implement this general distinction, not just IE. In fact, I suspect that the only reason that left click takes on some move functions is because some people have mouses with a single button.
  In addition:
   - A minor thing, but in particular, if you want to walk really close to a travel region, only right click will allow to do that. Left click with take you _through_ it instead.
   - Also, of course, only right click allows and can allow rotating party formations.
4. I mentioned BG2 because that's the only engine I care about, but now I fired up BG1, IWD, BGEE - it's all the same. I'm reasonably sure that's true for all IE games.

bradallred:
re code complexity: I mean my preference is to make things simple and not rely on the number of PCs that are selected. I think we are in agreement there. I didn't realize that there was a subtle difference (interaction vs only movement) between movements directed by left vs right click; that is good to know and is useful behavior.

[...]

PR #960, opened 2020-10-06
Improve formation movement and right-click moves in general

bernds:
## Description
A version of my formations patch against the subviews branch.
This is actually a little more complete than the previous one. The maths is simplified a bit more, and now right click, left click, and Alt-Leftclick all behave consistently, and as far as I can tell the behaviour matches the IE behaviour pretty well.

I modified the gestures code to be consistent as well, but I can't test it.

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -184,6 +184,9 @@ Point GameControl::GetFormationPoint(const Point& origin, size_t pos, double ang
 	Point stepVec;
 	int direction = (pos % 2 == 0) ? 1 : -1;
 
+	/* Correct for the innate orientation of the points.  */
+	angle += M_PI / 2;
+
 	if (pos < FORMATIONSIZE) {
 		// calculate new coordinates by rotating formation around (0,0)
 		vec = RotatePoint(formations[formation][pos], angle);
@@ -608,7 +611,9 @@ void GameControl::DrawSelf(Region screen, const Region& /*clip*/)
 	Point gameMousePos = GameMousePos();
 	// draw reticles
 	if (isFormationRotation) {
-		double angle = AngleFromPoints(gameMousePos, gameClickPoint);
+		double angle = formationBaseAngle;
+		if (Distance (gameMousePos, gameClickPoint) > 5)
+			angle = AngleFromPoints(gameMousePos, gameClickPoint);
 		DrawFormation(game->selected, gameClickPoint, angle);
 	} else {
 		int max = game->GetPartySize(false);
@@ -1469,7 +1474,7 @@ bool GameControl::OnMouseDrag(const MouseEvent& me)
 		Scroll(me.Delta());
 		return true;
 	}
-	
+
 	if (target_mode != TARGET_MODE_NONE) {
 		// we are in a target mode; nothing here applies
 		return true;
@@ -1479,23 +1484,8 @@ bool GameControl::OnMouseDrag(const MouseEvent& me)
 		return true;
 	}
 
-	if (me.ButtonState(GEM_MB_ACTION)) {
-		// PST uses alt + left click for formation rotation
-		// is there any harm in this being true in all games?
-		if (EventMgr::ModState(GEM_MOD_ALT)) {
-			isFormationRotation = true;
-		} else {
-			isSelectionRect = true;
-		}
-	}
-
-	if (me.ButtonState(GEM_MB_MENU)) {
-		isFormationRotation = true;
-	}
-
-	if (core->GetGame()->selected.size() <= 1) {
-		isFormationRotation = false;
-	}
+	if (me.ButtonState(GEM_MB_ACTION) && !isFormationRotation)
+		isSelectionRect = true;
 
 	if (isFormationRotation) {
 		SetCursor(core->Cursors[IE_CURSOR_USE]);
@@ -1558,8 +1548,8 @@ bool GameControl::OnTouchGesture(const GestureEvent& gesture)
 			if (core->GetGame()->selected.size() <= 1) {
 				isFormationRotation = false;
 			} else {
-				isFormationRotation = true;
 				screenMousePos = gesture.fingers[1].Pos();
+				InitFormation (screenMousePos);
 			}
 		} else { // scroll viewport
 			MoveViewportTo( vpOrigin - gesture.Delta(), false );
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
 	return false;
 }
 
+// Calculate the angle between a clicked point and the first selected character,
+// so that we can set a sensible orientation for the formation
+void GameControl::InitFormation (const Point &clickPoint)
+{
+	/* We get called on initial click, but also on drag.  Perform the setup only once.  */
+	if (isFormationRotation)
+		return;
+
+	if (core->GetGame()->selected.size() == 0) {
+		isFormationRotation = false;
+		return;
+	}
+
+	isFormationRotation = true;
+	Game* game = core->GetGame();
+	int max = game->GetPartySize(false);
+	formationBaseAngle = 0;
+	for (int idx = 1; idx <= max; idx++) {
+		Actor *act = game->FindPC(idx);
+		if (act->IsSelected()) {
+			formationBaseAngle = AngleFromPoints (clickPoint, act->Pos);
+			break;
+		}
+	}
+	SetCursor(core->Cursors[IE_CURSOR_USE]);
+}
+
 /** Mouse Button Down */
 bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
 {
@@ -2048,10 +2065,18 @@ bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
 	case GEM_MB_MENU: //right click.
 		if (core->HasFeature(GF_HAS_FLOAT_MENU) && !Mod) {
 			core->GetGUIScriptEngine()->RunFunction( "GUICommon", "OpenFloatMenuWindow", false, p);
+		} else {
+			InitFormation (gameClickPoint);
 		}
 		break;
 	case GEM_MB_ACTION:
 		isDoubleClick = me.repeats == 2;
+
+		// PST uses alt + left click for formation rotation
+		// is there any harm in this being true in all games?
+		if (!isDoubleClick && EventMgr::ModState(GEM_MOD_ALT))
+			InitFormation (gameClickPoint);
+
 		break;
 	}
 	return true;
@@ -2163,6 +2188,7 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 			ClearMouseState();
 			return true;
 		}
+		InitFormation(p);
 	}
 
 	// handle movement/travel
@@ -2249,7 +2275,10 @@ void GameControl::CommandSelectedMovement(const Point& p, unsigned short Mod)
 
 	double angle = 0.0;
 	if (isFormationRotation) {
-		angle = AngleFromPoints(GameMousePos(), p);
+		angle = formationBaseAngle;
+		Point mp = GameMousePos ();
+		if (Distance (mp, p) > 5)
+			angle = AngleFromPoints(mp, p);
 	}
 
 	bool doWorldMap = ShouldTriggerWorldMap(party[0]);
--- a/gemrb/core/GUI/GameControl.h
+++ b/gemrb/core/GUI/GameControl.h
@@ -92,6 +92,7 @@ private:
 	Point screenMousePos;
 	Point vpOrigin;
 	bool updateVPTimer;
+	double formationBaseAngle;
 
 	// currently selected targeting type, such as talk, attack, cast, ...
 	// private to enforce proper cursor changes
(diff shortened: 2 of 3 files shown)


gemrb/core/GUI/GameControl.cpp
```diff
@@ -184,6 +184,9 @@ Point GameControl::GetFormationPoint(const Point& origin, size_t pos, double ang
 	Point stepVec;
 	int direction = (pos % 2 == 0) ? 1 : -1;
 
+	/* Correct for the innate orientation of the points.  */
+	angle += M_PI / 2;
```
lynxlynxlynx, line 188:
```suggestion
	angle += M_PI_2;
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -608,7 +611,9 @@ void GameControl::DrawSelf(Region screen, const Region& /*clip*/)
 	Point gameMousePos = GameMousePos();
 	// draw reticles
 	if (isFormationRotation) {
-		double angle = AngleFromPoints(gameMousePos, gameClickPoint);
+		double angle = formationBaseAngle;
+		if (Distance (gameMousePos, gameClickPoint) > 5)
+			angle = AngleFromPoints(gameMousePos, gameClickPoint);
```
lynxlynxlynx, line 616: No spaces before opening parens and braces to avoid two types of warnings (ambiguity due to following statement)
```suggestion
		if (Distance(gameMousePos, gameClickPoint) > 5) {
			angle = AngleFromPoints(gameMousePos, gameClickPoint);
		}
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -1558,8 +1548,8 @@ bool GameControl::OnTouchGesture(const GestureEvent& gesture)
 			if (core->GetGame()->selected.size() <= 1) {
 				isFormationRotation = false;
 			} else {
-				isFormationRotation = true;
 				screenMousePos = gesture.fingers[1].Pos();
+				InitFormation (screenMousePos);
```
lynxlynxlynx, line 1552:
```suggestion
				InitFormation(screenMousePos);
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
 	return false;
 }
 
+// Calculate the angle between a clicked point and the first selected character,
+// so that we can set a sensible orientation for the formation
+void GameControl::InitFormation (const Point &clickPoint)
```
lynxlynxlynx, line 2033:
```suggestion
void GameControl::InitFormation(const Point &clickPoint)
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+{
+	/* We get called on initial click, but also on drag.  Perform the setup only once.  */
+	if (isFormationRotation)
+		return;
```
lynxlynxlynx, line 2037:
```suggestion
	if (isFormationRotation) return;
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+
+	if (core->GetGame()->selected.size() == 0) {
```
lynxlynxlynx, line 2039: Isn't this a regression from the current behaviour?
```suggestion
	if (core->GetGame()->selected.size() <= 1) {
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+		isFormationRotation = false;
+		return;
+	}
+
+	isFormationRotation = true;
+	Game* game = core->GetGame();
```
lynxlynxlynx, line 2045:
```suggestion
	Game *game = core->GetGame();
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+	int max = game->GetPartySize(false);
+	formationBaseAngle = 0;
+	for (int idx = 1; idx <= max; idx++) {
+		Actor *act = game->FindPC(idx);
+		if (act->IsSelected()) {
+			formationBaseAngle = AngleFromPoints (clickPoint, act->Pos);
```
lynxlynxlynx, line 2051:
```suggestion
			formationBaseAngle = AngleFromPoints(clickPoint, act->Pos);
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2163,6 +2188,7 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 			ClearMouseState();
 			return true;
 		}
+		InitFormation(p);
```
bradallred, line 2191: is this not redundant with `ClearMouseState`?

gemrb/core/GUI/GameControl.h
```diff
@@ -92,6 +92,7 @@ class GEM_EXPORT GameControl : public View {
 	Point screenMousePos;
 	Point vpOrigin;
 	bool updateVPTimer;
+	double formationBaseAngle;
```
bradallred, line 95: I think we of a mind to shift toward initializing member variables at their declaration (now that we use C++11)  so lets do that here by appending ` = 0.0`

```suggestion
	double formationBaseAngle = 0.0;
```

gemrb/core/GUI/GameControl.cpp
bernds: No - why would it be? The point is that right-click should always move, regardless of how many party members are selected. This is the #629 issue, and this change makes it behave more like the Infinity engine.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+		return;
+	}
+
+	isFormationRotation = true;
+	Game* game = core->GetGame();
+	int max = game->GetPartySize(false);
+	formationBaseAngle = 0;
+	for (int idx = 1; idx <= max; idx++) {
+		Actor *act = game->FindPC(idx);
+		if (act->IsSelected()) {
```
lynxlynxlynx, line 2050: What happens if no pc is selected (basically `OnlyNPCsSelected()`)? Can happen if you start the rotation with only summons. In that sense it would be better to iterate over `selected` from Game.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2048,10 +2065,18 @@ bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
 	case GEM_MB_MENU: //right click.
 		if (core->HasFeature(GF_HAS_FLOAT_MENU) && !Mod) {
 			core->GetGUIScriptEngine()->RunFunction( "GUICommon", "OpenFloatMenuWindow", false, p);
+		} else {
+			InitFormation (gameClickPoint);
```
lynxlynxlynx, line 2069:
```suggestion
			InitFormation(gameClickPoint);
```

gemrb/core/GUI/GameControl.cpp
bernds: I can change that, but there is absolutely no consistent style in this file, and this is just copied from another location.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2048,10 +2065,18 @@ bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
...
 		}
 		break;
 	case GEM_MB_ACTION:
 		isDoubleClick = me.repeats == 2;
+
+		// PST uses alt + left click for formation rotation
+		// is there any harm in this being true in all games?
+		if (!isDoubleClick && EventMgr::ModState(GEM_MOD_ALT))
+			InitFormation (gameClickPoint);
```
lynxlynxlynx, line 2078:
```suggestion
		if (!isDoubleClick && EventMgr::ModState(GEM_MOD_ALT)) {
			InitFormation(gameClickPoint);
		}
```

gemrb/core/GUI/GameControl.cpp
bernds: No, this ensures that we get a proper rotation even for left-click moves. I'm adding a comment.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2249,7 +2275,10 @@ void GameControl::CommandSelectedMovement(const Point& p, unsigned short Mod)
 
 	double angle = 0.0;
 	if (isFormationRotation) {
-		angle = AngleFromPoints(GameMousePos(), p);
+		angle = formationBaseAngle;
+		Point mp = GameMousePos ();
+		if (Distance (mp, p) > 5)
```
lynxlynxlynx, line 2280:
```suggestion
		Point mp = GameMousePos();
		if (Distance(mp, p) > 5)
```

gemrb/core/GUI/GameControl.cpp
bernds: Hmm, in that case we get the old behaviour. I'll change it to use the code from PerformSelectedAction instead.

gemrb/core/GUI/GameControl.cpp
bernds: I don't see any ambiguity, but ok.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: It's a bit obtuse, since there's no formation rotation going on when you're ordering around just one actor — is it just simpler to do it this way or what? If so, it deserves a comment.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: This is the style from most of the project, so the most consistent you can get. It will be enforced soon (still working on #161).

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2249,7 +2275,10 @@ void GameControl::CommandSelectedMovement(const Point& p, unsigned short Mod)
...
+			angle = AngleFromPoints(mp, p);
```
lynxlynxlynx, line 2281:
```suggestion
		Point mp = GameMousePos();
		if (Distance(mp, p) > 5) {
			angle = AngleFromPoints(mp, p);
		}
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -608,7 +611,9 @@ void GameControl::DrawSelf(Region screen, const Region& /*clip*/)
 	Point gameMousePos = GameMousePos();
 	// draw reticles
 	if (isFormationRotation) {
-		double angle = AngleFromPoints(gameMousePos, gameClickPoint);
+		double angle = formationBaseAngle;
+		if (Distance (gameMousePos, gameClickPoint) > 5)
```
lynxlynxlynx, line 615: This 5px dead zone appears twice, so it'd be nice if you created a named constant for it, with a comment on top.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: Not sure what you mean, but let's see.

gemrb/core/GUI/GameControl.cpp
bernds: I find it easier to think about when all movement is handled identically. The MouseUp handler tests isFormationRotation to determine whether we are moving or not, which I assume is what was leading to the various oddities, and that goes away once we eliminate the special case here.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+
+	if (core->GetGame()->selected.size() == 0) {
+		isFormationRotation = false;
+		return;
+	}
+
+	isFormationRotation = true;
+	Game* game = core->GetGame();
+	int max = game->GetPartySize(false);
+	formationBaseAngle = 0;
```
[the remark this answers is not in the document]
bernds: Err.. ok, but why?

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2028,33 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
 	return false;
 }
 
+// Calculate the angle between a clicked point and the first selected character,
+// so that we can set a sensible orientation for the formation
+void GameControl::InitFormation (const Point &clickPoint)
+{
+	/* We get called on initial click, but also on drag.  Perform the setup only once.  */
```
bradallred, line 2035: this comment appears to be incorrect

bernds: Ah, that was from before I moved the Alt-modifier stuff to also happen on mouse down.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: Ok, so the problem is that isFormationRotation is overloaded in the common/chess sense of the word then.

gemrb/core/GUI/GameControl.h
lynxlynxlynx: Yes, indeed. It will enable us to remove most of the ugly constructor initializations for the classes with tons of fields. :)

PR #961, opened 2020-10-06
Formations changes v3

bernds:
## Description
This should hopefully address all the comments in the previous pull request (#960) so far.

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -50,6 +50,10 @@
 
 namespace GemRB {
 
+// A deadzone when rotating the formation, to prevent erratic movement near
+// the initial click.
+const int rotate_deadzone = 5;
+
 #define FORMATIONSIZE 10
 typedef Point formation_type[FORMATIONSIZE];
 ieDword formationcount;
@@ -183,6 +187,9 @@ Point GameControl::GetFormationPoint(const Point& origin, size_t pos, double ang
 
 	Point stepVec;
 	int direction = (pos % 2 == 0) ? 1 : -1;
+
+	/* Correct for the innate orientation of the points.  */
+	angle += M_PI_2;
 
 	if (pos < FORMATIONSIZE) {
 		// calculate new coordinates by rotating formation around (0,0)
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks and another thanks for the responsiveness! LGTM now. 👍🏾

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2038,6 +2034,31 @@ bool GameControl::HandleActiveRegion(InfoPoint *trap, Actor * actor, const Point
...
+// so that we can set a sensible orientation for the formation.
+void GameControl::InitFormation(const Point &clickPoint)
+{
+	if (isFormationRotation) return;
+
+	// Of course single actors don't get rotated, but we need to ensure
+	// isFormationRotation is set in all cases where we initiate movement,
+	// since OnMouseUp tests for it.
+	if (core->GetGame()->selected.size() == 0) {
+		isFormationRotation = false;
```
bradallred, line 2047: this can't possibly be true given the check on line 2041. you could just combine the 2 checks.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -50,6 +50,10 @@
 
 namespace GemRB {
 
+// A deadzone when rotating the formation, to prevent erratic movement near
+// the initial click.
+const int rotate_deadzone = 5;
```
bradallred, line 55: now that we use c++11 we should favor `constexpr` in cases like this

```suggestion
constexpr int rotate_deadzone = 5;
```

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2249,7 +2282,11 @@ void GameControl::CommandSelectedMovement(const Point& p, unsigned short Mod)
 
 	double angle = 0.0;
```
bradallred, line 2283: shouldn't `angle` be initialized to `formationBaseAngle` instead? seems like without that non rotation events will always orient to `0.0` instead of in the direction of the party. so now regular clicks and rotations behave differently I think.

bernds: All regular clicks should go through a call to InitFormation now. In fact, in the previous pull request there was a question about one of these calls.
CommandSelectedMovement can still be called from other paths, such as walking to a door in PerformSelectedAction.

PR #963, opened 2020-10-07
Make '1'..'6' hotkeys for the portrait buttons

bernds:
## Description

I noticed a regression from master: the keys '1'..'6' no longer work in the inventory (or on the character screen). The problem is that they are handled in GameControl, but that gets disabled when a top window is added.
Adding hotkeys to the portrait buttons seems to fix it.

Out of curiousity, where exactly are the hotkeys 'I' and 'R' added? No matter how hard I grep I just can't find the location. I'm guessing it's InitOptionButton, but if so, where does it get the keys 'I' and 'R'?

### DIFF
--- a/gemrb/GUIScripts/GUICommonWindows.py
+++ b/gemrb/GUIScripts/GUICommonWindows.py
@@ -1707,7 +1707,7 @@ def OpenPortraitWindow (needcontrols=0, pos=WINDOW_RIGHT|WINDOW_VCENTER):
 	SelectionChanged ()
 	return Window
 
-def UpdatePortraitWindow ():
+def UpdatePortraitWindow (indialog = False):
 	"""Updates all of the portraits."""
 
 	Window = PortraitWindow
@@ -1718,9 +1718,13 @@ def UpdatePortraitWindow ():
 	PortraitButtons = GetPortraitButtonPairs (Window)
 	for i, Button in PortraitButtons.iteritems():
 		pcID = i + 1
+		if indialog:
+			Button.SetHotKey(None)
 		if (pcID <= GemRB.GetPartySize()):
 			Button.SetAction(lambda btn, val, pc=pcID: GemRB.GameControlLocateActor(pc), IE_ACT_MOUSE_ENTER);
 			Button.SetAction(lambda: GemRB.GameControlLocateActor(-1), IE_ACT_MOUSE_LEAVE);
+			if (i < 6 and not indialog):
+				Button.SetHotKey(chr(ord('1') + i), 0, True)
 
 		if GameCheck.IsPST():
 			UpdateAnimatedPortrait(Window, i)
--- a/gemrb/GUIScripts/GUIWORLD.py
+++ b/gemrb/GUIScripts/GUIWORLD.py
@@ -68,7 +68,7 @@ def DialogStarted ():
 	GemRB.GameSetScreenFlags(GS_DIALOG, OP_OR)
 
 	if GUICommonWindows.PortraitWindow:
-		GUICommonWindows.UpdatePortraitWindow ()
+		GUICommonWindows.UpdatePortraitWindow (True)
 
 	ContinueWindow = OpenDialogButton(9)
 
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -762,18 +762,12 @@ bool GameControl::OnKeyPress(const KeyboardEvent& Key, unsigned short mod)
 				case '=':
 					SelectActor(-1);
 					break;
-				case '1':
-				case '2':
-				case '3':
-				case '4':
-				case '5':
-				case '6':
-					game->SelectPCSingle(keycode-'0');
-					SelectActor(keycode-'0');
-					break;
 				case '7': // 1 & 2
 				case '8': // 3 & 4
 				case '9': // 5 & 6
+					// We do not handle the range 1..6, these are handled as hotkeys
+					// for the portrait buttons, so that they remain working when the
+					// inventory screen is open.
 					game->SelectActor( NULL, false, SELECT_NORMAL );
 					i = game->GetPartySize(false);
 					pc = 2*(keycode - '6')-1;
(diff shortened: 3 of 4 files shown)


bradallred:
> Out of curiousity, where exactly are the hotkeys 'I' and 'R' added? No matter how hard I grep I just can't find the location. I'm guessing it's InitOptionButton, but if so, where does it get the keys 'I' and 'R'?

Yes, it is done via `InitOptionButton `. When you pass an integer to `SetHotKey` instead of a string character it performs a lookup in our `KeyMap` (keymap.ini) for the appropriate hot key.

bradallred:
I'm trying to think of the consequences of using a hotkey instead... do the dialog options still work? We could use `GameControl::DispatchEvent` like we do with tab and escape if we need to.

gemrb/GUIScripts/GUICommonWindows.py
```diff
@@ -1721,6 +1721,8 @@ def UpdatePortraitWindow ():
 		if (pcID <= GemRB.GetPartySize()):
 			Button.SetAction(lambda btn, val, pc=pcID: GemRB.GameControlLocateActor(pc), IE_ACT_MOUSE_ENTER);
 			Button.SetAction(lambda: GemRB.GameControlLocateActor(-1), IE_ACT_MOUSE_LEAVE);
+		        if (i < 6):
+				Button.SetHotKey(chr(ord('1') + i), 0, True)
```
lynxlynxlynx, line 1725: I'm surprised python didn't exit complaining here, since the indentation is all wrong.

bernds: I'm not the greatest Python expert - how is it wrong?

bernds:
Hmm, good point about the dialog options. I didn't think of that, and they do in fact no longer work. I'll try to think of a solution.

bradallred:
> I'll try to think of a solution.

Is the `DispatchEvent` suggestion no good either? should work fine.

gemrb/GUIScripts/GUICommonWindows.py
lynxlynxlynx: You used spaces on the first line, while everything uses tabs. I guess you have a tab width of 2 set, so it's easier to make the mistake.

bernds: Oh wow. That's emacs autoindenting it, and using spaces despite the fact that I thought I'd set it up not to (and only in place of one tab apparently).

bradallred:
Looking at it, I actually think `GameControl::DispatchEvent` should simply be:

```
bool GameControl::DispatchEvent(const Event& event)
{
    return OnKeyPress(event.keyboard, 0); // maybe 0 should be some call to EventMgr to get the modifier state
}
```

and all the key logic should be handled in `GameControl:: OnKeyPress`. this is the pattern we use in `TextArea::SpanSelector` too.

bernds:
Presumable with a check for event.type == Event::KeyDown, and event.mod for the modifier? That seems to work, but from what I can tell we'll be getting two copies of the OnKeyPress call. The normal one still goes through, with a backtrace that looks like this:

#0  GemRB::GameControl::OnKeyPress (this=0x7e5200, Key=..., mod=0) at GameControl.cpp:691
#1  0x00007ffff7d91afd in GemRB::View::KeyPress (this=0x7e5200, key=..., mod=0) at View.cpp:667
#2  0x00007ffff7d979ff in GemRB::Window::DispatchKey (this=0x64b910, keyView=0x7e5200, event=...) at Window.cpp:424
#3  0x00007ffff7d96811 in GemRB::Window::DispatchEvent (this=0x64b910, event=...) at GUI/Window.cpp:524
#4  0x00007ffff7d9c57f in GemRB::WindowManager::DispatchEvent (this=0x6ea800, event=...) at WindowManager.cpp:392
etc.

So, then I tried renaming OnKeyPress to KeyPressed so that we only call it from DispatchEvent, and doing the same for KeyReleased for consistency - but then I found that KeyUp events never seem to make it to DispatchEvent.

The other oddity I noticed is that the DispatchEvent return value appears to be unused. I had imagined the caller would stop event processing if it returns true.

bernds:
Ok, figured out a few more things - the DispatchEvent method is just misnamed, it's a local callback which is registered only for KeyDown events. So, new idea is to have KeyPressed and KeyReleased methods in place of DispatchEvent and register them both in the constructor.

That seems to work as expected. If that sounds good to you I'll try pushing to this branch/pull request.

bradallred:
I _think_ that sounds good :)

bernds:
New version force-pushed. What I still don't quite get is how the dialog window manages to intercept the keyboard events so that these handlers never see them.

bradallred:
> dialog window manages to intercept the keyboard events

It shouldn't, `SpanSelector` also uses an event tap and all taps should get invoked for all matching events.

I'm leery about this change. It seems like now `GameControl` will get key events for these keys even if it is disabled or hidden. Can't we just bind the callback to `View::KeyPress` and `View::KeyRelease` to maintain consistency and keep `GameControl::OnKeyPress` and `GameControl::OnKeyRelease` overrides?

bernds:
Well, I believe I just implemented what was suggested, and the whole point was to make it get key events when it is disabled. A number of these key events (1-6, or spacebar for pause) should work while the inventory is shown.
But on the whole I would also prefer the previous hot key solution; we could just unregister them when dialog is initiated (although I haven't quite found the place where that happens).

bradallred:
> the whole point was to make it get key events when it is disabled

but all (global) key events and all ignoring the control state?

>  I just implemented what was suggested

my suggestion was to have the event tap callback invoke `OnKeyPress`, but i can see now that you would get duplicate events since taps can't stop normal event propagation.

bernds:
Yes - the original suggestion would have essentially the same effect as what I ended up with, but with duplicate OnKeyPress calls. I just eliminated those and made KeyUp and KeyDown consistent.
If you can point me to the location where we enter dialog mode I can try to improve the hotkeys solution.

bernds:
Reverted back to the original plan, and added code to clear the hotkeys while a dialog is in progress. Seems to work well now on my end.

bradallred:
yes, I like this solution. if its not too much trouble could you rebase it onto the change I just pushed up for `Button::SetHotKey`? we don't need to add a new GUIScript function for this.

bernds:
Had to make one additional change to actually call SetHotKey, but otherwise it seems to work.

bradallred:
arg that is my fault. the `Py_RETURN_NONE` shouldn't be there. I'll merge and fix

PR #964, opened 2020-10-07
Ensure the swap cursor does not stay around after clicking a portrait

bernds:
## Description

Observed behaviour: clicking on a portrait and moving the mouse a little
switches to the portrait swap cursor.  If the mouse is released while
still over the same portrait, the cursor does not revert back.

This fixes it by adding a SetCursor call to OnMouseUp to match the one in
OnMouseDrag.

This change also fixes the inconsistent indentation in this function.


bradallred:
I think this ought to be moved to the `IE_GUI_BUTTON_PRESSED` case since it should only happen for the "dragged" button. Real drag operations are already cleaned up in `CompleteDragOperation` which would get called if you had actually dragged to another `View`, the "drag" in `OnMouseUp` is a misnomer and is for picked up inventory items if that wasn't clear.

bernds:
Do you mean a few lines down in the same function? The actual State at this point seems to be IE_GUI_BUTTON_LOCKED_PRESSED, however. If you can confirm that this is as expected, I'll make the change.

bradallred:
I guess that is right. im not actually sure about what `IE_GUI_BUTTON_LOCKED_PRESSED` is all about. the comments around its declaration seem... wrong? I don't know how a portrait button is getting a locked state.

PR #966, opened 2020-10-07
Make hitpoint display go away when Tab is released

bernds:
## Description
This fixes what seems to be a simple oversight and brings behaviour in line with the IE. When the Tab key is released, the overhead hitpoint display should disappear.

There were two problems, fixed in two separate commits - there was no code to hide the hitpoints in the KeyUp handler, and SDL didn't properly translate its keycodes. The code in SDL could probably be improved further, but I think what I have is the minimum safe fix.


bradallred:
great change thank you

Issue #952, opened 2020-10-01
IWD: Pathing problems in AR8011 


fizzet:
#### Bug description

The new pathfinder apparently has problem with bigger actors. It's pretty obvious generally if you move through this level, as most of the elementals present won't be able to approach the group or attack properly.

The fire giant at 800/1410 shows this problem as well since his script starts a cutscene and then moves him towards the group. If you approach him from the east that works about half the time in my tests. When it doesn't, the cutscene simply hangs.

lynxlynxlynx:
We were testing mostly with dragons, so it's unexpected that something smaller would have problems. 
I have to reinstall my iwd, but a few things come to mind immediately:
- are their circle sizes correct?
  - if so, perhaps we're constructing too big patches to block on the searchmap
- does he get stuck because of a narrow search map or something else? Anything interesting in the log?

fizzet:
Log output differs. Sometimes I get spammed with:

`[FindPath/DEBUG]: s = (1060, 1397), d = (1221, 1404), caller = Feuerriese, dist = 0, size = 4`

but I sometimes also get

`FindPath/DEBUG]: Feuerriese can't fit in destination`

There are somewhat similar problems when moving towards big enemies. If I order the whole party to attack a big foe, at least some party members will stop short of the enemy and not attack. Others might circle around him to the opposite side before attacking.

lynxlynxlynx:
The circling back was noticed before and we decided to treat it as a feature. Dragon bias. :)

Actors should definitely not stop short though. Perhaps its hitting the simplistic weapon range calculation we use. We just multiply by 10, but it's probably a proper range like everything else, taking perspective into account (so the correct factor is between 12-16).

@kmfrick any thoughts? Can you try the area yourself?

kmfrick:
Ground circles look okay, trying to reproduce stopping short right now.

fizzet:
No. Just approaching that giant should do it. I can provide a save if necessary.

kmfrick:
Are you referring to the scripted dialogue with the elf woman? Because if it's that, I can't reproduce it getting stuck - but will try harder - and if it's not, then I can't get the cutscene to trigger. In any case, a save would be really helpful, thank you!!

fizzet:
No, it's not about the elf, it's about the fire giant I mentioned above.

Here's a save (IWD + HoW). Just move the party a bit to the southwest. Chances for a hang-up are higher if you stay south of the rails.
[...]

PR #967, opened 2020-10-08
Tweak distance checks

kmfrick:
## Description
- Tweaks a check that gives up on pathfinding to use `MAX_OPERATING_DISTANCE` in order to be consistent with the `Actions` code.
- Tweaks a "too big to fit" check in `Map::FindPath` to account for maximum allowed distance. Tested with the Draconis battle that prompted it in the first place.

Prompted by #952 

## Checklist

- [x] I extended the documentation, if necessary (NA)

### DIFF
--- a/gemrb/core/Scriptable/Scriptable.cpp
+++ b/gemrb/core/Scriptable/Scriptable.cpp
@@ -22,6 +22,7 @@
 #include "win32def.h"
 #include "strrefs.h"
 #include "ie_cursors.h"
+#include "voodooconst.h"
 
 #include "DialogHandler.h"
 #include "DisplayMessage.h"
@@ -2222,8 +2223,6 @@ void Movable::BumpBack()
 // for a random time (inspired by network media access control algorithms) or just stops if
 // the goal is close enough.
 void Movable::DoStep(unsigned int walkScale, ieDword time) {
-	const int XEPS = 72;
-	const int YEPS = 36;
 	Actor *actor = nullptr;
 	if (Type == ST_ACTOR) actor = (Actor*)this;
 	// Only bump back if not moving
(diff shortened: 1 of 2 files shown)


gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -2265,11 +2264,12 @@ void Movable::DoStep(unsigned int walkScale, ieDword time) {
 
 		if (BlocksSearchMap() && actorInTheWay && actorInTheWay != this && actorInTheWay->BlocksSearchMap()) {
 			// Give up instead of bumping if you are close to the goal
-			if (!(step->Next) && std::abs(nmptStep.x - Pos.x) < XEPS && std::abs(nmptStep.y - Pos.y) < YEPS) {
+			if (!(step->Next) && PersonalDistance(nmptStep, this) < MAX_OPERATING_DISTANCE) {
```
lynxlynxlynx, line 2267: Did you use PersonalDistance deliberately? It will shrink the distance for actors and considering the sizes involved, I'm surprised dragons can still move.

kmfrick: I used it for consistency with WithinPersonalRange which is used by MoveNearerTo, as well as with the PersonalDistance checks in e.g. UseItem. I tested it with the draconis battle and did not find any issues, but I still have some more tests to conduct which I plan to do today.

lynxlynxlynx: If you can move him around after ctrl-q, then it should be fine.

kmfrick: Yup, works with very close moves too.

lynxlynxlynx: Thanks, I'll merge it then.

PR #968, opened 2020-10-08
Drag/drop improvements

bernds:
## Description
One of the things I noticed very soon is that inventory drag/drop behaves differently than in BG2 running on the real Infinity Engine (and every other drag/drop implementation). GemRB always requires an extra click, which makes things cumbersome.

This set of changes fixes most of the problem by improving item drag support in the Button class and in Window::DispatchMouseUp. The old behaviour seems to be still supported - if you click and release without dragging the mouse, the item is picked up as previously.

One problem remains, which is dragging to a portrait: for some reason the MouseUp event doesn't find the portrait button, and an extra click is still required there.

The final change is purely optional - I originally modified some Python code as well and ended up unifying four blocks of code identical in three cases and near identical in IWD2. Feel free to drop it if there's some reason not to do this. I tested with BG2 only.

### DIFF
--- a/gemrb/GUIScripts/InventoryCommon.py
+++ b/gemrb/GUIScripts/InventoryCommon.py
@@ -544,6 +544,39 @@ def OpenItemAmountWindow (btn, slot):
 	Window.ShowModal (MODAL_SHADOW_GRAY)
 	return
 
+# count goes up to 5 for all games except IWD2, which uses 6, and
+# PST, which does not use this function.
+def UpdateGroundSlots (Window, pc, count):
+	"""Updates the five (or six, for IWD2) inventory ground slots."""
+
+	Container = GemRB.GetContainer (pc, 1)
+	TopIndex = GemRB.GetVar ("TopIndex")
+	for i in range (count):
+		if i<5:
+			Button = Window.GetControl (i+68)
+		else:
+			Button = Window.GetControl (i+76)
+
+		if GemRB.IsDraggingItem ()==1:
+			Button.SetState (IE_GUI_BUTTON_FAKEPRESSED)
+		else:
+			Button.SetState (IE_GUI_BUTTON_ENABLED)
+		Button.SetAction (OnDragItemGround, IE_ACT_DRAG_DROP_DST)
+		Slot = GemRB.GetContainerItem (pc, i+TopIndex)
+
+		if Slot == None:
+			Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, None)
+			Button.SetEvent (IE_GUI_BUTTON_ON_RIGHT_PRESS, None)
+			Button.SetEvent (IE_GUI_BUTTON_ON_SHIFT_PRESS, None)
+		else:
+			Button.SetAction (OnDragItemGround, IE_ACT_DRAG_DROP_CRT)
+			Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, OnDragItemGround)
+			Button.SetEvent (IE_GUI_BUTTON_ON_RIGHT_PRESS, OpenGroundItemInfoWindow)
+			Button.SetEvent (IE_GUI_BUTTON_ON_SHIFT_PRESS, OpenGroundItemAmountWindow)
+
+		UpdateInventorySlot (pc, Button, Slot, "ground")
+	return
+
 def UpdateSlot (pc, slot):
 	"""Updates a specific slot."""
 
--- a/gemrb/GUIScripts/InventoryCommon.py
+++ b/gemrb/GUIScripts/InventoryCommon.py
@@ -43,7 +43,7 @@ UpdateInventoryWindow = None
 
 def OnDragItemGround (btn, slot):
 	"""Drops and item to the ground."""
-	
+
 	pc = GemRB.GameGetSelectedPCSingle ()
 	slot = slot + GemRB.GetVar ("TopIndex") - btn.ControlID
 
@@ -85,7 +85,7 @@ def OnDragItem (btn, slot):
 
 	pc = GemRB.GameGetSelectedPCSingle ()
 	slot_item = GemRB.GetSlotItem (pc, slot)
-	
+
 	if not GemRB.IsDraggingItem ():
 		item = GemRB.GetItem (slot_item["ItemResRef"])
 		GemRB.DragItem (pc, slot, item["ItemIcon"], 0, 0)
@@ -565,6 +565,7 @@ def UpdateGroundSlots (Window, pc, count):
 		Slot = GemRB.GetContainerItem (pc, i+TopIndex)
 
 		if Slot == None:
+			Button.SetAction (None, IE_ACT_DRAG_DROP_CRT)
 			Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, None)
 			Button.SetEvent (IE_GUI_BUTTON_ON_RIGHT_PRESS, None)
 			Button.SetEvent (IE_GUI_BUTTON_ON_SHIFT_PRESS, None)
@@ -636,7 +637,7 @@ def UpdateSlot (pc, slot):
 	UpdateInventorySlot (pc, Button, slot_item, "inventory", SlotType["Type"]&SLOT_INVENTORY == 0)
 
 	if slot_item:
-		Button.SetAction(OnDragItem, IE_ACT_DRAG_DROP_CRT)
+		Button.SetAction (OnDragItem, IE_ACT_DRAG_DROP_CRT)
 		Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, OnDragItem)
 		Button.SetEvent (IE_GUI_BUTTON_ON_RIGHT_PRESS, OpenItemInfoWindow)
 		Button.SetEvent (IE_GUI_BUTTON_ON_SHIFT_PRESS, OpenItemAmountWindow)
@@ -664,6 +665,7 @@ def UpdateSlot (pc, slot):
 			Button.SetTooltip ("")
 			itemname = ""
 
+		Button.SetAction (None, IE_ACT_DRAG_DROP_CRT)
 		Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, None)
 		Button.SetEvent (IE_GUI_BUTTON_ON_RIGHT_PRESS, None)
 		Button.SetEvent (IE_GUI_BUTTON_ON_SHIFT_PRESS, None)
--- a/gemrb/GUIScripts/bg1/GUIINV.py
+++ b/gemrb/GUIScripts/bg1/GUIINV.py
@@ -244,29 +244,7 @@ def RefreshInventoryWindow (Window):
 	Color = GemRB.GetPlayerStat (pc, IE_MINOR_COLOR, 1) & 0xFF
 	Button.SetBAM ("COLGRAD", 0, 0, Color)
 
-	# update ground inventory slots
-	Container = GemRB.GetContainer (pc, 1)
-	TopIndex = GemRB.GetVar ("TopIndex")
-	for i in range (5):
-		Button = Window.GetControl (i+68)
-		if GemRB.IsDraggingItem ()==1:
-			Button.SetState (IE_GUI_BUTTON_FAKEPRESSED)
-		else:
-			Button.SetState (IE_GUI_BUTTON_ENABLED)
-		Button.SetAction (InventoryCommon.OnDragItemGround, IE_ACT_DRAG_DROP_DST)
-		Slot = GemRB.GetContainerItem (pc, i+TopIndex)
-
-		if Slot == None:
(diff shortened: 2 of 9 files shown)


bradallred:
how do you feel about changing `Interface::GetDraggedItem` to return a `DragOp`? so that we can just treat it the same way as a real drag instead of introducing another code path. we also wouldn't need `AcceptsItemDrag` as we already have `AcceptsDragOperation`

bernds:
I feel like in an ideal world, there would be only one drag mechanism. But I also felt like I couldn't think through all the potential consequences, and I don't really understand DragOp yet. Hence I came up with this solution in order to make an improvement while risking little breakage.
Unifying the two might be a nice followup project.

bradallred:
I've gone ahead and just changed the item drag into a dummy `DragOp` for now. In case somebody is compelled in the future to actually properly implement item dragging. Should eliminate the need for the things mentioned above.

bernds:
I'm trying to understand this. There's a new comment:

+// TODO: there is no reason why this can't be generated directly from
+// the inventory button drag event using the button value as the slot id
+// to get the appropriate CREItem

Do you mean in the Button:OnMouseDrag handler? I don't think we can create a dragged item there, since some items may not be draggable (think Imoen's belt).

bradallred:
I mean `Button::DragOperation`which is invoked automatically in response to a drag event. Non draggable items are easily solved by disabling the button preventing the drag. This is all for future reference. in order to properly do this would require a lot more changes.

bernds:
Disabling the button would prevent the message "This item cannot be removed" when the player tries to initiate a drag and might cause player confusion.

Do you want me to pursue a variant of my changes rebased on what is there now, or should I just drop it? It's a bit of a shame to leave the usability issue in there when a simple fix is available.

bradallred:
we're getting off into the weeds here. We aren't trying to pursue a full drag implementation at this time. The changes I made were so you could use the same code path in `Window::DispatchMouseUp` for all `DragOp`s and use `AcceptsDragOperation` instead of `AcceptsItemDrag` so that in the future implementing a proper drag capability isn't any more difficult.

instead of:
```
// Drag and item drag are two separate mechanisms.
 bool item_drag = core->GetDraggedItem() != NULL;
 if (drag) {
 	if (target->AcceptsDragOperation(*drag) && drag->dragView != target) {
 		target->CompleteDragOperation(*drag);
 	}
} else if (item_drag && target && target->AcceptsItemDrag()) {
 	target->MouseUp(me, mod);
 } else if (trackingView) {
```

just do:
```
auto dragop = (drag) ? drag : core->GetDraggedItem();
 if (dragop) {
 	if (target->AcceptsDragOperation(* dragop) && drag->dragView != target) {
 		target->CompleteDragOperation(* dragop);
 	}
 } else if (trackingView) {
```

and change `Button:: AcceptsDragOperation` to be able to handle the item drags.

bernds:
The Button can quite easily determine if it is an inventory button by checking whether an Action::DragDrop was registered.

I'm still puzzling over the change you committed, and I still feel it just makes things more awkward. Beyond the issues I have with the comment, I also wonder why there is a View inside ItemDragOp (which seems unused and brings its sizeof to 224).
The other thing is that the suggested
  auto dragop = (drag) ? drag : core->GetDraggedItem();
does not compile, since these are different types. I could use get() on each arm of the conditional to just get a DragOp * pointer if that is the intention. However, then the currently available information from DragOp seems insufficient for AcceptsDragOperation to work out what we're dealing with (we should not allow a portrait to be dragged to an inventory slot, for example). Do you want to add a type field to DragOp?

bradallred:
> determine if it is an inventory button by checking whether an Action::DragDrop was registered.
this is not a future proof method tho

> I also wonder why there is a View inside ItemDragOp (which seems unused and brings its sizeof to 224)
There is a comment explaining it. who cares about 224 bytes in this context?

> Do you want to add a type field to DragOp?
just use `dynamic_cast` for now

lynxlynxlynx:
> One of the things I noticed very soon is that inventory drag/drop behaves differently than in BG2 running on the real Infinity Engine (and every other drag/drop implementation). GemRB always requires an extra click, which makes things cumbersome.

What do you mean by that? Item dragging requires a click to complete both in the originals and EEs. Reordering the inventory would be a PITA if an item was dropped when over the first compatible slot. It's portrait dragging that uses a typical drag and finishes as you release the mouse, but that works fine already as well.

chilvence:
I've actually only noticed it now that it's been pointed out, but the original games have 'click and hold' style dragging, which drops the item on mouse button release. GemRB requires you to click and release to drag an item.

lynxlynxlynx:
Ah, I see where the confusion stems from now: it supports both. If you're the type to hold the mouse button down, it will drop on release (I guess it checks if the underlying slot is different). Otherwise, with a normal click, it attaches the item to the cursor and you are forced to click to drop it.

bradallred:
I've implemented item dragging generically (still some work to do) so I dont think we need the `Button`/`Window` changes anymore, but the Python stuff would be good to have still.

bernds:
I have to say I'm still somewhat unhappy with this new implementation. Besides the bugs, which I've hopefully fixed, it has a number of ugly hacks that were unnecessary with my original patch - control name comparison, timer abuse, and the dummy view thing.

bradallred:
please stop trying to make changes to the core drag and drop code. I've already stated im in the process of finishing it. we already have issues open for some of these. if you see a bug that doesn't have an issue please open one for discussion and how to address it.

bernds:
It sounded like you were done, and it's been left broken for a week. Well, I guess no point in trying to help.

bradallred:
Sorry, I should have been more clear about it. Don't get me wrong I do appreciate both help and feedback. All your Python changes are welcome. I would merge them in absolutely.

Subviews is a development branch. Expect it to both have bugs and for bugs to not be immediately addressed. I welcome help in continuing the work, but it would be much better if you would open issues about things you find on that branch so we aren't in conflict and approaches can be discussed. I welcome your ideas and contributions.

bernds:
I find it pointless to open issues when I can just as easily go in and fix things. That should save everyone time. Did you look at the changes, is there anything you actually object to? To me everything looks like an obvious fix in the first three patches at least.

bradallred:
> I have to say I'm still somewhat unhappy with this new implementation. Besides the bugs, which I've hopefully fixed, it has a number of ugly hacks that were unnecessary with my original patch - control name comparison, timer abuse, and the dummy view thing.

One of the objectives with this branch is to give us the tools and framework for properly implementing event handling and provide generic event callbacks so that we can handle events via Python scripting instead of hardcoding things in the engine. This goes beyond the specifics of inventory item dragging. To me, it makes little sense to not steer towards that goal even if it requires a bit more work.

This is short list of the main issues I'm trying to address:

* a singular event handler so you were either limited in what you could accomplish, or had to rely on core side hacks (which i'm still in the process of removing)
* drag events were arbitrarily limited to buttons
* drag cursor management was a hardcoded hack
* each drag operation needed its own subclass. Again, this limits the ability to handle this generically and easily with our current Python bindings.

I wanted to address those issues. It's a large change and a work in progress. Yes, there are bugs. I'm sorry this displeases you, but bugs in new features on a development branch of an open source project where people volunteer their valuable time should be expected :)

I do understand your objections to hacks, truly. However, I believe your judgement is clouded by the hacks that were simply already in place. Everywhere in `Button.cpp` that has `IS_PORTRAIT` is a hack :) Buttons shouldn't care one iota about what a character portrait is. We have that because we originally lacked a way to handle and express it in Python. Likewise, the code that instantiates an `ItemDragOp` would never have been written that way had we possessed the framework we have now.

You may not realize this about me, but I'm a professional at refactoring old and large codebases. The "dummy `View`" hack you called out is a popular refactoring technique to push the hacks out of the code your are refactoring (in this back that would be the GUI/rendering/Event systems) and into the code that you would design differently if given the chance. You usually do this with an isolated adaptor, which is what I would label `ItemDragOp` as. The hack is a contained private member so it shouldn't leak out into other parts of the code. I could have avoided the whole thing with using `nullptr` checks and likely nobody would have said anything. However, having a NULL `dragView` makes no logical sense and those checks would in fact be a hack while masking the real issues which are quite clear when using a dummy.

I also understand the bad taste of `VarName`. It's easy to see why you would mistake it as simply the control name. One of its purposes is actually grouping controls (checkboxes, radios, etc) like I am doing here. There also is no readily available alternative to allow us to perform this association from Python so I'm afraid we are stuck for the time being. I think the pros outweigh the cons.

Now the "timer abuse" thing I simply disagree with you. The reason simply boils down to having an implementation flexible enough to handle a callback when a drag operation begins; it makes logical sense `DragOp` to exist prior to invoking the callback. It also would be a bug to _not_ invalidate `actionTimer` when doing a drag operation, else you can end up with repeat click events while dragging.

I opposed your original patch simply because it perpetuated the problems from before, not because you yourself caused any issues. Adding new methods specifically for an item drag, entrenching more drag code into `Button`, and retaining 2 distinct code paths for dragging things. None of these things were your fault and I don't blame you for making those choices. It was the shortest most obvious path to the goal. If this branch for refactoring the GUI for a more modern version didn't exist I'm sure I would have merged them. I'm not in anyway deriding you or your abilities.

> I find it pointless to open issues when I can just as easily go in and fix things. That should save everyone time. Did you look at the changes, is there anything you actually object to? To me everything looks like an obvious fix in the first three patches at least.

You seem frustrated which is why I suggested opening issues. That would allow for discussion early on. You write beautifully detailed commit messages, so I didn't think writing issues would be an anathema. Nobody here is going to insist on it, but I at least find them useful for historical reasons. Especially when the same issue reappears or when screenshots and a save help more than a 1000 work commit log :)

I'm not sure that I object to any of the fixes, I'm not the one that closed the pull request. Worst case is I merge it in and resolve conflicts with my ongoing work. It's not a big deal, but I would also like to keep you informed and prevent larger issues.

> It sounded like you were done, and it's been left broken for a week. Well, I guess no point in trying to help.

I'm sorry you feel that way. I wish I had enough free time to address everything in my life that quickly.
Things will break from time to time while development is in progress. It's not abnormal, especially on a development branch of a reverse engineered engine supporting 5 different games :) My free time is valuable and difficult to come by. It's hard to spend all of it here. I hope you understand.

bernds:
Well, since you spent a lot of time explaining your perspective, I'll add something about my viewpoint. My original idea was that GemRB has been around long enough that it must be reasonably far along, so I'll try a BG2 playthrough and fix any small issues that occur along the way. I made it to the third group of goblins before giving up on that idea because there are just too many problems, and new ones keep appearing.

I'm no stranger to open source projects, and "development branch" does not have to mean "constantly broken". I'm used to a fairly strict testing and no regression policy, which requires a little more care upfront but in the end helps everyone and improves the quality of the product. I'd say GemRB is somewhat unusual in that it seems to rapidly accumulate breakage from changes that are untested at best and known broken at worst. That not only moves it further away from the goal of being both playable and enjoyable, it also makes it hard to work with because you're constantly wondering "I didn't see this yesterday, did my last change break it?". And when an attempt to fix some of it is met with hostility, as in this case, the motivation to work on it approaches zero, at least for the time being.

As for a generic drag&drop framework - I guess I would see the point if this were Qt, but it's not a GUI library, it's an Infinity Engine clone, and the requirements are known. It has enough serious issues that overengineering that particular corner doesn't seem like the best way to spend time.

Feel free to fetch & cherry pick the branch if there's anything you see there of value.

lynxlynxlynx:
What you're describing as typical is the master branch; that's the one that's conservative and tries harder to be regression-free. We have the unfortunate problem that most of the systems can't be automatically tested, even when interactivity is not needed, otherwise we'd have test suites for everything. We're stuck with integration tests and manual testing on seven+ dynamic sets of data.

Development branches are there for experimenting and large-scale changes that can't easily be done iteratively. It's not ideal than we haven't been able to merge back faster in this case, but that's the way it is. Complaining about stability in something that is clearly WIP is a bit silly, though concrete reports are of course welcome. Is the problem that it *wasn't* clearly a WIP branch? If we weren't so close to rounding it up, it would make sense to add more docs to the site or contributing.md, but it's not clear when we'll be in a similar situation again, so I'm on the fence.

bradallred:
I'm positive that we could use and would welcome many improvements to our virtually non existent automated testing (we have exactly 0 unit tests). I don't think that we are completely unaware of these points, simply that we have limited time to dedicate to the project and there are many areas that require attention. I definitely agree we would be on better footing if we had good test coverage.

I don't recall being hostile towards you :) I'm not sure what I said that came off that way, but I certainly don't feel that way. I'm actually quite grateful for your contributions; I simply think that issues and conversations are a good thing to have.

lynxlynxlynx:
We actually do have a few unit tests for the text subsystem, but they're limited.

bernds:
Wouldn't be the first time the internet hasn't transported tone correctly, but I read "stop trying to change this code" as "go away", especially since it did not come with a conversation about the merits of the changes. Anyhow, never mind.

PR #969, opened 2020-10-08
Use Holder/Held refcounting for Palette

bernds:
I left GemRB running overnight, and in the morning the system was deep
into swap.  Running with -fsanitize=address suggests there are large
leaks of Palette structures.

Examining the code, it uses its own refcounting mechanism, and requires
its users to get acquire/release calls correct.  This all seems
unnecessarily fragile.

This change changes Palette to use Holder/Held.  All instances of raw
Palette * pointers are replaced by PaletteHolder, and (hopefully all)
manual acquire/release calls are removed.  One Cache was replaced with
a std::unordered_map, since it appears Cache can't hold non-pointers
(it uses hard casts to void *).

Tested with BG2: playing movies, going through the various panes
(inventory, character, world map, etc.), and resting.

### DIFF
--- a/gemrb/core/FontManager.h
+++ b/gemrb/core/FontManager.h
@@ -37,7 +37,7 @@ public:
 	FontManager();
 	virtual ~FontManager(void);
 
-	virtual Font* GetFont(ieWord pxSize, FontStyle style, Palette* = NULL)=0;
+	virtual Font* GetFont(ieWord pxSize, FontStyle style, PaletteHolder = nullptr)=0;
 };
 
 }
--- a/gemrb/core/GUI/TextSystem/Font.cpp
+++ b/gemrb/core/GUI/TextSystem/Font.cpp
@@ -141,7 +141,7 @@ void Font::GlyphAtlasPage::DumpToScreen(const Region& r)
 #endif
 
 
-Font::Font(Palette* pal, ieWord lineheight, ieWord baseline)
+Font::Font(PaletteHolder pal, ieWord lineheight, ieWord baseline)
 : palette(NULL), LineHeight(lineheight), Baseline(baseline)
 {
 	CurrentAtlasPage = NULL;
@@ -529,7 +529,7 @@ Sprite2D* Font::RenderTextAsSprite(const String& string, const Size& size,
 	return canvas;
 }
 
-void Font::SetAtlasPalette(Palette* pal) const
+void Font::SetAtlasPalette(PaletteHolder pal) const
 {
 	GlyphAtlas::const_iterator it;
 	for (it = Atlas.begin(); it != Atlas.end(); ++it) {
@@ -540,7 +540,7 @@ void Font::SetAtlasPalette(Palette* pal) const
 }
 
 size_t Font::Print(Region rgn, const String& string,
-				   Palette* color, ieByte alignment, Point* point) const
+				   PaletteHolder color, ieByte alignment, Point* point) const
 {
 	if (rgn.Dimensions().IsEmpty()) return 0;
 
@@ -570,7 +570,7 @@ size_t Font::Print(Region rgn, const String& string,
 		}
 	}
 
-	Palette* restore = NULL;
+	PaletteHolder restore = nullptr;
 	if (color) {
 		// we set palette in this way because we want all the Atlas pages to inherit the change
 		restore = palette;
@@ -720,10 +720,8 @@ Holder<Palette> Font::GetPalette() const
 	return palette;
 }
 
-void Font::SetPalette(Palette* pal)
+void Font::SetPalette(PaletteHolder pal)
 {
-	if (pal) pal->acquire();
-	if (palette) palette->release();
 	palette = pal;
 }
 
--- a/gemrb/core/GUI/TextSystem/Font.h
+++ b/gemrb/core/GUI/TextSystem/Font.h
@@ -50,6 +50,7 @@ enum FontStyle {
 };
 
 class Palette;
+using PaletteHolder = Holder<Palette>;
 
 #define IE_FONT_ALIGN_LEFT   0x00
 #define IE_FONT_ALIGN_CENTER 0x01
@@ -149,7 +150,7 @@ private:
 	GlyphAtlas Atlas;
 
 protected:
-	mutable Palette* palette;
+	mutable PaletteHolder palette;
 
 public:
 	const int LineHeight;
@@ -164,10 +165,10 @@ private:
 	size_t RenderLine(const String& string, const Region& rgn,
 					  Point& dp, ieByte** canvas = NULL) const;
 
-	void SetAtlasPalette(Palette* pal) const;
+	void SetAtlasPalette(PaletteHolder pal) const;
 
 public:
-	Font(Palette*, ieWord lineheight, ieWord baseline);
+	Font(PaletteHolder, ieWord lineheight, ieWord baseline);
 	virtual ~Font();
 
 	const Glyph& CreateGlyphForCharSprite(ieWord chr, const Sprite2D*);
@@ -178,8 +179,8 @@ public:
 	//allow reading but not setting glyphs
 	virtual const Glyph& GetGlyph(ieWord chr) const;
 
-	Holder<Palette> GetPalette() const;
-	void SetPalette(Palette* pal);
+	PaletteHolder GetPalette() const;
+	void SetPalette(PaletteHolder pal);
 
 	virtual int GetKerningOffset(ieWord /*leftChr*/, ieWord /*rightChr*/) const {return 0;};
 
@@ -190,7 +191,7 @@ public:
 	// the "point" parameter can be passed with a start point for rendering
 	// it will be filled with the point inside 'rgn' where the string ends upon return
 	size_t Print(Region rgn, const String& string,
-				 Palette* hicolor, ieByte Alignment, Point* point = NULL) const;
+				 PaletteHolder hicolor, ieByte Alignment, Point* point = nullptr) const;
 
 	/** Returns size of the string rendered in this font in pixels */
 	struct StringSizeMetrics {
--- a/gemrb/core/GUI/TextSystem/GemMarkup.h
+++ b/gemrb/core/GUI/TextSystem/GemMarkup.h
@@ -37,12 +37,12 @@ public:
 	};
 
 	GemMarkupParser();
-	GemMarkupParser(const Font* ftext, Palette* textPal = NULL,
-					const Font* finit = NULL, Palette* initPal = NULL);
+	GemMarkupParser(const Font* ftext, PaletteHolder textPal = nullptr,
+					const Font* finit = nullptr, PaletteHolder initPal = nullptr);
 	~GemMarkupParser() {};
 
-	void ResetAttributes(const Font* ftext = NULL, Holder<Palette> textPal = NULL,
-						 const Font* finit = NULL, Holder<Palette> initPal = NULL);
+	void ResetAttributes(const Font *ftext = nullptr, PaletteHolder textPal = nullptr,
+						 const Font *finit = nullptr, PaletteHolder initPal = nullptr);
 
 	void Reset();
 
@@ -60,8 +60,8 @@ private:
 		const Font* SwapFont;
 
 		public:
-		TextAttributes(const Font* text, Holder<Palette> textPal = NULL,
-					   const Font* init = NULL, Holder<Palette> initPal = NULL)
+		TextAttributes(const Font* text, PaletteHolder textPal = nullptr,
+					   const Font *init = nullptr, PaletteHolder initPal = nullptr)
 		: palette(textPal), swapPalette(initPal)
 		{
 			TextFont = text;
--- a/gemrb/core/GUI/WorldMapControl.cpp
+++ b/gemrb/core/GUI/WorldMapControl.cpp
@@ -91,11 +91,10 @@ void WorldMapControl::DrawSelf(Region rgn, const Region& /*clip*/)
 		Sprite2D* icon = m->GetMapIcon(worldmap->bam, OverrideIconPalette);
 		if( icon ) {
 			if (m == Area && m->HighlightSelected()) {
-				Palette *pal = icon->GetPalette();
+				PaletteHolder pal = icon->GetPalette();
 				icon->SetPalette(pal_selected.get());
 				video->BlitSprite( icon, xOffs, yOffs, &rgn );
 				icon->SetPalette(pal);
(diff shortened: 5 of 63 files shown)


bradallred:
I routinely run with `fsanitize=address` and have never seen any such thing

gemrb/core/Cache.h
```diff
@@ -32,6 +34,13 @@ namespace GemRB {
 typedef void (*ReleaseFun)(void *);
 #endif
 
+typedef std::array<char, 9> ieResRefKey;
```
bradallred, line 37: let's not use `typedef`. now that we use C++11 `using` is the way.

moot for this, since we ought to just use the `ResRef` class here too.

bradallred (changes requested):

bernds:
Might be driver-specific? As far as I remember, the reported leaks were somewhere in SDLVideo.

bernds:
Also - what exactly do you see when running with the sanitizer? I get a very large amount of output about lost allocations, most of them small. SDLTextureSprite2D seems to be the biggest offender after this patch.

bradallred:
non of my leak detectors report anything, nor does my memory use exhibit any symptoms of leaks.

bernds:
Just starting up the program and closing it reports many issues. Should we try to figure out whose leak checker is broken? Here's the largest one I'm seeing:

Indirect leak of 10529280 byte(s) in 13 object(s) allocated from:
    .0 0x7fca5061d628 in __interceptor_malloc 
    .1 0x7fca49e97cef in GemRB::MOSImporter::GetSprite2D() plugins/MOSImporter/MOSImporter.cpp:79
    .2 0x7fca493bb12a in GemRB::CHUImporter::GetWindow(unsigned long) const plugins/CHUImporter/CHUImporter.cpp:143
    .3 0x7fca500cf196 in GemRB::Interface::LoadWindow(unsigned long, GemRB::ResRef const&, GemRB::Window::WindowPosition) core/Interface.cpp:2579
    .4 0x7fca494024bc in GemRB_LoadWindow plugins/GUIScript/GUIScript.cpp:644

And I cannot see a corresponding free call for the pixels, but since this is all C-style manual memory management, it may just be hard to find. The data seems to be passed to SDL_CreateRGBSurfaceFrom, and the manpage says the pixels should be called after the corresponding SDL_FreeSurface. None of the calls to SDL_FreeSurface seem to have a corresponding free call near them?

bradallred:
I'm 100% positive my leak checker isn't broken. I doubt yours is either.

> And I cannot see a corresponding free call for the pixels

the `free` call for that particular example is in `Sprite2D::~Sprite2D`

bernds:
Ok, I can see the idea. But printf debugging shows that some of the memory allocated there reaches this destructor, but not everything. Investigating a bit further, I can trace the returned sprite to CHUImporter::GetWindow, where we have
	win->SetBackground(bg);
After that, the sprite has a refcount of 2. I can add
	bg->release ();
directly afterwards, and only then does the memory gets freed at program exit. That suggests to me that sprites have the same issue I was seeing with palettes: manual refcounting that is just too error prone.

bradallred:
yes, I definitely agree with that assessment. Something we have all complained about for a long time and never gotten around to addressing.

bernds:
Ok, well, in case it got lost in the noise: the branch was updated to use the ResRef class. If it's accepted I might try to tackle sprites as well.

bradallred:
yeah, we may as well get both now. That would be awesome.

FWIW I do see the leaks when I actually exit the program normally. I should stop using the debugger stop button ha ha.

gemrb/core/FontManager.h
```diff
@@ -37,7 +37,7 @@ class GEM_EXPORT FontManager : public Resource {
 	FontManager();
 	virtual ~FontManager(void);
 
-	virtual Font* GetFont(ieWord pxSize, FontStyle style, Palette* = NULL)=0;
+	virtual Font* GetFont(ieWord pxSize, FontStyle style, PaletteHolder = NULL)=0;
```
bradallred, line 40:
```suggestion
	virtual Font* GetFont(ieWord pxSize, FontStyle style, PaletteHolder = nullptr)=0;
```

gemrb/core/GUI/TextSystem/Font.h
```diff
@@ -190,7 +191,7 @@ class GEM_EXPORT Font {
 	// the "point" parameter can be passed with a start point for rendering
 	// it will be filled with the point inside 'rgn' where the string ends upon return
 	size_t Print(Region rgn, const String& string,
-				 Palette* hicolor, ieByte Alignment, Point* point = NULL) const;
+				 PaletteHolder hicolor, ieByte Alignment, Point* point = NULL) const;
```
bradallred, line 194:
```suggestion
				 PaletteHolder hicolor, ieByte Alignment, Point* point = nullptr) const;
```

gemrb/core/GUI/TextSystem/GemMarkup.h
```diff
@@ -37,12 +37,12 @@ class GemMarkupParser {
 	};
 
 	GemMarkupParser();
-	GemMarkupParser(const Font* ftext, Palette* textPal = NULL,
-					const Font* finit = NULL, Palette* initPal = NULL);
+	GemMarkupParser(const Font* ftext, PaletteHolder textPal = NULL,
+					const Font* finit = NULL, PaletteHolder initPal = NULL);
```
bradallred, line 41:
```suggestion
	GemMarkupParser(const Font* ftext, PaletteHolder textPal = nullptr,
					const Font* finit = nullptr, PaletteHolder initPal = nullptr);
```

gemrb/core/GUI/TextSystem/GemMarkup.h
```diff
@@ -37,12 +37,12 @@ class GemMarkupParser {
...
 	~GemMarkupParser() {};
 
 	void ResetAttributes(const Font* ftext = NULL, Holder<Palette> textPal = NULL,
-						 const Font* finit = NULL, Holder<Palette> initPal = NULL);
+						 const Font* finit = NULL, PaletteHolder initPal = NULL);
```
bradallred, line 45:
```suggestion
						 const Font* finit = nullptr, PaletteHolder initPal = nullptr);
```

gemrb/core/GUI/WorldMapControl.cpp
```diff
@@ -129,7 +128,7 @@ void WorldMapControl::DrawSelf(Region rgn, const Region& /*clip*/)
 		if (!m->GetCaption())
 			continue;
 
-		Palette* text_pal = pal_normal.get();
+		PaletteHolder text_pal = pal_normal.get();
```
bradallred, line 131:
```suggestion
		PaletteHolder text_pal = pal_normal;
```

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
```diff
@@ -106,9 +106,9 @@ void SDLSurfaceSprite2D::UnlockSprite() const
 }
 
 /** Get the Palette of a Sprite */
-Palette* SDLSurfaceSprite2D::GetPalette() const
+PaletteHolder SDLSurfaceSprite2D::GetPalette() const
 {
-	Palette* palette = surface->palette.get();
+	PaletteHolder palette = surface->palette.get();
```
bradallred, line 111:
```suggestion
	PaletteHolder palette = surface->palette;
```

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
```diff
@@ -119,27 +119,25 @@ Palette* SDLSurfaceSprite2D::GetPalette() const
...
 
-void SDLSurfaceSprite2D::SetPalette(Palette* pal)
+void SDLSurfaceSprite2D::SetPalette(PaletteHolder pal)
 {
-	Palette* palette = surface->palette.get();
+	PaletteHolder palette = surface->palette.get();
 
 	// we don't use shared palettes because it is a performance bottleneck on SDL2
-	assert(pal != palette);
+	assert(pal.get() != palette.get());
```
bradallred, line 138: we should have `operator==` overloads defined for this

```suggestion
	assert(pal != palette);
```

gemrb/core/GUI/TextSystem/Font.cpp
```diff
@@ -570,7 +570,7 @@ size_t Font::Print(Region rgn, const String& string,
 		}
 	}
 
-	Palette* restore = NULL;
+	PaletteHolder restore = NULL;
```
bradallred, line 573:
```suggestion
	PaletteHolder restore = nullptr;
```

gemrb/core/GameData.cpp
```diff
@@ -64,12 +64,16 @@ static void ReleaseEffect(void *poi)
 	delete ((Effect *) poi);
 }
 
-static void ReleasePalette(void *poi)
+// Very similar to Cache::MyHashKey, but returns size_t and can be used from
+// things like unordered_map.
+size_t ieResRefHash::operator() (const ResRef k) const
```
bradallred, line 69: I think it would be nice to move this to the `ResRef` class

gemrb/core/GameData.cpp
```diff
@@ -239,55 +243,31 @@ bool GameData::DelTable(unsigned int index)
...
-		return NULL;
-	}
+	auto iter = PaletteCache.find(resname);
+	if (iter != PaletteCache.end())
+		return iter->second;
+
 	ResourceHolder<ImageMgr> im = GetResourceHolder<ImageMgr>(resname);
 	if (im == nullptr) {
-		PaletteCache.SetAt(resname, NULL);
+		PaletteCache[resname] = NULL;
```
bradallred, line 254:
```suggestion
		PaletteCache[resname] = nullptr;
```

gemrb/core/GameData.cpp
```diff
@@ -239,55 +243,31 @@ bool GameData::DelTable(unsigned int index)
...
-		error("Core", "Corrupted Palette cache encountered (reference count went below zero), Palette name is: %.8s\n", name);
-	}
-	if (!res) {
-		pal->release();
-	}
+	// This was previously much hairier, trying to keep track of two different
+	// palette refcounts.  Now we just rely on Holder/Held to make sure memory
+	// is freed, while not bothering about freeing named palettes from the
+	// map.
 	pal = NULL;
```
bradallred, line 271: are "named" `Palettes` even a thing anymore? this function is just clearing a `PaletteHolder` and is no different from just assigning `nullptr` to the holder. I would prefer to just do that and nix this function.

gemrb/core/GameData.h
```diff
@@ -58,6 +60,11 @@ struct Table {
 	unsigned int refcount;
 };
 
+struct ieResRefHash
```
bradallred, line 63: let's move this to `ResRef`

gemrb/core/Projectile.cpp
```diff
@@ -1913,7 +1913,7 @@ void Projectile::Draw(Sprite2D* spr, const Point& p,
 		  unsigned int flags, Color tint) const
 {
 	Video *video = core->GetVideoDriver();
-	Palette* pal = (spr->Bpp == 8) ? palette : NULL;
+	PaletteHolder pal = (spr->Bpp == 8) ? palette : NULL;
```
bradallred, line 1916:
```suggestion
	PaletteHolder pal = (spr->Bpp == 8) ? palette : nullptr;
```

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -618,7 +618,7 @@ void Actor::SetName(int strref, unsigned char type)
 void Actor::SetAnimationID(unsigned int AnimID)
 {
 	//if the palette is locked, then it will be transferred to the new animation
-	Palette *recover = NULL;
+	PaletteHolder recover = NULL;
```
bradallred, line 621:
```suggestion
	PaletteHolder recover = nullptr;
```

gemrb/plugins/TTFImporter/TTFFontManager.h
```diff
@@ -46,7 +46,7 @@ Public methods
 	void Close();
 
 	Font* GetFont(unsigned short pxSize,
-				  FontStyle style, Palette* pal = NULL);
+				  FontStyle style, PaletteHolder pal = NULL) override;
```
bradallred, line 49:
```suggestion
				  FontStyle style, PaletteHolder pal = nullptr) override;
```

bradallred (changes requested):
mostly just changing `NULL` -> `nullptr`

you don't need to worry about the `SDL20GLVideo` files. those are now unmaintained garbage. At this point we should probably delete them.

gemrb/core/GUI/WorldMapControl.cpp
bernds: That's a really good spot, actually.

gemrb/core/GameData.cpp
bernds: That's not something I can answer. To me it looks like palettes can be looked up by name, making them named, and then also copied into unnamed ones, or just created unnamed. I suspect the Cache here was trying to do two things, keep a cache of named palettes, and also provide refcounting for other palettes. But it was all quite opaque.

I'll commit to making another pass to remove this function in a followup patch (possibly after tackling sprites). Ok to leave as is for now?

bradallred: sure, thats fine.

gemrb/core/GameData.h
bernds: Done, and I also found a few more NULLs. Squashed and pushed the update.

bernds:
Ah, I might not have caught everything - Github was hiding some suggestions. Let me go through them again real quick.

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
bernds: I actually tried that first, and changed it when it errored out. The overload for operator!= of two Holders is missing, I'll add it to the patch.

gemrb/core/GameData.h
bradallred: thanks

bradallred:
thanks. I look forward to the `Sprite` treatment :)

bernds:
I noticed the clang build shows some issues, I'm working on fixing those. I thought I was making an improvement when I was adding a missing override, but I didn't realize that there was another method that needed it, or that clang would warn about the inconsistency.

bradallred:
my bad. go ahead and submit another PR.

Issue #539, opened 2019-12-18
[bug] PST AR0511 : Stuck in Endless circular conversation with guards in AR0511


chilvence:
[000000196-GRB Nihilistic Nitwit.zip](https://github.com/gemrb/gemrb/files/3980448/000000196-GRB.Nihilistic.Nitwit.zip)

Go through the door, the guard will start a conversation with you and pass you on to the man who has locked himself  in his room. After your conversation is finished, the guard will start a conversation with you and pass you on to the man who has locked himself in his room. After your conversation is finished, the guard will start a conversation with you...

This appears to be a problem in running 0511ING1.BCS

```
IF
	WasInDialog()
THEN
	RESPONSE #100
		GlobalSet("0511_Talked_Guards","AR0511",1)
END

IF
	See([PC])
	Global("G_Test3","GLOBAL",1)
	Global("0511_Talked_Guards","AR0511",0)
THEN
	RESPONSE #100
		Dialogue([PC])
END
```

Checking a savegame made afterwards, global **0511_Talked_Guards** never becomes set , so the WasInDialog() action could be faulty, or the fact that the dialog is passed to a second actor could be confusing the situation?

lynxlynxlynx:
Does it work if you change the order of the two blocks? WasInDialog is true for only a short while.

chilvence:
It didn't seem to make any difference. Tried moving it various places.

lynxlynxlynx:
There are 240 uses in pst, so it's important to get this right. It seems mostly used to ensure proper placement (in combination with ReturnToSavedPlace).

chilvence:
This is probably the reason that Lothar does not return to his study after tasking you to find a skull, I would wager. I also think that Mantuok is supposed to physically block you from escaping before you come back with results. I will have to shadow my own steps again with the real game to be sure, but I think this one will be interesting :)

By the way, can you tell me how you are finding the usages of each script function? That was one of my first thoughts to look at, but NI doesn't seem to be helping me do that.

lynxlynxlynx:
I dumped all of them using weidu to plain text, so they're easy to grep through.
Unless they have area script overrides, those two don't have scripts that use WasInDialog.

anewusername:
I tried printing out the Scriptable in both `LeaveDialog` and `WasInDialog`; I suspect it's just not getting applied to the guard during the handoff:
```
[Unknown]: wasn't in dialog 0x621002274100
[Unknown]: wasn't in dialog 0x621002274100
[...]

PR #972, opened 2020-10-09
Call LeaveDialog() when changing conversation target

anewusername:
## Description
Fix for #539 (endless conversation loop in PST AR0511).

## Checklist


chilvence:
This appears to fix the very similar annoyance when returning to Curst and approaching the gate to Carceri 👍

lynxlynxlynx:
Thanks!

Issue #630, opened 2020-02-12
[bug] Autopause on enemy sighted triggers on every new enemy


burner1024:
**Describe the bug**
"Enemy sighted" autopause triggers every time a new enemy appears in sight. In original, it doesn't trigger if you already see another enemy.
Basically, with this option enabled it's a non-stop autopause fest.

GemRB version: 0.8.6-git

**To Reproduce**
1. Go to Government discrict.
2. Attack an Amn soldier in the top of the area.
3. See the game being paused again and again as backup arrives.

**Expected behavior**
The game gets paused once, when the party goes from "we see no enemies" to "we see some enemies".

**Screenshots**
![Captura de pantalla de 2020-02-13 03-40-21](https://user-images.githubusercontent.com/1279852/74376049-034e5880-4e14-11ea-8aaf-45c12defac42.png)

lynxlynxlynx:
I'm not sure I'd consider this a bug — in many other cases the pause would be welcome. Don't have any spectacular ideas on how to avoid your case though. A timer would work (minimum time has to pass before we consider retriggering), but maybe we can do better.

burner1024:
How do you mean? Why it would be welcome?

lynxlynxlynx:
New enemies later join an ongoing battle; reinforcements, domination, friendly fire, something else ...

burner1024:
I don't see the appeal, but I guess it wouldn't hurt as an additional option/setting.
However, surely it makes sense to implement the option the way it works in the original first, no?

lynxlynxlynx:
Not really, if we blindly copied every stupid thing the original did, we'd go crazy. If you're familiar with vanilla as a version concept, most engine reimplementation projects are actually stracciatellas.

burner1024:
I see, that makes sense.
I disagree that this one in particular is a stupid thing, though. Is it your final decision that there will no possibility to have it work vanilla-like?

lynxlynxlynx:
You shouldn't even notice the difference.

burner1024:
But that's definition of difference - if you can tell it, there is one, and if you can't, there isn't, for practical purposes. I mean, what does that mean making it different, if users can't notice it? (Meaning UX, different "under the hood" doesn't count - the whole GemRB is "different" in that case).

[...]

PR #974, opened 2020-10-09
Fix autopause on enemy sighted

bernds:
There were a number of problems here.  The first one is that directly
after creating a character, BG2 goes into pause mode, which looks like
a lockup to the user, since the GUI isn't shown at that point.  The
issue is that Actor::UpdateActorState doesn't check for visibility
and triggers autopause for the Cambion, which is somewhere else
entirely.  I believe this code isn't needed at all, and deleted it.

The next problem is the check in Map::InitActor.  This has an
isVisible check using actor->Pos, but the position isn't set at that
point, it's 0,0.  Most callers of AddActor appear to follow up with
SetPosition (and you can't swap the two).  I removed the
ActorSpottedByPlayer call in this function, and added a log message
to show an error in case the condition ever triggers, which really
ought not to be possible.

The only semi-reasonable autopause code was in GenerateQueues, and now
we generate all enemy-sighted pauses from there.  It was necessary to
separate the autopause code from ActorSpottedByPlayer, so that we can
also call it for already IF_ACTIVE actors.

The final issue was the use of AnyPCInCombat.  That seems to just be
a countdown, which was observable in Chateau Irenicus.  After
defeating the Ogre Mage spawned in by the genie, and then walking to
the first group of kobolds.  If you do it quickly, AnyPCInCombat
returns true, and no autopause is generated, which is incorrect.
Instead of this, we now keep track of whether hostile actors are
visible on the map, and suppress new autopause events when we already
saw some.  That is bug #630.

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -1677,13 +1677,21 @@ void Map::ActorSpottedByPlayer(Actor *actor)
 			core->GetGame()->SetBeastKnown(avatar->Bestiary);
 		}
 	}
+}
 
-	if (!(actor->GetInternalFlag()&IF_STOPATTACK) && !core->GetGame()->AnyPCInCombat()) {
-		if (actor->Modified[IE_EA] > EA_EVILCUTOFF && !(actor->GetInternalFlag() & IF_TRIGGER_AP)) {
-			actor->SetInternalFlag(IF_TRIGGER_AP, OP_OR);
+// Call this for any visible actor.  do_pause can be false if hostile
+// actors were already seen on the map.  We used to check AnyPCInCombat,
+// which is less reliable.  Returns true if this is a hostile enemy
+// that should trigger pause.
+bool Map::HandleAutopauseForVisible(Actor *actor, bool do_pause)
+{
+	if (actor->Modified[IE_EA] > EA_EVILCUTOFF && !(actor->GetInternalFlag() & IF_STOPATTACK)) {
+		if (do_pause && !(actor->GetInternalFlag() & IF_TRIGGER_AP))
 			core->Autopause(AP_ENEMY, actor);
-		}
+		actor->SetInternalFlag(IF_TRIGGER_AP, OP_OR);
+		return true;
 	}
+	return false;
 }
 
 //call this once, after area was loaded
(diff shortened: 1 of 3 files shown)


bradallred:
I think it makes more sense to issue this PR against master branch. It doesn't appear to be in conflict with changes made in subviews.

bernds:
I thought this was going to become the new master branch. How is this managed - do all commits on master get merged to subviews?
Also, I thought I saw differences in behaviour when I went from master to subviews (for example I don't recall seeing the pause on a new game), so there may already be differences in the code.

bradallred:
yes, I've been periodically merging master into this branch. 

let's see what @lynxlynxlynx says

bernds:
Also, there are in fact rather massive differenes in Actor.cpp between master and subviews. The UpdateActorState method I'm patch looks completely different, which I think explains why I saw different behaviour between the two branches. Given this, I really don't see value in adapting this to master (which I remember as behaving better overall anyway).

bradallred:
you're right. I'll just merge it here, but would you mind opening an issue (or searching for an existing one) and adding the subviews-recheck label to it?

bradallred:
maybe issue #630

bernds:
Ok, I was going to try tagging that one (if I have permissions - don't know that Github feature yet), but I saw you already did so.

gemrb/core/Map.cpp
```diff
@@ -1704,8 +1712,16 @@ void Map::InitActor(Actor *actor)
...
+	if (IsVisible(actor->Pos, false)) {
+		// This case is probably not possible.  At this point, the actor's position
+		// ought to be 0,0 and the caller of AddActor is expected to follow up with
+		// SetPosition to move the actor to its real position.
+		// The Schedule call here preserves previous behaviour, while also logging
+		// the event if something unexpected happens.  It's unclear whether an
+		// actor can be visible at 0,0 by accident and whether the Schedule call
+		// is meaningful for that case.
+		Log(ERROR, "Map", "Actor visible in InitActor before real position was set");
+		actor->Schedule(gametime, true);
```
lynxlynxlynx, line 1724: This is a noop, since you're not using the return value (Schedule is const). I question the need for the whole block.

lynxlynxlynx:
This is now too eager in avoiding the auto pause; the whole point of the report was to improve the behaviour, not just copy it.

gemrb/core/Map.cpp
bernds: Ah, ok. The name Schedule suggested that an action takes place. The whole block can really be deleted then.

bernds:
Please elaborate? The game was unplayable for me on the subviews branch because every goblin in a group would pause the game, apparently multiple times, and that is now fixed. How would you improve the behaviour?

lynxlynxlynx:
See the discussion in the bug, we can do better than the original. The easiest way would be to only trigger at most once per round of GameTime (mod it with core->Time.round_size).

bernds:
Well, the reporter of that bug very much wanted vanilla-like behaviour, and I agree with that. I also agree with his assessment that the timer thing is unlikely to work well: some characters pop in and out of invisibility all the time, and too short a timer would lead to issues like pausing once per goblin again.

I think it might be feasible to add additional autopause calls to cases where an invisible enemy becomes visible. But already in vanilla there are situation where you try to pause, and an autopause just came in before, so the game doesn't actually pause. I think there is a high risk of making that kind of behaviour worse.

lynxlynxlynx:
Like I said, invisibility is not a problem, since every actor can trigger the pause only once. That's the internal flag bit in the code you've been modifying. A round takes 5-7 seconds, depending on the game, so it's a good limiting time.

PR #979, opened 2020-10-10
Use a deadzone when going into rectangle select mode

bernds:
Currently, it can be difficult to initiate movement or combat, since a
slight movement of the mouse before releasing triggers rectangle select
mode.  I checked with the IE, and it also applies a deadzone so that
you can move the mouse a little between click and release.  This change
implements that.

The deadzone in IE is pretty large, around half of a reticle, so I bumped
ours as well.

One issue remains: IE shows feedback immediately on mouse press, rather
than waiting for mouse release, which makes for a better user experience.


lynxlynxlynx:
Thanks, LGTM.

PR #980, opened 2020-10-10
Clear trackingWin on mouse up

bernds:
Observed behaviour: Click on the rest button, wait for the movie to start,
then click on the same screen position again.  The game aborts with

Window.cpp:464: bool GemRB::Window::DispatchEvent(const GemRB::Event&): Assertion `target == NULL || target->IsVisible()' failed.

This seems to be caused by leaving trackingWin set after a MouseUp event,
an inconsistency with TouchUp.

Please note that the return value of this DispatchEvent is unused.  While
changing this code, I also removed a return value that looked meaningful
but wasn't, and changed it plain false.


bradallred:
thanks!

PR #981, opened 2020-10-10
RespawnNameless():  Reselect TNO after death

chilvence:
There is a small ease of life feature in the original game, when Nameless dies, he stays selected, since you come back to life moments later (well, from TNO's point of view, I guess)

I made this tweak a while ago, after noticing a convenient place here https://github.com/gemrb/gemrb/commit/b7cbeb87df04b55b25585c1c97fec5f0ae799661, but I was wondering if it could potentially cause any problems

It seems to work alright, however, there is a worrying comment in Scriptable.cpp at Selectable::Select https://github.com/gemrb/gemrb/blob/e14fe1ac692a999d6acb0203bf1b632152141a5e/gemrb/core/Scriptable/Scriptable.cpp#L1816-L1817

This is only really a fix for a minor annoyance, but I was hoping to get some feedback so I can scratch it off the list. Can the condition that the comment warns about actually occur? Or is there a better place to add this line or a better way to achieve the same result?

### DIFF
--- a/gemrb/core/IniSpawn.cpp
+++ b/gemrb/core/IniSpawn.cpp
@@ -636,6 +636,11 @@ void IniSpawn::RespawnNameless()
 	// resurrect leaves you at 1hp for raise dead, so manually bump it back to max
 	nameless->RefreshEffects(NULL);
 	nameless->SetBase(IE_HITPOINTS, 9999);
+
+	// reselect nameless, since he didn't really 'die'
+	// this matches the unconditional reselect behavior of the original
+	core->GetGame()->SelectActor(nameless, true, SELECT_NORMAL);
+
 	//hardcoded!!!
 	if (NamelessState==36) {
 		nameless->SetStance(IE_ANI_PST_START);


lynxlynxlynx:
You should just use Game::SelectActor.

chilvence:
Awesome, does this version look ok?

lynxlynxlynx:
No; wrong branch and I think SelectActor also obviates the need to manually invoke the change handler.

chilvence:
Ugh, No idea how I ended up with branches overlapping, but I think that is fixed now.

The reason I added the line to force a selection changed event was because the portrait window was not updating, it seemed to need a nudge. But changing to SELECT_NORMAL seems to work, I guess I assumed the purpose of that flag wrongly, I thought it was something to do with voice feedback.

So I think the line https://github.com/gemrb/gemrb/blob/7e23f0e36c3a6a879cd58b416a4084ffcda31b3b/gemrb/core/Game.cpp#L694 sets the pieces in motion to get the UI updated, is that right?

lynxlynxlynx:
Yes.

PR #982, opened 2020-10-10
FreePalette followup

bernds:
## Description
As promised in the other PR, here is a followup patch to clean up a bit more after the PaletteHolder changes. These changes are optional, they are not expected to have a user-visible impact.

FreePalette still remains, since the CharAnimations is hard to follow and I wanted to get the simple cases out of the way first. The second patch in this set contains an initial cleanup for CharAnimations to fix name shadowing.

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -3571,7 +3571,6 @@ AreaAnimation::AreaAnimation()
 {
 	animation=NULL;
 	animcount=0;
-	palette=NULL;
 	appearance = sequence = frame = transparency = height = 0;
 	Flags = originalFlags = startFrameRange = skipcycle = startchance = 0;
 	unknown48 = 0;
@@ -3616,7 +3615,6 @@ AreaAnimation::~AreaAnimation()
 		}
 	}
 	free(animation);
-	gamedata->FreePalette(palette, PaletteRef);
 }
 
 Animation *AreaAnimation::GetAnimationPiece(AnimationFactory *af, int animCycle)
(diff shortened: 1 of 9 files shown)


gemrb/core/Map.cpp
```diff
@@ -3679,9 +3677,7 @@ void AreaAnimation::InitAnimation()
 void AreaAnimation::SetPalette(ieResRef Pal)
 {
 	Flags |= A_ANI_PALETTE;
-	gamedata->FreePalette(palette, PaletteRef);
-	strnlwrcpy(PaletteRef, Pal, 8);
-	palette = gamedata->GetPalette(PaletteRef);
```
lynxlynxlynx, line 3684: PaletteRef is a member variable, so this would make it go out of sync.

Best to also keep strnlwrcpy in, since we'll have to remove it from ResRef if we'll want to use it more often. Changing case of stored strings have savegame compatibility implications, so it's better to be safe.

bernds: This is fixed in the latest push.

lynxlynxlynx:
Thanks, looks fine now. Have you verified with valgrind, ASAN or somesuch that there are no new leaks?

bernds:
It's a little hard to tell, since asan reports are still quite noisy. I'm looking into reducing that if possible, but some of it comes from Python apparently. But the amount leaked is comparable and there's nothing that jumps out at me as a new problem.

lynxlynxlynx:
Ok, thanks. We also inherit some small ones from other libraries, but the whole thing has been in the kilo range last time I checked, so nothing to worry about.

I leave this to @bradallred to merge, in case he has some other comments.

bradallred:
Excellent, thanks again!

PR #983, opened 2020-10-10
Subviews MSVC build issues

chilvence:
I was just checking to see if I could still build Subviews with visual studio and found some new issues.

1: Apparently, using #define _USE_MATH_DEFINES before cmath is useless under MSVC, people report they can only get it to work if it is set as a command line directive, which is why there have to be M_PI defines in win32def.h

I worked around it by adding it to the core CMakeLists.txt, since it seems that is where the defines are needed, but I don't know if it will cause any issues, I don't really know why it is optional

2: The WEDimporter refused to build because back_inserter was not declared, I found you have to include iterator

3: The window.rc moved, so I cherry picked the commit from master that updates the location of the icon

This gets me as far as  40ce03e , but after that I am lost because I don't understand what is actually happening to be able to investigate the errors I am getting from b8224c3

## Checklist

### DIFF
--- a/gemrb/core/GameData.h
+++ b/gemrb/core/GameData.h
@@ -136,7 +136,7 @@ private:
 	Cache ItemCache;
 	Cache SpellCache;
 	Cache EffectCache;
-	std::unordered_map<const ResRef, PaletteHolder, ResRef::Hash> PaletteCache;
+	std::unordered_map<ResRef, PaletteHolder, ResRef::Hash> PaletteCache;
 	Factory* factory;
 	std::vector<Table> tables;
 	typedef std::map<const char*, Store*, iless> StoreMap;
--- a/gemrb/includes/win32def.h
+++ b/gemrb/includes/win32def.h
@@ -61,6 +61,9 @@
 #ifndef M_PI_2
 #define M_PI_2  1.57079632679489661923 // pi/2
 #endif
+#ifndef M_PI_4
+#define 	M_PI_4   0.78539816339744830962
+#endif
 
  /* WinAPI collision. GetObject from wingdi.h conflicts with a type used in the GUIScript source (core\ScriptEngine.h) */
 #ifdef GetObject
--- a/gemrb/plugins/WEDImporter/WEDImporter.cpp
+++ b/gemrb/plugins/WEDImporter/WEDImporter.cpp
@@ -28,6 +28,7 @@
 #include "TileSetMgr.h"
 
 #include <cmath>
+#include <iterator>
 
 #if HAVE_UNISTD_H
 #include <unistd.h>


lynxlynxlynx:
1: This sounds the same. Did you have an actual problem somewhere?
The official docs have the same usage:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=vs-2019

2: That's fine.

3: This would have gotten in with the next merge, but it also doesn't hurt.

4: If you don't share the errors, nobody can help you ...

chilvence:
1: The first problem was there was a new usage of M_PI_4, which wasn't accounted for in win32def.h, at first I just added it from a reference similar to your link, but then I felt like finding out what was actually happening - so, if adding _USE_MATH_DEFINES to cmakelists is undesirable, I can always just define M_PI_4 the same way in win32def

2: Yay 👍 

3: My main motivation is rebasing the PST fixes I am working on to the latest iteration of subviews, so I had to throw that in

4: I did make a comment on the actual commit, but I'll copy the error here: 

```Error	C2280	'std::pair<const GemRB::ResRef,GemRB::PaletteHolder> &std::pair<const GemRB::ResRef,GemRB::PaletteHolder>::operator =(volatile const std::pair<const GemRB::ResRef,GemRB::PaletteHolder> &)': attempting to reference a deleted function	B:\msvc\src\chiv\out\build\subviews-ninja\chiv	B:\VS\VC\Tools\MSVC\14.27.29110\include\list	1284```

That's starting from b8224c3 , it's all a bit over my head unfortunately

**Comment by lynxlynxlynx** [comment]
1. Safer to add the same redefinition in the header, since this is not standardized anywhere sensible (only Unix98 afaik). 
4. Anything before that? This error alone doesn't show what triggered it and we use that particular pair a lot now.

**Comment by chilvence** [comment]
It's probably better if I paste my whole build log for subviews, since I can't figure out what to look for

https://pastebin.com/r45p5rGu

**Comment by MarcelHB** [comment]
I found some of the criteria in "Deleted implicitly-declared copy assignment operator" [here](http://naipc.uchicago.edu/2014/ref/cppreference/en/cpp/language/as_operator.html), looks like something older of [that](https://en.cppreference.com/w/cpp/language/copy_assignment), maybe that's what MSVC is about:

> T has a non-static data member that cannot be copy-assigned (has deleted, inaccessible, or ambiguous copy assignment operator) 

That sounds like `char ref[9]` of `ResRef` to be such a candidate, and thus MSVC protests for that to be used in an `std::unordered_map`. So, without testing, I guess we need to restore that assignment operator on our own, like (to be put into `struct ResRef` in `Resource.h`):

```cpp
void operator=(const ResRef& other) {
   operator=(other.ref);
}
```

chilvence:
I think I understand, I mean what you said makes sense, but in practice I am not able to get it to work. 

Does it make any difference that it already has this line?

https://github.com/gemrb/gemrb/blob/f0fa8d40230580fac734b1f605554e32287bfb16/gemrb/core/Resource.h#L85

@bernds do you have any suggestions?

MarcelHB:
That is just one overload of the assignment operator, but not what matches the signature of an implicit copy assignment operator. It is what you need to override (or forbid) to do something like this:

```cpp
ResRef a, b;
b = a; // <-- that
b = "SPL123"; // <-- the overload for C strings
```

bernds:
This perhaps? Does this do anything?

```diff
diff --git a/gemrb/core/Resource.h b/gemrb/core/Resource.h
index f277f9c13..70bf94f3b 100644
--- a/gemrb/core/Resource.h
+++ b/gemrb/core/Resource.h
@@ -82,7 +82,8 @@ public:
 		return (ref[0] == '\0') ? NULL : ref;
 	}
 
-	void operator=(const char* str) {
+	ResRef &operator=(const ResRef &other) = default;
+	ResRef &operator=(const char* str) {
 		if (str == NULL) {
 			Clear();
 		} else {
@@ -92,6 +93,7 @@ public:
 			strnlwrcpy(ref, str, sizeof(ref)-1 );
 			ref[sizeof(ref)-1] = '\0';
 		}
+		return *this;
 	}
 
 	// Case insensitive
```

bernds:
Perhaps also this?
```diff
diff --git a/gemrb/core/Resource.h b/gemrb/core/Resource.h
index f277f9c13..8cfb2ff81 100644
--- a/gemrb/core/Resource.h
+++ b/gemrb/core/Resource.h
@@ -61,7 +61,7 @@ public:
 	ResRef() {
 		Clear();
 	}
-
+	ResRef(const ResRef &other) = default;
 	ResRef(const char* str) {
 		operator=(str);
 	};
```

chilvence:
Sadly not, sorry, it still gives me the c2280 error with both those patches applied... https://pastebin.com/YG6ehKwP

I wonder why other compilers don't make a fuss, I had no trouble at all with literally everything else I tried.

bernds:
Okay, let's try to also add move constructor/assignment, add noexcept, and tighten up Holder as well. If this doesn't work I'm probably out of ideas.

```diff
diff --git a/gemrb/core/Holder.h b/gemrb/core/Holder.h
index 9cd2491ca..5330bbee8 100644
--- a/gemrb/core/Holder.h
+++ b/gemrb/core/Holder.h
@@ -58,7 +58,7 @@ private:
 template <class T>
 class Holder {
 public:
-	Holder(T* ptr = NULL)
+	Holder(T* ptr = NULL) noexcept
 		: ptr(ptr)
 	{
 		if (ptr)
@@ -69,16 +69,32 @@ public:
 		if (ptr)
 			ptr->release();
 	}
-	Holder(const Holder& rhs)
+	Holder(const Holder& rhs) noexcept
 		: ptr(rhs.ptr)
 	{
 		if (ptr)
 			ptr->acquire();
 	}
-	
-	Holder& operator=(Holder rhs)
+	Holder(Holder &&rhs) noexcept
+		: ptr (rhs.ptr)
 	{
-		std::swap(rhs.ptr, ptr);
+		rhs.ptr = nullptr;
+	}
+
+	Holder& operator=(const Holder &rhs) noexcept
+	{
+		T *old = ptr;
+		ptr = rhs.ptr;
+		if (ptr)
+			ptr->acquire();
+		if (old)
+			old->release();
+		return *this;
+	}
+
+	Holder& operator=(Holder &&rhs) noexcept
+	{
+		std::swap (rhs.ptr, ptr);
 		return *this;
 	}
 
diff --git a/gemrb/core/Resource.h b/gemrb/core/Resource.h
index f277f9c13..d0f033a4f 100644
--- a/gemrb/core/Resource.h
+++ b/gemrb/core/Resource.h
@@ -61,7 +61,8 @@ public:
 	ResRef() {
 		Clear();
 	}
-
+	ResRef(const ResRef &other) = default;
+	ResRef(ResRef &&other) = default;
 	ResRef(const char* str) {
 		operator=(str);
 	};
@@ -82,7 +83,9 @@ public:
 		return (ref[0] == '\0') ? NULL : ref;
 	}
 
-	void operator=(const char* str) {
+	ResRef &operator=(const ResRef &other) = default;
+	ResRef &operator=(ResRef &&other) = default;
+	ResRef &operator=(const char* str) {
 		if (str == NULL) {
 			Clear();
 		} else {
@@ -92,6 +95,7 @@ public:
 			strnlwrcpy(ref, str, sizeof(ref)-1 );
 			ref[sizeof(ref)-1] = '\0';
 		}
+		return *this;
 	}
 
 	// Case insensitive
```

chilvence:
Unfortunately still no luck. I will keep trying though, at least now I have some insight into what the problem is and some stuff to read up on. There is bound to be a solution.

bernds:
The last thing you could try is to add noexcept to the ResRef default constructor, which I overlooked in the last patch.

chilvence:
I wasn't sure which one you meant to change, but I tried them changing them all one by one and nothing seemed to be different.

I'm interested in this line of the build output:
```
 B:\VS\VC\Tools\MSVC\14.27.29110\include\utility(273): note: see declaration of 'std::pair<const GemRB::ResRef,GemRB::PaletteHolder>::operator ='
```

It doesn't seem to be talking about the ResRef struct by itself , also the PaletteHolder

line 273 in include/utility has this: 
```    pair& operator=(const volatile pair&) = delete; ```

Could that be throwing a spanner in the works?

**Comment by MarcelHB** [comment]
The docs [say ](https://en.cppreference.com/w/cpp/utility/pair/operator%3D) that this `operator=` of a pair is to be deleted when one of its types has no such operator.

**Comment by bradallred** [comment]
I haven't (yet) read this thread, but both `ResRef` and `PaletteHolder` _ought_ to be assignable so if they aren't we should fix that.

**Comment by bradallred** [comment]
I've added copy assignment for `ResRef`. try now?

**Comment by chilvence** [comment]
Sadly, it still doesn't want to work:

```
B:\VS\VC\Tools\MSVC\14.27.29110\include\list(1284): error C2280: 'std::pair<const GemRB::ResRef,GemRB::PaletteHolder> &std::pair<const GemRB::ResRef,GemRB::PaletteHolder>::operator =(volatile const std::pair<const GemRB::ResRef,GemRB::PaletteHolder> &)': attempting to reference a deleted function
```

I wish I could wrap my head around it, I can't even think of things to try. Thoroughly vexed.

**Comment by bernds** [comment]
Hold on, const ResRef as first argument - that's probably why you can't assign to it.
Remove const from the unordered_map, line 139 in GameData.h, and try again.

**Comment by chilvence** [comment]
Aha! You hit the nail on the head, that was what was stopping it. Thanks!

Although that's good news, I still have another error left:

```
Error	C2027	use of undefined type 'GemRB::Sprite2D'	B:\msvc\src\chiv\out\build\subviews-ninja\chiv	B:\msvc\src\chiv\gemrb\core\Holder.h	76	
```

```
B:\msvc\src\chiv\gemrb\core\Holder.h(70): error C2027: use of undefined type 'GemRB::Sprite2D'
  B:\msvc\src\chiv\gemrb\core\Animation.h(35): note: see declaration of 'GemRB::Sprite2D'
  B:\msvc\src\chiv\gemrb\core\Holder.h(68): note: while compiling class template member function 'GemRB::Holder<GemRB::Sprite2D>::~Holder(void)'
  ..\..\..\gemrb\core\AnimationMgr.cpp(34): note: see reference to function template instantiation 'GemRB::Holder<GemRB::Sprite2D>::~Holder(void)' being compiled
  B:\VS\VC\Tools\MSVC\14.27.29110\include\vector(1484): note: see reference to class template instantiation 'GemRB::Holder<GemRB::Sprite2D>' being compiled
  B:\VS\VC\Tools\MSVC\14.27.29110\include\vector(1482): note: while compiling class template member function 'unsigned __int64 std::vector<GemRB::Holder<GemRB::Sprite2D>,std::allocator<GemRB::Holder<GemRB::Sprite2D>>>::size(void) noexcept const'
  B:\msvc\src\chiv\gemrb\core\AnimationFactory.h(52): note: see reference to function template instantiation 'unsigned __int64 std::vector<GemRB::Holder<GemRB::Sprite2D>,std::allocator<GemRB::Holder<GemRB::Sprite2D>>>::size(void) noexcept const' being compiled
  B:\msvc\src\chiv\gemrb\core\Animation.h(41): note: see reference to class template instantiation 'std::vector<GemRB::Holder<GemRB::Sprite2D>,std::allocator<GemRB::Holder<GemRB::Sprite2D>>>' being compiled
  [4/216] Building CXX object gemrb\core\CMakeFiles\gemrb_core.dir\ArchiveImporter.cpp.obj
```

I think the Sprite2D needs to be defined before referenced by Holder? I think this one might be simpler, but it's now 3:30 am...

**Comment by bradallred** [comment]
try it now. I don't know why it requires the definition. A forward declaration should have been sufficient.

**Comment by chilvence** [comment]
Thanks, it finally builds with no errors!

It is frustrating when it seems to complain over nothing, I think part of the problem is that the latest version of Visual Studio throws out their MSBuild system and replaces it with Ninja. It was the ultimate cause of many of the previous errors I had just trying to get it to build GemRB, if I just forced it to use the older version of the tools, it would stop complaining. Sadly, they always seem to try to make that hard work as well. I originally had some instructions written up on how to do so for the build guide, but it was just ugly... and in the end, once it was possible to nail down what it was complaining about, it usually turned out to be just as tiny a change needed as the ones here.

File: gemrb/core/GameData.h [hunk 1]
```diff
@@ -136,7 +136,7 @@ class GEM_EXPORT GameData : public ResourceManager
 	Cache ItemCache;
 	Cache SpellCache;
 	Cache EffectCache;
-	std::unordered_map<const ResRef, PaletteHolder, ResRef::Hash> PaletteCache;
+	std::unordered_map< ResRef, PaletteHolder, ResRef::Hash> PaletteCache;
```
lynxlynxlynx, line 139:
```suggestion
	std::unordered_map<ResRef, PaletteHolder, ResRef::Hash> PaletteCache;
```

chilvence: Sorry, sloppy habit, I fixed and updated

lynxlynxlynx:
Thanks!

chilvence:
Thanks again everyone for the help, I'd be pretty lost without it!

PR #985, opened 2020-10-12
Correct proficiencies scrollbar

bernds:
This one seems like an odd issue to remain for so long, but it really looks to me like the +1 is just wrong. Maybe someone wants to double check they are seeing the same problem before merging? Below is the commit message.

When creating a character in BG2, the weapon proficiencies can be scrolled
down beyond the end of the list, with an empty entry at the bottom.

Adding a few Log calls, the first entries of the table are all the
actual proficiencies, ending in Two Weapon Style. Then follow a number of
entries that are recognized as invalid, and ProfCount is reduced for
these.  Finally, there is one entry that is not recognized as invalid,
but calling GetString on it only produces an empty string - and when
scrolling past the end, there is indeed an empty row.

The RowCount calculation has a "+ 1" that just looks incorrect, and
removing it fixes BG2.  BG1 is unaffected either way, it does not
show a scrollbar (and in fact calls into this code with scroll=False).

### DIFF
--- a/gemrb/GUIScripts/LUProfsSelection.py
+++ b/gemrb/GUIScripts/LUProfsSelection.py
@@ -255,7 +255,7 @@ def SetupProfsWindow (pc, type, window, callback, level1=[0,0,0], level2=[1,1,1]
 			ProfsColumn = ProfsTable.GetColumnIndex (ClassName)
 
 	#setup some basic counts
-	RowCount = ProfsTable.GetRowCount () - ProfsTableOffset + 1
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Easy to test everywhere, let's see.

lynxlynxlynx:
Line 266 should also be changed to use `<= 0`. This whole thing surfaced because the default table value is not -1 in either form.

Last touched in 2009. ;)

bernds:
I made the change I think you requested.

lynxlynxlynx:
Thanks!

Issue #804, opened 2020-06-12
[subviews] Breakages in PST/GUIINV.py


chilvence:
There's lots of broken functionality in the PST inventory window in the subviews branch which works fine in the other games. Currently it is impossible to drag and drop anything, there may be other glitches too.

I did a bisect and traced the point where it broke to [this commit](https://github.com/gemrb/gemrb/commit/1b324807e4edb538a09eecb81453dcac137addaa) but it didn't really make sense why that would cause a problem so instead I started updating the python scripts based on the already working examples.

https://github.com/gemrb/gemrb/compare/subviews...chilvence:subviews-pst-GUIINV

These changes restore basic function, but it's not enough because for some bizarre reason if you try and drag the top left eyeball/earring slot it picks up your default weapon instead. Everything else seems fine though.

So, I'll open this bug report for now until I can figure that part out, at least then its a known issue.

chilvence:
Ach, never mind, in the initialiser I should have used >= instead of >, because the control ID's start at 0 instead of 1 like in BG.... I thought it was something completely unrelated.

I'll do some more in game testing and submit a PR

lynxlynxlynx:
You can also just use `if Button:`, since on subviews a missing control for GetControl does not raise an exception any more.

chilvence:
That works as well.

I also need to fix the mouse highlight handlers, do we want to use the cool fancy item highlighting behavior from the other games or just keep it as is?

lynxlynxlynx:
What  do you mean? Just maintain parity with master for now.

chilvence:
^ There's some new color pulse on mouse hover that looks pretty cool, but my priority is just straightening it out, so I'll do that.

I wouldn't expect any of the bugs I've found to actually affect many people, since its subviews branch, but I hope fixing pre-emptively will help

I do feel like I have pulled on the proverbial loose thread of the sweater though. I've got dragging, mouse highlighting, item splitting, item using and talking back to working properly, but I will give it another few days to see if anything else springs up.

bradallred:
the hover effect wasn't intentional in the inventory, it was a change so that hovering an actor on `GameControl` would pulse the portrait button for the actor (like the original does).

It just didn't bother me for it to be on the inventory buttons (in fact I found it useful) so I never bothered creating a way to disable it. If we are going to keep it I would like to set the borders for the non-magic items as well so that it is a consistent behavior for all items.

chilvence:
Ah right, so serendipitous :)

[...]

PR #986, opened 2020-10-13
Subviews PST inventory update/fixes

chilvence:
This branch is a collection of fixes for the inventory management in PST under Subviews branch, relating to #804 (sorry it took so long to get back to this!)

Currently drag and drop, item usage and info panel cycling is broken, also the stack split window seems to have never been functional.

I have tried to reorganise it so that everything works more like the other games. I would have liked to fully use InventoryCommon, but there were some instances where I had to stick with duplicated functions because there are still minor differences.  This is basically just triage, it was possible to remove a lot of special cases here, but the fact that characters can have different slots for different things makes it difficult to unify. If I use too much of InventoryCommon directly, the special slot icons start getting messed up and unpredictable things happen. If I start changing too much of it, then I also stray into the uncomfortable area of "fixing things that ain't broke" :)

I am just making this pull request because I finally managed to get the inventory to basically work again so I can play the game, but I'd welcome any feedback on how to improve it more.

## Checklist


gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -385,14 +385,20 @@ def OnAutoEquip ():
...
+	if ItemAmountWindow != None:
+		return
+
+	pc = GemRB.GameGetSelectedPCSingle ()
+	slot_item = GemRB.GetSlotItem (pc, slot)
 
 	if not GemRB.IsDraggingItem ():
-		ResRef = slot_item['ItemResRef']
-		item = GemRB.GetItem (ResRef)
+		if not slot_item:
```
lynxlynxlynx, line 399: When does this happen?

gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -467,36 +544,24 @@ def OpenItemAmountWindow ():
 	# 6 text
 
 	Window.ShowModal (MODAL_SHADOW_GRAY)
-
-def MouseEnterSlot ():
+	Window.SetFlags (WF_BORDERLESS, OP_OR)
+	return
+def MouseEnterSlot (btn,slot):
```
lynxlynxlynx, line 549: Either remove the return or add another newline.

gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -467,36 +544,24 @@ def OpenItemAmountWindow ():
...
-	if slot not in ItemHash:
-		return
-	slot = ItemHash[slot][0]-1
 	if slot == OverSlot or not GemRB.IsDraggingItem ():
 		OverSlot = None
-
-	for i in range(len(slot_list)):
-		if slot_list[i]==slot:
-			UpdateSlot (pc, i)
+	UpdateSlot (pc, slot-1)
```
lynxlynxlynx, line 564: Why the -1? Looks unnatural.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -237,7 +237,14 @@ def DisplayItem (slotItem, type):
 	global ItemInfoWindow
 
 	item = GemRB.GetItem (slotItem["ItemResRef"])
-	ItemInfoWindow = Window = GemRB.LoadWindow (5)
+	
+	#window can be refreshed by cycling to next/prev item, so it may still exist
+	if ItemInfoWindow:
+		print "DisplayItem: Reusing existing window: ", ItemInfoWindow
```
lynxlynxlynx, line 243: Basically a debug print, not really needed.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -381,32 +395,48 @@ def DisplayItem (slotItem, type):
...
 		Button = Window.GetControl (14)
-		Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, lambda: DisplayNext(1))
+		Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, lambda: CycleDisplayItem(1))
 
 	ItemInfoWindow.ShowModal(MODAL_SHADOW_GRAY)
 	return
 
+def CycleDisplayItem(dir):
+
+	ib = int(GemRB.GetVar('ItemButton'))
```
lynxlynxlynx, line 412: ib should really be named slot.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -363,32 +395,48 @@ def DisplayItem (slotItem, type):
...
+
+	#print "nextid: ", ib
+
+	pc = GemRB.GameGetSelectedPCSingle ()
+
+	slot_item = None
+	
+	#try the next slot for an item. if the slot is empty, loop until one is found.
+	while not slot_item:
+		ib+=dir
```
bradallred, line 423:
```suggestion
		ib += dir
```

bradallred (changes requested):

gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -34,8 +34,6 @@
 ItemAmountWindow = None
 OverSlot = None
```
bradallred, line 35: can we please get rid of this global and just pass it directly to `UpdateSlot`?

or probably better yet move
```
if GemRB.CanUseItemType (SlotType["Type"], itemname):
			Button.SetState (IE_GUI_BUTTON_SELECTED)
		else:
			Button.SetState (IE_GUI_BUTTON_ENABLED)
```

from `UpdateSlot` directly to  `MouseEnterSlot`

AFICT, every other interaction with `OverSlot` becomes useless with that change and can just be deleted

bradallred (changes requested):

gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -406,36 +412,97 @@ def OnDragItem ():
...
+def OpenItemAmountWindow (btn, slot):
+	"""Open the split window."""
+
+	global UsedSlot, OverSlot
+	global ItemAmountWindow, StackAmount
+
+	pc = GemRB.GameGetSelectedPCSingle ()
 
 	if ItemAmountWindow != None:
 		if ItemAmountWindow:
```
bradallred, line 473: this is completely redundant with the check above

gemrb/GUIScripts/pst/GUIINV.py
bradallred: Actually, I can see this is done in `InventoryCommon`... I'll fix it there and you can probably just use the `InventoryCommon` code.

gemrb/GUIScripts/pst/GUIINV.py
```diff
@@ -406,36 +412,97 @@ def OnDragItem ():
...
+	"""Increases the split size."""
+
+	Text = ItemAmountWindow.GetControl (6)
+	Amount = Text.QueryText ()
+	number = int ("0"+Amount)+1
+	if number>StackAmount:
+		number=StackAmount
+	Text.SetText (str (number))
+	return
+
```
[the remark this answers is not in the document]
chilvence: Its the ItemAmountWindow.GetControl that causes an error, if I use InventoryCommon.IncreaseStackAmount, it looks for the ItemAmountWindow there, whereas it is actually loaded in pst/GUIINV.py line 503

Although looking at it again, it is probably an easy canditate for consolidation, I think I might be able to actually end up with less rather than more here, I will try again tonight.

gemrb/GUIScripts/pst/GUIINV.py
chilvence: IIRC, I placed this here to silence a python error when you try and 'drag' from an empty slot, but its possible that the root cause that no longer exists, so I will test again without it.

gemrb/GUIScripts/pst/GUIINV.py
lynxlynxlynx: ok, let's leave it for later.

gemrb/GUIScripts/pst/GUIINV.py
chilvence: This function I barely touched, just in case it looked like it might be possible to merge back - which I think now it is, I was just stumped on something else so I didn't see it at first.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -381,32 +395,48 @@ def DisplayItem (slotItem, type):
...
+		if ib > 53:
+			ib = 0
+		elif ib < 0:
+			ib = 53
+
+		slot_item = GemRB.GetSlotItem (pc, ib)
+		GemRB.SetVar('ItemButton', ib)
+
+		if ib == thisitem:
+			break
```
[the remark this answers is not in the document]
chilvence: I wasn't sure about that either, I was thinking about the possibility of having only 1 inventory item, looking at it, and then a script takes it away or something like that, either button could trigger a runaway loop. I realise it is statistically unlikely though...

gemrb/GUIScripts/InventoryCommon.py
chilvence: I'll squash this one in when I update it

lynxlynxlynx:
When you update, please rebase on latest subviews, so the style checker will get updated.

gemrb/GUIScripts/pst/GUIINV.py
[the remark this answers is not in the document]
bradallred: you can easily get rid of `ItemAmountWindow` altogether by using `AddAlias` and `GetView` instead of a global variable.

gemrb/GUIScripts/pst/GUIINV.py
bradallred: I've changed `InventoryCommon`. This function is no longer used to close the window.

chilvence: No worries, I've always found that to be a bit counter-intuitive, it's odd how it can get confusing even when you know what it is doing

It turns out I don't need to clone any of this, it all folded in nicely with InventoryCommon with a slight tweak

chilvence:
Updated, but I think I can still do more here

I will try to also eliminate the duplicated mouseover action handlers, but I will need to make InventoryCommon.UpdateSlot play nice with the per-character inventory slot assignment.

gemrb/GUIScripts/pst/GUIINV.py
lynxlynxlynx: Deserves a comment then.

gemrb/GUIScripts/pst/GUIINV.py
bradallred: Can you please pull in the updates from `InventoryCommon`. This global is undesireable.

chilvence: Did the latest push catch it? I think I had missed a step in the rebase attempt

bradallred: well I didn't change the PST version. I was hoping you would just use the `MouseEnter` and `MouseLeave` handlers from the common version. Anything that still has `OverSlot` needs to be killed.

chilvence: Ah of course, apologies, it's a bit late here, I misunderstood. 

That is what I will try to do next, there is just a bit more I need to straighten out before I can do that, for now please consider this as still WIP

bradallred:
eh, I'd just hack it for now by doing `InventoryCommon.UpdateSlot = UpdateSlot` at the top of the script...

chilvence:
> eh, I'd just hack it for now by doing `InventoryCommon.UpdateSlot = UpdateSlot` at the bottom of the script...

I wanted that to work, but I think python is too clever, it might be because it's still in the InventoryCommon scope, certain other needed variables aren't found.

I'm still working on a full merge, but it's awkward to get everything just right without making a mess... for now I've patched it, just in case that plan goes awry

gemrb/GUIScripts/InventoryCommon.py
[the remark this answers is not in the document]
lynxlynxlynx: Things can't happen in parallel like that, so it's safe. But you can also test it, since it's a very simple exercise.

chilvence:
I used your suggestion to create a compromise - the logic to update the inventory slots is now handled by InventoryCommon, however UpdateInventorySlot is overridden when playing PST, so that the stylistic difference of the inventory icons is preserved. I had to add a game check to updateslot to deal with the custom per-character slot indexes, but it shouldn't affect other games. 

pst/GUIINV.py now has much fewer duplicated functions, and the few that are still left can probably also be pruned.

This is still quite rough though, and I really need to do some solid testing with the other games to make sure I didn't break anything

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -549,7 +549,7 @@ def UpdateSlot (pc, slot):
 	SlotType = GemRB.GetSlotType (slot+1, pc)
 	ControlID = SlotType["ID"]
 
-	if not ControlID:
+	if ControlID is None or ControlID is -1:
```
lynxlynxlynx, line 577: ID is a number and `None != 0`, so I don't know what you're trying to achieve here. It also appears to be unsigned, so checking for -1 is odd as well.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -546,7 +549,29 @@ def UpdateSlot (pc, slot):
 
 	Window = GemRB.GetView("WIN_INV")
 
-	SlotType = GemRB.GetSlotType (slot+1, pc)
+	if not GameCheck.IsPST:
```
lynxlynxlynx, line 552:
```suggestion
	if not GameCheck.IsPST ():
```

lynxlynxlynx: Also, does python not care about scope here (slot_item, SlotType, using_fists)?

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -620,6 +620,14 @@ def UpdateSlot (pc, slot):
 			Button.SetBAM (SlotType["ResRef"],0,0)
 			Button.SetTooltip (SlotType["Tip"])
 
+		if GameCheck.IsPST():
+			#SlotMap won't be defined if pst/GUIINV is not loaded
```
lynxlynxlynx, line 624: You can just check if SlotMap, it's not true that it won't be defined — you added it to the top of the file.

chilvence: Oh yeah, of course - although I still have to have the indented checks I think, or I'll get an index out of range error from python from trying to read SlotMap[slot]

lynxlynxlynx:
The first six commits are all fine and dandy, so if you want, you can open a separate PR for that, so we don't keep dragging them along so much (some of them are months old by now).

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Yes, you do. Just merge the two ifs into one and it'll be even better.

chilvence:
Sorry, I know this has gone on a bit, trouble is that as soon as I fixed one bug, I seemed to immediately find another. Some of them had me really stumped. 

Half the purpose was to work out what the PST code actually needed to do different compared to the main, which turned out to be not a great deal, so eventually I naturally ended up with something very similar just by shuffling things about. At that point, it seemed logical to eliminate what are now basically cloned functions, motive being that they can't get out of date again if they are no longer there :)

lynxlynxlynx:
It's fine, I'm just trying to ease review and get some closure.

gemrb/GUIScripts/InventoryCommon.py
chilvence: The -1 here came from GetSlotType, I haven't resolved the cause yet so I included a safety check

```
[Python]: control id:  21
[Python]: control id:  22
[Python]: control id:  23
[Python]: control id:  -1
[Python]: No control found for :  {'Slot': 16, 'ResRef': '-', 'Tip': 0, 'Effects': 5, 'Type': 0, 'ID': -1}
```

gemrb/GUIScripts/InventoryCommon.py
chilvence: That's just me being really dumb, fixed over here locally

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: It's from gemrb/unhardcoded/pst/slottype.2da. So checking for -1 is fine, but the first half still needs to go.

gemrb/GUIScripts/pst/GUIINV.py
chilvence: It appears UpdateSlot is careful enough to avoid this, I removed the check

chilvence:
I've tested BG1/2 and IWD1/2 now, this doesn't appear to cause any problems there (not that it should have, but always best to make sure)

lynxlynxlynx:
Here's a clean look at the other half:
https://github.com/chilvence/gemrb/compare/3208fc2...subviews-pst-inventory-update

I have to look at both UpdateInventorySlot variants — are they really that incompatible? The rest are just minor quips for tomorrow.

chilvence:
No they are not really functionally very different, only that the colour choices really clash with the PST art style. 

It also never had that "magic item" highlight that BG/IWD has, I think it looks out of place with it, since it isn't really the same game of "you found the super awesome longsword +5!"

The Button.SetBorder also uses different border indexes(?) in places, since I am not 100% sure why I thought it safer to preserve how it was set previously.

It could probably be merged if I change it to set color1, color2, color3 etc on load, so it won't have to have 50 extra game checks every time the inventory refreshes.

lynxlynxlynx:
Took a look and they definitely can be merged without much fuss.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -546,7 +549,30 @@ def UpdateSlot (pc, slot):
 
 	Window = GemRB.GetView("WIN_INV")
 
-	SlotType = GemRB.GetSlotType (slot+1, pc)
+	using_fists = slot_item = SlotType = None
+
+	if not GameCheck.IsPST():
+		SlotType = GemRB.GetSlotType (slot+1, pc)
+		slot_item = GemRB.GetSlotItem (pc, slot+1)
+	else:
```
lynxlynxlynx, line 557:
```
if not:
else:
```

is an antipattern; just invert the order to get rid of the negation.

chilvence:
Turned out to be even less fuss than not much, the only meaningful difference needed was avoiding the blue magical item border

lynxlynxlynx:
All fine, thanks!

chilvence:
Nice, now to move on to the fun stuff... Leveling up :)

PR #997, opened 2020-10-18
[subviews] Fixes for Windows and GCC 10

MarcelHB:
I had two problems compiling a recent subviews:

1. Nothing ever sets `USE_OPENGL_API`, so we had erroneously trying to load OpenGL _ES_
2. GCC 10.2 does not like memsetting the array of `Holder`, minor refactoring.


lynxlynxlynx:
Thanks!

PR #1004, opened 2020-10-20
Pst Message Window Close button

### LINKED ISSUE #445: [bug] Nitpick: PST Dialogue: Click to close is only meant to appear when reviewing the history
**Describe the bug**
Currently in regular dialogue, the click to close text appears, while the button cannot actually be used because there must be a dialogue response chosen instead.

GemRB version: 
any
**To Reproduce**
Talk to a party member or npc
**Expected behavior**
The button should appear blank until needed, either when reviewing the history or when the dialogue window is overflowing with text (click to continue/click for more in EE)

chilvence:
This is to fix the erroneous enabled appearance of the close button when in normal dialog, should close #445


lynxlynxlynx:
Thanks!

Issue #293, opened 2019-10-05
[subviews] Some missing text centering in PST


MarcelHB:
In master, it looks as in the originals, but in subviews (without recognizing a structural pattern), some texts appear misaligned when center is expected. I found two occurrences so far:

1. The game menu (see screenshot) some buttons - the blue one looks good.
2. In the wizard/priest spell book where it says _Nth level_ between the pages.

![pst_main_menu_buttons](https://user-images.githubusercontent.com/238558/66259857-0fb40c00-e7b7-11e9-91d5-938859c69d56.jpg)

bradallred:
Are you sure the center flag is actually set? it looks like if it is then its not centered because it's a single word that has a break.

MarcelHB:
I don't know, I can check the GUIscripts then.

bradallred:
Looking at the history of `Font` on this branch everything looks unrelated. the only thing I saw as a potential candidate is 70eda9c and that was clearly added to fix a bug somewhere else. You can try reversing that change and at least identify if that is the cause.

I'm assuming the comparison of _all_ alignment flags (not just the center flag) between master and this branch is the same?

MarcelHB:
Revert that commit fixes the issue, yes.

Other than that, the flags in the GUI script looks untouched (`IE_GUI_BUTTON_MULTILINE`).

bradallred:
I meant the value of `align` in `font->Print( r, Text, ppoi, (ieByte) align );`, but if reverting that commit fixes it then it is moot.

reverting that commit of course isn't the right thing to do. We would have to look at the history around that to get a better idea of what exact use case it was fixing.

its possible that simply reordering these 2 if blocks would fix it without breaking the other:
```
if (eos && stop->w && wordW < stop->w ) {
	// its possible the last word of string is longer than any of the previous lines
	w = std::max(w, wordW);
}
if (stop->h && (LineHeight * (lines + 1)) > stop->h ) {
	break;
}
```

bradallred:
actually, probably not since it looks like we don't pass any metrics to the print call in `Button`.  That leaves the `else` case or passing metrics as the possible fixes. Still need to look and see what issue this was fixing to know what to check for regression.

I suspect deleting the else would fix this but break the other. I'm thinking passing the metrics is a good solution, but I don't know if that would cause issues in yet a 3rd unknown case. I think its unlikely to be a problem tho.

bradallred:
[...]

PR #1009, opened 2020-10-20
[subviews] Improve whitespace handling

MarcelHB:
The original has some bad strings with two whitespaces (German PST) that are ignored by its rendering. We need to do this as some texts otherwise hit their containment borders.

about #293

## Checklist


bradallred:
I've mulled it over now and must say I think I was wrong to suggest we should just eat double spaces.

1. I hadn't realized that `StringSize` would need to be changed which I think definitely ought to not ignore whitespace.
2. its just unexpected behavior from a `Font`

we should find another way to address this irritant. There are lots of options and I'm not sure which is best.

1. Add a new flag to `GetString`
2. handle this in `Button::SetText`
3. handle it on the Python side

PR #1013, opened 2020-10-23
Speed up program start when in gdb by reducing plugin use

bernds:
This is more of a RFD - the problem of slow start is real, so I want to gauge whether there's interest in converting plugins to core code when the core depends on them anyway. This change handles the ResourceSource plugins and one which depends on them, but I can imagine more changes along these lines. Commit message below:

gdb can be very slow to start gemrb; it is probably reevaluating
breakpoints after every load of a plugin.
This moves all the ResourceSource code into core.  This simplifies the
code, and reduces the amount of plugins to be processed.  There's no
reason for them to be plugins, since it's not like the program can
function without them: Interface.cpp has hard dependencies on all of
them.


lynxlynxlynx:
I use gdb regularly and haven't noticed any unbearable slowdowns. Are you doing something special?

How much does your proposal speed things up? Most of the plugins still remain plugins after all.

bernds:
It's mostly noticeable if you have a lot of breakpoints and conditions set and restart the program.

This patch by itself isn't much of a speedup. As I said, it's mostly an RFD to see whether changes like this are welcome. I think it makes sense even outside of speedup considerations since it simplifies things and I can't see a good reason why any of these things should be plugins.

lynxlynxlynx:
The plugins related to file types I definitely don't support as moved to core, since that just makes them harder to disentangle and extend (eg. adding another music manager compatible source). NullSource and DirectoryImporter I'm indifferent on, but it doesn't look like much of a cleanup either way.

PR #1021, opened 2020-10-25
Pst level up subviews

chilvence:
I recently started rebasing the changes to PST level up I have been working on top of the subviews branch - it gives me a chance to look at everything twice, it avoids having to worry about merge conflicts with master, and also it means I can test for new bugs and regressions in the game along the way (of course not just in the level up code, but anything else I encounter)

I already found and fixed two new problems. One was an unintended consequence of my last skills patch which broke in subviews because the CreateTextArea function seems to be different. Another is that NewLife.py was hiding the various in game windows, but not setting them visible again after level up.

I also included here one small change for discussion, patching the common level up code to work with TNO's separate XP values, so that the code can be used generally to detect ability to level up. 

Before I can really continue though, I need some advice about how to best deal with TNO's three different XP/LEVEL stats

Unlike a multi class character, who gets evenly distributed XP divided by number of classes, each class XP value is individually tracked in the original game. You only gain XP in the class you are currently playing as. Currently the main problem in GemRB, there is nothing handling the adding XP gained to the correct IE_XP_[_____] stat depending on the active class. So although you can change your class in game, the XP you gain always goes into the first XP stat, which is only meant to save his fighter XP value. 

On my game just for testing, I wrote a workaround in python to shift any gained XP to the correct stat based on the active class of TNO . It worked very well in practice, I could switch classes and level them up individually, but it is also just a little bit hacky.

Another option is to handle it in Actor.cpp, right here in AddExperience: https://github.com/gemrb/gemrb/blob/d5da8d109a65d6b55e1ae4f660c9e6a7d213492f/gemrb/core/Scriptable/Actor.cpp#L8115

This is the only real major blocker I think, everything else should fall into place easily after deciding what to do here.

### DIFF
--- a/gemrb/GUIScripts/GUICommon.py
+++ b/gemrb/GUIScripts/GUICommon.py
@@ -558,6 +558,20 @@ def IsMultiClassed (actor, verbose):
 	# return the tuple
 	return (NumClasses, Classes[0], Classes[1], Classes[2])
 
+def NamelessOneClass (actor):
+	# simple check to identify if the actor is TNO
+	# and return his active class (by name)
+
+	if not GameCheck.IsPST():
+		return None
+
+	Specific = GemRB.GetPlayerStat (actor, IE_SPECIFIC)
+	if Specific == 2:
+		Class = GetClassRowName(actor)
+		return Class
+
+	return None
+
 def CanDualClass(actor):
 	# race restriction (human)
 	RaceName = CommonTables.Races.FindValue ("ID", GemRB.GetPlayerStat (actor, IE_RACE, 1))
--- a/gemrb/GUIScripts/LUCommon.py
+++ b/gemrb/GUIScripts/LUCommon.py
@@ -59,6 +59,21 @@ def CanLevelUp(actor):
 		next = GUIREC.GetNextLevelExp(levelsum, GUIREC.GetECL(actor))
 		return next <= xp
 
+	# hardcoded special case to handle TNO, who is usually a single class 
+	# but with three separate Levels/XP values and the ability to switch between them
+	# it returns the active class if true
+	SwitcherClass = GUICommon.NamelessOneClass(actor) 
+	if SwitcherClass:
+		xp = { "FIGHTER" : GemRB.GetPlayerStat (actor, IE_XP), "MAGE" : GemRB.GetPlayerStat (actor, IE_XP_MAGE), "THIEF" : GemRB.GetPlayerStat (actor, IE_XP_THIEF) }
+		lvls = { "FIGHTER" : Levels[0] , "MAGE": Levels[1], "THIEF": Levels [2] }
+
+		tmpNext = GetNextLevelExp (lvls[SwitcherClass], SwitcherClass)
+		if (tmpNext != 0 or lvls[SwitcherClass] == 0) and tmpNext <= xp[SwitcherClass]:
+			return 1
+		#ignore the rest of this function, to avoid false positives
+		#other classes can only be achieved by hacking the game somehow
+		return 0
+
 	if Multi[0] > 1: # multiclassed
 		xp = xp/Multi[0] # divide the xp evenly between the classes
 		for i in range (Multi[0]):
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -805,7 +805,6 @@ def OpenBiographyWindow ():
 
 
 def AcceptLevelUp():
-	LevelUpWindow.Close()
 	#do level up
 	pc = GemRB.GameGetSelectedPCSingle ()
 	GemRB.SetPlayerStat (pc, IE_SAVEVSDEATH, SavThrows[0])
@@ -833,7 +832,7 @@ def AcceptLevelUp():
 	
 	LUSkillsSelection.SkillsSave (pc)
 
-	UpdateRecordsWindow(RecordsWindow)
+	LevelUpWindow.Close()
 	NewLife.OpenLUStatsWindow()
 
 def RedrawSkills():
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -67,6 +67,7 @@
 import GemRB
 import GUICommon
 import CommonTables
+import LUCommon
 import GUICommonWindows
 import NewLife
 from GUIDefines import *
@@ -148,17 +149,10 @@ def UpdateRecordsWindow (Window):
 
 	# Checking whether character has leveled up.
 	Button = Window.GetControl (9)
-	if avatar_header['SecoLevel'] == 0:
-		if avatar_header['XP'] >= avatar_header['PrimNextLevXP']:
-			Button.SetState (IE_GUI_BUTTON_ENABLED)
-		else:
-			Button.SetState (IE_GUI_BUTTON_DISABLED)
+	if LUCommon.CanLevelUp (pc):
+		Button.SetState (IE_GUI_BUTTON_ENABLED)
 	else:
-		# Character is Multi-Class
-		if avatar_header['XP'] >= avatar_header['PrimNextLevXP'] or avatar_header['XP'] >= avatar_header['SecoNextLevXP']:
-			Button.SetState (IE_GUI_BUTTON_ENABLED)
-		else:
-			Button.SetState (IE_GUI_BUTTON_DISABLED)
+		Button.SetState (IE_GUI_BUTTON_DISABLED)
 
 	# name
 	Label = Window.GetControl (0x1000000a)
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -805,7 +805,7 @@ def OpenBiographyWindow ():
 
 
 def AcceptLevelUp():
-	OpenLevelUpWindow()
+	LevelUpWindow.Close()
 	#do level up
 	pc = GemRB.GameGetSelectedPCSingle ()
 	GemRB.SetPlayerStat (pc, IE_SAVEVSDEATH, SavThrows[0])
@@ -851,13 +851,6 @@ def OpenLevelUpWindow ():
 	global HPGained
 	global WeapProfType, CurrWeapProf, WeapProfGained
 	global NumOfPrimLevUp, NumOfSecoLevUp
-
-	if LevelUpWindow != None:
-		if LevelUpWindow:
-			LevelUpWindow.Unload ()
-		LevelUpWindow = None
-
-		return
 
 	LevelUpWindow = Window = GemRB.LoadWindow (4, "GUIREC") # since we get called from NewLife
 
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -102,7 +102,7 @@ def InitRecordsWindow (Window):
 	# Level Up
 	Button = Window.GetControl (9)
 	Button.SetText (4246)
-	Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, NewLife.OpenLUStatsWindow)
+	Button.SetEvent (IE_GUI_BUTTON_ON_PRESS, OpenLevelUpWindow)
 
 	statevents = (OnRecordsHelpStrength, OnRecordsHelpIntelligence, OnRecordsHelpWisdom, OnRecordsHelpDexterity, OnRecordsHelpConstitution, OnRecordsHelpCharisma)
 	# stat buttons
@@ -832,7 +832,9 @@ def AcceptLevelUp():
 			GemRB.SetPlayerStat (pc, IE_LEVEL2, GemRB.GetPlayerStat (pc, IE_LEVEL2)+NumOfSecoLevUp)
 	
 	LUSkillsSelection.SkillsSave (pc)
-	UpdateRecordsWindow ()
+
+	UpdateRecordsWindow(RecordsWindow)
+	NewLife.OpenLUStatsWindow()
 
 def RedrawSkills():
 	DoneButton = LevelUpWindow.GetControl(0)
--- a/gemrb/GUIScripts/pst/NewLife.py
+++ b/gemrb/GUIScripts/pst/NewLife.py
@@ -142,7 +142,10 @@ def OpenLUStatsWindow(Type = 1):
 		Button.SetActionInterval (200)
 
 	NewLifeLabel = NewLifeWindow.GetControl(0x10000023)
-	NewLifeLabel.SetText(1899)
+	if LevelUp:
+		NewLifeLabel.SetText(19356)
+	else:
+		NewLifeLabel.SetText(1899)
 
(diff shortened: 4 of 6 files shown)


gemrb/GUIScripts/LUCommon.py
```diff
@@ -59,6 +59,28 @@ def CanLevelUp(actor):
 		next = GUIREC.GetNextLevelExp(levelsum, GUIREC.GetECL(actor))
 		return next <= xp
 
+	# hardcoded special case to handle TNO, who is usually a single class 
+	# but with three separate Levels/XP values and the ability to switch between them
+	# it returns the active class if true
+	Switcher = GUICommon.IsNamelessOne(actor) 
+	if Switcher:
+		xp = [ GemRB.GetPlayerStat (actor, IE_XP), GemRB.GetPlayerStat (actor, IE_XP_MAGE), GemRB.GetPlayerStat (actor, IE_XP_THIEF) ]
```
lynxlynxlynx, line 67: You can add a dict to map the class names to the xp/level index, use it to get the correct index regardless of results and then avoid the code triplication below. Less chance of mistakes that way too — now you have too many Levels[0] checks.

lynxlynxlynx:
I'm not sure this will help with merge conflicts, but I doubt it can make things work.

As far as xp stats go, this will have to be checked for when applying the xp granted in Actor::AddExperience.

chilvence:
> I'm not sure this will help with merge conflicts

Sure it will - if I make any changes in the master branch, I also have to think about whether or not it breaks after subviews is merged, like the two problems fixed above. If I just target subviews branch directly and forget master, there is no need to consider the changes that have happened between master and subviews at all. Also from my POV, since subviews branch is pegged to be the next release version, the less I interfere with the code it is based on, the better. 

Not to mention the fact that I can only help playtest if I am actually using this branch. Just trying to find the overall most effective way to use my time :)

Back to the original point though, I will try modifying Actor::AddExperience based on what I have in my python prototype. The result should be neater.

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -838,7 +832,13 @@ def AcceptLevelUp():
 			GemRB.SetPlayerStat (pc, IE_LEVEL2, GemRB.GetPlayerStat (pc, IE_LEVEL2)+NumOfSecoLevUp)
 	
 	LUSkillsSelection.SkillsSave (pc)
-	UpdateRecordsWindow ()
+
+	GUICommonWindows.OptionsWindow.SetVisible (True)
```
bradallred, line 836: I'm trying to figure out why this is required, but even if it is I believe `SetGameGUIHidden(False)` will do this all in one line

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -838,7 +832,13 @@ def AcceptLevelUp():
...
+	GUICommonWindows.PortraitWindow.SetVisible (True)
+	GUICommonWindows.ActionsWindow.SetVisible (True)
+	RecordsWindow.SetVisible (True)
+
+	UpdateRecordsWindow (RecordsWindow)
```
bradallred, line 841: You wouldn't need to manually call `UpdateRecordsWindow` if you set `UpdateRecordsWindow` as a handler for  the `ACTION_WINDOW_FOCUS_GAINED` action

bradallred (changes requested):

gemrb/GUIScripts/pst/GUIREC.py
chilvence: NewLife.py hides all the windows because it loads the stat selection window for the Nameless One (regardless of which character is levelling). That's a fullscreen window like the start screen or bio screen.

Although, that also needs maintenance, so I can add that to the todo list

gemrb/GUIScripts/pst/GUIREC.py
chilvence: Hmm, I see that should be covered by CreateTopWinLoader, so perhaps this part is redundant? I think I just patched it from what was already there, to avoid a python error.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8184,14 +8184,36 @@ void Actor::AddExperience(int exp, int combat)
 		bonus += adjustmentPercent;
 	}
 	bonus += GetFavoredPenalties();
-	exp = ((exp * (100 + bonus)) / 100) + BaseStats[IE_XP];
+
+	int classid = GetActiveClass() -1; //-1 to allow lookup in xpcap array
```
lynxlynxlynx, line 8188: You can check IE_CLASS directly and also only for pst. Plus, please use a consistent naming style. I think the most common one is camel case for vars, which is true at least for this function.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8184,14 +8184,36 @@ void Actor::AddExperience(int exp, int combat)
...
+	int xp_which = IE_XP;
+
+	// decide which particular XP stat to add to (only for TNO's switchable classes)
+	if (pstflags) {
+		Game* game = core->GetGame();
+		// rule only applies to the protagonist
+		if ( GetScriptName() == game->GetPC(0, false)->GetScriptName() ) {
```
lynxlynxlynx, line 8195: You can compare the pointers directly, no need to look up the script name.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8184,14 +8184,36 @@ void Actor::AddExperience(int exp, int combat)
...
+			switch (classid+1) {
```
lynxlynxlynx, line 8196: -1+1=0

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8184,14 +8184,36 @@ void Actor::AddExperience(int exp, int combat)
 		bonus += adjustmentPercent;
 	}
 	bonus += GetFavoredPenalties();
-	exp = ((exp * (100 + bonus)) / 100) + BaseStats[IE_XP];
+
+	int classid = GetActiveClass() -1; //-1 to allow lookup in xpcap array
+	int xp_which = IE_XP;
```
lynxlynxlynx, line 8189: xpStat sounds clearer to me.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: and you can do it one level higher, so there's less nesting.

gemrb/core/Scriptable/Actor.cpp
chilvence: The classid var is used at the end of the function to lookup the xp cap from an array, hence the starting at 0 - I didn't change that part, I just moved the initialisation of it upward because I wanted to reuse the value

But it sounds like it isn't needed at all, if I can just use BaseStats[IE_CLASS] in both cases?

chilvence: Actually it looks like it's important to keep this as a call to GetActiveClass, since it checks for dual classing in the other games.

lynxlynxlynx: You can internalize the +1 to the case values.

lynxlynxlynx: Right, didn't see you moved it — thought it was new.

gemrb/core/Scriptable/Actor.cpp
chilvence: I was basically looking around for anywhere else that checks if the actor is TNO, and that's the first example I found:

https://github.com/gemrb/gemrb/blob/d5da8d109a65d6b55e1ae4f660c9e6a7d213492f/gemrb/core/Scriptable/Actor.cpp#L5849-L5853

Not sure if there is a less convoluted check somewhere else, if not, should I come up with one?

lynxlynxlynx: It's not convoluted, it can just be improved (which will also make it work in iwds). You can add a `Game::IsProtagonist(Actor *pc)` method to abstract it away.

gemrb/GUIScripts/pst/GUIREC.py
bradallred: That doesn't appear to apply here because `OpenLevelUpWindow` doesn't use that mechanism (tho it probably should).

gemrb/core/Scriptable/Actor.cpp
chilvence: Out of curiosity, is there any reason not to just use "this" to compare to PC 0 ? I pushed a modified example that seems to work fine

chilvence: Also, I think IWD is fine in the lines I linked to, because PM_NO is only set for PST, IWD gets PM_TEAM

lynxlynxlynx: this: that's what I meant by the first comment.
iwd: the function wouldn't check for protagonist mode though, so it would break if it checked for script name.

chilvence: I'm not sure I'm following 100%

I'm not really trying to add a check to detect the protagonist, I have to specifically identify the Nameless one is the actor here, since he is the only character with this XP gain style. That's what I meant the `pstflags` part of the check to achieve.

According to IESDP, the other games don't even have these extra stats, not even IWD2. They keep track of the different levels, sure, but the XP is one stat divvied between number of classes (or whatever other method for dual class chars)

Unless we're talking about completely different things?

lynxlynxlynx: I was talking about a Game::IsProtagonist.

chilvence: Ah right, I'm not even sure how I would decide that in IWD, I thought it wasn't even supposed to treat any one character as the 'main' character. 

But again, the intention is just to isolate this block to PST and TNO specifically, it shouldn't go near any of the other games actors, or even the other PST npc's which already work fine.

gemrb/GUIScripts/pst/GUIREC.py
bradallred: im not sure I understand. why does that translate into needing to hide the GUI? you said it is a fullscreen window so who cares what else is underneath it.

gemrb/GUIScripts/GUICommon.py
```diff
@@ -558,6 +558,20 @@ def IsMultiClassed (actor, verbose):
...
+	if not GameCheck.IsPST():
+		return False
+
+	Specific = GemRB.GetPlayerStat (actor, IE_SPECIFIC)
+	if Specific == 2:
+		Class = GetClassRowName(actor)
+		return Class
+
+	return False
+
```
bradallred, line 574: the name `IsNamelessOne` suggest a boolean return value which is not the case. I would return `None` instead of `False`.

It's also not clear to me how you can hit the last return. shouldn't it be an exception if `Specific != 2`? maybe I just dont understand PST leveling tho.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: The protagonist is always the first character, that's what the check is about.

gemrb/GUIScripts/GUICommon.py
lynxlynxlynx: Re 2: no, other pcs will run this code as well.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: You're right, it works fine without it, so I will remove it all.

chilvence:
Oh lord, what have I made git do now.

gemrb/GUIScripts/pst/NewLife.py
```diff
@@ -63,17 +63,12 @@ def OpenLUStatsWindow(Type = 1):
...
-		GUICommonWindows.OptionsWindow.SetVisible (False)
-		GUICommonWindows.PortraitWindow.SetVisible (False)
-		GUICommonWindows.ActionsWindow.SetVisible (False)
-		GUIREC.RecordsWindow.SetVisible (False)
 		# only TNO gets the main stat boosts
 		pc = GemRB.GameGetSelectedPCSingle ()
 		Specific = GemRB.GetPlayerStat (pc, IE_SPECIFIC)
 		if Specific != 2:
-			GUIREC.OpenLevelUpWindow ()
+			GUIREC.UpdateRecordsWindow(GUIREC.RecordsWindow)
```
bradallred, line 71: this is useless. `OpenLUStatsWindow` is called from `InitRecordsWindow` which is part of the `CreateTopWinLoader` that will also call `UpdateRecordsWindow`

as an additional comment that has nothing to do with your changes, but maybe you feel like addressing anyway:

I hate that `OpenLUStatsWindow` does things other than open the window. I wish that other stuff would be moved to the singular location where `OpenLUStatsWindow` is called.

gemrb/GUIScripts/pst/NewLife.py
```diff
@@ -250,9 +250,10 @@ def AcceptPress():
 	GemRB.SetPlayerStat(1, IE_CHR, Stats[5])
 
 	if LevelUp:
-		# hp is handled in GUIREC
+		# Return to the RecordsWindow
+		NewLifeWindow.Close()
 		import GUIREC
-		GUIREC.OpenLevelUpWindow ()
+		GUIREC.UpdateRecordsWindow(GUIREC.RecordsWindow)
```
bradallred, line 256: this too _should_ be useless, and if it isn't then there is a bug. 

`UpdateRecordsWindow` is already bound to the `ACTION_WINDOW_FOCUS_GAINED` event for the records window so it should be getting called automatically anytime the records window gains focus.

gemrb/GUIScripts/pst/NewLife.py
chilvence: That's not a problem, I can split it out - I just don't usually think too much about changing stuff unless it isn't working, everyone else is way more qualified than me to make design decisions :)

gemrb/GUIScripts/pst/NewLife.py
chilvence: I tried first just with GUIREC.RecordsWindow.Focus() , but for some reason that wasn't working. But I'll try and figure out why, if that is how it should be done.

bradallred: I think you misunderstand me. You shouldn't have to do anything. when `NewLifeWindow` closes the records window will automatically gain focus and automatically call `UpdateRecordsWindow`

chilvence: The records window opens the level up sub window, but looking at AcceptLevelUp I think I might know where it is going wrong, just need to try something...

lynxlynxlynx:
Looks fine now to me.

gemrb/GUIScripts/pst/NewLife.py
bradallred: I'm not sure why you are scratching your head. I think the problem is that you are closing the window at the top of the function instead of the bottom.

chilvence: Doh, that's exactly it. I just assumed that the function was being called by closing the NewLife window...

Issue #998, opened 2020-10-18
[PST] Regression in 0.8.7: Item "Use"/"Talk to Item" not working anymore 


rfht:
#### Bug description

In PST (built on OpenBSD), I can't use items anymore - neither from inventory screen, nor from the quick menu. I encountered this first with the note on Zombie number 1201 in the Morgue, then found the Copper Ring in the same place and any item (Charcoal Charm) can't be used anymore. A "Use" Button is shown in the lower left corner of the item description in the inventory, but nothing happens when I click it and the console log shows: `[GUIScript/WARNING]: Empty slot used?`

Compared it to 0.8.6 - there, the item description for the Zombie 1201 note and the Copper Earring show "Talk to Item" rather than "Use" in the lower left corner, and pressing it opens the dialog to manipulate it.

GemRB version:  0.8.7

#### Steps to reproduce
1. Obtain the note from Zombie 1201 in the Morgue. Go to inventory.
2. Right click on the note. Then click on "Use".
3. Nothing happens. Take a look at console log.
4. See error: `[GUIScript/WARNING]: Empty slot used?`

#### Expected behavior

Dialog to interact with the item opens as it did in 0.8.6.

lynxlynxlynx:
Thanks for the report. I think there was a misunderstanding when fixing one of the bugs and #986 will take care of it. If you want a quick fix, cherry-pick the DisplayItem change from e903950.

chilvence:
Is it worth me splitting this commit out to master? I think the rest of the bugs are subviews specific, but this one isn't

lynxlynxlynx:
We're getting close, so I don't think it's needed.

Northfear:
@chilvence is it still possible to split it into master? Changes from DisplayItem did help, but I'm still getting "Empty slot used?" when using  Jar of Embalming Fluid  for example.

chilvence:
Are you building from source? If so you can probably get away with just "git cherry-pick"-ing the relevant commit ( which is 
bd235ad ) . I don't manage the repository at all, I just send patches sometimes, so I can't split it out myself, but the commit should merge in without any problems.

Northfear:
Yeah, I did exactly that. Changes from "DisplayItem" fixed the note from 1201, but I'm still getting
`[GUIScript/WARNING]: Empty slot used?`
when using  Jar of Embalming Fluid from inventory, so I'm just confused whether it's working as expected or not quite.

chilvence:
[...]

PR #1022, opened 2020-10-26
InventoryCommon.DialogItemWindow: Update the inventory window closing method

chilvence:
This should hopefully fix the rest of #998

The last inventory PR fixed the ability to enter the dialog mode, but I missed the fact that the inventory window does not close like it should in subviews branch, so it was overall still not functional.

This changes the close function to what I think is the newer method (please correct me if I'm wrong, but it does work in game)

### DIFF
--- a/gemrb/GUIScripts/InventoryCommon.py
+++ b/gemrb/GUIScripts/InventoryCommon.py
@@ -873,9 +873,7 @@ def DialogItemWindow ():
 	dialog=item["Dialog"]
 	if ItemInfoWindow:
 		ItemInfoWindow.Unload ()
-		
-	import GUIINV
-	GUIINV.OpenInventoryWindow ()
+
 	GemRB.ExecuteString ("StartDialogOverride(\""+dialog+"\",Myself,0,0,1)", pc)
 	return
 


bradallred:
yes, this is an appropriate method. I'm assuming its not possible to get here without the inventory window still being around? otherwise we need an `if InventoryWindow` check

bradallred:
hmmm, i'm trying to understand the need for this at all. `DialogStarted` should automatically close the window.

chilvence:
1: It should be impossible to reach the ItemInfoWindow unless the Inventory is open, I think it's one of the subwindows there

2: I'm not so sure that would automatically happen, there are various other sub windows you can reach via the ItemInfoWindow which shouldn't close the inventory - this is just to handle PST dialog items and BG2 Lilarcor, which only has to exit the inventory to load the message window really.

bradallred:
can you see if that last commit fixed everything you are seeing? We should still merge in the removal of 

```
import GUIINV
 GUIINV.OpenInventoryWindow ()
```

chilvence:
That actually fixes the DialogStart function and makes the need to close the Inventory window redundant, so I will change this to reflect that.

bradallred:
thanks!

PR #1023, opened 2020-10-28
Pst level up subviews continued

chilvence:
A continuation of bugfixes to PST level up:

1: Level up was not considering which TNO class level to upgrade, it would only upgrade the fighter level stat

2: In many cases, the assumption TNO's specific stat was wrong. It is actually 2, not 1, when you inspect a save game character or look at CHARBASE

3: The weapon proficiency table was not properly returning a value because it had no 'default' at the top. All calls to it were simply returning "ID", the name of the first column, so character proficiencies were not being detected

4: HasGainedWeapProf was given the wrong class name type as a parameter - the code was passing it the display name, which comes from a string and changes by language, and then comparing it to the internal name in all caps, so that was never functional

5: AcceptLevelUp was not accounting for the fact that WeapProfType is just an offset from the first weapon proficiency stat. Although this code would never have been run because of the above issues, it would have just done ```SetPlayerStat(pc, 0, etc``` , which would be bad.

6: There was also no accounting for adding TNO's unused slots for the dialog based combat training

## Checklist

### DIFF
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -819,12 +819,13 @@ def AcceptLevelUp():
 	#increase weapon proficiency if needed
 	if WeapProfType!=-1:
 		GemRB.SetPlayerStat (pc, WeapProfType, CurrWeapProf + WeapProfGained )
-	Specific = GemRB.GetPlayerStat (pc, IE_SPECIFIC)
-	if Specific == 1:
-		#TODO:
-		#the nameless one is dual classed
-		#so we have to determine which level to increase
-		GemRB.SetPlayerStat (pc, IE_LEVEL, GemRB.GetPlayerStat (pc, IE_LEVEL)+NumOfPrimLevUp)
+
+	SwitcherClass = GUICommon.NamelessOneClass(pc) 
+	if SwitcherClass:
+		# Handle saving of TNO class level in the correct CRE stat
+		Levels = { "FIGHTER" : GemRB.GetPlayerStat (pc, IE_LEVEL) , "MAGE": GemRB.GetPlayerStat (pc, IE_LEVEL2), "THIEF": GemRB.GetPlayerStat (pc, IE_LEVEL3) }
+		LevelStats = { "FIGHTER" : IE_LEVEL , "MAGE": IE_LEVEL2, "THIEF": IE_LEVEL3 }
+		GemRB.SetPlayerStat (pc, LevelStats[SwitcherClass], Levels[SwitcherClass]+NumOfPrimLevUp)
 	else:
 		GemRB.SetPlayerStat (pc, IE_LEVEL, GemRB.GetPlayerStat (pc, IE_LEVEL)+NumOfPrimLevUp)
 		if avatar_header['SecoLevel'] != 0:
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -899,7 +899,14 @@ def OpenLevelUpWindow ():
 				break
 
 	if WeapProfType!=-1:
+		# NPC: Just count the points in the favoured weapon proficiency
 		CurrWeapProf = GemRB.GetPlayerStat (pc, IE_PROFICIENCYBASTARDSWORD + WeapProfType)
+	else:
+		# TNO: Count the total amount of assigned and unassigned proficiencies
+		CurrWeapProf = GemRB.GetPlayerStat(pc, IE_FREESLOTS)
+		for i in range (6):
+			CurrWeapProf += GemRB.GetPlayerStat (pc, IE_PROFICIENCYBASTARDSWORD + i)
+		CurrWeapProf -= 4 #adjust for the 4 slots already given at level 1
 
 	# Recording this avatar's current proficiency level
 	# Since Nameless one is not covered, hammer and club can't occur
@@ -947,6 +954,11 @@ def OpenLevelUpWindow ():
 		while avatar_header['XP'] >= GetNextLevelExp (NextLevel, Class):
 			NextLevel = NextLevel + 1
 		NumOfPrimLevUp = NextLevel - avatar_header['PrimLevel'] # How many levels did we go up?
+
+		#How many weapon procifiencies we get
+		for i in range (NumOfPrimLevUp):
+			if HasGainedWeapProf (pc, CurrWeapProf + WeapProfGained, avatar_header['PrimLevel'] + i):
+				WeapProfGained += 1
 
 		# Is avatar Nameless One?
 		if Specific == 2:
@@ -998,10 +1010,6 @@ def OpenLevelUpWindow ():
 						SavThrUpdated = True
 			# Cleaning up
 		else:
-			#How many weapon procifiencies we get
-			for i in range (NumOfPrimLevUp):
-				if HasGainedWeapProf (pc, CurrWeapProf + WeapProfGained, avatar_header['PrimLevel'] + i):
-					WeapProfGained += 1
 
 			# Saving Throws
 			# Loading the right saving throw table
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -396,13 +396,13 @@ def GetCharacterHeader (pc):
 
 	Class = GemRB.GetPlayerStat (pc, IE_CLASS) - 1
 	Multi = GUICommon.HasMultiClassBits (pc)
-	Specific = "%d"%GemRB.GetPlayerStat (pc, IE_SPECIFIC)
+	Specific = GemRB.GetPlayerStat (pc, IE_SPECIFIC)
 
-	#Nameless is Specific == 1
+	#Nameless is Specific == 2
 	avatar_header['Specific'] = Specific
 
 	# Nameless is a special case (dual class)
-	if Specific == 1:
+	if Specific == 2:
 		avatar_header['PrimClass'] = CommonTables.Classes.GetRowName (Class)
 		avatar_header['SecoClass'] = "*"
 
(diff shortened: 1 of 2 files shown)


gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -398,11 +398,11 @@ def GetCharacterHeader (pc):
 	Multi = GUICommon.HasMultiClassBits (pc)
 	Specific = "%d"%GemRB.GetPlayerStat (pc, IE_SPECIFIC)
```
lynxlynxlynx, line 399: does anything depend on this being a string? Now we use a mix and you're changing it further, so it's a good chance to standardize.

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -1185,9 +1185,10 @@ def GetThac0 (Class, Level):
 	return Thac0Table.GetValue (Class, str (Level))
 
 #apparently the original code doesn't follow profsmax/profs tables
-def HasGainedWeapProf (pc, currProf, currLevel, Class):
+def HasGainedWeapProf (pc, currProf, currLevel):
 	#only fighters gain weapon proficiencies
-	if Class!="FIGHTER":
+ 	Class = GUICommon.GetClassRowName(pc).split("_")
+	if Class[0]!="FIGHTER":
```
lynxlynxlynx, line 1194: Something cleaner like:
```
	Class = GUICommon.GetClassRowName (pc)
	if not "FIGHTER" in Class:
```

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -818,7 +818,7 @@ def AcceptLevelUp():
 	GemRB.SetPlayerStat (pc, IE_HITPOINTS, HPGained+oldhp)
 	#increase weapon proficiency if needed
 	if WeapProfType!=-1:
-		GemRB.SetPlayerStat (pc, WeapProfType, CurrWeapProf + WeapProfGained )
```
lynxlynxlynx, line 821: If you don't like the addition, just set up WeapProfType properly. After all it's read from a table, so it can be fixed in the data.

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -819,6 +819,9 @@ def AcceptLevelUp():
 	#increase weapon proficiency if needed
 	if WeapProfType!=-1:
 		GemRB.SetPlayerStat (pc, IE_PROFICIENCYBASTARDSWORD+WeapProfType, CurrWeapProf + WeapProfGained )
+	else:
+		freeSlots = GemRB.GetPlayerStat(pc, IE_FREESLOTS)
+		GemRB.SetPlayerStat (pc, IE_FREESLOTS, freeSlots + WeapProfGained )
```
lynxlynxlynx, line 824: WeapProfGained isn't decreased any where, so shouldn't this be a subtraction, capped at 0?

gemrb/GUIScripts/pst/GUIREC.py
chilvence: Glad you said so, because I thought it was odd. I searched for any instance where it was required to be a string, but couldn't find any, so changed it to a regular number.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: That is way better and works just as well in practice, done.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: I think it just reads weird when you look at it because of the name ```IE_PROFICIENCYBASTARDSWORD``` which is kind of just an alias for eg "IE_PROFICIENCYFISTS" if the definitions had PST based names. The result is exactly the same though, and this method is used various other places.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: WeapProfGained should just be the total new prof points gained from the ```HasGainedWeapProf``` loop. You don't get to actually assign them immediately on level up. I thought to just treat it like any other weapon proficiency, because you can potentially stack them up if you can't find an NPC that can train you. But it's quite possible I have this sussed all wrong...

lynxlynxlynx:
Are you still working on something?

chilvence:
I just didn't expect fixing one thing to immediately break something else , but I had already started the PR and it was late. I am pretty sure HP and Thac0 calculation is fine again, just wanted to do a bit more testing to be sure.

lynxlynxlynx:
Well just let us know then. Since you're marking your PRs as drafts, it's never clear when you're done.

chilvence:
I'm mainly using drafts for where I'm genuinely not sure I'm doing the right thing - much better to rewrite as many times as necessary than merge it and regret later

I might actually retract some of this though, unless you think it would be fine to have an temporary solution to the weapon proficiencies not being given.

I was only trying to preserve this automatic weapon proficiency code because it is one of the few PST specific level up quirks that can't be handled by just plugging it into the common level up code. It is actually working and doing something now, which is better than it was before. However, the more I test it, the less I find it matches up to what actually happens in PST. It never accounted for the different limits/starting profs of any character, and when I tweak the formula it uses to make it right for one character, it ends up wrong for a different one.

I would have guessed that even though it says that the original game doesn't use the 2da tables in a comment there, that the progression is still supposed to be straight D&D rules. But the number of proficiencies each character starts with in the original seems to be much less than the rules would have. I think they were nerfed because they can only use one weapon type and they wanted to space out the level up progression.

I'm pretty sure where it said "only fighters get proficiencies" is also a bit of red herring, since it's only Grace that gets no points on account of being a pacifist with no ability to carry weapons. I think the secondary class is also supposed to be taken into account for Dakkon and Annah, otherwise they seem to get short-changed compared to the original. As it is, it only cares about the primary class.

Or it could just have been completely arbitrary on a per character basis. Should I draw up a character proficiency table based on whatever I observe hacking the original chars and use that instead?

Alternatively, should I just use the BG/IWD method, add any missing tables, and then tweak the results where necessary?

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -1188,15 +1188,29 @@ def GetThac0 (Class, Level):
...
+	# Limits obtained from testing original game
+	limit = 1000
+	if Class == "FIGHTER":
+		# Morte, Nordom amd Vhailor are limited to 5, but not TNO
+		if not GUICommon.NamelessOneClass(pc):
+			limit = 5
+	elif Class == "FIGHTER_MAGE" or Class == "FIGHTER_THIEF":
+		# Dakkon and Annah are limited to 4
+		limit = 4
+	else:
```
lynxlynxlynx, line 1203: What about Ignus the mage?

lynxlynxlynx:
Any improvement is welcome, even if it's not the 100 % correct one yet. Perfect is the enemy of good and all that.

Proficiencies: I guess a table would help understand what's going on, even if we then end up using formulae.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: That's a good point, I've never really included Ignus in my party so I can't remember, but it could be possible he is meant to have some progression. Easy to test though.

chilvence: Just checked, even at level 20 he only has one prof point and a thac0 of 14

chilvence: It seems to be a little bit arbitrary, just drew this up for reference:
```
Dakkon
Fighter		4	5	5	6	6	6	7	7	8	8	9	9	10	11
Mage		4	4	5	5	6	7	7	8	8	9	10	11	11	12
Points		1	2	2	2	2	2	2	2	3	3	3	3	3	4
Special	Blade shift at level 7 (+2), 10 (+3)

Annah
Fighter		4	5	5	5	6	6	7	8	8	8	9	9	10	10	11
Thief		4	4	5	6	6	7	7	9	10	11	11	12	12	13	13
Points		1	2	2	2	2	2	2	3	3	3	3	3	3	3	4

Morte
Fighter		2	3	4 	5	6	7	8	9	10	11
Points		2	2	2	3	3	3	4	4	4	5

Vhailor
Fighter		9	10	11	12	13	14	15	16	17
Points		2	2	3	3	4	4	4	4	5

Nordom
Fighter		6	7	8 	9	10	11	12	13	
Ponts		2	2	3	3	3	4	4	5
```

Dakkon and Annah get a point when their fighter level hits 5, 8, 11, capped at 4
They do only get points for fighter levels, but the progression is moved ahead one level so they get them sooner

Morte starts with 2 points, and also gets one at 5,8,11, capped at 5

Vhailor starts with 2 at level 9, gets a point at 11, 14 then 17 capped at 5

Then Nordom spoils the pattern by getting a point at lv 13 instead of 14 :)

Grace and Ignus get nothing. You'd think Ignus would practice throwing his fireballs just for the fun of it...

TNO starts with 2 spent and 3 unspent points, gains one at level 6 and every three fighter levels thereafter, regardless of how many levels in other classes. However his ability to spend proficiencies in a particular weapon is capped according to his class specialisation, I believe.

lynxlynxlynx: A table is the way to go then and it will nicely encode at least FFG. Are you saying TNO doesn't get any extra points by just maxing out a thief or a mage?

chilvence: No, you can be level 125 mage or thief and still get nothing

Interestingly though, for TNO his proficiency progression continues after level 20, seemingly still 1 point every three levels. I racked it up to 34 unspent points and stopped testing

chilvence:
I'm going to close this for now while I re-consider the approach - It's helped me figure out the nitty details, but I think I'll do this prof part again with a blank slate.

PR #1026, opened 2020-10-29
pst/GUIREC: Fix annoyance of missing species display on record sheet

### LINKED ISSUE #1025: PST Record Sheet: Character species label broken due to original save bug
I've known about this one for a while, but not sure the best way to solve it.  It isn't really super important, but an issue on the tracker means it won't be forgotten

The original game seems to not save the player characters species stat properly, they all get set to ```0``` so when you load a game with GemRB, which uses that stat to decide which text label to display on the record sheet, they all show ```*``` instead.

https://github.com/gemrb/gemrb/blob/d43ce277445eddf2a6a7a5f3ce5026264fdd7543/gemrb/GUIScripts/pst/GUIREC.py#L256

Changing this stat has no effect on the character species label when you load in the original, so I can only guess that it was hardcoded.

GemRB version:  Any

#### Screenshots
unknown species

I can probably solve this one easily, but I don't know what the preferred method is.

Perhaps it can be solved by adding or extending an unhardcoded 2da file and using that instead of this stat?

chilvence:
This only applies to original save games, which don't seem to respect the IE_SPECIES stat of player characters

Should close #1025

### DIFF
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -250,10 +250,18 @@ def UpdateRecordsWindow ():
 		Label.SetText (str (stats[i]))
 
 	# race
-	# HACK: for some strange reason, Morte's race is 1 (Human), instead of 45 (Morte)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #1034, opened 2020-11-03
pst/GUIREC: Handle saving of TNO level in the correct Stat

chilvence:
This commit was actually meant to be the last part of #1021 and the handling of each different class level for TNO, but got left out before that was merged.

I was just going bundle it with #1023 at first, but since I'll be rethinking the weapon proficiency handling and that part of GUIREC in general, I've split it out from there so that it doesn't end up getting lost.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/pst/GUIREC.py
+++ b/gemrb/GUIScripts/pst/GUIREC.py
@@ -819,12 +819,13 @@ def AcceptLevelUp():
 	#increase weapon proficiency if needed
 	if WeapProfType!=-1:
 		GemRB.SetPlayerStat (pc, WeapProfType, CurrWeapProf + WeapProfGained )
-	Specific = GemRB.GetPlayerStat (pc, IE_SPECIFIC)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
There's a bunch of other good stuff in that PR, so this is a weird choice.

chilvence:
I still have a ton more to come, I am just breaking it into small parts instead of sending one huge PR.

Basically I did a load of work on the completely unimplemented stuff over the week I was on holiday in october, but the weapon proficiencies and the XP stuff from the last PR were the only things I didn't get time to scrutinise that deeply.

lynxlynxlynx:
Why not include the table and specifics fixes?

chilvence:
I can do, but they only fix part of the proficiency problem, and since I don't know quite yet what I'll end up with, I thought it better to roll them up with the next set of fixes that revolve around the same theme.

I wont be getting rid of any of that in any case, just putting it on ice for a few days.

PR #1041, opened 2020-11-08
New purple.mos image (yellow borders with blue gems in the corners)

m7600:
## Checklist

- [ ] purple.mos
- [ ] source/purple.bmp


lynxlynxlynx:
Thanks, waiting for the fixed scrollbars before merging. Reminds me we should add PNG support also for these.

lynxlynxlynx:
Actually it's already there, so please instead just submit a `purple.png` and there'll be no need for also something in source/.

m7600:
I added the purple.png file (I think, I'm not sure) but I can't delete the .bmp file. I'm doing this from the github website, but right now I'm installing Git Cola (I'm on Ubuntu, so no TortoiseGit).
Using github from the website is a bit of a nightmare, and there's a lot of things that I don't remember how to do.

I couldn't solve the transparency problem for the scrollbar, I left you a comment on Gibberlings.

lynxlynxlynx:
I can cherry-pick the commits, but if you manage to improve the workflow in the future, that's even better.

m7600:
Yeah, sorry about that lynx. I'll try to have a better workflow from now on. Here is the scrlbar1.bam. Do you need the source .bmp files for it as well?
The Gibberlings forum seems to be down at the moment.

m7600:
I deleted the purple.bmp file as you requested. Is this correct now, or did I do something incorrect with this pull request?

lynxlynxlynx:
I've cherry-picked the good stuff, don't worry. Looking into the transparency now. Thanks!

lynxlynxlynx:
Fixed. The transparency index was bad, since it actually wasn't the first color that was green (as displayed in NearInfinity, but DLTCEP did fine). Might have worked in the original, but we're stricter.

You'll notice there are a few artifacts in some of the frames, stray pixels. Feel free to clean those up.

lynxlynxlynx:
And congratulations for your first accepted contribution! 👍🏾

m7600:
Thanks for the kind words! I didn't know the engine had support for .png files. That's awesome! Much better to use those than .mos and .bmp

PR #1043, opened 2020-11-11
Demo: improved fog of war

m7600:
An improved fog of war for the demo. It's not perfect, but I think it works better than the current black blocks. Here's a [screenshot](https://raw.githubusercontent.com/m7600/GemRB-demo-extension/main/Screenshot%20from%202020-11-11%2000-26-21.png).


lynxlynxlynx:
Nice, thanks!

chilvence:
I'm really curious about this, would it be feasible to have some sort of alpha blending on the fog of war bam? It's practically the only feature I really miss from the EE versions of the game.

lynxlynxlynx:
It'd be easiest to just use an image format that has an alpha channel. With partial transparency you get blending for free and there's no need for dithering.

I don't remember if it was done on master or subviews, but we do support the alpha channel on BAMv1 bams already, at least partially.

chilvence:
Oh cool, I didn't even realise BAM files were capable of storing an alpha channel, I thought the game just used palette index transparency. Actually I just looked at https://gibberlings3.github.io/iesdp/file_formats/ie_formats/bam_v1.htm and it says they only added it support for it in EE .... were they planning to have it originally?

@m7600 does the tool you used to make this say anything about alpha channels?

m7600:
@chilvence I don't know. I used Bam Worshop 2 for the fog of war, but I'm not really an expert on that program, or on .bam files. What I did was I exported the fog of war from Icewind Dale 2 and I tried to make something similar to that (same number of frames, each frame pointing in the right direction, etc.), and I tried to make a different look for each of the frames (although there are similarities, there's not a whole lot you can do to make a 32 x 32 black-and-transparent fog of war frame look unique)

PR #1044, opened 2020-11-11
Demo: remove the old fogowar.bam

m7600:
Apparently, when I uploaded the new FOGOWAR.BAM for the demo yesterday, it didn't replace the old one, fogowar.bam. Not sure why the case sensitivity is making a difference here, but I propose to remove the old fogowar.bam


lynxlynxlynx:
Thanks, didn't notice it either.

PR #1046, opened 2020-11-11
Demo: fixed the clown colors

m7600:
I fixed the clown colors for the demo, there was a bad line in avatars.2da
But there's another problem now, the avatar looks completely yellow. Is this normal?


lynxlynxlynx:
This is wrong as explained, closing. I don't think we can fix this bug without a good look at the core code. Your effort is best spent on other parts of the dataset.

m7600:
But did you see the screenshot? The change definately did something, though surely not the desired result.

m7600:
You're right, my apologies. It was from changing the palette column like you said.

Issue #167, opened 2019-08-19
[bug] PST area animation clipping by wall groups


lynxlynxlynx:
See #163 for previous discussion. Neither the animation in question nor the wallgroup has any cover bits set. Should PST wallgroups cover by default?

AR0508 example:
![ar0508_103_set](https://user-images.githubusercontent.com/238558/63229656-28666380-c203-11e9-8949-d9cb5b7eeb62.jpg)

_Originally posted by @MarcelHB in https://github.com/gemrb/gemrb/issues/163#issuecomment-522350499_

lynxlynxlynx:
It is relatively simple to write an https://github.com/gemrb/iesh script to check all PST areas and see how many area animations and wall polygons overlap. This would make answering the question about WF_COVERANIMS default value almost trivial, as the impact of the change could be exhaustively checked.

lynxlynxlynx:
Added some notes on the WED format from bg2 (91b4964ffb) — perhaps it will be useful for pst as well.

anewusername:
I took a look at this -- not sure if I did it right, but it looks like roughly ~500 individual overlaps across 89 area files.
I just took the 0th frame of each BAM, formed a rectangle based on (animation's center + frame's xy) and frame's dimensions, and then checked for intersection with wall polygons.

In any case, it doesn't seem like forcing (or inverting) WF_COVERANIMS would fix this; I get the below effect in either case:
![4BciWQ](https://user-images.githubusercontent.com/12633865/95168790-5b6e0c80-07a1-11eb-980c-ee8e2c0a9b70.png)
(on subviews, I don't see any effect from WF_COVERANIMS)

lynxlynxlynx:
Thanks for investigating! Can you list some of the other combinations, just so we have more test cases?

And huh, that's an interesting result. Looks like the two bottom animations get misplaced (handlebar gap and bottom).

Just for reference, the wall polygons:
![image](https://user-images.githubusercontent.com/121515/95170122-3a62e700-07b4-11eb-916f-63cce8e65ec7.png)

bradallred:
> (on subviews, I don't see any effect from WF_COVERANIMS)

You need to be using SDL1 or SDL2 + GL, plain SDL 2 can't blend using r/g/b channels.

anewusername:
Looks (close to) correct with SDL2+GL and forced WF_COVERANIMS:
[...]

PR #1049, opened 2020-11-13
PST-specific animation- and wall-flag changes

anewusername:
## Description
Fix for #167 and some further progress on #163 .

Many bits/flags are still totally ignored -- notably A_ANI_BACKGROUND or whatever replaces it.
I saw improvements in most PST animations with these changes, but they could definitely use some more testing.

## Checklist

### DIFF
--- a/gemrb/plugins/AREImporter/AREImporter.cpp
+++ b/gemrb/plugins/AREImporter/AREImporter.cpp
@@ -1575,13 +1575,34 @@ void AREImporter::AdjustPSTFlags(AreaAnimation *areaAnim) {
 	 * break things (like stopping early, hiding under FoW).
 	 *
 	 * So far, a better approximation towards handling animations is:
-	 * - always set A_ANI_BLEND, A_ANI_SYNC,
-	 * - unset A_ANI_PLAYONCE, A_ANI_NOT_IN_FOG, A_ANI_BACKGROUND.
+	 * - zero everything
+	 * - always set A_ANI_SYNC
+	 * - copy/map known flags (A_ANI_ACTIVE, A_ANI_NO_WALL, A_ANI_BLEND)
+	 *
+	 * Note that WF_COVERANIMS is enabled by default for PST, so ANI_NO_WALL
+	 *   is important.
 	 *
 	 * The actual use of bits in PST map anims isn't fully solved here.
 	 */
-	areaAnim->Flags |= (A_ANI_BLEND | A_ANI_SYNC);
-	areaAnim->Flags &= ~(A_ANI_PLAYONCE | A_ANI_NOT_IN_FOG | A_ANI_BACKGROUND);
+
+	#define PST_ANI_NO_WALL 0x0008  // A_ANI_PLAYONCE in other games
+	#define PST_ANI_BLEND 0x0100  // A_ANI_BACKGROUND in other games
+
+	areaAnim->Flags = 0;             // Clear everything
+
+	// Set default-on flags (currently only A_ANI_SYNC)
+	areaAnim->Flags |= A_ANI_SYNC;
+
+	// Copy still-relevant A_ANI_* flags
+	areaAnim->Flags |= areaAnim->originalFlags & A_ANI_ACTIVE;
+
+	// Map known flags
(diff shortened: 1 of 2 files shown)


gemrb/plugins/WEDImporter/WEDImporter.cpp
```diff
@@ -372,6 +372,9 @@ void WEDImporter::ReadWallPolygons()
 			if (flags&WF_BASELINE) {
 				polygonTable[i]->SetBaseline(base0, base1);
 			}
+			if (core->HasFeature(GF_PST_STATE_FLAGS)) {
```
anewusername, line 375: Not sure if this is the correct way to check for PST here.

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -1575,13 +1575,33 @@ void AREImporter::AdjustPSTFlags(AreaAnimation *areaAnim) {
...
 	 *
 	 * The actual use of bits in PST map anims isn't fully solved here.
 	 */
-	areaAnim->Flags |= (A_ANI_BLEND | A_ANI_SYNC);
-	areaAnim->Flags &= ~(A_ANI_PLAYONCE | A_ANI_NOT_IN_FOG | A_ANI_BACKGROUND);
+
+	#define PST_ANI_ACTIVE 0x0001
+	#define PST_ANI_NO_WALL 0x0008
+	#define PST_ANI_BLEND 0x0100
+
```
anewusername, line 1591: Probably not the right location for the PST_ANI_* definitions, but I think it's clearer to have them here until the mapping is more complete/obvious.

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -1575,13 +1575,33 @@ void AREImporter::AdjustPSTFlags(AreaAnimation *areaAnim) {
...
+	 *   is important.
 	 *
 	 * The actual use of bits in PST map anims isn't fully solved here.
 	 */
-	areaAnim->Flags |= (A_ANI_BLEND | A_ANI_SYNC);
-	areaAnim->Flags &= ~(A_ANI_PLAYONCE | A_ANI_NOT_IN_FOG | A_ANI_BACKGROUND);
+
+	#define PST_ANI_ACTIVE 0x0001
+	#define PST_ANI_NO_WALL 0x0008
+	#define PST_ANI_BLEND 0x0100
```
lynxlynxlynx, line 1590: How about something like this:
```suggestion
	#define PST_ANI_NO_WALL 0x0008 // A_ANI_PLAYONCE in other games
	#define PST_ANI_BLEND 0x0100 // A_ANI_BACKGROUND in other games
```
Documents the mismatch, while PST_ANI_ACTIVE is the same as A_ANI_ACTIVE, so there's no need for it.

gemrb/plugins/WEDImporter/WEDImporter.cpp
lynxlynxlynx: Ideally it would be a new feature flag, but if we were strict, we'd have too many already. Another candidate is GF_AREA_OVERRIDE, but it's not any better, so let's just leave it.

lynxlynxlynx:
Thanks, looks promising!

lynxlynxlynx:
Was there any where you noticed a regression?

anewusername:
I didn't notice any regressions; most of the always-on animations look pretty good to me with this patch.
- The face on the wall in the alley (AR0405 I think) is broken, but I think that was true before, too. It shows up while moving but then disappears afterwards, and doesn't seem blended quite right while moving.
- The catacombs might have some broken animations, but I didn't notice any changes from the patch. Same for Undersigil.
- The foundry, print shop, art gallery, and brothel improved.
- I didn't see any issues in the fortress and mortuary.

chilvence:
The alley face animation problem is an old issue, but I'm not sure if it's on the tracker anywhere, I think it just got mentioned in another discussion. I think it is supposed to just play once and freeze on the last frame.

lynxlynxlynx:
Thanks!

Issue #313, opened 2019-10-07
[enhancement] flesh out the included demo


lynxlynxlynx:
This is basically a modding todo. We include a very basic demo with open-licensed data, but it's currently just an area with a simplistically animated character. More assets and use of features would be cool.

# Stage 1
This has already been achieved, a basic runnable demo. The only record we have is the [git history](https://github.com/gemrb/gemrb/commits/master?after=7306dbe9e4ef65b2b8e0bdd5d1cb92b9135b74bb+34&branch=master&path%5B%5D=demo).

# Stage 2

These things can be worked on right now and will gel together well. It's a TODO for the first area and some interface stuff.

Area 1 (grove exterior):
- add some area animations
  - [x] ~some small insect swarm somewhere~
    - [x] indicating a hidden cache - add container
- [x] add a rabbit or other small animal (or two) to area
  - [x] create animation (standing and walking is enough): can reuse [rabbit](https://opengameart.org/content/rabbit-0) or  [rat](https://opengameart.org/content/evil-giant-rat)
  - [x] script it to random walk and flee when approached (don't forget to add used actions and triggers to the ids files)
- [x] shorten shadows in the area background to about a quarter or less
  - [x] fix lightmap to match
- [x] add a forcefield "door", something blocking the entrance to the interior
  - [x] add actual movement blocking
- [x] Infopoints on the monoliths (old language / for deciphering, for riddle)
  - [x] decide if the area graphics need a touch up, so it's clear something is on them
- area script:
  - [x] advance time to night time after reentering and clear the weather if it doesn't phase out on its own
- add puzzle
  - [x] add NPC granting it
  - [x] design the riddles
  - [x] create dialog with all the puzzles and connective tissue
  - [x] fix monolith texts to make sense
  - [x] on final success, trigger a cutscene: rain, lightning, dispel the forcefield enabling access to inside
- create a story:
  - [x] design it
    - perhaps the player going to their village elders to learn of the feats of famous ancestors … one of which you then relive interactively in the demo
  - [x] write the texts
  - [x] displayed in chapter Textscreen for the intro setting of the scene
  - [x] provide motive to go to first area, solve the puzzle (and enter the interior)
  - [x] as a rite of passage, blocking entry into the next area + cutscene ... through one or more npc dialogs

Misc:
- [x] create a nice avatar for the pc
- [x] create nice fog of war
- [x] add ground pile drawing(s) for dropped loot

[...]

PR #1050, opened 2020-11-13
Demo: add a start window

m7600:
## Description
This is a modest start window for the demo. Here's a [screenshot](https://raw.githubusercontent.com/m7600/GemRB-demo-extension/main/Screenshot%20from%202020-11-13%2016-49-37.png). 

This pull requests includes:
A background (start.png)
Graphics for a New Game and Quit buttons, in a single file (bstart.bam)
An updated version of Start.py for functionality (It's probably not the most elegant Python code you've seen, but it works)
A new .chu file (guiconn.chu)

EDIT: Whoops, I forgot to list the .chu file. Give me a second... Done.


lynxlynxlynx:
Looks ok, thanks! If you want to improve it, the SetVisible call shouldn't be needed and LoadGame belongs in NewGamePress, which then also means the QuitGame call can be removed.

What did you have in mind for future work? Might be a good idea to sit on this for a while, so it's easier to work.

m7600:
Thanks lynx, I'll make those changes you mentioned. 

Future work: I'm not sure. There's a lot of things that I'd like to add to the demo. I admit that I'm more comfortable making art assets than writing code. I was thinking about making a portrait for the protagonist, and maybe add an NPC, a few enemies, and a new area. What do you think I should focus on?

chilvence:
Oh wow, if you reckon you can make some npc's and enemies, that's the start of a game right there! :)

I've always wanted to do something with this demo if I would ever be able to finish what I can on PST stuff - can't promise much, but can I support you in any way?

m7600:
@chilvence Thanks. Yeah, here's a modest [orc animation](https://raw.githubusercontent.com/m7600/GemRB-demo-extension/main/orc.gif) I made, it kinda sucks, but I can try to improve it. I made it with Blender and Gimp.

m7600:
BTW Planescape: Torment is my favorite game

chilvence:
That's a really nice start, model actually looks pretty good, I wouldn't say it sucks. If I were to criticise I would just say the animation could be more fluid.

I know hand animation is a really tricky skill though, cos I tried it once upon a time...

lynxlynxlynx:
Oh, you can also add another label and then `Label.SetText(GemRB.Version)`, which will do the obvious thing. Or change the existing ones to include the version.

The overall bug for this is #313. 

What I think is needed most, since it's probably the hardest to do, but you actually know how, is more character animations. Ideally we'd have at least a documented process if not an automated workflow to export from Blender (for all the required orientations, correct camera angle, scale and all stances) and perhaps a good starting project file for further creation. iesh can be used to automate stitching into bams. I don't know which animation type you used for the orc, but there are several choices.

Don't think about combat yet: that will require quite some work, since most of the data is stubbed. Neutral NPCs are fine though.

Btw, I've already committed the bg for the interior area, but so far I haven't managed to make the WED for it successfully, so that's why it's not in yet. Currently I have time just for quick diversions, but that should change in a few weeks. But even then, if I get some proper time, I try to work on polishing subviews, so it's best also for me if you focus on content. The windowing code will need to be updated after the merge anyway, so let's deal with new gui stuff last.

m7600:
@chilvence Thanks for the feedback, I agree

@lynxlynxlynx Ok, I'll make some character animations then. I'll have to make them pre-shaded until we can figure out the issue of the clown colors, or do you prefer that I give them clown colors? I can make some monsters like giant spiders, goblins, orcs, etc. For the automated part in Blender, I use a Python script that was written by Clint Bellanger in his FLARE project, here's [his tutorial with the Python code](http://clintbellanger.net/articles/isometric_tiles/). 
I'll take a look at the interior area, maybe I can make the WED file and make it work using a combination of DLTCEP and IE Tileset Map Editor.

m7600:
BTW lynx I looked at the TODO for the demo and it says that the searchmap should be improved, but I think that has already been fixed, right? I'm asking because the protagon doesn't seem to be able to walk through the pillars anymore.

chilvence:
Heartily agree on the need for automated workflow, some sort of glue between Blender and Iesh that anyone can use. I actually feel a sense of dread when thinking about making a character sprite, which is not an ideal feeling :)

I haven't used Blender for a long time, but from what I know of it, it's quite capable of doing the sprite exporting legwork. It's iesh that I have never really used, I always rely on NearInfinity to look at stuff

Shall I open a tracker issue to carry on this train of thought? I mean, I'm hearing the corn whisper "if you build it, they will come" right now...

lynxlynxlynx:
searchmap: yes, it's stale and should be removed from the list.

nice that that flare script is in python — makes it easier to fix (missing and excessive orientations) and integrate with our tools — iesh's bam_composer in particular. You write a simple project file, run it and it does the magic. Docs:
https://github.com/gemrb/iesh/blob/master/bam_composer
https://github.com/gemrb/iesh/blob/master/infinity/bam_composer.py#L19

chiv: for the workflow? Yes please and feel free to copy over the above.

chilvence:
Ok, I will condense the discussion on that so far and see if I can find anything else useful

lynxlynxlynx:
Oh yeah, forgot to answer one thing: monsters don't need to be clown coloured — they won't be customizable.

lynxlynxlynx:
I've added a new tag for demo/new game reports and filed a few more. You can click on the yellow button to get to the listing.

m7600:
I'll close this pull request for now, if that's ok with you.

m7600:
Testing the new cave area, I'll upload the files soon. 

Screenshot from 2020-11-15 18-18-38

lynxlynxlynx:
I've now added an intro window based on your work. One can skip it via cfg option if desired and it also sports a textedit for the character name.

PR #1066, opened 2020-11-15
LevelUp.py Character Generation label

chilvence:
A change from a few months back to clean up the BG2 level up window interferes with the other games, BG for example doesn't have this control, so the UI breaks down on the Level Up screen.


lynxlynxlynx:
Thanks

Issue #819, opened 2020-06-15
berzerker to mage dual class goes wrong 


lynxlynxlynx:
If you go through the process, you end up with a mage X/berzerker 1 instead of the other way around.

Don't know if this combo ever worked properly, but it should be for all fighter -> mage paths, while mage -> fighter probably works. This has all to do with how the levels and class names are stored in the data.

Nykakin:
Experienced the same issue after dualing kensai to a thief.

lynxlynxlynx:
So F/T. Looks like we have a problem with all the normal variants (only a fighter/thief class exists and not the reverse), so either we don't flip them around enough or too much. The relevant code starts in DCMainDonePress of DualClass.py.

chilvence:
I will put #1067 on pause and see if I can figure out what is going on here.

chilvence:
Testing save games in both GemRB and BG2 original, no matter whether I go from fighter to thief or thief to fighter, the resulting order of the level stats is the same, they are in FIGHTER_THIEF order. Same with Mage -> cleric, cleric -> mage, the levels are always stored according to the relevant multi class combination order, IE CLERIC_MAGE.

That appears to be the intention of this part:

https://github.com/gemrb/gemrb/blob/4ae154d6f1a534b6362536a5894729338ae0f414/gemrb/GUIScripts/DualClass.py#L230-L234

The correct 'original class' bit also seems to be set correctly, which is telling the game what the correct active class is, whether 1st or 2nd.

It appears to only break down when kits get involved. The GUICommon.IsDualSwap function is probably to blame -

https://github.com/gemrb/gemrb/blob/4ae154d6f1a534b6362536a5894729338ae0f414/gemrb/GUIScripts/GUICommon.py#L515-L523

Here's a debug output from my Kensai -> Thief attempt

```
[Python]: dualswap multiclass combo:  ['FIGHTER', 'THIEF']
[Python]: baseclass kit id: 3  #Kensai
[Python]: baseclass kitlist value id: 2 #Fighter
[Python]: old kit class:  PALADIN
```

The call here to GetClassRowName literally returns whatever the row of the [classes.2da](https://github.com/gemrb/gemrb/blob/master/gemrb/unhardcoded/bg2/classes.2da) table, rather than looking for the row with the correct original class ID. 

https://github.com/gemrb/gemrb/blob/4ae154d6f1a534b6362536a5894729338ae0f414/gemrb/GUIScripts/GUICommon.py#L407-L420

[...]

PR #1067, opened 2020-11-15
PST Level up Learn Spells support

chilvence:
This is nearly finished, but I need to make sure I haven't broken anything

What I have done is further split some of LevelUp.py into different functions, so that I can call them from pst/GUIREC

This allows learning of new spells for TNO, Dakkon and Grace, and as a side effect fixes the skill points given  on multiple level up

What I really need to test is Dual Class still works in BG/BG2, because I am not 100 sure I understand it (see comment)

### DIFF
--- a/gemrb/GUIScripts/LUCommon.py
+++ b/gemrb/GUIScripts/LUCommon.py
@@ -37,6 +37,57 @@ def GetNextLevelExp (Level, Class):
 	# if you change it, check that all callers can handle it
 	return 0
 
+def GetNextLevels (actor, Classes):
+	"""Returns the next level the PC will attain based on current XP"""
+
+	Level = [0]*3
+
+	# Used only for TNO, all other games divide XP/Number of classes
+	XPStat = [IE_XP, IE_XP_MAGE, IE_XP_THIEF]
+
+	# Dual class character only care about the first class
+	# However the inactive class is needed to handle reactivation
+	if (GUICommon.IsDualClassed(actor, 0)):
+		NumClasses = 1
+		if GUICommon.IsDualSwap(actor):
+			Level[1] = GemRB.GetPlayerStat (actor, IE_LEVEL)
+		else:
+			Level[1] = GemRB.GetPlayerStat (actor, IE_LEVEL2)
+	else:
+		NumClasses = len (Classes)
+
+	for i in range(NumClasses):
+		# Get the next level we will use to look up new stats from the 2da tables
+		if GUICommon.IsNamelessOne(actor):
+			#TNO is the only character in all games that tracks individual XP stats
+			Level[i] = GetNextLevelFromExp (GemRB.GetPlayerStat (actor, XPStat[i]), Classes[i])
+		else:
+			# In all other cases, XP cut exactly by number of classes
+			Level[i] = GetNextLevelFromExp (GemRB.GetPlayerStat (actor, IE_XP)/NumClasses, Classes[i])
+
+	return Level
+
+def GetLevelDiff (actor, Level):
+	"""Calculate the number of new levels based on current levels"""
+
+	LevelDiff = [0]*3
+	LevelStat = [IE_LEVEL, IE_LEVEL2, IE_LEVEL3]
+
+	# Dual Swap: since dual class characters are based on mc combinations
+	# the level stat used by the 'new' class must be the one that
+	# matches the multiclass combo. EG a Fighter that duals to a Mage
+	# will now advance the IE_LEVEL2 stat since he will be FIGHTER_MAGE
+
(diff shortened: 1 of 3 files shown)


gemrb/GUIScripts/LUCommon.py
```diff
@@ -37,6 +37,45 @@ def GetNextLevelExp (Level, Class):
...
+			Level[i] = GetNextLevelFromExp (GemRB.GetPlayerStat (actor, IE_XP)/NumClasses, Classes[i])
+
+	return Level
+
+def GetLevelDiff (actor, Level):
+	"""Calculate the number of new levels based on current levels"""
+
+	LevelDiff = [0]*3
+	LevelStat = [IE_LEVEL, IE_LEVEL2, IE_LEVEL3]
+
```
chilvence, line 66: Have I understood and commented this behaviour correctly? The Dualswap thing gives me a bit of a headache so I am not sure.

gemrb/GUIScripts/LUCommon.py
```diff
@@ -37,6 +37,45 @@ def GetNextLevelExp (Level, Class):
...
+	# Reverse the order of levels for dual class characters
+	# if their original class is the second class in a mc combination
+	# eg a FIGHTER that duals to MAGE will be a FIGHTER_MAGE, not a MAGE_FIGHTER
+	# so we find his Mage level in IE_LEVEL2
+	if (GUICommon.IsDualSwap(actor)):
+			LevelStat = [IE_LEVEL2, IE_LEVEL, IE_LEVEL3]
+
```
chilvence, line 73: Referring to this part here of course, I wasn't sure how the github comments would display...

gemrb/GUIScripts/LUCommon.py
lynxlynxlynx: Kinda, you just reversed the example.
a FIGHTER that duals to MAGE will have FIGHTER as the original class, so FIGHTER_MAGE fits fine. But the other way around it would require a swap.

chilvence: So, a mage that duals to fighter will keep his mage level in IE_LEVEL, and fighter level in IE_LEVEL2?

gemrb/GUIScripts/pst/GUIREC.py
```diff
@@ -848,11 +859,13 @@ def AcceptLevelUp():
 		GemRB.SetPlayerStat (pc, IE_LEVEL, GemRB.GetPlayerStat (pc, IE_LEVEL)+NumOfPrimLevUp)
 		if avatar_header['SecoLevel'] != 0:
 			GemRB.SetPlayerStat (pc, IE_LEVEL2, GemRB.GetPlayerStat (pc, IE_LEVEL2)+NumOfSecoLevUp)
-	
-	LUSkillsSelection.SkillsSave (pc)
 
-	LevelUpWindow.Close()
-	NewLife.OpenLUStatsWindow()
+	if LevelUpWindow:
```
lynxlynxlynx, line 871: Can this really be false?

gemrb/GUIScripts/LUCommon.py
lynxlynxlynx: no, the current active class' level is always in IE_LEVEL.

gemrb/GUIScripts/pst/GUIREC.py
chilvence: No, but it's just part of a work in progress merge - I'm inching toward fully eliminating the need for the separate PST level up logic, sometimes I shuffle things around on a branch so they are easier to compare side by side. It's done like that in LevelUp.py, but I guess it doesn't necessarily have to be there either.

gemrb/GUIScripts/LUCommon.py
chilvence: I think I know why I am getting confused...

mc

 I guess if you start with a class on the left of the multi-class list, once you dual class the one on the right will be your primary class.. then the game will know which one is active. The old class is stored in the MC bits part of the cre file, then internally the game figures out which MC combination it should be to decide what abilities you can use... does that all sound right?

I've tested a few now and it seems to work ok, I just wanted to make sure I was reading it right. I need to fix some bugs and test BG2/IWD before I'm happy though.

lynxlynxlynx: The game will know which one is active just by comparing levels. For DC chars it's always one or the other, regardless where in the process you are. "primary" is not a good word to use, since it's ambiguous.

When testing, do note that something is already broken for DCs: #819

chilvence: Ok, I've reworded the comment, hopefully it is a bit clearer now

Issue #1053, opened 2020-11-14
 PST verbal constants should only be audible


lynxlynxlynx:
They contain text not in tags, so now we print this garbage to the message window. Eg.:
```
@46009 = ~Nameless One Action 2 1~ [NAM032]
```

`Actor::VerbalConstant` should disable printing in PST. You can check for it with `core->HasFeature(GF_FLOATING_TEXT)`.

chilvence:
Hmm, I chased it to here, and disabling the first if statement stops it spamming the message window:

https://github.com/gemrb/gemrb/blob/bb2515f815f9905e96990dfe7eb326a47056b9ae/gemrb/core/GameScript/GSUtils.cpp#L503-L511

I think it's probably safe to assume that no text was ever destined to show up in the message window via this method, since it is always closed, does that seem fair?

lynxlynxlynx:
Ignoring DS_CONSOLE sounds like a good first attempt.

chilvence:
I think the flag is supposed to be GF_ONSCREEN_TEXT

Does this look alright? 

https://github.com/gemrb/gemrb/commit/8d54623c5a737210bfb99820a398c160b33d3f6c

lynxlynxlynx:
Yep. Sound is still ok, right?

chilvence:
Yes, selection feedback still works (so I assume everything else will)

PR #1068, opened 2020-11-15
GSUtils: Disable the printing of verbal constants to the console in PST

chilvence:
This is to fix the annoyance in #1053


lynxlynxlynx:
Thanks!

PR #1073, opened 2020-11-17
Dualclass kitindex

chilvence:
For #819

Although I've found other issues with dual class worth chasing while testing, this particular fix for DC from kitted classes seems to be ok, I tested with a kensai, berserker, swashbuckler, beastmaster, avenger and cleric of lathander to various different combinations, and levelled each one up with 250000 XP to be sure the correct class was being used.

## Checklist


lynxlynxlynx:
Thanks!

Issue #841, opened 2020-06-25
[subviews] Can pick up some of the holy PC items in PST


MarcelHB:
#### Bug description

There are some inventory observations in PST that do not match master and I think should be considered broken; when you try a bit, you can completley drop the following holy PC items, including the default weapons that appear when the upgrades are taken:

1. The Touch by FFG (see below)
2. Annah's knifes
3. Morte's teeth

What looks unaffected:
1. The sword of Dak'kon and the ranged weapon of Nordom
2. Minsc and Boo

#### Screenshots

![store](https://user-images.githubusercontent.com/238558/85788214-e6fce980-b72c-11ea-85a0-c7b9130767c4.gif)

bradallred:
Is this different from #804?

MarcelHB:
> Currently it is impossible to drag and drop anything, there may be other glitches too.

First the part sounds like the opposite of this issue, the latter one - sure. The comment of @chilvence to the suggested changes - no idea. Feel free to close if this is related.

chilvence:
I think this is a consequence of the fact that the script displays the fist weapon if no other weapon is equipped, but does not disable the drag handler (I have actually just commited a fix for that in my WIP

Incidentally, I finally realised why I couldn't get the item cycling working again, which means I have a chance to finish these pst GUIINV fixes now (for both branches)

MarcelHB:
Cool, then let's wait for the PR, then we'll check for closing this.

chilvence:
This branch should fix the dragging in general, but I am uncertain about a couple of things

https://github.com/chilvence/gemrb/tree/841-pst-fistweapons-subviews

bradallred:
> the passing of arguments from button controls got changed somewhere along the line?

I'm not sure if this is what you mean, but control callbacks on subviews can take 2 optional parameters. The first is the control that invoked the callback and the second is the "value" of that control.

chilvence:
That's what I thought, I just probably should have focused on that difference first before turning the scripts upside down, hah

Oh well, I fixed lots of other stuff :)

bradallred:
I don't understand the continued use of `GemRB.GetVar ("ItemButton")` that value should be passed directly to the callback

chilvence:
I was just using it as it was in place elsewhere, I didn't really want to start changing anything, just trying to get PST back in synch functionality wise with the other games.

[...]

PR #1095, opened 2020-12-08
PST: Enable item info display for the 'default' weapon

chilvence:
This should fix the item info display for PST's default weapon. (#841) It appears to work for all NPCs. However I am not sure how this is handled in the other games. Do they even have this functionality? Just opening as a draft PR in case it needs amending.

### DIFF
--- a/gemrb/GUIScripts/InventoryCommon.py
+++ b/gemrb/GUIScripts/InventoryCommon.py
@@ -418,17 +418,24 @@ def CycleDisplayItem(direction):
 def OpenItemInfoWindow (btn, slot):
 	pc = GemRB.GameGetSelectedPCSingle ()
 
-	slot_item = GemRB.GetSlotItem (pc, slot)
-	item = GemRB.GetItem (slot_item["ItemResRef"])
+	slotItem = GemRB.GetSlotItem (pc, slot)
+	slotType = GemRB.GetSlotType (slot, pc)
 
-	if TryAutoIdentification(pc, item, slot, slot_item, True):
+	# PST: if the slot is empty but is also the first quick weapon slot, display the info for the "default" weapon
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
In other games fists don't have any descriptions, so this should somehow be pst-only. That 53 is mighty ugly. How about using `GemRB.GetSlotType`? Looks like the effects column should be 2 to mark fists (see slottype.2da).

@bradallred looks like you didn't change all the users ...

bradallred:
I'm sorry, I don't know what you mean by "change all the users" or how it relates to this PR

lynxlynxlynx:
Look at the buildbot build error.

chilvence:
Ok, updated to be PST only. Those are the conditions that need to be true to trigger display of the default weapon in the inventory in the original also, it really does disappear and reappear depending on what you have equipped. Not sure if that is worth keeping that way however? Would it be more useful if it actually allowed you to select the melee weapon as well?

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -418,17 +418,23 @@ def CycleDisplayItem(direction):
...
 	pc = GemRB.GameGetSelectedPCSingle ()
 
-	slot_item = GemRB.GetSlotItem (pc, slot)
-	item = GemRB.GetItem (slot_item["ItemResRef"])
+	slotItem = GemRB.GetSlotItem (pc, slot)
+	slotType = GemRB.GetSlotType (slot, pc)
+
+	# PST: if the slot is empty but is also the first quick weapon slot, display the info for the "default" weapon
+	if GameCheck.IsPST() and slotItem is None and slotType["ID"] == 10 and GemRB.GetEquippedQuickSlot(pc) == 10:
+		slotItem = GemRB.GetSlotItem (pc, 0)
```
lynxlynxlynx, line 426: You can call DisplayItem here and return early, since none of the other code is relevant for these items.

lynxlynxlynx:
I'm happy with this, thanks!

PR #1113, opened 2020-12-22
Fix crash due to race condition in music playback

Northfear:
## Description
SDLAudio plugin was crashing sometimes during the change of audio tracks (at the end of the combat).
It was mostly caused by `MusicReader->read_samples`that was still executing in audio thread when `CreateStream` was called. Locking `CreateStream` with mutexes stopped the crashes, but introduced deadlocks, since this method can be called from both main and audio thread (with `GetMusicMgr()->PlayNext()` function).

Basically methods CreateStream, Stop, Play and ResetMusics can be called from both main and audio thread of audio plugin. The proposed change adds a bool parameter to them (and a few methods from musimporter), that tells the method if it should lock the audio thread with mutex (if it's called from the outside of the audio thread) or not (it it's called from the audio thread).

OpenAL plugin looked affected too, but detecting this bug on PC was probably much harder due to faster IO.

I've also removed mutexes in `Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel, int XPos, int YPos, unsigned int flags, unsigned int *length)` method of SDLAudio plugin. I don't think that they're needed, since it's only called from the main thread anyway and has nothing to do with music.

## Checklist


lynxlynxlynx:
Thanks, it looks fine to me. @bradallred and/or @MarcelHB, can you give it a scan as well?

bradallred:
I'm also skeptical this was needed by OpenAL. I've ran with Tsan enabled many times without it complaining about either deadlock potential or race conditions, but then again perhaps the use of `SDL_mutex` is opaque to it (but if it were I would expect complaints about data races). We really should be using `std::mutex` and `std::thread` now anyway, but thats another issue entirely.

I do not like adding a parameter to a public method for locking. Outside callers shouldn't need to know anything about locking at all. We should probably just make these `bool` parameter methods private and rename them then add a public implementation like this:
```
bool Play()
{
  if (SDL_TryLockMutex(mutex) == 0) {
    return PrivatePlay(false);
  }
  return false;
}
```

Alternatively, remove the mutex locking from the private methods and just put the mutex juggling in the hands of the private caller sites.

Obviously, this can fail, but that appears to have been the case all along since these return `bool`. The callers need to be aware of the possibility of failure and retry the next run though the run loop. If we have different kinds of failure then we should change the return type to `int` and designate return values for the callers to distinguish if they need to retry next frame, or something else. I would imagine failure should always mean to retry tho since this is a game engine.

There appear to still be some eyebrow raising concerns that existed before this patch. Take `OpenALAudioDriver::Play` as an example:

```
bool OpenALAudioDriver::Play()
{
	if (!MusicReader) return false;

	SDL_mutexP( musicMutex );
	if (!MusicPlaying)
		MusicPlaying = true;
	SDL_mutexV( musicMutex );

	return true;
}
```

This implementation is just using a mutex to set a boolean. If that were really all it needed to do then all we would need to do is make `MusicPlaying` atomic and drop the mutex. However, it is clear that `MusicReader` needs mutex protection, which it currently lacks.

It is nice to see those Vita hacks gone tho : )

Northfear:
>We should probably just make these bool parameter methods private and rename them then add a public implementation like this

SDL_TryLockMutex was my first thought, but it's SDL2 only. Not sure what to do with SDL1 then. Replace everything with std stuff altogether?

>Alternatively, remove the mutex locking from the private methods and just put the mutex juggling in the hands of the private caller sites.

I'm not sure if I understand this approach. Would be really grateful for some kind of explanation.

bradallred:
I didn't realize that `SDL_TryLockMutex` was SDL 2 only, but yes, that sounds like another reason to use `std::thread` and `std::mutex`

What I mean is, instead of this:
```
bool PlayImp(bool lock) {
  if (lock) {
    SDL_LockMutex(mutex);
  }
  ...
   if (lock) {
    SDL_UnLockMutex(mutex);
  }
}

bool Play()
{
  if (SDL_TryLockMutex(mutex) == 0) {
    return PlayImp(false);
  }
  return false;
}

void SomeOtherMethod() {
  ...
  PlayImp(true);
  ...
}
```

we can do this:
```
bool PlayImp() {
  // must lock mutex before calling this method
  ...
}

bool Play()
{
  if (SDL_TryLockMutex(mutex) == 0) {
    return PlayImp();
  }
  return false;
}

void SomeOtherMethod() {
  ...
  SDL_LockMutex(mutex);
  ...
  PlayImp(true);
  ...
  SDL_UnLockMutex(mutex);
  ...
}
```

this is often desirable because usually you need to mutex held for additional reasons in the calling method anyway

Northfear:
@bradallred the real problem is that all of the locking calls are done from the outside of the SDLAudio/OpenALAudio class (they're done in MUSImporter), so there's no methods that would call `PlayImp()` in our case.
Just using `SDL_TryLockMutex` w/o private method would probably stop crashes/deadlocks, but I suspect that it would introduce some bugs too. I'm honestly not sure how to deal with this one properly.

bradallred:
I'm not sure what you mean, but I've gone ahead and updated the audio drivers to use `std::thread` and `std::mutex` so `SDL_TryLockMutex` is no longer a concern.

lynxlynxlynx:
@Northfear: is the state now good for you?

Northfear:
@lynxlynxlynx std threads were messed up on Vita last month, so I kinda postphoned the whole thing for now (and it's really hard to test this properly on PC due to much faster IO)

Northfear:
Ok, std::threads seems to be working again and I'm back at it..

It looks like the real problem with `std::mutex::try_lock` is that
> If try_lock is called by a thread that already owns the mutex, the behavior is undefined. 

Which I guess is the part of the problem. Like I've said before `music_callback` (that's executed from audio thread) can call `SDLAudio::CreateStream`, which could lead to undefined behavior with try_lock in our case.
And missing mutex in either of those places could lead to crash in `MusicReader->read_samples`

Northfear:
Seems to be working fine so far. Looks like recursive_mutex lock in `CreateStream` fixed everything.

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -104,6 +104,13 @@ bool SDLAudio::Init(void)
 #endif
 		return false;
 	}
+
+	int result = Mix_AllocateChannels(MIXER_CHANNELS);
+	if( result < 0 ) {
```
lynxlynxlynx, line 109:
```suggestion
	if (result < 0) {
```

lynxlynxlynx:
That's great to hear, thanks for grinding it. @bradallred, do you want to take another look before I squash this in?

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -133,9 +143,9 @@ void SDLAudio::music_callback(void *udata, uint8_t *stream, int len)
 	int mixerLen = len;
 
 	SDLAudio *driver = (SDLAudio *)udata;
+	driver->MusicMutex.lock();
```
bradallred, line 146: why did you make this change? never yielding a lock in an infinite loop is a code smell at best.

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -340,32 +345,33 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
 
 int SDLAudio::CreateStream(Holder<SoundMgr> newMusic)
 {
+	std::lock_guard<std::recursive_mutex> l(MusicMutex);
```
bradallred, line 348: Adding this lock implies `MusicReader` needs to be protected by the lock, yet we don't do that everywhere.

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -340,32 +345,33 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
...
 {
-	// TODO
 	MusicPlaying = false;
 	Mix_HookMusic(NULL, NULL);
 	return true;
 }
 
 bool SDLAudio::Play()
 {
+	if (!MusicReader) {
```
bradallred, line 365: shouldn't this lock `MusicMutex`?

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -402,9 +408,9 @@ void SDLAudio::buffer_callback(void *udata, uint8_t *stream, int len)
 	int mixerLen = len;
 
 	SDLAudio *driver = (SDLAudio *)udata;
-	std::lock_guard<std::recursive_mutex> l(driver->OurMutex);
-
 	unsigned int remaining = len;
+	driver->MusicMutex.lock();
```
bradallred, line 412: please use a `lock_guard` instead of manually locking/unlocking so that changes to the code behave expectedly. there is just a lot of code between the 2.

bradallred (changes requested):

bradallred:
Can we get an explanation for the removal of `SDL_LockAudio` stuff? I don't know anything about SDL audio. is it just superfluous now?

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: But we're doing that in `music_callback`

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: I was like that before, I can change it. But isn't there a chance of `MusicReader` change during that `while(true)` check?

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: But there's no need to lock anything at `SetAudioStreamVolume` anymore. Or even at `memset` before that. Is it still better to use `lock_guard` in that case?

Northfear:
> Can we get an explanation for the removal of `SDL_LockAudio` stuff? I don't know anything about SDL audio. is it just superfluous now?

I've added it while trying to imitate `Mix_FreeChunk` that was kinda broken for me in the first place, but there's really no need for it, since removal of currently playing chunk is impossible anyway (`evictBuffer` and `Play` are in the same thread)

gemrb/plugins/SDLAudio/SDLAudio.cpp
bradallred: thats why `MusicReader` is _supposed_ to be protected by a mutex. I believe I noted this in my other comments.

gemrb/plugins/SDLAudio/SDLAudio.cpp
bradallred: right, but we need to hold the lock _every_ time we touch it and we are not doing that. So I'm not suggesting a change here, I'm suggesting that we add a lock to the other locations.

gemrb/plugins/SDLAudio/SDLAudio.cpp
bradallred: this is another one of those infinite loops where a `lock_guard` should be at the top of the loop instead of outside the loop.

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: Changed. Still not quite sure why it's better to keep the guard inside the loop. Isn't it slower, since it's locking/unlocking each iteration? And what about possible change of `MusicReader` during condition check on every iteration? That could mess up things too, since we'd want to continue reading stuff from the same buffer.
If locking/unlocking is a bad thing to do maybe it's better to place guard and while loop inside additional `{ }` code block?

bradallred: its not about being _slow_ or _fast_, it is about being _correct_. The reason why you practically never should lock a mutex outside of an infinite loop is you've now locked every other thread out of reading/writing whatever that mutex is protecting. If you are looping that _implies_ you expect some condition to change, but don't know when (eg you expect another thread to do something to change this state to break from the loop). Almost certainly this thing you expect to change _requires_ the very mutex you locked and now you either have a race condition or a deadlock.

Your concern about `MusicReader` changing is exactly why you should have the `lock_guard`; `MusicReader` _shouldn't_ be read/written by anything not owning the mutex. The other thread will block waiting to make the change while this thread holds the mutex. then the mutex becomes unlocked at the bottom of the loop and the other thread can grab it then this thread blocks at the top of the loop waiting for the mutex (eg the `MusicReader` change) to become available.

Northfear: >then the mutex becomes unlocked at the bottom of the loop and the other thread can grab it then this thread blocks at the top of the loop waiting for the mutex (eg the MusicReader change) to become available.

`music_callback` doesn't rely on any other thread for switching MusicReader (since it can do this itself or break the loop if there's no songs left to play), but yeah, it shouldn't really matter if it's finishing reading stuff from the next song on the same playlist or from a different one. So guard lock should work here fine indeed.

Lock in the `buffer_callback` should still be outside the loop tho. I think that `FreeBuffers()` can mess it up pretty badly if it's run on the loop condition check.

Also fixed one more bug while looking through the code.
`stream = stream + cnt;`
cnt is number of short samples and stream is  uint_8*. This resulted in empty ending and some overlapping of the buffer on song changes.

bradallred:
Thank you sir

PR #1143, opened 2021-01-22
Modernize code to use override (available since C++11)

fenuks:
## Description
Hello, I've been browsing issues list, and found information about modernizing code. I thought it would be fun to give clang-tidy a try.

This MR adds `override` support. I've disabled all other checks, and used only [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html). Worked remarkably well, there were few issues (`GLuint` was replaced with `uint`, and one file had dos endings, so I converted it to unix format). I checked all changes, and didn't notice further problems. It builds cleanly with gcc and clang. And since gemrb already requires C++11 compiler, I hope it builds without problem with MS compiler as well.

## Checklist

### DIFF
--- a/apple/AppleLogger.h
+++ b/apple/AppleLogger.h
@@ -26,11 +26,11 @@ namespace GemRB {
 class GEM_EXPORT AppleLogger : public StdioLogger {
 public:
 	AppleLogger();
-	virtual ~AppleLogger();
+	~AppleLogger() override;
 
-	void textcolor(log_color);
+	void textcolor(log_color) override;
 protected:
-	void LogInternal(log_level level, const char* owner, const char* message, log_color color);
+	void LogInternal(log_level level, const char* owner, const char* message, log_color color) override;
 };
 
 Logger* createAppleLogger();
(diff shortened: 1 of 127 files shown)


bradallred:
this is going to cause too many conflicts with the dev branch...

fenuks:
Not a problem. Perhaps it would be wiser to resubmit it again when it will be merged, or submit it against dev branch?

bradallred:
I wouldn't mind it being applied against the subviews branch if @lynxlynxlynx has no objections

lynxlynxlynx:
At this point, not really.

fenuks:
More fixes can be easily applied (NULL -> nullptr one is certainly valuable), but I think it's best to have each one in separate PR.

lynxlynxlynx:
That one will be orders of magnitude more noisy, so I'm not sold on it. Plus, it can be avoided in several cases.

fenuks:
I'll check, even if just out curiosity, how large changeset will be for this fix.

fenuks:
It touches 191 files.

About python style linter, it has hardcoded `master` branch, hence it lints all changes made in GUI scripts in `subviews` branch.

lynxlynxlynx:
It's not hardcoded, but it looks like github didn't pick up you changing the target branch. Still, might not work regardless, but it's also not important here.

fenuks:
You are right, and yes, it is irrelevant here.

bradallred:
thanks!

PR #1144, opened 2021-01-23
Handle inserting non-ASCII characters in text edit

### LINKED ISSUE #679: Non ASCII keystrokes have no effect
I'm happy to inform  that subviews version works ideally, in par with original engine. Non-ASCII letters cannot be inserted (Right ALT + l, for example in Polish keyboard layout), but original didn't allow to do that as well. Only difference is that in case of IE ‘plain’ ASCII letter is inserted, whereas in case of GemRB no letter is inserted at all.

_Originally posted by @fenuks in https://github.com/gemrb/gemrb/issues/663#issuecomment-612668426_

fenuks:
## Description
This fixes #679. According to SDL wiki [TextInputEvent](https://wiki.libsdl.org/SDL_TextInputEvent) is always encoded in UTF-8. Converting it to encoding used by game fixes problem. Note that convention is from string with multi byte characters, to one that has single-byte ones (at least in my case), which means not all characters can be represented… Without changing GemRB internals to use UTF-8 or equivalent, not much can be done. Such failures are handled gracefully, i.e. some gibberish is presented like before.

I also trimmed trailing whitespaces in files I edited, and one mistake I found by occasion.

## Checklist

### DIFF
--- a/gemrb/core/GUI/EventMgr.cpp
+++ b/gemrb/core/GUI/EventMgr.cpp
@@ -57,11 +57,11 @@ MouseEvent MouseEventFromTouch(const TouchEvent& te, bool down)
 MouseEvent MouseEventFromController(const ControllerEvent& ce, bool down)
 {
 	Point p = EventMgr::MousePos();
-	
+
 	MouseEvent me;
 	me.x = p.x;
 	me.y = p.y;
-	
+
 	if (ce.axis % 2) {
 		me.deltaX = ce.axisDelta;
 		me.deltaY = 0;
@@ -69,7 +69,7 @@ MouseEvent MouseEventFromController(const ControllerEvent& ce, bool down)
 		me.deltaX = 0;
 		me.deltaY = ce.axisDelta;
 	}
-	
+
 	EventButton btn = 0;
 	switch (ce.button) {
 		case CONTROLLER_BUTTON_A:
@@ -92,7 +92,7 @@ MouseEvent MouseEventFromController(const ControllerEvent& ce, bool down)
 KeyboardEvent KeyEventFromController(const ControllerEvent& ce)
 {
 	KeyboardEvent ke;
-	
+
 	// TODO: probably want more than the DPad
 	switch (ce.button) {
 		case CONTROLLER_BUTTON_DPAD_UP:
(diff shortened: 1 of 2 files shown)


bradallred:
Thanks. good catch on that typo!

fenuks:
Typo was tricky, I wish compiler could warn about it. As far I can see, compilers have no such warning, but [this](https://clang.llvm.org/extra/clang-tidy/checks/readability-implicit-bool-conversion.html) clang-tidy rule detects it (it needs to be enabled).

```cpp
int test() {
    return 1 + true;
}
```
```
$ clang-tidy --checks=readability-implicit-bool-conversion a.cpp
1 warning generated.
/home/fenuks/a.cpp:2:16: warning: implicit conversion bool -> 'int' [readability-implicit-bool-conversion]
    return 1 + true;
               ^~~~
               1
```
There are also few interesting flags that could be enabled for GCC (e.g. `-Wduplicated-branches`, it is not enabled by `-Wall -Wextra`). I'd like to give it a shot in the future.

lynxlynxlynx:
I'm pretty sure the static analyzers catch duplicate branches, since that's basically dead/unreachable code. If it's not unreliable (why isn't it in -Wextra?), we can add it to the cmake list via CHECK_CXX_COMPILER_FLAG.

fenuks:
I'll experiment with these flags later, and they will be useful, I'll open new PR. It is also worth considering using clang-tidy, if none of the services gemrb uses for static analysis doesn't have it baked in.

lynxlynxlynx:
Sonar does. :)

fenuks:
Sonarqube is pretty neat. ;) Certain checks (like this with implicit bool -> integer conversion) are not enabled by default, so possibly settings could be polished a bit. And for compilers, clang has `-Weverything` flag that turns all warnings (as opposed to `-Wall`), but it's very strict. Still, turning it on, and disabling by hand all warnings that are useless for the project could work. =)

lynxlynxlynx:
Maybe once we clear the easier stuff, as -Weverything is overkill now. Same for the implicit conversion warnings, if they can't be specified in great detail, since we use a lot of implicit conversion *to* bool.

Issue #1105, opened 2020-12-12
Defer logging I/O to a background thread


lynxlynxlynx:
It can be slow, so we don't want it to affect core performance.

bradallred:
It should be noted that our logging is not thread safe at all currently. Logging messages from the OpenAL thread definitely sets off ThreadSanitizer

PR #1155, opened 2021-02-16
Multithreaded Logging

bradallred:
## Description
Address #1105 along with additional enhancements. Logging is now thread safe and all I/O is handled on a separate thread.

This is big and changes Android and Windows which I have no ability to test. I'm hoping others will help do some testing there prior to merge.

## Checklist


MarcelHB:
I'll test Windows.

gemrb/core/System/Logger/Stdio.cpp
```diff
@@ -107,18 +102,12 @@ void StdioLogger::LogInternal(LogMessage&& msg)
 	print("\n");
 }
 
-void StdioLogger::destroy()
-{
-	textcolor(DEFAULT);
```
lynxlynxlynx, line 115: Is this handled somewhere else now? I didn't see it. It's was added to ensure we don't leave the vterm in the wrong color if we terminate or stop (gdb) at an inopportune time.

lynxlynxlynx:
Android changes look light, but we don't have a working build any way, so let's not worry about it here.

gemrb/core/System/Logger.h
```diff
@@ -27,11 +27,28 @@
...
+#include <thread>
 
 namespace GemRB {
 
-enum log_color {
+// !!! Keep this synchronized with GUIDefines !!!
+enum log_level : int {
+	INTERNAL = -1, // special value that can only be used by the logger itself. these messages cannot be supressed
+	FATAL = 0,
+	ERROR = 1,
```
MarcelHB, line 44: This seems to be conflicting with some `ERROR` enum value in `windows.h`. `enum class` to the rescue?

bradallred: if `enum class` does the job, sure. Else there is always scoping it. However, `GemRB::log_level::ERROR` is already distinct from `::ERROR`. What exactly is the error? A symbol ambiguity error?

gemrb/core/System/Logger/Stdio.cpp
bradallred: ah, good point, I'll add it to the destructor then

gemrb/core/System/Logger.h
MarcelHB:
```
D:/Sources/gemrb/gemrb/core/System/Logger.h:44:2: error: expected identifier before numeric constant
   44 |  ERROR = 1,
      |  ^~~~~
```
First thought was these C enum tokens overlap. But actually it looks like it's a macro from inside of the Windows headers that can be suppressed by `#define NOGDI` right before loading `windows.h`. I tested this and it works.

MarcelHB: Wait before applying this, there something else down the line in `Win32Console.cpp` and I'm not sure yet wheter this related or independent from this workaround. Will check later.

lynxlynxlynx: If that's problematic, an undef should work.

MarcelHB (commented):
Below, you'll find a patch that makes it compile and link on Windows/MSYS2. But there is something I don't have time today to dig deeper into it:

- Using `Win32Console` works good, at least I don't see anything wrong at first sight.
- But when opting for `StdioLogWriter`, the application hangs before emitting the first log statement, does not get to launch.

But since it's stdio, I can try again to debug under Linux tomorrow or so.

```diff
diff --git a/gemrb/core/System/Logger.h b/gemrb/core/System/Logger.h
index e575d691d..934c08dae 100644
--- a/gemrb/core/System/Logger.h
+++ b/gemrb/core/System/Logger.h
@@ -121,7 +121,7 @@ public:
 
 extern const char* log_level_text[];
 
-extern Logger::LogWriter* (*createDefaultLogWriter)();
+extern GEM_EXPORT Logger::LogWriter* (*createDefaultLogWriter)();
 
 }
 
diff --git a/gemrb/core/System/Logger/Win32Console.cpp b/gemrb/core/System/Logger/Win32Console.cpp
index e63d0bfb5..d3ca4430b 100644
--- a/gemrb/core/System/Logger/Win32Console.cpp
+++ b/gemrb/core/System/Logger/Win32Console.cpp
@@ -18,7 +18,6 @@
 
 #include "System/Logger/Win32Console.h"
 
-#include "System/Logging.h"
 
 #include <cstdio>
 
@@ -31,15 +30,19 @@ namespace GemRB {
 # define cprintf _cprintf
 #endif
 
-Win32ConsoleLogger::Win32ConsoleLogger(bool useColor)
-: StdioLogger(useColor)
+Win32ConsoleLogger::Win32ConsoleLogger(log_level level, bool useColor)
+: StdioLogWriter(level, useColor)
 {
 	hConsole = GetStdHandle( STD_OUTPUT_HANDLE );
 }
 
-void Win32ConsoleLogger::print(const char *message)
+void Win32ConsoleLogger::WriteLogMessage(const Logger::LogMessage& msg)
 {
-	cprintf("%s", message);
+	cprintf("[%s", msg.owner.c_str());
+	if (log_level_text[msg.level][0]) {
+		cprintf("/%s", log_level_text[msg.level]);
+	}
+	cprintf("]: %s\n", msg.message.c_str());
 }
 
 static int colors[] = {
@@ -68,9 +71,9 @@ void Win32ConsoleLogger::textcolor(log_color c)
 		SetConsoleTextAttribute(hConsole, colors[c]);
 }
 
-Logger::LogWriter* createWin32ConsoleLogger()
+Logger::LogWriter* createWin32ConsoleLogWriter()
 {
-	return new Win32ConsoleLogger(true);
+	return new Win32ConsoleLogger(DEBUG, true);
 }
 
 }
diff --git a/gemrb/core/System/Logger/Win32Console.h b/gemrb/core/System/Logger/Win32Console.h
index 279cf982f..8b23648a3 100644
--- a/gemrb/core/System/Logger/Win32Console.h
+++ b/gemrb/core/System/Logger/Win32Console.h
@@ -20,22 +20,23 @@
 #ifndef LOGGER_WIN32_H
 #define LOGGER_WIN32_H
 
+#include "System/Logging.h"
 #include "System/Logger/Stdio.h"
 
 namespace GemRB {
 
-class GEM_EXPORT Win32ConsoleLogger : public StdioLogger {
+class GEM_EXPORT Win32ConsoleLogger : public StdioLogWriter {
 public:
-	Win32ConsoleLogger(bool useColor);
-
-	void print(const char* message) override;
-	void textcolor(log_color);
+	Win32ConsoleLogger(log_level, bool useColor);
 
+	void WriteLogMessage(const Logger::LogMessage& msg) override;
 private:
 	HANDLE hConsole;
+
+	void textcolor(log_color) override;
 };
 
-Logger::LogWriter* createWin32ConsoleLogger();
+Logger::LogWriter* createWin32ConsoleLogWriter();
 
 }
 
diff --git a/gemrb/includes/win32def.h b/gemrb/includes/win32def.h
index eb2919d38..f961f9445 100644
--- a/gemrb/includes/win32def.h
+++ b/gemrb/includes/win32def.h
@@ -42,6 +42,9 @@
 #define UNICODE
 #define _UNICODE
 
+// Otherwise, MS claims the ERROR macro, conflicting with logging.
+#define NOGDI
+
 # include <windows.h>
 
 #else //WIN32
```

gemrb/core/System/FileStream.h
```diff
@@ -40,17 +40,64 @@ namespace GemRB {
...
+struct File {
+private:
+	FILE* file = nullptr;
+public:
+	File(FILE* f) : file(f) {}
+	File() = default;
+	File(const File&) = delete;
+	File(File&&) = default;
+	~File() {
+		fclose(file);
```
MarcelHB, line 52: On Linux, this gives me an exception when this is `nullptr`.

bradallred:
@MarcelHB I think if you do a review and propose changes that way they can be accepted and you can properly be attributed

MarcelHB:
Please, don't mind this. It's just that I have little time and mind to do this thoroughly at the moment, so the best I can do is throwing some code snippets from time to time.

bradallred:
no problem, just didn't want to take credit from ya 🤷 thanks for the patch.

lynxlynxlynx:
In such cases you can set the author manually when committing (`--author "bla bla <bla@bla.bla>"`).

MarcelHB:
@lynxlynxlynx Can you check this on Linux? I do not see any logging, but just some errors and a non-terminating process:
```
➜  gemrb git:(99c004fc3) ✗ ./gemrb -c GemRB_linux.cfg 
close failed in file object destructor:
sys.excepthook is missing
lost sys.stderr
```

lynxlynxlynx:
Something is wrong for sure, since the buildbots can't manage to run the minimal test; it just segfaults.

bradallred:
Have a stack trace for me?

bradallred:
is there anyway to get github/travis to spit out a backtrace when this happens?

MarcelHB:
The only thing I can provide is a an error when shutting down GemRB on Linux, as there is obviously some `join` that never returns.

bradallred:
> here is obviously some join that never returns
why do you think that is what is happening? It's not obvious to me

you can see on Travis it segfaults when running `gemrb/gemrb -c ../gemrb/GemRB.cfg.noinstall.sample`

MarcelHB:
I'll post it later, but 1. the process never exits for me and 2. when interrupting with GDB, it noticed the `join` is where it's stuck.

bradallred:
not exiting can be any deadlock, but I did notice we aren't notifying the condition_variable prior to the join so it is possible we are stuck on that sometimes. I've added that.

MarcelHB:
Aside from the compiler issue [mentioned above](https://github.com/gemrb/gemrb/pull/1155#discussion_r578559491), Linux now works me as expected.

I'll have a 2nd look at Windows soon.

MarcelHB:
Also, with the patch above, Windows now works with stdio output.

MarcelHB:
Ok, Windows works good now with stdio output too, but hangs when terminating on something what I suspect to be related to that cond. variable, for now I just see that:
```gdb
(gdb) t 1
[Switching to thread 1 (Thread 1168.0x15f0)]
#0  0x00007ff83d9ed174 in ntdll!ZwWaitForMultipleObjects () from C:\Windows\SYSTEM32\ntdll.dll
(gdb) bt
#0  0x00007ff83d9ed174 in ntdll!ZwWaitForMultipleObjects () from C:\Windows\SYSTEM32\ntdll.dll
#1  0x00007ff83b350d40 in WaitForMultipleObjectsEx () from C:\Windows\System32\KernelBase.dll
#2  0x00007ff83b350c3e in WaitForMultipleObjects () from C:\Windows\System32\KernelBase.dll
#3  0x00007ff830a51e48 in ?? () from C:\MSYS2\mingw64\bin\libwinpthread-1.dll
#4  0x00007ff830a52186 in ?? () from C:\MSYS2\mingw64\bin\libwinpthread-1.dll
#5  0x00007ff830a5220c in ?? () from C:\MSYS2\mingw64\bin\libwinpthread-1.dll
#6  0x00007ff823f665d9 in ?? () from C:\MSYS2\mingw64\bin\libstdc++-6.dll
#7  0x00007ffff17235e1 in GemRB::Logger::~Logger (this=0x1bc39c43f50, __in_chrg=<optimized out>)
    at D:/Sources/gemrb/gemrb/core/System/Logger.cpp:44
```

bradallred:
what is the trace for the logging thread? also might be more useful with a debug build with optimizations disabled.

MarcelHB:
I guess there is no such thread at this point, the other one in the list is probably related to gdb:
```
0x00007ff8a60f0111 in ntdll!DbgBreakPoint () from C:\Windows\SYSTEM32\ntdll.dll
(gdb) info threads
  Id   Target Id           Frame
  1    Thread 10096.0x2774 0x00007ff8a60ed174 in ntdll!ZwWaitForMultipleObjects () from C:\Windows\SYSTEM32\ntdll.dll
* 16   Thread 10096.0x27d4 0x00007ff8a60f0111 in ntdll!DbgBreakPoint () from C:\Windows\SYSTEM32\ntdll.dll
```
Also, no helpful backtrace:
```
(gdb) bt
#0  0x00007ff8a60f0111 in ntdll!DbgBreakPoint () from C:\Windows\SYSTEM32\ntdll.dll
#1  0x00007ff8a611c94e in ntdll!DbgUiRemoteBreakin () from C:\Windows\SYSTEM32\ntdll.dll
#2  0x00007ff8a4797034 in KERNEL32!BaseThreadInitThunk () from C:\Windows\System32\kernel32.dll
#3  0x00007ff8a609d241 in ntdll!RtlUserThreadStart () from C:\Windows\SYSTEM32\ntdll.dll
#4  0x0000000000000000 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

bradallred:
does it get more useful with a debug build and optimizations disabled? I'm not a Windows dev so I have no idea what to make of these symbols.

you can add a `return` on Logger.cpp:40 with a breakpoint there and see if the thread is terminating that way.

you can try adding this to the top of `~Logger` just to rule out the writer
:
```
std::lock_guard<std::mutex> l(writerLock);
writers.clear()
```

MarcelHB:
I tried it even harder by putting assertion failures at the end, but no: no GDB breaks, no auxiliary printings to see, and every thread termination is reported as a zero return code (but that does not necessarily reveal useful information as I read). I can even remove `running = false` so that there is something else that terminates these threads before `~Logger` is reached.

To me, it looks as if something on Windows (or libwinpthread) terminates the child threads for process cleanup reasons before the global destructor gets there, and then it smells like undefined when notifying into the void. I'll do some tests.

MarcelHB:
Heh, I guess it's exactly that. You just need to make sure to terminate when still somewhere inside of `main`, and voilà. There is probably some higher concept behind hidden in MSDN but in the meantime, this probably does the work as simple as it can be:

```diff
--- a/gemrb/GemRB.cpp
+++ b/gemrb/GemRB.cpp
@@ -151,6 +151,7 @@ int main(int argc, char* argv[])
 #endif
        core->Main();
        delete( core );
+       ToggleLogging(false);
 #ifdef VITA
        sceKernelExitProcess(0);
 #endif
```

bradallred:
while it is normal behavior on all platforms to terminate all child threads after returning from `main()` this shouldn't happen until after the static destructors run (which would have cleaned up our thread in the normal way). I wonder what odd thing Windows does. I've seen lots of examples of static function objects creating their own `std::thread`s and this doesnt seem to work on Windows if you can't guarantee those threads exit before the return from main it would seem. I have no idea what MS docs to look at, but I thought the C++ standard required static/global destructors to run as if they were called from the bottom of main right before the closing `}`

 The weirdest thing tho was that the stdio logger seemed to be ok. oh well 🤷

bradallred:
ah, the "[problem](https://stackoverflow.com/questions/42535990/worker-threads-are-forcibly-terminated-prior-to-the-destruction-of-static-object)" is that this thread comes from a DLL. interesting. I'm surprised that MS doesnt do the DLL unloading very last.

it does seem then that we should be sure to turn logging off before main returns. thanks for finding that!

PR #1158, opened 2021-02-17
TextArea conflict with windows winuser.h

chilvence:
I was just trying to get up to date again and found another thing that Visual studio does not like

It appears that COLOR_BACKGROUND is already defined in winuser.h as 1 , so the the enumeration here in COLOR_TYPE https://github.com/gemrb/gemrb/blob/d8822e12c2f264fbd6fb5462bac459535b006547/gemrb/core/GUI/TextArea.h#L166-L172
fails because it inherits the number from that header instead of making a new one when trying to build I guess.

I just came up with a random alternative from ye trusty thesaurus, because I actually don't know which use case this color type is intended for and was hastily trying to make it build - please disregard this PR, if you have a more preferable label for it. I just wanted to highlight the build issue that's all.

Other than this , no other build issues encountered with Visual Studio from latest Subviews branch. Noice. 


bradallred:
did d7c9379 fix it?

chilvence:
Yep, all good, closing

PR #1159, opened 2021-02-18
SDLAudio volume control

Northfear:
## Description
Working volume control for SDLAudio plugin. All volume sliders seems to be working now (except ambient, since it's not implemented anyway). Music/Movie volume control is a bit hacky (since `Mix_VolumeMusic` is not supported due to custom music playback function), but functional.

## Checklist

### DIFF
--- a/gemrb/plugins/SDLAudio/SDLAudio.cpp
+++ b/gemrb/plugins/SDLAudio/SDLAudio.cpp
@@ -34,6 +34,18 @@
 
 using namespace GemRB;
 
+static void SetChannelPosition(int listenerXPos, int listenerYPos, int XPos, int YPos, int channel)
+{
+	int x = listenerXPos - XPos;
+	int y = listenerYPos - YPos;
+	int16_t angle = atan2(y, x) * 180 / M_PI - 90;
+	if (angle < 0) {
+		angle += 360;
+	}
+	uint8_t distance = std::min(static_cast<int32_t>(sqrt(x * x + y * y) / AUDIO_DISTANCE_ROLLOFF_MOD), 255);
+	Mix_SetPosition(channel, angle, distance);
+}
+
 void SDLAudioSoundHandle::SetPos(int XPos, int YPos)
 {
 	if (sndRelative)
@@ -42,15 +54,7 @@ void SDLAudioSoundHandle::SetPos(int XPos, int YPos)
 	int listenerXPos = 0;
 	int listenerYPos = 0;
 	core->GetAudioDrv()->GetListenerPos(listenerXPos, listenerYPos);
-	
-	int x = listenerXPos - XPos;
-	int y = listenerYPos - YPos;
-	double angle = atan2(y, x);
-	if (angle < 0) {
-		angle += 360;
-	}
-	uint8_t distance = std::min(static_cast<int32_t>(sqrt(x * x + y * y) / AUDIO_DISTANCE_ROLLOFF_MOD), 255);
-	Mix_SetPosition(chunkChannel, angle * 180 / M_PI - 90, distance);
+	SetChannelPosition(listenerXPos, listenerYPos, XPos, YPos, chunkChannel);
 }
 
 bool SDLAudioSoundHandle::Playing()
@@ -122,7 +126,7 @@ void SDLAudio::music_callback(void *udata, uint8_t *stream, int len)
 	ieDword volume = 100;
 	core->GetDictionary()->Lookup("Volume Music", volume);
 
-	// No point of bothering if it's off anyway. Hope nothing breaks..
+	// No point of bothering if it's off anyway
 	if (volume == 0) {
 		return;
 	}
@@ -329,17 +333,8 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
 		return Holder<SoundHandle>();
 	}
 
-	// Positional audio
-	if (!(flags & GEM_SND_RELATIVE))
-	{
-		int x = listenerPos.x - XPos;
-		int y = listenerPos.y - YPos;
-		double angle = atan2(y, x);
-		if (angle < 0) {
-			angle += 360;
-		}
-		uint8_t distance = std::min(static_cast<int32_t>(sqrt(x * x + y * y) / AUDIO_DISTANCE_ROLLOFF_MOD), 255);
-		Mix_SetPosition(chan, angle * 180 / M_PI - 90, distance);
+	if (!(flags & GEM_SND_RELATIVE)) {
+		SetChannelPosition(listenerPos.x, listenerPos.y, XPos, YPos, chan);
 	}
 
 	return new SDLAudioSoundHandle(chunk, chan, flags & GEM_SND_RELATIVE);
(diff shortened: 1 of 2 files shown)


gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -73,7 +73,28 @@ bool SDLAudio::Init(void)
...
 }
 
-void SDLAudio::music_callback(void *udata, unsigned short *stream, int len) {
+void SDLAudio::SetAudioStreamVolume(uint8_t *stream, int len, int volume)
+{
+	uint8_t *mixData = new uint8_t[len];
+	memcpy(mixData, stream, len * sizeof(uint8_t));
+	memset(stream, 0, len); // mix audio data against silence
+	SDL_MixAudio(static_cast<uint8_t*>(stream), static_cast<const uint8_t*>(mixData), len, volume);
+	delete[] mixData;
```
bradallred, line 82: this all seems useless to me. am I misunderstanding?

looks like...
1. create a new buffer
2. copy existing stream into said buffer
3. set contents of current buffer to 0
4. mix new buffer into old buffer

can you really not do `SDL_MixAudio(stream, stream, len, volume);`

at the very lease those `static_cast`s are useless

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -73,7 +73,28 @@ bool SDLAudio::Init(void)
...
+}
+
+void SDLAudio::music_callback(void *udata, uint8_t *stream, int len)
+{
+	ieDword volume = 100;
+	core->GetDictionary()->Lookup("Volume Music", volume);
+
+	// No point of bothering if it's off anyway. Hope nothing breaks..
+	if (volume == 0) {
+		return;
```
bradallred, line 92: isn't this just going to leave the stream filled with whatever was in it before? I guess it doesn't matter?

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: `SDL_MixAudio` is probably the only way of setting volume here that I know of
And to control the balance is a right way we have to mix against silence (mixing silence against existing stream results in the same stream, since silence * volume is well.. silence.
Changing `stream` pointer results in crash, so creation of empty mixData and doing some std::swap is not an option either.
I'm honestly not sure if it can be done any differently (except from modifying stream yourself in some way?).

static_casts are useless leftovers indeed tho. Removing them..

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: SDL_mixer 2 resets it to internally 0 before music callback.
SDL_mixer 1.2 doesn't, but it seems to be working fine, so my guess is that it's taken care of somewhere in SDL too.

gemrb/plugins/SDLAudio/SDLAudio.cpp
bradallred: sounds good, thanks

gemrb/plugins/SDLAudio/SDLAudio.cpp
bradallred: cool

Northfear:
Added basic positional audio support too

gemrb/plugins/SDLAudio/SDLAudio.h
```diff
@@ -79,7 +80,7 @@ class SDLAudio : public Audio {
 	void clearBufferCache();
 	Mix_Chunk* loadSound(const char *ResRef, unsigned int &time_length);
 
-	int XPos, YPos;
+	int listenerXPos, listenerYPos;
```
lynxlynxlynx, line 83: You can init them here instead, now that we're on c++11.

bradallred: also, since you're changing it anyway, maybe make it a `Point` instead

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -252,22 +283,29 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
...
+	if (!(flags & GEM_SND_RELATIVE))
+	{
+		int x = listenerXPos - XPos;
+		int y = listenerYPos - YPos;
+		int16_t angle = atan2(y, x) * 180 / M_PI - 90;
+		if (angle < 0) {
+			angle += 360;
+		}
+		uint8_t distance = std::min(static_cast<int32_t>(sqrt(x * x + y * y) / AUDIO_DISTANCE_ROLLOFF_MOD), 255);
+		Mix_SetPosition(chan, angle, distance);
```
bradallred, line 306: I believe all our math is in radians elsewhere. It would be nice to keep it so and converting to degrees when passing into `Mix_SetPosition`

Northfear:
Added some kind of basic `SDLAudioSoundHandle` too. Seems to be working, but if I got anything wrong - let me know

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: Done I guess

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -252,24 +319,30 @@ Holder<SoundHandle> SDLAudio::Play(const char* ResRef, unsigned int channel,
...
+	{
+		int x = listenerPos.x - XPos;
+		int y = listenerPos.y - YPos;
+		double angle = atan2(y, x);
+		if (angle < 0) {
+			angle += 360;
+		}
+		uint8_t distance = std::min(static_cast<int32_t>(sqrt(x * x + y * y) / AUDIO_DISTANCE_ROLLOFF_MOD), 255);
+		Mix_SetPosition(chan, angle * 180 / M_PI - 90, distance);
+	}
```
lynxlynxlynx, line 343: This chunk is the same as in SDLAudioSoundHandle::SetPos, so I suggest you extract it to a common private method.

Northfear: These are two different classes tho. Should I just make some separate function in the file or some public static method in SDLAudio?

lynxlynxlynx: File static sounds fine, since you only have a few params.

Northfear: Done

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: Nope. Rolled it back to angles. Mix_SetPosition has 90 degrees shift vs atan2, so we have to fix negative values after converting it to angles. It would be just cleaner too keep in degrees here

gemrb/plugins/SDLAudio/SDLAudio.cpp
lynxlynxlynx: Thanks! Is this PR done or are you still working on something (like GEM_SND_VOL_AMBIENTS)?

Northfear: I think it's done. No plans for GEM_SND_VOL_AMBIENTS for now

lynxlynxlynx:
Thanks!

PR #1161, opened 2021-02-20
wrap win32def.h in a Platform.h

bradallred:
## Description
trying to unravel the tangled web of includes/dependancies

Provide `Platform.h` as an umbrella file for defining platform specifics which should greatly reduce the `#ifdef` clutter surrounding windows.h or unistd.h

## Checklist

### DIFF
--- a/gemrb/includes/Platform.h
+++ b/gemrb/includes/Platform.h
@@ -1,5 +1,5 @@
 /* GemRB - Infinity Engine Emulator
- * Copyright (C) 2004 The GemRB Project
+ * Copyright (C) 2021 The GemRB Project
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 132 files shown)


bradallred:
hmmm I dont understand the source of these redefinition errors. We have the `#define` wrapped in `#ifndef`

gemrb/includes/Platform.h
```diff
@@ -0,0 +1,50 @@
+/* GemRB - Infinity Engine Emulator
+ * Copyright (C) 2003 The GemRB Project
```
lynxlynxlynx, line 2:
```suggestion
 * Copyright (C) 2021 The GemRB Project
```

gemrb/includes/Platform.h
```diff
@@ -0,0 +1,50 @@
...
+	#endif
+#endif
+
+#ifdef WIN32
+	#include "win32def.h"
+#elif defined(HAVE_UNISTD_H)
+	#include <unistd.h>
+#endif
+
+#include <cstdio>
```
lynxlynxlynx, line 44: Nice, no more including VFS.h everywhere. Once we fully move to nullptr, cstdio can be inlined as well. :)

lynxlynxlynx:
The redefinition problem shows something else is wrong. The ie_types ones are there only as a fallback. Since you removed the  include of config.h in it, the order is bad.

bradallred:
ok, that seems to be it AFICT. master has unresolved symbol errors for Windows that somebody needs to figure out. hard to say if there is a Windows problem on this branch without getting past that.

gemrb/includes/win32def.h
```diff
@@ -18,51 +18,22 @@
...
-
-#ifdef WIN32
-# define WIN32_LEAN_AND_MEAN
-
-# ifndef NOMINMAX
-# define NOMINMAX
-# endif
-
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
```
MarcelHB, line 27: Well, we need this conditionally, otherwise MSYS2 complains:
```
C:/MSYS2/mingw64/include/c++/10.2.0/x86_64-w64-mingw32/bits/os_defines.h:45: note: this is the location of the previous definition
   45 | #define NOMINMAX 1
```

MarcelHB (commented):
Other than the issue mentioned, I currently observe errors on a missing `consoleapi.h` but this is obviously not related to this PR. Will check tomorrow.

lynxlynxlynx:
Maybe it wants an explicit destructor. Sonar complains about undefined behaviour in the parent destructor:
https://sonarcloud.io/project/issues?id=gemrb_gemrb&issues=AXe_XyYoAP2uxvofo8JN&open=AXe_XyYoAP2uxvofo8JN

bradallred:
C++ has never required a destructor to be defined except for the specifying a `virtual` one for a base class, which we have. Not saying that adding one wouldn't fix it, but it's certainly not a standard requirement.

The Sonar warning is because we are calling a virtual method in a destructor which will not be able to be virtually dispatched. meaning it will always call the `StdioLogWriter` version which of course we don't want for the windows logger, but that wouldn't be a compile error, that would just be gibberish on the windows console and not reset the colors. So yes, something to address, but it is well defined C++ code.

bradallred:
looks like that finally satisfied the windows build so ill go ahead and merge this. @MarcelHB would you mind opening an issue or a PR for the build issues you have for mingw?

MarcelHB:
Positive, will do soon.

PR #1171, opened 2021-03-06
add proper handling for SIGINT

bradallred:
otherwise the signal is handled on an arbitrary thread and calls std::terminate which isn't always a clean exit

### DIFF
--- a/gemrb/GemRB.cpp
+++ b/gemrb/GemRB.cpp
@@ -25,14 +25,18 @@
 #include "Interface.h"
 #include "System/Logger/Stdio.h"
 
-#ifdef HAVE_MALLOC_H
-#include <malloc.h>
-#endif
+#include <csignal>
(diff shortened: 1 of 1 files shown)


bradallred:
actually its not clear to me if on Linux ctrl+c just sends `SDL_QUIT`. if it does we don't actually need this.

MarcelHB:
From some research: [Yes, it does](https://discourse.libsdl.org/t/ctrl-c-to-quit-program-doesnt-work/12002/6), and does your proposal [by default](https://wiki.libsdl.org/SDL_HINT_NO_SIGNAL_HANDLERS).

Doing it like this, the process is not terminating here because `join` is waiting indefinitely without sending an additional Ctrl+C.

bradallred:
awesome, I'll close this then. thanks!

PR #1172, opened 2021-03-07
Win32Console: fixes building on MSYS2

MarcelHB:
As mentioned in #1161, I cannot build this with MSYS2 WinAPI headers as there is no discrete `consoleapi.h`. The general `windows.h` does the job, and all the exotic color stuff isn't there anymore anyway.

## Checklist

### DIFF
--- a/platforms/windows/Win32Console.h
+++ b/platforms/windows/Win32Console.h
@@ -22,7 +22,7 @@
 
 #include "System/Logger/Stdio.h"
 
-#include <consoleapi.h>
+#include <windows.h>
 
 namespace GemRB {
 


bradallred:
I'm curious why our AppVeyor build works. I know nothing about windows development.

The colors should still work too according to the [MS documentation](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) ansi color codes are supported. Can you figure out where we are going wrong? I notice we supply the `esc` character in octal instead of hexadecimal, but that should be encoded by the compiler the same.

MarcelHB:
No, it's fine. I see colors in the native Windows shells, it's just that at some point in recent time, before I noticed your latest refactoring, there was more trouble related to some Console API calls that I remember were related to color settings, but this is gone now.

bradallred:
thanks!

Issue #1175, opened 2021-03-11
[demo] Protagonist looks like he moves sideways when he walks north


m7600:
#### Bug description
When the protagon moves north, there is a problem with the animation, it looks like he is moving to a side. Here's a screenshot to show what I mean

#### Screenshots
![Screenshot from 2021-03-11 09-20-51](https://user-images.githubusercontent.com/56371872/110787277-ea4cec80-824b-11eb-89ce-582ab5e7b7ac.png)

lynxlynxlynx:
Interesting, all the upward directions have some problems, while the bottom ones don't. The bams have all the needed orientations too.

Ah, they're set up wrong for the animation type they use:
https://gibberlings3.github.io/iesdp/appendices/avatar_naming/iwdstyle/wk.htm
Eg. the 4th cycle is pointing west, not north.

It should follow the "BG2/IWD style" one, documented here, otherwise less orientations are available:
https://gibberlings3.github.io/iesdp/appendices/avatarnaming.htm

m7600:
Thanks for the links. I'll try to fix the bams for the protagon and the rabbit.

lynxlynxlynx:
I think it's easiest if you look at eg. CHMF from bg2. Perhaps it will be enough to rename and set the TYPE in avatars.2da to 0.

m7600:
Here's a simple test. Included in the .zip are one .bam file (for the walking animation) and the avatars.2da file. 

The original name for the walk animation was ZHMW1WK, I had to replace the last part with G11, so now it's called ZHMW1G11. This fixes the problem, but only for that .bam file. I don't know what other names the rest of the animations should have (I don't have the original BG2, I have the EE version, so there are animations that Near Infinity doesn't show me)

[protagon_walk.zip](https://github.com/gemrb/gemrb/files/6124983/protagon_walk.zip)

lynxlynxlynx:
Here are all the suffixes, since G ones aren't detailed on IESDP:
https://github.com/gemrb/gemrb/blob/2515887fcdbd6929ff3ef8903f5e209fdbcfb33d/gemrb/core/CharAnimations.cpp#L1906

You can see g11 is for walking, g17 for normal standing and so on.

m7600:
For the standing animation (SD, or G17), I can't see it, and I get an error about cycles 63-65:

```
[CharAnimations/ERROR]: Couldn't load animation: zhmw1g17, cycle 63
[...]

PR #1176, opened 2021-03-12
[demo] Fixed the orientations for the protagon animation

m7600:
## Description
Fixed the orientations for two of the protagon's animations (standing and walking), see #1175 . 
Now, to be perfectly honest, I have no idea why I had to put more than 60 empty cycles in the .bam file for the walking animation. But, if that's what [the line in the cpp file](https://github.com/gemrb/gemrb/blob/2515887fcdbd6929ff3ef8903f5e209fdbcfb33d/gemrb/core/CharAnimations.cpp#L1938) says that we have to do, then that's what we have to do.


lynxlynxlynx:
Thanks!

Issue #431, opened 2019-11-25
[bug] BG2 GUICG4.py: Faulty max stats for dwarves and halflings


caillean7:
**Description**
Dwarves can have a maximum dex of 18 on GemRB while it should be capped at 17. Also, according to the [BG2 Wiki](https://baldursgate.fandom.com/wiki/Halfling) Halflings should get -1 to wis besides +1 dex and -1 str (which works). All other races are fine, I compared them to the stat modifiers listed on the Wiki.

**GemRB version**
0.86-git

**Expected behavior**
- max dex for dwarves: 17
- max wis for halflings: 17

lynxlynxlynx:
This is all in:
    https://gibberlings3.github.io/iesdp/files/2da/2da_tob/abracerq.htm
    https://gibberlings3.github.io/iesdp/files/2da/2da_tob/abracead.htm

Which we already read and use, so it's a bug in gemrb/GUIScripts/bg2/GUICG4.py .

PR #1213, opened 2021-04-01
GUICG:Fixed faulty stats calculation of min and max values

Alan1616:
abracerq.2da values for races should always be treated as base for calculation.
abclasrq.2da should be last thing to to consider as it is required boundary.

## Description
see #431 
It looks like there still might be an issue with Abclsmod it probably shouldn't impact min values, at least it look like it works that way in originals.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg2/GUICG4.py
+++ b/gemrb/GUIScripts/bg2/GUICG4.py
@@ -49,30 +49,30 @@ def CalcLimits(Abidx):
 	Minimum = 3
 	Maximum = 18
 
-	tmp = Abclasrq.GetValue(KitIndex, Abidx)
-	if tmp!=0 and tmp>Minimum:
-		Minimum = tmp
-
 	Race = Abracerq.GetRowIndex(RaceName)
 	tmp = Abracerq.GetValue(Race, Abidx*2)
-	if tmp!=0 and tmp>Minimum:
+	if tmp != 0:
 		Minimum = tmp
 
 	tmp = Abracerq.GetValue(Race, Abidx*2+1)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Looks good, thanks! Nice first patch. :)

If you want to verify your abclsmod hypothesis, I suggest you edit the values in the table(s) and then it should be obvious how the original behaves.

gemrb/GUIScripts/bg2/GUICG4.py
```diff
@@ -49,30 +49,30 @@ def CalcLimits(Abidx):
...
+	if tmp != 0:
 		Minimum = tmp
 
 	tmp = Abracerq.GetValue(Race, Abidx*2+1)
-	if tmp!=0 and tmp>Maximum:
+	if tmp != 0:
 		Maximum = tmp
 
+	Add = Abracead.GetValue(Race, Abidx)
 	Race = Abracead.GetRowIndex(RaceName)
```
fizzet, line 62: These two lines look like they might need to be swapped?

lynxlynxlynx: The two tables have the same rows, so in practice there's no difference. But better be consistent, I agree.

Alan1616:
After some tweaking of Abclsmod table it looks like it works this way in original though it obviously it's hard to check how this table affects the dice roll it look like it does so there's that Add modified at the end.

lynxlynxlynx:
Thanks! If you have the other games as well, this bug is likely also present in bg1 (GUICG4.py) and iwd1 (CharGen.py).

Issue #1215, opened 2021-04-02
Need for alternative to existing GetTickCount


MarcelHB:
#### Bug description
For as long as I've been compiling GemRB on Windows, there was always one mystery for me: The game does not _feel_ running smoothly, regardless of the SDL backend settings. And despite seeing 30FPS, it never was the case on Linux -- regardless how cheap the CPU/GPU. There it's how it is supposed to feel.

Over the time, I ocassionally debugged into SDL, did profiling but nothing could reveal what's wrong. Unless the whole thing runs in some different time-continuum (like showing 30FPS where seconds are not what we know to be seconds), on Windows at least.

Then [I got a suspicion](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount):

> The resolution of the GetTickCount function is limited to the resolution of the system timer, which is typically in the range of 10 milliseconds to 16 milliseconds. 

From what I see, this can end up in all kind of aliasing problems when sampling the time on Windows, including wobbly FPS capping and not-so-smooth animations.

So in contrast to the non-Windows implementation of `GetTickCount` in GemRB that gives 1ms precision, Windows needs something [else](https://github.com/libsdl-org/SDL/blob/bd06538778102f72bad8393ef07da5a1ec444217/src/timer/windows/SDL_systimer.c#L129-L136). 

I did at quick implementation and that's it -- a whole new experience of playing GemRB on Windows (it's like getting a new, higher Hz display). But there are some things to address first:

- [SDL_GetTicks](https://wiki.libsdl.org/SDL_GetTicks) picks what's best and available on a per-platform basis, including fallbacks but it also requires changes to the linking since it would require SDL to be linked against the core.
- We can copy something like this on our own but this requires some refactoring, since BIK/MVE player also need a counter with the surrounding tick variables (currently inlinable code from `globals.h`).

What's your preference? Better addressing this in subviews, or after that?

bradallred:
this is easy now that we require C++11. just use `std::chrono::high_resolution_clock `

MarcelHB:
Yeah, good idea.

bradallred:
it's possibly important to provide a fallback to `std::chrono::steady_clock` if `std::chrono::high_resolution_clock::is_steady` is for some reason `false`. I can't imagine that they are different clocks under the hood tho, but esoteric platforms exist. we could always `assert(std::chrono::high_resolution_clock::is_steady )` and address it if anybody complains.

[...]

PR #1217, opened 2021-04-02
Obtain ticks from an std::chrono implementation

MarcelHB:
See #1215

I even found a good example that makes the difference at the same FPS indication:

before:
fps2
after:
fps

## Checklist


MarcelHB:
Checking Linux ...

lynxlynxlynx:
LGTM. Also checked and the types are big enough to avoid 2038 problems. :)

bradallred:
fantastic. I didn't realize GetTickCount was a win32 api function.

Issue #1210, opened 2021-03-27
demo: showcase more subsystems / add more data


lynxlynxlynx:
This is a continuation of #313, where stage 1 and 2 were laid out and implemented. To keep it more manageable, I've split it apart. This process will probably need to be repeated as we figure out everything that needs to be done — the previous list grew and changed organically as features were implemented and better understood.

# Stage 3

PVRZ support  #1099:
- [ ] change placeholder forcefield "door" bam to [this beauty](https://opengameart.org/content/animated-portal-or-wormhole-several-variants) or a non-dithered version of the current one
  - [ ] add a "dispel" sort of visual to the disabling cutscene
- [ ] add puzzle step completion effects (eg. buzzing light over related monolith)

Area 2 (interior):
- [ ] **TODO: design the content, expand this list**
- [ ] create mini-map (MOS image) by scaling the base map down by 10.6666666666667
- [ ] add entrance and fix the travel region from outside to point to it
- [ ] add travel region to outside, another entrance to the cave mouth if needed and connect the two
- [ ] add to worldmap (use index 1 for the icon)
- [ ] add general area music (main ambient); ~perhaps [cave ambiance](https://opengameart.org/content/fantasy-sound-effects-library) or [this](https://opengameart.org/content/trembling-mists)~ already selected from a studio album
  - [ ] maybe a subambient in one of the caverns?
- [ ] add overground container / chest, [eg.](https://opengameart.org/content/treasure-pack)
- [ ] on exit: `SetGlobal("ReturnedOutside","GLOBAL",1)`

# Stage X - for later / to be fleshed out

Not fleshed out yet and out of scope until most of Stage 2 is done.

Area 1 (exterior):
- [ ] night wed?
- [ ] night area animations
- [ ] night music

GUI:
- World map travelling
- Add tooltips to relevant elements
  - [x] tooltip background image
-  [ ] Stats window ala simple GUIREC (just add to the designated area in the inventory window)
-  [ ] journal window
  -  [ ] add journal entries for events in the demo
- [ ] Saving and loading?
- [ ] item pickup and drop sound (copy over itemsnd.2da and amend it) once we have more items
- cursors:
  - [x] open/close door
- [ ] character generation?

**TODO: COMBAT**
- [ ] set up protagonist's stats
- [ ] add player portrait to indicate hp/damage
- [x] attack cursor
- [ ] add some traps for boss fight in Area 2?
- [ ] add injured status stuff, so tooltips will display it
[...]

Issue #1070, opened 2020-11-15
Demo: connect ar0100 and ar0110


m7600:
![Screenshot from 2020-11-15 18-18-38](https://user-images.githubusercontent.com/56371872/99199395-8757b700-277d-11eb-939c-e03caaae9b41.png)

I've got both areas working, and I can travel between them using the console command MoveToArea("").
But when I try to move the protagonist from one area to another using point-and-click, I get the following error:

```
[Map/ERROR]: Invalid entrance 'Trav0110' for area ar0110
[Map/WARNING]: WARNING!!! EntryPoint 'Trav0110' does not exist and direction 0 is invalid
[GameScript/ERROR]: Invalid scripting action: LeaveArea("ar0110",[2560.1920],-1)
[Scriptable/WARNING]: AA: NULL action encountered for protagon!
```

lynxlynxlynx:
First, there's a shortcut you can use: mta("ar0110")

You need to add this to gemact.ids:
93 LeaveAreaName(O:Target*)

As far as the entrance goes, it looks like you didn't set it up properly.

lynxlynxlynx:
If you send me the files over, I can take a look.
I also mentioned the wrong action, since I apparently can't read:
91 LeaveArea(S:Area*,P:Point*,I:Face*Dir)

m7600:
That fixed it! Thanks! I get a few warnings but it works, the character moves to the interior of the cave (though not to a correct location, but to a black area, because I didn't configure it properly). [Here are the files](https://github.com/m7600/GemRB-demo-extension/tree/main/areas), but I haven't made the search map (or the light map) for the new area yet.

m7600:
You probably don't need the .slh files, but I uploaded everything just in case.

lynxlynxlynx:
Thanks, though in that case I'll look at it later, once we're done with the exterior.

PR #1218, opened 2021-04-02
[demo] Add a minimap for the cave area 

m7600:
This commit adds a minimap for the demo's cave area, see #1210. The .mos file was made with DLTCEP.


lynxlynxlynx:
Thanks! This supplements the files in #1070. Can you make the searchmap as well (I did it with DLTCEP)? Then we have all the basics for adding the area.

m7600:
I'm trying to make it following [this tutorial](http://www.simpilot.net/~sc/dltcep/areamaking/main.htm#overlay), but something goes wrong and I don't know how to fix it. I'll attach the screenshot showing the problem. I'll try doing this with IETME instead of DLTCEP.

Screenshot from 2021-04-03 17-07-43

m7600:
Nevermind, I found out what the problem was, the frame count was wrong. OK, I'll keep using DLTCEP for this.

lynxlynxlynx:
You can look at the one of the outside area, to get the correct number for walls (I don't remember off the top of my head).

m7600:
Yes, I'm trying to set it up for the flood fill now. Also known as "Left-click Hell".

Screenshot from 2021-04-03 17-22-50

lynxlynxlynx:
That doesn't look that good though. Near the bookcase, there's too much ground marked as 4 (assuming it means wall) and the same for the opposite walls that are "tilted" away from the viewer — only its base needs to be inaccessible, the hidden part marked as a wall. The bookcases will need the same approach.

Some things are easier to fixed before the fill, since the bg is more visible, so it's worth it to make the outlines good from the start.

m7600:
4 represents the floor in this case. 0 would be the impassable sections. It's possible to toggle between them in DLTCEP. But yeah, I'll try to make it as clean as possible. One thing that I'm not entirely convinced of is the speed of this workflow. It may be faster with IETME, I'll have to do some tests.

Screenshot from 2021-04-03 20-10-11

Screenshot from 2021-04-03 20-09-56

m7600:
Ok, I updated the pull request. DLTCEP crashed while I was using it, but I managed to make the search map before it crashed. It's a black and blue .bmp file. It's not perfect, but it's a start.

lynxlynxlynx:
Cool, thanks! Can you teleport to the area? I get crashes, looks like because the searchmap and tilemap (TIS) sizes are different. Please attach the latest TIS and WED.

m7600:
When I use mta("ar0110"), I'm getting some serious crashes here as well, and possibly also a memory leak, since my computer starts processing like crazy and everything freezes, I have to actually press the power off button to stop it. I tried this with the latest TIS and WED files, also with the old ones, and I even made everything from scratch again with IETME, it seems to be the same crash in each case. This may have something to do with subviews. I've updated the PR with the new files that you requested.

lynxlynxlynx:
Thanks, I'll check them out. In the meantime I've added an assert, so the crash will always happen at the same place and not write to random memory.

lynxlynxlynx:
Your files worked fine for me, coupled with my previous changes. :) Added a bunch of other basic stuff. Now you can travel inside and out or cheat your way in by rightclicking on the map icon.

m7600:
Awesome, thanks! I just tested it. The search map definitely needs some refinement, but it's good to see that the cave area is now part of the demo.

PR #1248, opened 2021-05-05
Don't stop movement after setting a new path

Northfear:
## Description
Prevents actors from stopping after setting a new path (they we're stopping in place with walking active animation).

## Checklist


lynxlynxlynx:
Thanks!

Issue #232, opened 2019-09-10
[enhancement] Convert `Point` from short -> int32_t


bradallred:
There are several issues stemming from the use of the `short` type in `Point`.

1. incompatible with `Region` and `Size` (they use `int32_t`)
2. cannot cast to `SDL_Point` which is a performance bottleneck in some places (issue bradallred#123)
3. IWD2 (possibly EE) uses 4 bytes for them
4. PST Journal must be truncated due to overflow
5. some controls `WorldMapControl`, `ProgressBar`, possibly others are encoded in the CHU with them

there is also a lot of code that could be cleaned up where we want to assign a point/size to a `Region`, but cannot because the use of short prevents defining `Region` in terms of a `Point` + `Size`

please wait till after #95 to avoid conflicts and possible subtle bugs on merge

bradallred:
note to future self: search for explicit c style casts to `(short)`

bradallred:
while we are at it and having to change/scrutinize the construction sites... we may as well make `Point` conform to `std::is_trivial` as it is not much extra work on top of this.

kmfrick:
I'm not sure if this is the right place to point this out, but there are two different kinds of `Point`: those using searchmap coordinates (where pathfinding is done) and those using navmap coordinates (where e.g. scriptable positions are memorized). 
The searchmap is a downscaled version of the navmap (16x on the x axis and 12x on the y axis), which makes sense for a pathfinder and doubly so with the new one I've been working on for too long which decouples pathfinding and actual movement.
However, while working on said pathfinder I often inadvertently passed navmap points to functions expecting searchmap points and the reverse, causing much woe. 

I added a couple typedefs and used tons of Hungarian notation in my code to make it easier to understand, but while I'm at it, post-merge of my PR, I could split Navmap and Searchmap points and add automatic conversions or something like that (Navmap to Searchmap is `x/16` and `y/12`, while Searchmap to Navmap is `x * 16 + 8` and `y * 12 + 6` because the searchmap points map to centers of the navmap). I know which functions expect what pretty much by heart now and it would remove quite a lot of `q.x = p.x / 16 ...` and so on.

lynxlynxlynx:
Besides the increase in memory use, I don't see a connection.

kmfrick:
Okay, disregard the post then, we'll discuss this after I'm done with that pathfinder

kmfrick:
[...]

PR #1264, opened 2021-06-04
convert Point to a pair of ints

bradallred:
## Description
This is to address the issues in #232

## Checklist


gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -2079,164 +2079,100 @@ Trigger *GenerateTriggerCore(const char *src, const char *str, int trIndex, int
...
+	if (strnicmp( newVarName, "LOCALS", 6) == 0) {
 		Sender->locals->SetAt( poi, value, NoCreate );
 		return;
 	}
 	Game *game = core->GetGame();
-	if (HasKaputz && !stricmp(newVarName,"KAPUTZ") ) {
+	if (HasKaputz && !strnicmp(newVarName, "KAPUTZ", 6)) {
+		// FIXME?: NoCreate used to ony be passed when Context was null
+		// It was changed during consolidation because I couldnt think of why that difference was purposeful
+		// it felt like a bug, but thats only a hunch
```
lynxlynxlynx, line 2111: It's fine, you can remove the fixme. PST doesn't want any new variables and if we've been lucky not to hit it here, it doesn't matter — let's be consistent.

lynxlynxlynx:
Since it's a bit bigger and all over the place, can we wait with merging for right after 0.9.0?

bradallred:
certainly, I submitted as a PR precisely because I expect something, somewhere, is subtly broken.

Issue #1282, opened 2021-06-12
[demo] Add the PortraitWindow 


m7600:
I'm trying to add the PortraitWindow to the demo. I modified guiw.chu with DLTCEP (added a window and a control button). Here is what I added to MessageWindow.py:

```
	PortraitWindow = GemRB.LoadWindow(10, GUICommon.GetWindowPack(), WINDOW_RIGHT)
	PortraitWindow.AddAlias("PORTWIN")
	
	PortraitButton = MessageWindow.GetControl (24)
```
	
But when I run the demo, only the windows are visible, the player screen is completely black, as you can see in the attached screenshot. In the terminal, I get this error:

```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUICommonWindows.py", line 1721, in UpdatePortraitWindow
[Python/ERROR]:     Button.SetAction(lambda btn, val, pc=pcID: GemRB.GameControlLocateActor(pc), IE_ACT_MOUSE_ENTER);
[Python/ERROR]: AttributeError: 'NoneType' object has no attribute 'SetAction'
```

I'm obviously doing something wrong.

Here's a screenshot (I need to fine-tune the position of the portrait and the size of the column, it's a work in progress for now).

![Screenshot from 2021-06-12 02-10-22](https://user-images.githubusercontent.com/56371872/121765826-064e7b80-cb24-11eb-9ada-af1e7819e138.png)

lynxlynxlynx:
It just expects a bunch of extra controls, since the common code to run portrait windows detected the window is now present. In this case it expects 6 slots for party members.

m7600:
I added this to MessageWindow.py:

```
	PortraitWindow = GemRB.LoadWindow(10, GUICommon.GetWindowPack(), WINDOW_RIGHT)
	PortraitWindow.AddAlias("PORTWIN")
	
	PortraitButton = MessageWindow.GetControl (24)
	PortraitButton = MessageWindow.GetControl (25)
	PortraitButton = MessageWindow.GetControl (26)
	PortraitButton = MessageWindow.GetControl (27)
	PortraitButton = MessageWindow.GetControl (28)
	PortraitButton = MessageWindow.GetControl (29)
```
What else should I do?

lynxlynxlynx:
The CHU file needs the buttons — did you add them there?

m7600:
[...]

PR #1287, opened 2021-06-13
[demo] Add the PortraitWindow

m7600:
This commit adds the PortraitWindow to the demo, please see #1282 

Screenshot from 2021-06-13 16-47-57


gemrb/GUIScripts/demo/MessageWindow.py
```diff
@@ -43,6 +43,16 @@ def OnLoad():
...
+	
+	PortraitWindow = GemRB.LoadWindow(10, GUICommon.GetWindowPack(), WINDOW_RIGHT)
+	PortraitWindow.AddAlias("PORTWIN")
+	
+	PortraitButton = MessageWindow.GetControl (1)
+	PortraitButton = MessageWindow.GetControl (2)
+	PortraitButton = MessageWindow.GetControl (3)
+	PortraitButton = MessageWindow.GetControl (4)
+	PortraitButton = MessageWindow.GetControl (5)
+	PortraitButton = MessageWindow.GetControl (6)
```
lynxlynxlynx, line 55: This is useless.

lynxlynxlynx:
Please move the portrait controls, so it will look horizontally centered.

It also takes away even more game space. Maybe it would be better if we added it over the right bottom side, which is full of placeholder buttons anyway. And using much smaller portraits, since they don't do anything anyway.

m7600:
Ok, will do. If we're going to move the placeholder buttons, would it be a good idea to add the OptionsWindow now for them? To the left side of the screen.

lynxlynxlynx:
No, that would just take more space. I kinda like the pst/iwd2 approach here. We can move them in the place for dialog buttons and just hide them when in dialog. Then the now free grid space can be for the portraits

m7600:
Ok. Then we don't really need a new a PortraitWindow either, just a new button for MessageWindow with the player's potrait. Right? Or an I missing something here?

lynxlynxlynx:
Yeah, the code expects it, so if we provide it, things will just work, otherwise tons of hacks in shared code would be required, which would be bad.

There's no problem with overlapping, though we can also cut the region out from the messagewindow.

lynxlynxlynx:
No, I meant the buttons for the map etc., the fake options window. Then the portraits can be bigger (as if there were only max 4) in their place.

m7600:
Ok, but then where would the buttons go (map, inventory, center)?

m7600:
@lynxlynxlynx  is this what you mean? Are the buttons for the map, inventory. etc. ok where I placed them at the bottom left?
Screenshot from 2021-06-14 09-21-06

lynxlynxlynx:
Pretty much yeah. :) The portrait buttons can now be bigger. But also the messagewindow needs to be shortened and is bg cut, so things will work as expected on higher resolutions — where the windows otherwise wouldn't overlap perfectly.

PR #1297, opened 2021-06-18
change Point struct to use ints instead of shorts

bradallred:
## Description
to resolve issues in #232

## Checklist


bradallred:
I assume there are no objections to this being merged now...

gemrb/core/GameScript/Actions.cpp
```diff
@@ -1471,7 +1471,7 @@ void GameScript::MoveToCenterOfScreen(Scriptable* Sender, Action* /*parameters*/
 	}
 	Region vp = core->GetGameControl()->Viewport();
 	Actor* actor = ( Actor* ) Sender;
-	Point p((vp.x+vp.w/2), (vp.y+vp.h/2) );
+	Point p(vp.x + vp.w / 2, vp.y + vp.h / 2);
```
lynxlynxlynx, line 1474: This could be simplified further via Center()

lynxlynxlynx:
I suggest you add an IsInvalid() to Point as well, as manually comparing against a global InvalidPoint looks very odd.

lynxlynxlynx:
LGTM! Not merging in case you're still working on something.

Issue #1314, opened 2021-06-23
Python error when loading a PST save game


MarcelHB:
When loading a save game in PST, a Python error appears. Caused by this line, I guess: https://github.com/gemrb/gemrb/blob/d6459e7fc81107ff850ab691e75eaeb06e1d5b87/gemrb/plugins/GUIScript/GUIScript.cpp#L3886

It's not immediately breaking anything (although there is something else, possibly unrelated happening later in that turn, I'll report in a minute), but we probably don't want it either:
```py
[GUIScript/ERROR]: Runtime Error:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts\pst/Game.py", line 31, in EnterGame
[Python/ERROR]:     MessageWindow.OnLoad()
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts\pst/MessageWindow.py", line 63, in OnLoad
[Python/ERROR]:     PortraitWindow = GUICommonWindows.OpenPortraitWindow (1, WINDOW_BOTTOM|WINDOW_HCENTER)
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts/GUICommonWindows.py", line 1673, in OpenPortraitWindow
[Python/ERROR]:     UpdatePortraitWindow ()
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts/GUICommonWindows.py", line 1700, in UpdatePortraitWindow
[Python/ERROR]:     UpdateAnimatedPortrait(Window, i)
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts/GUICommonWindows.py", line 1790, in UpdateAnimatedPortrait
[Python/ERROR]:     Button.SetAnimation ("")
[Python/ERROR]:   File "D:\Sources\gemrb\gemrb\GUIScripts/MetaClasses.py", line 61, in handler
[Python/ERROR]:     raise type(e)(str(e) + "\nMethod Docs:\n" + str(f.__doc__))
[Python/ERROR]: RuntimeError: af cannot be null.
````

MarcelHB:
Well, the segfault following a bit later looks like related fallout:
```gdb
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffc95fa9498 in GemRB::GlobalTimer::AddAnimation (this=0x18b78357d10,
[...]

PR #1316, opened 2021-06-24
Fixes to GUI animations

MarcelHB:
## Description

See two issues mentioned in #1314.

## Checklist

### DIFF
--- a/gemrb/core/GlobalTimer.cpp
+++ b/gemrb/core/GlobalTimer.cpp
@@ -29,6 +29,7 @@
 namespace GemRB {
 
 GlobalTimer::GlobalTimer(void)
+  : animations(animationComparator)
 {
 	//AI_UPDATE_TIME: how many AI updates in a second
 	interval = ( 1000 / AI_UPDATE_TIME );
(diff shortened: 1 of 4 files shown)


gemrb/core/GlobalTimer.cpp
```diff
@@ -249,58 +250,32 @@ void GlobalTimer::SetFadeFromColor(tick_t Count, unsigned short factor)
...
-			anim->UpdateAnimation(paused, thisTime);
-			first_animation++;
-			continue;
-		}
-		break;
+	for (auto it = animations.begin(); it != animations.end() && (*it)->Time() <= thisTime;) {
+		auto animation = *it;
+		it = animations.erase(it);
+		animation->UpdateAnimation(paused, thisTime);
+		animations.insert(animation);
```
bradallred, line 272: we should check `animation->HasEnded()` before reinserting

or perhaps make `UpdateAnimation` return a bool indicating it

gemrb/core/GlobalTimer.cpp
```diff
@@ -249,58 +250,32 @@ void GlobalTimer::SetFadeFromColor(tick_t Count, unsigned short factor)
...
+	animations.insert(anim);
 }
 
 void GlobalTimer::RemoveAnimation(SpriteAnimation* anim)
 {
-	// Animation refs for given control are not physically removed,
-	// but just marked by erasing ptr to the control. They will be
-	// collected when they get to the front of the vector
-	for (auto it = animations.begin() + first_animation; it != animations.end (); ++it) {
+	for (auto it = animations.lower_bound(anim); it != animations.end() && (*it)->Time() == anim->Time(); ++it) {
```
bradallred, line 258: I'm not sure we should be checking the time here. shouldnt we remove the animation regardless of the time?

bradallred (changes requested):

gemrb/core/GlobalTimer.cpp
MarcelHB: Looking up by time first brings us to the matching slice of the set in logarithmic time. Otherwise, we to potentially iterate through the whole container to find the pointer -- under the assumption that there will never be a discrepancy in the set's tree and the actual time in the struct.

bradallred: sure, you look up using `lower_bound` thats fine, I was talking about the `(*it)->Time() == anim->Time();` part. Is it valid to have the same animation in the collection multiple times? that seems like it would cause mysterious bugs with animations being advanced multiple times a frame, or out of sync with their internal timing.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -3877,7 +3877,12 @@ static PyObject* GemRB_Button_SetAnimation(PyObject* self, PyObject* args)
 
 	Button* btn = GetView<Button>(self);
 	ABORT_IF_NULL(btn);
-	
+
+	if (ResRef[0] == 0) {
```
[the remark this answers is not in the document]
MarcelHB: It's still a `char *`, not `ResRef` type.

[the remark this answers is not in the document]
bradallred: its fine then

gemrb/core/GlobalTimer.cpp
MarcelHB: I just wanted to make sure not to exceed the loop if `lower_bound` brings us above the timestamp we are looking for, because something isn't just in there.

But no, the same animation in terms of its location is not supposed to happe.

bradallred: that makes sense

bradallred:
thanks!

PR #1331, opened 2021-06-28
CMake: swab for MSYS2

MarcelHB:
## Description
The swab check does not work (like this) for Windows/MinGW: it's just not found then. And even if, there is still the problem of its use in `DataStream` violating the assumed call signature. So I see these options:

1. Solely relying on an own implementation that does not conflict with anything.
2. Doing it like in this PR (inspired from the VLC code).

## Checklist


bradallred:
how much of this is still needed after 8d3753b?

MarcelHB:
Still nothing addresses this case (we have `swab` but want `char*`, not `void*` -- both MSVC and MinGW expect that):
```
D:/Sources/gemrb/gemrb/core/System/swab.h:28:17: error: conflicting declaration of C function 'void swab(const void*, void*, ssize_t)'
   28 | GEM_EXPORT void swab(const void *bfrom, void *bto, ssize_t n);
      |                 ^~~~
C:/Program Files/MSYS2/mingw64/x86_64-w64-mingw32/include/stdlib.h:714:16: note: previous declaration 'void swab(char*, char*, int)'
  714 |   void __cdecl swab(char *_Buf1,char *_Buf2,int _SizeInBytes) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                ^~~~
```

cmake/modules/config.cmake
```diff
@@ -18,8 +18,13 @@ CHECK_FUNCTION_EXISTS("ldexpf" HAVE_LDEXPF)
...
 CHECK_FUNCTION_EXISTS("mmap" HAVE_MMAP)
 CHECK_FUNCTION_EXISTS("swab" HAVE_SWAB)
-IF(NOT HAVE_SWAB)
-	CHECK_FUNCTION_EXISTS("_swab" HAVE_SWAB)
+IF(NOT HAVE_SWAB AND MINGW)
+	# on Win/MinGW, this needs a different check, maybe returning a char*
+	# kind of version
+	INCLUDE (CheckSymbolExists)
+	CHECK_SYMBOL_EXISTS(swab "stdlib.h" HAVE_CSWAB)
+ELSEIF(NOT HAVE_SWAB)
```
bradallred, line 26: we can remove this else. it didnt help any

gemrb/core/System/swab.h
```diff
@@ -30,7 +30,13 @@ extern "C" {
 #endif
 
 #ifndef HAVE_SWAB
+
+#ifdef HAVE_CSWAB
+#define swab(a,b,c)  swab((char*)(a), (char*)(b), (c))
```
bradallred, line 35: lets use `static_cast` to avoid more sonar complaints

bradallred (changes requested):

gemrb/core/System/swab.h
bradallred: actually... I see we also want to cast away const. im unsure if that is wise here. how do we know the implementation isn't writing back to a? that would be bad for us. it would be better then to just expose our `swap_private` as some other name and use it instead.

MarcelHB: Yeah, all C++-casts I tried ended up in complaints. No objections, drop this shit alltogether.

Issue #1296, opened 2021-06-18
If you let a dialog start while on a different virtual desktop, the dialog window is not drawn 


lynxlynxlynx:
To continue with exotic reports ... the window is there, just not drawn, so you either have to mouse over some options or press enter in the case of continue windows — once you switch back to that window.
![image](https://user-images.githubusercontent.com/121515/122520368-cef43900-d013-11eb-8385-fd46c6c6d5d8.png)

It's not enough for the window to not have focus

bradallred:
IIRC this is actually an SDL bug I saw fixed in git. SDL_Renderer doesn't draw when the renderer is "hidden" causing us to believe the window is not dirty when in fact it has to be completely redrawn.

The SDL folks were trying to be clever and efficient, but fail to realize this breaks down when rendering to a texture, which we are doing here. If updating to the tip of SDL doesn't fix it, I will open a bug with SDL about it.

we could (and may have to for things like `SDL_RENDER_TARGETS_RESET` and `SDL_RENDER_DEVICE_RESET`) add add support for application type events such as `SDL_APP_DIDENTERFOREGROUND` in our `EventManager` and `WindowManager` can just call `MarkDirty()` on all the windows whenever we gain focus.

lynxlynxlynx:
Just noting this isn't fixed yet, though there's a nice TODO for SDL_RENDER_DEVICE_RESET present.

MarcelHB:
But this seems to have relevance only for some corner-cases of D3D11, like having a driver updated _while_ running a game.

bradallred:
I still think the fix lies in updating your libSDL2. If that doesn't fix it this is an issue that needs to be raised with SDL.

lynxlynxlynx:
I can reproduce it with the buildbot build that now uses 2.0.12. But you were mentioning unreleased code, which I haven't tried.

bradallred:
well the latest release is 2.0.16. I don't remember when that bug was resolved.

lynxlynxlynx:
Can confirm it works on 2.0.20.

Issue #1317, opened 2021-06-25
 UI elements disappear when resizing the window


MarcelHB:
Since nothing prevents me from resizing or maximizing the window, I do so -- convinced it once worked without flaws on SDL2 w.r.t. scaling -- but then UI elements disappear until they are dirty and get redrawn, obviously.

This results in black, missing or partially redrawn UI/HUD parts in virtually every screen where performed. Here is an example in BG2 when resizing the window and hitting Pause.

![screen_size](https://user-images.githubusercontent.com/238558/123476214-b16b3480-d5fc-11eb-886c-26df390d4827.jpg)

I guess something is now aware of the physical screen size, instead of the logical one (or not scaled in time?), or we better prevent window modification, since this is easy to run into.

lynxlynxlynx:
What's your sdl2 version? I had no problems with resizing when making a video, but the symptom is similar to #1296.

MarcelHB:
It's 2.0.14.

lynxlynxlynx:
Maybe a windows backend thing then.

MarcelHB:
Well, yes. Just tested Linux: cannot reproduce there. :unamused:

MarcelHB:
Latest git-SDL doesn't contain any out-of-the-box fixes either.

MarcelHB:
It's inside of `direct3d`. OGL is fine.

bradallred:
D3D invalidates the textures. you have to handle `SDL_RENDER_DEVICE_RESET` and `SDL_RENDER_TARGETS_RESET` to have the window recreate it's texture. Easy to do if you feel inclined.

PR #1335, opened 2021-07-02
D3D: Scaling on window resizing

MarcelHB:
## Description

Some findings:
* `SDL_RENDER_TARGETS_RESET` is exclusively fired for D3D9 (or `direct3d` backend), D3D11 is unaffected at all. Anyway, unlike the removed comment, the [SDL docs imply](https://wiki.libsdl.org/SDL_EventType) that existing textures can be reused. So I think it's the easiest way just to make every window redraw on the next occasion.
* `SDL_RENDER_DEVICE_RESET` is exlusively fired for D3D11 and is not so much related to the issue. From what I read in [MSDN](https://docs.microsoft.com/en-us/windows/uwp/gaming/handling-device-lost-scenarios), it covers some very, very edgy cases that I'd like to ignore here.

See #1317

## Checklist

### DIFF
--- a/gemrb/core/GUI/EventMgr.cpp
+++ b/gemrb/core/GUI/EventMgr.cpp
@@ -500,4 +500,11 @@ Event EventMgr::CreateControllerButtonEvent(EventButton btn, bool down)
 	return e;
 }
 
+Event EventMgr::CreateRedrawRequestEvent() {
+	Event e = {};
+	e.type = Event::RedrawRequest;
+
+	return e;
+}
+
 }
--- a/gemrb/core/GUI/EventMgr.h
+++ b/gemrb/core/GUI/EventMgr.h
@@ -165,6 +165,9 @@ struct GEM_EXPORT GestureEvent : public TouchEvent {
 	float dDist;
 };
 
+struct GEM_EXPORT ApplicationEvent : public EventBase {
+};
+
 struct GEM_EXPORT Event {
 	enum EventType {
 		MouseMove = 0,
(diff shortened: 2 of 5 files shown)


bradallred:
excellent. thank you.

a couple of things:

1. I think this could also fix #1296 if we do this for `SDL_APP_DIDENTERFOREGROUND` as well
2. we dont need to make `Interface` a middle man to everything. Just grab the `WindowManager` and call on it directly.

MarcelHB:
1. Ok, I'll try to have a look.
2. Had it like this first, but then I noticed nobody ever `GEM_EXPORT`ed the WM so far, which isn't so bad from a coupling perspective.

bradallred:
im saying `Interface` already has `GetWindowManager`

bradallred:
if you are concerned about coupling you can always just create a new "application" event type in the `EventManager` and let window manager handle all this internally.

MarcelHB:
I spotted it, but the linker complained when being used outside of core.

The event design sounds better, I'll check it out.

gemrb/core/Interface.cpp
```diff
@@ -1849,6 +1850,16 @@ int Interface::Init(InterfaceConfig* config)
...
+	EventMgr::EventCallback appEventCallback = [this](const Event& e) {
+		if (e.type == Event::EventType::RedrawRequest) {
+			winmgr->MarkAllDirty();
+		}
+
+		return true;
+	};
+
+	appEventMonitor = EventMgr::RegisterEventMonitor(appEventCallback, Event::EventTypeMask::AllApplicationMask);
+
```
bradallred, line 1862: you dont need to do this. Just handle the event in `WindowManager::DispatchEvent`

bradallred:
beautiful. thank you.

PR #1337, opened 2021-07-04
Improving load & save time of heavy save games

MarcelHB:
## Description
I noticed that save games of later stages in the game take notably longer: Every ARE is extracted into the cache onto the disk. Yet, we don't need all these ARE files right now, and given the progress, some of them never again.

BG1 spills around 260 ARE files into the cache folder, other games a little less for me.

Why not extract only when we really need it? On my very fast NVMe this _only_ saves half a second (and a lot of console entries), but surely some more on HDDs or RPIs -- unfortunately no test hardware at hand right now.

Feel free to discuss first, but I considered this an avoidable practice.

## Checklist

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -4513,6 +4513,7 @@ int Interface::CompressSave(const char *folder, bool overrideRunning)
 	PluginHolder<ArchiveImporter> ai = MakePluginHolder<ArchiveImporter>(IE_SAV_CLASS_ID);
 	ai->CreateArchive( &str);
 
+	tick_t startTime = GetTicks();
 	// If we override the savegame we are running to fetch AREs from, it has already dumped
 	// itself as "ares.blb" into the cache folder. Otherwise, just copy directly.
 	if (!overrideRunning && saveGameAREExtractor.copyRetainedAREs(&str) == GEM_ERROR) {
@@ -4550,6 +4551,9 @@ int Interface::CompressSave(const char *folder, bool overrideRunning)
 			dir.Rewind();
 		}
 	}
+
+	tick_t endTime = GetTicks();
+	Log(WARNING, "Core", "%lu ms (compressing SAV file)", endTime - startTime);
 	return GEM_OK;
 }
 
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -129,7 +129,7 @@ ItemDragOp::ItemDragOp(CREItem* item)
 	}
 
 	cursor = pic;
-	
+
 	// FIXME: this VarName is not consistant
 	strnlwrcpy(dragDummy.VarName, "itembutton", MAX_VARIABLE_LENGTH - 1);
 }
@@ -516,7 +516,7 @@ void Interface::HandleFlags()
 		// if Exitgame was set, we'll set Start.py too
 		QuitGame (QuitFlag&QF_EXITGAME);
 	}
-	
+
 	if (QuitFlag & (QF_QUITGAME | QF_EXITGAME | QF_LOADGAME | QF_ENTERGAME)) {
 		delete winmgr->GetGameWindow()->RemoveSubview(gamectrl);
 		gamectrl = nullptr;
@@ -528,7 +528,7 @@ void Interface::HandleFlags()
 
 	if (QuitFlag&QF_LOADGAME) {
 		QuitFlag &= ~QF_LOADGAME;
-		
+
 		LoadGame(LoadGameIndex.get(), VersionOverride );
 		LoadGameIndex.release();
 		//after loading a game, always check if the game needs to be upgraded
@@ -1335,7 +1335,7 @@ int Interface::Init(InterfaceConfig* config)
 		return GEM_ERROR;
 	}
 	if (!KeepCache) DelTree((const char *) CachePath, false);
-	
+
 	// potentially disable logging before plugins are loaded (the log file is a plugin)
 	value = config->GetValueForKey("Logging");
 	if (value) ToggleLogging(atoi(value));
@@ -1375,10 +1375,10 @@ int Interface::Init(InterfaceConfig* config)
 		Log(FATAL, "Core", "Cannot Initialize Video Driver.");
 		return GEM_ERROR;
 	}
-	
+
 	// ask the driver if a touch device is in use
 	EventMgr::TouchInputEnabled = touchInput < 0 ? video->TouchInputEnabled() : touchInput;
-	
+
 	ieDword brightness = 10;
 	ieDword contrast = 5;
 	vars->Lookup("Brightness Correction", brightness);
@@ -1836,7 +1836,7 @@ int Interface::Init(InterfaceConfig* config)
 		pathFile->Close();
 	}
 	delete pathFile;
-	
+
 	EventMgr::EventCallback ToggleConsole = [this](const Event& e) {
 		if (e.type != Event::KeyDown) return false;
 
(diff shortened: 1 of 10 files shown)


MarcelHB:
Let me check one gap tomorrow again: When we repack the SAV file by saving and then try to enter some older area, the look up cache probably needs to be rebuild.

lynxlynxlynx:
Interesting! 
Game::LoadMap gets called also from other (nonsave) places iirc - is that part fine?

MarcelHB:
That's not just fine but a requirement, since we don't really want to care who requests a map change, only that we may need to emit it from the running savegame.

MarcelHB:
Now, we have the following status:
* Loading is roughly constant.
* The first save gets the penalty of about what we saved in the loading, since every sleeping ARE is extracted before repacking. Any subsequent save is spared, but of course everything will be packed again, and again.

Room for even more improvement: Why extracting, and repacking? It involves CPU and filesystem for no reason. Instead, we could copy the remaining, compressed entries over into a new savegame before picking up the cache stuff. With some precaution of course, if we update the same savegame we are currently feeding from.

MarcelHB:
I'd like to invite you for some testing. Here is a BG save game to do some stress-testing if useful: 
[000000003-1.zip](https://github.com/gemrb/gemrb/files/6760737/000000003-1.zip)

So what's going on now:

- Extract ARE files only on demand from the save game that is know to have loaded the current session. Keep the rest in a registry that knows what's still in the box.
- When saving into a different save game: raw-copy the compressed data into the new save game. So no more decompressing, FS bothering and recompression. Continue with the ARE files from the cache folder as before.
- When saving into the same save game: drop a `ares.blb` into the cache folder (that is sequence of all unextracted AREs) and pick this up into the new SAV file. The registry is getting updated by the new locations, so we can safely change maps after the old SAV is no more. Continue with the ARE files from the cache folder as before. 

The results are quite promising to me.

gemrb/core/Interface.cpp
```diff
@@ -4492,18 +4494,31 @@ int Interface::WriteWorldMap(const char *folder)
 	return 0;
 }
 
-int Interface::CompressSave(const char *folder)
+static bool IsBlobSaveItem(const char *path) {
```
lynxlynxlynx, line 4497: Please also handle a null parameter.

lynxlynxlynx:
I think it would be worthwhile to add some timing logging, so we can test the benefits more easily. I'd just log the total duration once the process is over.

The ees strongly improved the save times, as they're practically instant there.

MarcelHB:
What do you want to see exactly? I can emit different pieces of ms spent for the steps in saving and loading (which need to be summed then) but to see the difference, you'd need to do some different logging in master as well.

Saving is now virtually instant here, too.

lynxlynxlynx:
Ideally it'd be a separate pr, that we can merge first, gather results without the patches and then also with them.

As for data, for me the totals would be enough. Like maybe the time spent in LoadGame.

bradallred:
We can just use perf or other profiling tools. Not really a need to add our own instrumentation.

MarcelHB:
So, benchmarks. I put my `Cache2` onto some old USB 2.0 FAT32 key to simulate something from real life, but slower than my NVMe. This is for amplifying the effect.

Loading the BG save game from above:
master: ~8.5s (!!)
this: ~900ms

Saving the same:
master: ~900ms
this: ~80ms

For an NVMe, loading:
master: ~900ms
this: ~300ms

Saving: neglectable

I've put a temporary commit on top of this branch, and if you like to see yourself: [for master](https://github.com/gemrb/gemrb/compare/master...MarcelHB:are_loading_old?expand=1.)

> 
> We can just use perf or other profiling tools. Not really a need to add our own instrumentation.
> […](#)
> On Jul 5, 2021, at 10:31, Jaka Kranjc ***@***.***> wrote: ﻿ Ideally it'd be a separate pr, that we can merge first, gather results without the patches and then also with them. As for data, for me the totals would be enough. Like maybe the time spent in LoadGame. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

You are right, but I don't know how to get precise slices on Windows (not sure if `perf` works here, and AMD's tool has its limits)  r/n, so I'll stick to this method.

lynxlynxlynx:
My test on a SSD and a good (something like 10k RPM) shows slight improvements even for saves with few areas.

However, when testing bg1, after the first resave, the rest took several seconds and I got a few invalid seeks: 
```
[Unknown]: [Streams]: Invalid seek position 2374 in file baldur.sav(limit: 0)
```
Could be from other recent changes perhaps, but in this worst case it slows down saving significantly. Restarting the game and loading that save also takes ages (decompresing the blb: 27661 ms).

Something is fishy — why did the size explode so?
```
$ ls -lh build/gemrb/Cache2/
total 4.0G
-rw-r--r-- 1 kajak kajak  43K Jul  9 16:33 ar3800.are
-rw-r--r-- 1 kajak kajak 4.0G Jul  9 16:33 ares.blb
-rw-r--r-- 1 kajak kajak   20 Jul  9 16:33 default.toh
-rw-r--r-- 1 kajak kajak    0 Jul  9 16:33 default.tot
```

MarcelHB:
So reproducing for you is: Loading a game, and saving to the same slot, or quicksaving?

lynxlynxlynx:
It was always manual saving.

MarcelHB:
Ok, one thing fell victim to my post-wiring of error return codes. The other thing I did not notice came up when resaving the 2nd time, since strlen and the value in zip entry are off by one.

lynxlynxlynx:
Thanks, appears to work fine now.

MarcelHB:
Cool. At times, you may revert 409d310 though :wink:

lynxlynxlynx:
A user reported a crash and it led me back here. Starting a new tob game is a simple way to reproduce. An autosave is scripted and it corrupts memory and boom.

```gdb
#0  0x00000000006b4311 in GemRB::ResourceManager::GetResource (this=0x5028, 
    ResRef=0x1010 <error: Cannot access memory at address 0x1010>, type=327680, silent=true)
    at /home/kajak/games/gemrb/gemrb/core/ResourceManager.cpp:102
#1  0x00000000006b8f25 in GemRB::SaveGame::GetSave (this=0x0)
    at /home/kajak/games/gemrb/gemrb/core/SaveGameIterator.cpp:171
#2  0x00000000006b567c in GemRB::SaveGameAREExtractor::copyRetainedAREs (this=0xa39bb8, 
    destStream=0x7fffffff7240, trackLocations=false)
    at /home/kajak/games/gemrb/gemrb/core/SaveGameAREExtractor.cpp:40
#3  0x0000000000653c42 in GemRB::Interface::CompressSave (this=0xa39780, 
    folder=0x7fffffffb4c0 "/media/kajak/games/bg2-gog/prefix/drive_c/GOG Games/Baldur's Gate 2/save/000000000-Auto-Save", overrideRunning=false) at /home/kajak/games/gemrb/gemrb/core/Interface.cpp:4510
#4  0x00000000006bb176 in GemRB::DoSaveGame (
    Path=0x7fffffffb4c0 "/media/kajak/games/bg2-gog/prefix/drive_c/GOG Games/Baldur's Gate 2/save/000000000-Auto-Save", overrideRunning=false) at /home/kajak/games/gemrb/gemrb/core/SaveGameIterator.cpp:436
#5  0x00000000006ba955 in GemRB::SaveGameIterator::CreateSaveGame (this=0xe23f90, index=0, mqs=false)
    at /home/kajak/games/gemrb/gemrb/core/SaveGameIterator.cpp:635
#6  0x000000000060b41d in GemRB::GameScript::SaveGame (parameters=0x3b972e0)
    at /home/kajak/games/gemrb/gemrb/core/GameScript/Actions.cpp:5921
```

MarcelHB:
Grml, but easy. Just the only place I forgot checking for a specific nullptr.

Issue #1311, opened 2021-06-23
BG2: All party moves opening containers 


rsantellan:
#### Bug description
When you click a container with all the party selected all move to the place instead of only the leader.

GemRB version: v0.9.0

#### Steps to reproduce
1. Go to a place with a container
2. Select all party
3. Click on the container

#### Expected behavior
Only the leader moves

#### Screenshots

https://user-images.githubusercontent.com/1820598/123025231-b0de5e00-d3b0-11eb-9643-a8ab5c2f79dc.mp4

lynxlynxlynx:
This is easy to fix. All the code is the mouse handler functions in GameControl.cpp and in one place we should clearly work with just the active actor, not the whole selection.

rsantellan:
For what I see the function HandleContainer is the one that invokes the opening the container.

![image](https://user-images.githubusercontent.com/1820598/123185166-a8972900-d46b-11eb-885d-6ba3d2ea88cc.png)

If I follow **UseContainer** to the file **core/GameScript/Actions.cpp** I found out that in that function I have the sender but can't figure out how to get the rest of the party to make them stop. Also I found out that the same applies to doors, is a race to see which one open it first!.

@lynxlynxlynx any clue on where I can look to stop the rest of the party ?

bradallred:
`HandleContainer` is not the problem, that clearly generates an action for a single actor. The problem is going to be here

https://github.com/gemrb/gemrb/blob/07a8a780f0bfb180e31a978f9afc8b3e4b531eda/gemrb/core/GUI/GameControl.cpp#L2215-L2218

bradallred:
also looks like the `ST_TRAVEL` part of things should be handled separately since then you do want everybody to move

rsantellan:
@bradallred that works like a charm now I'm trying to understand why :)

bradallred:
I added some more context to the diff for you so you can see how it relates :)

rsantellan:
So, you only set one actor to do the action instead of the CommandSelectedMovement that moves the hole party, then the next action is the UseContainer that if you are really far start moving the character to open the container. 
Now, when all where moving, it will open the character that was faster to open it ? Or still is the first selected ?

bradallred:
[...]

PR #1359, opened 2021-07-16
bug fix x4

Steamcrusher:
## Description
multiple fixes of bugs and crushes that were faced on a numerous test runs under fedora 34
## Checklist

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -59,6 +59,13 @@ static const ResRef TestSpell = "SPWI207";
 
 uint32_t GameControl::DebugFlags = 0;
 
+static Actor* GetMainSelectedActor()
+{
+	auto actor = core->GetFirstSelectedPC(false);
+	if (actor) return actor;
+	return core->GetFirstSelectedActor();
+}
+
 //If one of the actors has tracking on, the gamecontrol needs to display
 //arrow markers on the edges to point at detected monsters
 //tracterID is the tracker actor's global ID
@@ -2065,11 +2072,7 @@ void GameControl::InitFormation(const Point &clickPoint)
 		return;
 	}
 
-	const Game *game = core->GetGame();
-	const Actor *selectedActor = core->GetFirstSelectedPC(false);
-	if (!selectedActor) {
-		selectedActor = game->selected[0];
-	}
+	const Actor *selectedActor = GetMainSelectedActor();
 
 	isFormationRotation = true;
 	formationBaseAngle = AngleFromPoints(clickPoint, selectedActor->Pos);
@@ -2214,7 +2217,12 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 		
 		if (overContainer || overDoor || (overInfoPoint && overInfoPoint->Type==ST_TRAVEL && target_mode == TARGET_MODE_NONE)) {
 			// move to the object before trying to interact with it
-			CommandSelectedMovement(p, false, isDoubleClick);
+			Actor* mainActor = GetMainSelectedActor();
+			if (mainActor && overContainer) {
+				CreateMovement(mainActor, p, false, isDoubleClick);	// let one actor handle loot and containers
+			} else {
+				CommandSelectedMovement(p, false, isDoubleClick);
+			}
 		}
 		
 		if (target_mode != TARGET_MODE_NONE || overInfoPoint || overContainer || overDoor) {
@@ -2245,11 +2253,7 @@ void GameControl::PerformSelectedAction(const Point& p)
 	const Map* area = game->GetCurrentArea();
 	Actor* targetActor = area->GetActor(p, target_types & ~GA_NO_HIDDEN);
 
-	Actor* selectedActor = core->GetFirstSelectedPC(false);
-	if (!selectedActor) {
-		//this could be a non-PC
-		selectedActor = game->selected[0];
-	}
+	Actor* selectedActor = GetMainSelectedActor();
 
 	//add a check if you don't want some random monster handle doors and such
 	if (targetActor) {
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -8141,7 +8141,7 @@ int Actor::GetFavoredPenalties() const
 	classLevels.sort(); // ascending
 	if (flevel == -1) {
 		// any class - just remove the highest level
-		classLevels.erase((classLevels.end())--);
+		classLevels.erase(--classLevels.end());
 		classCount--;
 	} else {
 		// remove() kills all elements with the same value, so we have to jump through hoops
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -551,7 +551,7 @@ Holder<Sprite2D> SDL20VideoDriver::GetScreenshot(Region r, const VideoBufferPtr&
 
 	SDL_Texture* target = SDL_GetRenderTarget(renderer);
 	if (buf) {
-		auto texture = static_cast<SDLTextureVideoBuffer*>(drawingBuffer)->GetTexture();
+		auto texture = static_cast<SDLTextureVideoBuffer*>(buf.get())->GetTexture();
 		SDL_SetRenderTarget(renderer, texture);
 	} else {
 		SDL_SetRenderTarget(renderer, nullptr);
--- a/gemrb/plugins/BAMImporter/CMakeLists.txt
+++ b/gemrb/plugins/BAMImporter/CMakeLists.txt
@@ -1,1 +1,1 @@
-ADD_GEMRB_PLUGIN (BAMImporter BAMImporter.cpp BAMFontManager)
+ADD_GEMRB_PLUGIN (BAMImporter BAMImporter.cpp BAMFontManager.cpp)


gemrb/core/GUI/GameControl.cpp
```diff
@@ -2209,7 +2209,18 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 		
 		if (overContainer || overDoor || (overInfoPoint && overInfoPoint->Type==ST_TRAVEL && target_mode == TARGET_MODE_NONE)) {
 			// move to the object before trying to interact with it
-			CommandSelectedMovement(p, false, isDoubleClick);
+			//core->GetFirstSelectedPC(false)
```
lynxlynxlynx, line 2212: Please remove this.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2209,7 +2209,18 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
...
+			Actor* selectedActor = core->GetFirstSelectedPC(false);
+			if (!selectedActor) {
+				//this could be a non-PC
+				selectedActor = core->GetGame()->selected[0];
+			}
+			if ((overContainer || target_mode == TARGET_MODE_PICK ) && selectedActor)
+			{
```
lynxlynxlynx, line 2219: The brace should be on the previous line

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2209,7 +2209,18 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
...
+				CreateMovement(selectedActor, p, false, isDoubleClick);
+			}
+			else
+				CommandSelectedMovement(p, false, isDoubleClick);
```
lynxlynxlynx, line 2223: This should follow the general style, like above.

lynxlynxlynx:
Thanks, the changes look fine to me content-wise; the appveyor failure is not related.

gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -551,7 +551,7 @@ Holder<Sprite2D> SDL20VideoDriver::GetScreenshot(Region r, const VideoBufferPtr&
 
 	SDL_Texture* target = SDL_GetRenderTarget(renderer);
 	if (buf) {
-		auto texture = static_cast<SDLTextureVideoBuffer*>(drawingBuffer)->GetTexture();
```
bradallred, line 554: what makes you believe that `drawingBuffer` and `but` are interchangeable? even if sometimes true is it always? it would seem not or else you would have the same problem. In short I dont under stand how this fixes anything.

bradallred:
what was wrong with the fix proposed in #1311?

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2209,7 +2209,18 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 		
 		if (overContainer || overDoor || (overInfoPoint && overInfoPoint->Type==ST_TRAVEL && target_mode == TARGET_MODE_NONE)) {
 			// move to the object before trying to interact with it
-			CommandSelectedMovement(p, false, isDoubleClick);
+			//core->GetFirstSelectedPC(false)
+			Actor* selectedActor = core->GetFirstSelectedPC(false);
+			if (!selectedActor) {
+				//this could be a non-PC
+				selectedActor = core->GetGame()->selected[0];
+			}
```
bradallred, line 2217: we are now seeing this code duplicated 3x. It's time to make it a function.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8163,7 +8163,7 @@ int Actor::GetFavoredPenalties() const
 	classLevels.sort(); // ascending
 	if (flevel == -1) {
 		// any class - just remove the highest level
-		classLevels.erase((classLevels.end())--);
+		classLevels.erase(--classLevels.end());
```
bradallred, line 8166: I think `rbegin()` would be better

bradallred: TIL there is no `erase` overload for a "reverse" iterator.

gemrb/plugins/SDLVideo/SDL20Video.cpp
Steamcrusher: Even if they're not (which i highly doubt, because as i checked, screenshots and savefile screenshots are working as before), it's a bad idea to check for buf, but then cast to drawingBuffer. 
This function already has everything it needs as passed arguments, but doesn't know anything about the state the drawingBuffer is in. We *can* instead put a drawingBuffer check (and then we don't need buf argument at all), but then instead of crashes we will get to the SDL_SetRenderTarget(renderer, nullptr), which, i dunno, will make an entirely black screenshot maybe, or whatever it does.

Steamcrusher:
> what was wrong with the [fix proposed](https://github.com/gemrb/gemrb/issues/1311#issuecomment-867254118) in #1311?

nothing
haven't seen it
why it's still open then?

lynxlynxlynx:
We were coaching the other user to finalize the fix, got distracted, then you fixed it on your own. No real harm. :)

gemrb/plugins/SDLVideo/SDL20Video.cpp
bradallred: yup, seems like a mistake now that I look at it

gemrb/core/GUI/GameControl.cpp
bradallred: why is `TARGET_MODE_PICK` being checked anyway? just to match originals? It was my understanding the desired behavior was to just never sent the whole party along to a container

Steamcrusher: I modified it so only one character moves not only to the container but to pick up ground loot as well. From my perspective, it would suit gameplay. I haven't played this game for years, however, so i may be wrong. Should i cancel this addition?

bradallred: I dont think that answers why checking `TARGET_MODE_PICK` is desired

it might help to go read the discussion in the issue, but the door/infopoint stuff should probably be handled in a separate branch

lynxlynxlynx: Oh, then it should definitely be removed. Loot piles are also containers, while the pick mode is for lock picking, so definitely not what you intended. Bashing and disarming are two other ways, but yeah, makes little sense to limit.

gemrb/core/GUI/GameControl.cpp
Steamcrusher: done

gemrb/core/GUI/GameControl.cpp
Steamcrusher: done

gemrb/core/GUI/GameControl.cpp
Steamcrusher: done

gemrb/core/GUI/GameControl.cpp
Steamcrusher: done

gemrb/core/GUI/GameControl.cpp
```diff
@@ -59,6 +59,13 @@ static const ResRef TestSpell = "SPWI207";
 
 uint32_t GameControl::DebugFlags = 0;
 
+static Actor* GetMainActor()
+{
+	if (auto actor = core->GetFirstSelectedPC(false))
+		return actor;
+	return core->GetGame()->selected[0];
+}
+
```
lynxlynxlynx, line 68: I don't think we need this function. We can use `game->selected[0]` directly or more nicely, through `Interface::GetFirstSelectedActor()`. One of the places already ensures it's set, while the others can get asserts, since they should have it too.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: I still see the same problem, the last three lines don't match the style.

gemrb/core/GUI/GameControl.cpp
bradallred: the confusion (for me at least) is that `Interface::GetFirstSelectedPC` is doing something other than returning `selected[0]` and all the code in `GameControl` was following this pattern of try `Interface::GetFirstSelectedPC` and if that returns null then use `selected[0]`.

I dont believe we can just use `GetFirstSelectedActor`/`selected[0]` because that could be an NPC and the code calling this looks like it purposely wants to prioritize a PC over NPC. Wether it _should_ do that or not I do not know.

lynxlynxlynx: It's using `selected` like that for historic reasons. If there's only one selected thing, then it doesn't matter. If there's several, most of the special actions aren't available — only attack and defend are, which are both for everyone. Ergo using the first thing in the list is just fine.

bradallred: I'm confused. It still sounds like if you have multiple actors selected then `selected[0]` still may not be a PC so in this example you could have a group selected where the first actor is an NPC then direct the group to open a container and that NPC would then be directed to the container rather than a PC. Is that not accurate?

lynxlynxlynx: Ah, that's true, the odd one out. We could ensure the pcs are on top when manipulating the list, but so far it seems cleaner to have another lookup method. Then I have other comments.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -59,6 +59,13 @@ static const ResRef TestSpell = "SPWI207";
 
 uint32_t GameControl::DebugFlags = 0;
 
+static Actor* GetMainActor()
+{
+	if (auto actor = core->GetFirstSelectedPC(false))
+		return actor;
+	return core->GetGame()->selected[0];
```
lynxlynxlynx, line 66: Cleaner version:
```
auto actor = core->GetFirstSelectedPC(false);
if (actor) return actor;
return core->GetFirstSelectedActor();
```

bradallred: I actually disagree with a name that has `PC` in it as that implies the actor is a PC. It could just be `GetActorToPerformAction` or something along those lines.

bradallred: or `GetActorToCommand`

lynxlynxlynx: My suggestion doesn't have PC in it for the same reason.

bradallred: I clearly should have read the line below the code :)

gemrb/core/GUI/GameControl.cpp
Steamcrusher: should i just delete the function and make calls for GetFirstSelectedPC instead? if core->GetGame()->selected[0] can return some summoned unit or whatever, that's not what we need right here

lynxlynxlynx: But there might also be no pc selected, so that is not enough. As far as inlining back, I think it should stay. The function might come in handy some place else, but more importantly, since it's more thorough, it makes reasoning about the code easier, since we can't just use `selected[0]` everywhere.

Steamcrusher: done

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2214,7 +2217,12 @@ bool GameControl::OnMouseUp(const MouseEvent& me, unsigned short Mod)
 		
 		if (overContainer || overDoor || (overInfoPoint && overInfoPoint->Type==ST_TRAVEL && target_mode == TARGET_MODE_NONE)) {
 			// move to the object before trying to interact with it
-			CommandSelectedMovement(p, false, isDoubleClick);
+			Actor* mainActor = GetMainSelectedActor();
+			if (mainActor && (overContainer || target_mode == TARGET_MODE_PICK)) {
```
lynxlynxlynx, line 2221:
```suggestion
			if (mainActor && overContainer) {
```
Looks like you missed this a couple of days ago.

Steamcrusher: nah i just accidentally put this change in the next commit (that i haven't pushed yet, still waiting this current PR to be merged) 
fixed 10 secs ago
still learning how to use git

lynxlynxlynx: You're doing fine, don't worry. 👍🏾

lynxlynxlynx:
Ok, thanks, now it looks good and builds — the AppVeyor failure is from other changes. The only question remaining is if the container thing also works in practice. Will check in a bit, I'm in the middle of something.

lynxlynxlynx:
Thanks for the work and patience in making it perfect!

Issue #855, opened 2020-07-03
right click on non-pcs and get their selection line


lynxlynxlynx:
In fact, in original BG2 you can simply right click on anyone with blue circle and get their "existence" lines in the dialog window. That doesn't happen in GemRB.
AR0709:
![Captura de pantalla de 2020-02-18 14-45-14](https://user-images.githubusercontent.com/1279852/74714533-4c137080-525d-11ea-9ae4-fa0170b5c30c.png)

_Originally posted by @burner1024 in https://github.com/gemrb/gemrb/issues/634#issuecomment-587324595_

Looking at the data, it's clear the common select sounds are played, not the existence ones. Maybe those in cases where the soundset isn't present, but for this area, definitely not. Basically we need to call Actor::PlaySelectionSound on right-click in GameControl (and enhance the function if needed).

burner1024:
In fact, that's not just neutrals, enemies too.

PR #1370, opened 2021-07-22
Selection line when RCing actors

Steamcrusher:
More info: https://github.com/gemrb/gemrb/issues/855

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -325,11 +325,7 @@ void GameControl::CreateMovement(Actor *actor, const Point &p, bool append, bool
 // can we handle it (no movement impairments)?
 bool GameControl::CanRun(const Actor *actor) const
 {
-	if (!actor) return false;
-	if (actor->GetEncumbranceFactor(true) != 1) {
-		return false;
-	}
-	return true;
+	return (actor && actor->GetEncumbranceFactor(true) != 1);
 }
 
 bool GameControl::ShouldRun(const Actor *actor) const
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
 	SetCursor(core->Cursors[IE_CURSOR_USE]);
 }
 
+bool GameControl::TryDefaultTalk(Point p) const
+{
+	Actor *targetActor = core->GetGame()->GetCurrentArea()->GetActor(p,
+			target_types & ~(GA_NO_HIDDEN /*| GA_NO_ALLY | GA_NO_ENEMY*/ ));
+	if (targetActor)
+		targetActor->VerbalConstant(VB_SELECT, 4, DS_CIRCLE);
+	return true;
+}
+
 /** Mouse Button Down */
 bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
 {
@@ -2087,6 +2092,8 @@ bool GameControl::OnMouseDown(const MouseEvent& me, unsigned short Mod)
 
 	switch(me.button) {
 	case GEM_MB_MENU: //right click.
+		TryDefaultTalk(gameClickPoint);
+		//add a check if you don't want some random monster handle doors and such
 		if (core->HasFeature(GF_HAS_FLOAT_MENU) && !Mod) {
 			core->GetGUIScriptEngine()->RunFunction( "GUICommon", "OpenFloatMenuWindow", false, p);
 		}
--- a/gemrb/core/GUI/GameControl.h
+++ b/gemrb/core/GUI/GameControl.h
@@ -168,6 +168,8 @@ private:
 	bool CanLockFocus() const override { return true; };
 	void FlagsChanged(unsigned int /*oldflags*/) override;
 
+	bool TryDefaultTalk(Point p) const;
+
 public:
 	explicit GameControl(const Region& frame);
 	~GameControl(void) override;


gemrb/core/GUI/GameControl.cpp
```diff
@@ -325,11 +325,7 @@ void GameControl::CreateMovement(Actor *actor, const Point &p, bool append, bool
 // can we handle it (no movement impairments)?
 bool GameControl::CanRun(const Actor *actor) const
 {
-	if (!actor) return false;
-	if (actor->GetEncumbranceFactor(true) != 1) {
-		return false;
-	}
-	return true;
+	return (actor && actor->GetEncumbranceFactor(true) != 1);
 }
```
lynxlynxlynx, line 329: This chunk has nothing to do with the bug or its fix, please remove it. The golden rule is to do one thematic thing per commit, so it's easier to review, rereview and test when bisecting.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
 	SetCursor(core->Cursors[IE_CURSOR_USE]);
 }
 
+bool GameControl::TryDefaultTalk(Point p) const
+{
+	Actor *targetActor = core->GetGame()->GetCurrentArea()->GetActor(p,
+			target_types & ~(GA_NO_HIDDEN /*| GA_NO_ALLY | GA_NO_ENEMY*/ ));
```
lynxlynxlynx, line 2081: Also needs to check GA_NO_DEAD and GA_NO_UNSCHEDULED. The other two can go.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
 	SetCursor(core->Cursors[IE_CURSOR_USE]);
 }
 
+bool GameControl::TryDefaultTalk(Point p) const
+{
+	Actor *targetActor = core->GetGame()->GetCurrentArea()->GetActor(p,
```
lynxlynxlynx, line 2080: Should be const. But also, don't we have lastActor that holds what the mouse is over? Or is that just in other GameControl functions?

And if it's thus unavoidable, please split this into three lines, fetch the game, the map, then the actor, so the call itself won't have to be chopped up.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
...
+			target_types & ~(GA_NO_HIDDEN /*| GA_NO_ALLY | GA_NO_ENEMY*/ ));
+	if (targetActor)
+		targetActor->VerbalConstant(VB_SELECT, 4, DS_CIRCLE);
+	return true;
```
lynxlynxlynx, line 2084: The return value is never checked (or false), so you can make it void.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
 	SetCursor(core->Cursors[IE_CURSOR_USE]);
 }
 
+bool GameControl::TryDefaultTalk(Point p) const
```
lynxlynxlynx, line 2078: Pass the point by const reference, since it won't be touched.

gemrb/core/GUI/GameControl.cpp
```diff
@@ -2079,6 +2075,15 @@ void GameControl::InitFormation(const Point &clickPoint)
...
+{
+	Actor *targetActor = core->GetGame()->GetCurrentArea()->GetActor(p,
+			target_types & ~(GA_NO_HIDDEN /*| GA_NO_ALLY | GA_NO_ENEMY*/ ));
+	if (targetActor)
+		targetActor->VerbalConstant(VB_SELECT, 4, DS_CIRCLE);
```
lynxlynxlynx, line 2083: The number of selection sounds differs per game, so that's why I suggested using the Actor method. If the originals ignored the settings in this case, just add a parameter to it.

gemrb/core/GUI/GameControl.cpp
bradallred: tho do feel free to add it as an _additional_ commit

gemrb/core/GUI/GameControl.cpp
bradallred: yes, `GameControl::GetLastActor()` is the actor to use here

gemrb/core/GUI/GameControl.cpp
Steamcrusher: ok

gemrb/core/GUI/GameControl.cpp
Steamcrusher: i thought that in the future the return value could be of use somewhere, but ok

gemrb/core/GUI/GameControl.cpp
Steamcrusher: good idea, i'll make it

gemrb/core/GUI/GameControl.cpp
Steamcrusher: so i need to make a new actor method just to get rid of the magic constant?

gemrb/core/GUI/GameControl.cpp
Steamcrusher: ok got it

gemrb/core/GUI/GameControl.cpp
Steamcrusher: can go you mean i need to delete the comments or uncomment them?

lynxlynxlynx: Yeah, delete the commented stuff.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: Is the existing one I suggested not fit for this purpose?

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: Easier to add it then, if it's ever needed and we know what is actually needed. Perhaps it would even require an int to differentiate between more states.

lynxlynxlynx:
Closing, I've added a cleaned up version and some follow-up work to polish the experience. Thanks anyway!

Issue #102, opened 2019-02-04
[feature] door highlights are of different color


lynxlynxlynx:
**Description**
We currently highlight doors with the same color as containers. TAB was used in some games for this highlight (we use ALT for everything).

The color was light purple (red in iwd2), not the container blue. We need to externalize these colors into a table and then use that in core. This has been done — the table is colors.2da and one can use gamedata->GetColor("color-row-name") to get a specific entry.

And PST is just annoying, but this can be ignored for now:

> Highlighting of containers, doors etc. in PS:T was done differently
> 
> - no edge line
> - instead of almost opaque polygon drawn over it, in PS:T the object changes hue from whitish to cyanish, as if the red channel was set to 0 or something like that. Possibly it's done by calculating pixel lightness in HLS space and then mapping it to cyan gradient or st. like that.

bradallred:
the PST portion of this is fixed by #618

rsantellan:
Hi @lynxlynxlynx I have a question about the color format.

![image](https://user-images.githubusercontent.com/1820598/124688667-8eb70680-dead-11eb-9e2d-e9d01e53b90e.png)

That is the color in the original, in html will be a **3abeb7**

In which format is on the table ? 
https://github.com/gemrb/gemrb/blob/6cdc73df9e9c7a35cfa87cb0891d8c9327e2d412/gemrb/unhardcoded/shared/colors.2da#L6 

Should I add to all the games or if I add it on the shared will be overridden on the game specifics ? 

I found where to change it: https://github.com/gemrb/gemrb/blob/6cdc73df9e9c7a35cfa87cb0891d8c9327e2d412/gemrb/core/Map.cpp#L1183
And do some testing using: **d->outlineColor = gamedata->GetColor("MAPTXTBG");**
And is working... Only missing is the color format.

rsantellan:
For answering some of my questions...
If I put it on the shared it will be **shared**
![image](https://user-images.githubusercontent.com/1820598/124690667-ec008700-deb0-11eb-9e9c-a27d3acb9dcb.png)
Now I'm only missing the color... it turn out red my test...
```
2DA V1.0
0xffffffff
				COLOR
TOOLTIP			0xffffffff
TOOLTIPBG		0x00000000
MAPICNBG		0xffffffff
MAPTXTBG		0x00000000
DOORTEST		0x3ABEB700
```

lynxlynxlynx:
[...]

PR #1371, opened 2021-07-23
[feature] door highlights are of different color #102

rsantellan:
## Description
Implementation of the door highlights on different color

## Checklist


gemrb/unhardcoded/bg1/colors.2da
```diff
@@ -5,3 +5,4 @@ TOOLTIP			0x00000000
 TOOLTIPBG		0xf0b080ff
 MAPICNBG		0xf0b080ff
 MAPTXTBG		0xf0b080ff
+ALTDOOR		    0xff00ffff
```
lynxlynxlynx, line 8: The tables appear to use tabulators as separators, that's why the lines look misaligned in the web diff. Please match this in the new lines, so the columns will be nice and pretty.

gemrb/unhardcoded/shared/colors.2da
```diff
@@ -5,3 +5,4 @@ TOOLTIP			0xffffffff
 TOOLTIPBG		0x00000000
 MAPICNBG		0xffffffff
 MAPTXTBG		0x00000000
+ALTDOOR		    0xff0000ff
```
lynxlynxlynx, line 8: Since it's the shared file, this assume the color was red in all other games, so iwd2 *and* iwd1. I've now checked iwd1 and it matches, so all good. :)

lynxlynxlynx:
Thanks! Can you also take care of the highlight color difference in iwds on mouseover? It's that deeper blue shown in the bug screenshot.

rsantellan:
@lynxlynxlynx I don't iwd to test, it should be applied to all or only to that one ? Looking for the correct color I have:
image
But I don't think is quite there..

lynxlynxlynx:
iwds = iwd1, how, iwd2

looks pretty close already:
image

rsantellan:
@lynxlynxlynx  I have used the check ** core->HasFeature(GF_IWD2_SCRIPTNAME) ** but I don't know if is better way of doing it.

lynxlynxlynx:
The whole point of having one data file per game is to not have to add ifdefs to code, so no, that's not good. Add an iwd2 colors.2da. Also, related, it looks like we all forgot or you deleted it, but all the door colors are the same now.

gemrb/unhardcoded/shared/colors.2da
```diff
@@ -5,4 +5,5 @@ TOOLTIP			0xffffffff
 TOOLTIPBG		0x00000000
 MAPICNBG		0xffffffff
 MAPTXTBG		0x00000000
-ALTDOOR		    0xff0000ff
+ALTDOOR			0xff0000ff
+HOVERDOOR               0x1919e1ff
```
lynxlynxlynx, line 9: And again the table entries are misaligned.

lynxlynxlynx:
Do you need any further guidance?

rsantellan:
> Do you need any further guidance?

Not really, I create all the colors.2da in all the folders, but after thinking about it I will put the default for all the idws there so I don't have to copy the shared one to the others. Don't know if it is the best thing. My biggest problem is time... Don't have much..

lynxlynxlynx:
You're right, since all the other games already have colors.2da, you can just edit the shared one to cover the iwd games without creating new files.

rsantellan:
> You're right, since all the other games already have colors.2da, you can just edit the shared one to cover the iwd games without creating new files.

Hi @lynxlynxlynx I think I have it the only thing that I think is a little different is that the on the original the color is softer in the inside.
doors
Apart from that I did all the changes. Is that ok ?

lynxlynxlynx:
Yeah, looks good, thanks! For the last thing you mention, if you want to make it look closer to the original, try increasing the transparency of the color. You can do that by changing the last, "ff", part to something smaller, let's say "dd" or even lower. It will also affect the outline of the outline, but there should be a sweet spot where it's the closest that we can get to being the same.

bradallred:
To be me it looks like the outline is opaque and the inside is 50%. Just add the half trans flag when we draw the inner poly.

rsantellan:
@lynxlynxlynx I will try that. @bradallred I don't know how to do that.

bradallred:
dont worry about it. I see thats what we already do.

rsantellan:
@lynxlynxlynx I have tried putting a dd at the end: **0xff00ffdd**

image

And it look the same.
Also with an aa the same: **0xff00ffaa**

image

For what I can see the issue is that the inner part of the door should have another color that the one from the edges, I don't know if that is the case.
Also in the original the door only has four lines, and in here it have at least six. 
Don't know how to continue....

bradallred:
changing the alpha wont do anything unless you add `BlitFlags::ALPHA_MOD` here

https://github.com/gemrb/gemrb/blob/b76b1dcf5711a39aa73e477bacfcb7f5c8cb15fd/gemrb/core/Scriptable/Scriptable.cpp#L1827-L1832

rsantellan:
@bradallred setting 
**core->GetVideoDriver()->DrawPolygon( outline.get(), origin, outlineColor, true, BlitFlags::ALPHA_MOD|BlitFlags::HALFTRANS );** Makes the door look like this.

image

Don't know what I'm touching with this.... I will have to read the code a lot more to understand.

bradallred:
its not important. we can work on that later.

rsantellan:
> its not important. we can work on that later.

Ok, so I just have to wait for the build to work? Or have to do something more?

lynxlynxlynx:
No, that's it. :) Thanks and congratulations on your first code contribution!

Issue #1389, opened 2021-07-28
Windows build still broken 


lynxlynxlynx:
CreateWindow is still a problem:
https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/40165893#L439

There are some other errors further down, but at least 1 should be fixed by the same thing, since it's just a parsing problem.

bradallred:
@MarcelHB was on the case

MarcelHB:
Looking at Win right now and fixed a lot of things already, but I'm currently stuck on plugins and a linker error:
```
C:/Program Files/MSYS2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/AREImporter.dir/objects.a(AREImporter.cpp.obj): in function `GemRB::ImporterPlugin<GemRB::AREImporter>::ImporterPlugin()':
D:/Sources/gemrb/gemrb/core/Plugin.h:88: undefined reference to `__imp__ZTVN5GemRB14ImporterPluginINS_11AREImporterEEE'
C:/Program Files/MSYS2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/AREImporter.dir/objects.a(AREImporter.cpp.obj): in function `GemRB::ImporterPlugin<GemRB::AREImporter>::~ImporterPlugin()':
D:/Sources/gemrb/gemrb/core/Plugin.h:92: undefined reference to `__imp__ZTVN5GemRB14ImporterPluginINS_11AREImporterEEE'
C:/Program Files/MSYS2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/AREImporter.dir/objects.a(AREImporter.cpp.obj): in function `GemRB::ImporterPlugin<GemRB::AREImporter>::~ImporterPlugin()':
D:/Sources/gemrb/gemrb/core/Plugin.h:92: undefined reference to `__imp__ZTVN5GemRB14ImporterPluginINS_11AREImporterEEE'
```

that translates to: `__imp_vtable for GemRB::ImporterPlugin<GemRB::AREImporter>` but I haven't found the missing virtual or its missing definition ...

bradallred:
[...]

PR #1390, opened 2021-07-28
Making Windows compile again

MarcelHB:
## Description
Bring Windows back to the game. #1389 

## Checklist

### DIFF
--- a/gemrb/core/CharAnimations.cpp
+++ b/gemrb/core/CharAnimations.cpp
@@ -1466,7 +1466,7 @@ void CharAnimations::GetAnimResRef(unsigned char StanceID,
 			NewResRef = AvatarTable[AvatarsRowNum].Prefixes[Part];
 			break;
 		default:
-			error("CharAnimations", "Unknown animation type in avatars.2da row: %ld\n", AvatarsRowNum);
+			error("CharAnimations", "Unknown animation type in avatars.2da row: %lu\n", static_cast<unsigned long>(AvatarsRowNum));
 	}
 }
 
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -3224,7 +3224,7 @@ void Map::TriggerSpawn(Spawn *spawn)
 	//create spawns
 	int difficulty = spawn->Difficulty * core->GetGame()->GetTotalPartyLevel(true);
 	unsigned int spawncount = 0;
-	size_t i = RAND(0LU, spawn->Creatures.size() - 1);
+	size_t i = RAND(size_t(0), spawn->Creatures.size() - 1);
 	while (difficulty >= 0 && spawncount < spawn->Maximum) {
 		if (!SpawnCreature(spawn->Pos, spawn->Creatures[i], 0, 0, spawn->rwdist, &difficulty, &spawncount)) {
 			break;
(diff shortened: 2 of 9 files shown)


MarcelHB:
I just want to whether everything else and AppVeyor/MSVC builds. Still working on GCC as discussed in #1389

lynxlynxlynx:
Looks good so far and hopefully msvc's diagnostic is a bit more useful to you.

MarcelHB:
@bradallred Can you maybe take a look at MSVC errors? To me, it even looks valid since I don't see how https://github.com/gemrb/gemrb/blob/078f266ca8ea56b394f4d268e39a74a769fe8239/gemrb/core/Plugin.h#L90

ever gets a non-abstract plugin type. And then again, `GetImporter` (as in `DialogHandler.cpp`) -> ... -> `MakePluginHolder` makes me think that this `MakeHolder<IMPORTER>()` should not be there like this, and just have the holder injected.

bradallred:
when do you ever see `IMPORTER` passed as an _abstract_ type? the only uses are `AREImporter`, `BAMImporter`, `CHUImporter`, `CREImporter`, `DLGImporter`, `GAMImporter`, and `ITMImporter` which are all concrete.

MarcelHB:
From what I see, if you follow this line for example https://github.com/gemrb/gemrb/blob/078f266ca8ea56b394f4d268e39a74a769fe8239/gemrb/core/DialogHandler.cpp#L110

You end up in something that is this:
```cpp
PluginHolder<ImporterPlugin<DialogMgr>> MakeImporterPluginHolder(PluginID id) {
	return MakePluginHolder<ImporterPlugin<DialogMgr>>(id);
}
```
here at:
https://github.com/gemrb/gemrb/blob/078f266ca8ea56b394f4d268e39a74a769fe8239/gemrb/core/PluginMgr.h#L144-L147

bradallred:
> Looking at this statically, there is a reason to complain then.

which is?

MarcelHB:
The compiler then is probably seeing an attemp to do this:
```cpp
class GEM_EXPORT ImporterPlugin final : public Plugin
{
	Holder<DialogMgr> importer = MakeHolder<DialogMgr>();
```
and this:
```cpp
inline Holder<DialogMgr> MakeHolder( ) noexcept
{
	return Holder<DialogMgr>(new DialogMgr());
}
```
And since `DialogMgr` is abstract, there is the respective error.

bradallred:
the definition of `MakePluginHolder` is nothing more than a cast. There is no object construction. how does MSVC implement `std::static_ptr_cast` thats all we are trying to do here.

https://github.com/gemrb/gemrb/blob/078f266ca8ea56b394f4d268e39a74a769fe8239/gemrb/core/PluginMgr.h#L139-L142

how do you figure

MarcelHB:
Not an invocation, but maybe MSVC is just putting in that type into `T` and discovers an illegal code path? Which then made me wonder if there is something that requires this line at all, given that we ask for a pointer from the plugin manager.

bradallred:
the most straight forward fix (and arguably an upgrade) would be to remove `Held<T>` from the plugin hierarchy and just use `std::shared_ptr` and `std::static_ptr_cast`

MarcelHB:
Ok, feel free to work on it and I'll rebase and continue with the other stuff then. Otherwise I'll try it out tomorrow for myself.

bradallred:
sounds good. can we just merge this now then?

PR #1395, opened 2021-07-30
Plugin: dropping Held

MarcelHB:
## Description
Discussed as possible solution for linker problems mentioned in #1389 -- unfortunately no help for this issue.

## Checklist

### DIFF
--- a/gemrb/core/Audio.h
+++ b/gemrb/core/Audio.h
@@ -111,7 +111,7 @@ public:
 	virtual bool Stop() = 0;
 	virtual bool Pause() = 0;
 	virtual bool Resume() = 0;
-	virtual int CreateStream(Holder<SoundMgr>) = 0;
+	virtual int CreateStream(std::shared_ptr<SoundMgr>) = 0;
 	virtual void UpdateListenerPos(const Point&) = 0;
 	virtual Point GetListenerPos() = 0;
 	virtual bool ReleaseStream(int stream, bool HardStop=false ) = 0;
--- a/gemrb/core/GameData.cpp
+++ b/gemrb/core/GameData.cpp
@@ -198,13 +198,13 @@ int GameData::GetTableIndex(const ResRef &resRef) const
 	return -1;
 }
 /** Gets a Loaded Table by its index, returns NULL on error */
-Holder<TableMgr> GameData::GetTable(size_t index) const
+std::shared_ptr<TableMgr> GameData::GetTable(size_t index) const
 {
 	if (index >= tables.size()) {
-		return NULL;
+		return {};
 	}
 	if (tables[index].refcount == 0) {
-		return NULL;
+		return {};
 	}
 	return tables[index].tm;
 }
@@ -224,7 +224,7 @@ bool GameData::DelTable(unsigned int index)
 	}
 	tables[index].refcount--;
 	if (tables[index].refcount == 0 && tables[index].tm) {
-		tables[index].tm.release();
+		tables[index].tm.reset();
 	}
 	return true;
 }
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -1205,7 +1205,7 @@ int Interface::Init(InterfaceConfig* cfg)
 
 	Log(MESSAGE, "Core", "GemRB Core Initialization...");
 	Log(MESSAGE, "Core", "Initializing Video Driver...");
-	video = Holder<Video>(( Video * ) PluginMgr::Get()->GetDriver(&Video::ID, config.VideoDriverName.c_str()));
+	video = std::shared_ptr<Video>(static_cast<Video*>(PluginMgr::Get()->GetDriver(&Video::ID, config.VideoDriverName.c_str())));
 	if (!video) {
 		Log(FATAL, "Core", "No Video Driver Available.");
 		return GEM_ERROR;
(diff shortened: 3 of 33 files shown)


gemrb/core/GameData.cpp
```diff
@@ -198,13 +198,13 @@ int GameData::GetTableIndex(const ResRef &resRef) const
 	return -1;
 }
 /** Gets a Loaded Table by its index, returns NULL on error */
-Holder<TableMgr> GameData::GetTable(size_t index) const
+std::shared_ptr<TableMgr> GameData::GetTable(size_t index) const
 {
 	if (index >= tables.size()) {
-		return NULL;
+		return {};
```
lynxlynxlynx, line 204: Is this just a shorthand for nullptr or does it actually construct an object? We null-check most of these calls later.

MarcelHB: It constructs an object without a managed pointer. But it has an [overloaded bool](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator_bool) so we can leave these checks as-is, since this is false then.

bradallred: `return {};` and `return nullptr` are equivalent, but `NULL` is an integer type.

gemrb/core/Audio.h
```diff
@@ -111,7 +111,7 @@ class GEM_EXPORT Audio : public Plugin {
 	virtual bool Stop() = 0;
 	virtual bool Pause() = 0;
 	virtual bool Resume() = 0;
-	virtual int CreateStream(Holder<SoundMgr>) = 0;
+	virtual int CreateStream(std::shared_ptr<SoundMgr>&) = 0;
```
bradallred, line 114: let's not pass a shared_ptr by reference to another thread. instead `std::move` at the place of assignment.

gemrb/core/Interface.cpp
```diff
@@ -1205,7 +1205,7 @@ int Interface::Init(InterfaceConfig* cfg)
 
 	Log(MESSAGE, "Core", "GemRB Core Initialization...");
 	Log(MESSAGE, "Core", "Initializing Video Driver...");
-	video = Holder<Video>(( Video * ) PluginMgr::Get()->GetDriver(&Video::ID, config.VideoDriverName.c_str()));
+	video = std::shared_ptr<Video>(static_cast<Video*>(PluginMgr::Get()->GetDriver(&Video::ID, config.VideoDriverName.c_str())));
```
bradallred, line 1208: we should be using `static_ptr_cast` instead

gemrb/core/Interface.cpp
```diff
@@ -1478,7 +1478,7 @@ int Interface::Init(InterfaceConfig* cfg)
 	QuitFlag = QF_CHANGESCRIPT;
 
 	Log(MESSAGE, "Core", "Starting up the Sound Driver...");
-	AudioDriver = Holder<Audio>((Audio*) PluginMgr::Get()->GetDriver(&Audio::ID, config.AudioDriverName.c_str()));
+	AudioDriver = std::shared_ptr<Audio>(static_cast<Audio*>(PluginMgr::Get()->GetDriver(&Audio::ID, config.AudioDriverName.c_str())));
```
bradallred, line 1481: we should be using static_ptr_cast instead

gemrb/core/Plugin.h
```diff
@@ -45,12 +46,11 @@ namespace GemRB {
  * Base class for all GemRB plugins
  */
 
-class GEM_EXPORT Plugin : public Held<Plugin>
+class GEM_EXPORT Plugin
 {
 public:
 	// declaring these here so we dont have to GEM_EXPORT every Held<T> for MSVC to be happy
```
bradallred, line 52: we can remove this false comment too

gemrb/core/Plugin.h
```diff
@@ -84,16 +84,31 @@ class GEM_EXPORT ImporterBase : public Held<ImporterBase>
 	}
 };
 
+/* MSVC must not see an abstract type _anywhere_ going into a `new` or `make_shared`. */
+template<class T, typename... ARGS>
```
bradallred, line 88: how are these different from `std::make_shared`?

MarcelHB: They are not. It just stops MSVC from seeing abstract `DialogMgr` & co types on a potential instantiation path.

bradallred: it sounds like you are saying they are _not_ different.

bradallred: we should only be using _any_ make shared (std or our own) on concrete types. The non concrete users should use `std::static_ptr_cast`

gemrb/core/Interface.cpp
MarcelHB: You mean like?
```cpp
std::shared_pointer_cast<Video>(std::shared_ptr<Plugin>(PluginMgr...));
```
I just wonder whether this makes any difference here.

gemrb/core/Plugin.h
MarcelHB: I can rename that thing into `MakeImporter` or so since I, for now, only see this as a the only working way (with MS) to compile `ImporterPlugin` with either type appearance, so there is nothing we'd wanna be doing with non-concrete types here anyway. If I wanted to generally use it across the codebase, sure.

bradallred: so this is only used on the line below then? yeah, lets rename it to not conflate it.

bradallred:
wanna just fix the conflicts and I can merge this and continue to play around?

MarcelHB:
Here we are.

lynxlynxlynx:
Now the build breaks also on mac:
> gemrb/core/Plugin.h:96:22: error: unused parameter 'args' [-Werror,-Wunused-parameter]
> MakeShared(ARGS&&... args) noexcept

MarcelHB:
That's fixed now.

PR #1396, opened 2021-07-31
Map: pack lightmap/heightmap/materialmap/searchmap

bradallred:
into a single 32bit sprite. This will give great cache locality for everything we need to lookup in a given frame.
if we want we could add a way to dump it to a file for even better debugging

add enhanced DebugOverlay capable of displaying all maps

## Checklist

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -647,14 +647,6 @@ void GameControl::DrawSelf(Region screen, const Region& /*clip*/)
 		}
 	}
 
-	// Draw lightmap
-	if (DebugFlags & DEBUG_SHOW_LIGHTMAP) {
-		Holder<Sprite2D> spr = area->LightMap;
-		video->BlitSprite(spr, Point());
-		Region point(gameMousePos.x / 16, gameMousePos.y / 12, 2, 2);
-		video->DrawRect(point, ColorRed);
-	}
-
 	if (core->HasFeature(GF_ONSCREEN_TEXT) && DisplayText) {
 		core->GetTextFont()->Print(screen, *DisplayText, IE_FONT_ALIGN_CENTER | IE_FONT_ALIGN_MIDDLE, {core->InfoTextColor, ColorBlack});
 		if (!(DialogueFlags & DF_FREEZE_SCRIPTS)) {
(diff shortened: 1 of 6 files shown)


bradallred:
light map needs some work still it seems

gemrb/core/GUI/GameControl.cpp
```diff
@@ -1086,12 +1078,34 @@ bool GameControl::OnKeyRelease(const KeyboardEvent& Key, unsigned short Mod)
...
+						0,
+						DEBUG_SHOW_LIGHTMAP,
+						DEBUG_SHOW_HEIGHTMAP,
+						DEBUG_SHOW_MATERIALMAP,
+						DEBUG_SHOW_SEARCHMAP,
+					};
+					constexpr uint32_t mask = (DEBUG_SHOW_LIGHTMAP | DEBUG_SHOW_HEIGHTMAP
+											   | DEBUG_SHOW_MATERIALMAP | DEBUG_SHOW_SEARCHMAP);
+					
+					static uint32_t flagIdx = 0;
```
lynxlynxlynx, line 1094: Please start at 4, since the searchmap is the only really useful one.

gemrb/core/Map.cpp
```diff
@@ -2400,14 +2431,39 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
...
+	if (mapSize.PointInside(tilePos)) {
+		// Heightmaps are greyscale images where the top of the world is white and the bottom is black.
+		// this covers the range -7 – +7
+		// since the image is grey we can use any channel for the mapping
+		int val = tileProps->GetPixel(tilePos).r;
+		constexpr uint8_t input_range = 255;
+		constexpr uint8_t output_range = 14;
+
+		auto ret = val * output_range / input_range - 7;
+		assert(ret >= -7 && ret <= 7);
```
lynxlynxlynx, line 2444: It doesn't look like this can ever false, so you can remove it.

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -329,13 +329,9 @@ bool AREImporter::ChangeMap(Map *map, bool day_or_night)
...
 
-	// night small map is *optional*!
-	if (!sm) {
-		//fall back to day minimap
-		sm = GetResourceHolder<ImageMgr>(map->WEDResRef);
-	}
-	
 	if (sm) {
+		// night small map is *optional*!
+		// keep the exising map if this one is null
```
lynxlynxlynx, line 334: Doesn't this break if you start in an area at night?

bradallred: I dont think so. If you start at night then you should already have the night map (or normal map if there is no night). My understanding was there is _always_ a day map, just not always a night map.

lynxlynxlynx: In the WED sense, yes, but it remains to be shown for the minimap. So it's easier to reinstate the fallback.

bradallred: I dont understand. if `sm` is found, great. if its not just keep the existing one. why is that broken? this is different code than loads the map initially.

bradallred:
I tested all the games I have and didnt notice anything unexpected

PR #1401, opened 2021-08-03
make ieVariable a typedef of FixedWidthString

bradallred:
IESDP says it is case insensitive. Is that accurate? we have lots of code that was doing lower/upper case copies.

### DIFF
--- a/gemrb/core/GameScript/Actions.cpp
+++ b/gemrb/core/GameScript/Actions.cpp
@@ -5834,7 +5834,7 @@ void GameScript::AttachTransitionToDoor(Scriptable* Sender, Action* parameters)
 		return;
 	}
 	Door* door = ( Door* ) tar;
-	strnspccpy(door->LinkedInfo, parameters->string0Parameter, 32);
+	strnspccpy(door->LinkedInfo.begin(), parameters->string0Parameter, 32);
 }
 
 /*getting a handle of a temporary actor resource to copy its selected attributes*/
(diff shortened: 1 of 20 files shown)


lynxlynxlynx:
I don't think that's true, where did you read it? What was true is that variables for script use were case insensitive, but we use the type in other places as well.

bradallred:
I read it [here](https://gibberlings3.github.io/iesdp/appendices/variables.htm)

but also our own code here
https://github.com/gemrb/gemrb/blob/f29154ee577d515a562796b36a21aa9329a8b2b4/gemrb/core/Variables.cpp#L31-L82

lynxlynxlynx:
Variables are not ieVariables, so that's a bit of a tangent. We use them for our own ini dict, token dict and some other places.

bradallred:
I understand that. What I'm saying is, we dont have to shoehorn in the same string type for all these uses. If we want a fixed width string for a case sensitive purpose we can do so easily: `using INIKey = FixedWidthString<N, strncmp>`

lynxlynxlynx:
Sure, though I don't see much benefit doing that on types with the same length. ieVariable has a small number of users and it's not even clear to me if I'd consider some fields variables or not. So I wouldn't fragment it further.

bradallred:
well if strings are the same length, but used in such different ways that their case sensitivity is not the same it is a good reason to distinguish between them.

What users do we have that _need_ case sensitivity? Specifically I mean case sensitive comparison, the string itself does nothing to upper/lower case the contents. That is up to the creator.

lynxlynxlynx:
Everything that we read and save back should be case-preserved, due to our longstanding issue of savegame (in)compatibility. In a sense this sort of work is too early.

bradallred:
yes, there is work to do (marked as a WIP now) but in general reading and writing are fine. The issues would come from intermediate modifications. Other than `Variables` and `GUIScript` I dont know where that might be happening, but if it is happening it is already an issue. I cannot think of anything that _requires_ case sensitive comparisons tho. I would certainly argue against INI keys being case sensitive. I'm not sure about tokens tho.

I can open an issue for deciphering all of this and documenting the various uses of `ieVariable` so we can better see the consequences of change and identify what, if anything, needs a different type. Even if we only need a single underlying `FixedSizeString` typedefs can still be useful for distinguishing use/intent and make code easier to reason about and change when needed.

Issue #1405, opened 2021-08-04
[py3.9] Fail to build


FrElvire:
#### Bug description
With python 3.9 (ok with 2.7), build failed

```
 [ 87%] Building CXX object gemrb/plugins/TLKImporter/CMakeFiles/TLKImporter.dir/TLKImporter.cpp.o
/home/elvire/gemrb/gemrb/gemrb/plugins/GUIScript/GUIScript.cpp: In function ‘const GemRB::ScriptingRefBase* GetScriptingRef(PyObject*)’:
/home/elvire/gemrb/gemrb/gemrb/plugins/GUIScript/GUIScript.cpp:189:44: error: conversion from ‘GemRB::PyStringWrapper’ to non-scalar type ‘GemRB::ScriptingGroup_t’ {aka ‘FixedSizeString<15, strncasecmp>’} requested
  189 |  ScriptingGroup_t group = PyString_AsString(attr);
      |  
                        ~~~~~~~~~~~~~~~~~^~~~~~
``` 

If this version of python is not supported please give me the one to prevent me from opening these sort of bug report.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

Issue #1285, opened 2021-06-13
Switch to python3 on build bots


lynxlynxlynx:
Something for after 0.9.0:

- [ ] github builders and coverity checker
- [ ] appveyor
- [ ] cmake: python code in the appimage target
- [ ] gemrb/vcpkg_deps.cmake

PR #1406, opened 2021-08-04
Add PYTHON_VERSION configure option

h3xx:
Allows one to request a specific Python version to link against using `-DPYTHON_VERSION=3`. Defaults to `Auto` which means to try 2.3, then 3, then fail, which is what it does currently.

**Does not change the default behavior if this option is not specified.**


h3xx:
> Let's take the opportunity and default to version 3 though. We might have a minimum requirement on 3.2 like many others, but I don't know and it doesn't really matter.

@lynxlynxlynx Good idea -- Python 2 *is* EOL since April 2020, so it makes sense to let it go.

Unfortunately, I just tried compiling gemrb against Python 3, and it's giving me errors. Working on a fix for a FTBFS right now.

lynxlynxlynx:
Yeah, it was just reported — is it the same as #1405?

h3xx:
@lynxlynxlynx Yes, that's the same issue I ran into. I have a fix that at least fixes the compile issue here: https://github.com/h3xx/gemrb/commits/fix-python3-type-error

lynxlynxlynx:
Please open a separate PR for that then. I suspect it won't build on py2, but the bots will test that for us.

h3xx:
Added #1407.

This PR (i.e. the page you're on, #1406) should be ready to merge as it doesn't change anything except give you the option to try compiling it against Python 3.

lynxlynxlynx:
I still think switching the default is a good idea, but I haven't thought about the build bots. We do plan to switch the bots over to py3 by default in this cycle as well (#1285), but until we fix the last known issue of broken state fonts, that's not going to happen. So it'd be pointless to now force py2 just to revert that later. So yeah, all good.

Issue #1425, opened 2021-08-11
Compiler deprecation complaints with newer Python 3 


MarcelHB:
With Python 3.9, I get a compiler problem:
```
In file included from C:/Program Files/MSYS2/mingw64/include/python3.9/unicodeobject.h:1026,
                 from C:/Program Files/MSYS2/mingw64/include/python3.9/Python.h:106,
                 from D:/Sources/gemrb/gemrb/plugins/GUIScript/GUIScript.h:25,
                 from D:/Sources/gemrb/gemrb/plugins/GUIScript/PythonConversions.h:23,
                 from D:/Sources/gemrb/gemrb/plugins/GUIScript/PythonConversions.cpp:19:
D:/Sources/gemrb/gemrb/plugins/GUIScript/PythonConversions.cpp: In function 'GemRB::PyStringWrapper GemRB::PyString_AsString(PyObject*)':
C:/Program Files/MSYS2/mingw64/include/python3.9/cpython/unicodeobject.h:279:45: error: 'Py_UNICODE* PyUnicode_AsUnicode(PyObject*)' is deprecated [-Werror=deprecated-declarations]
  279 |       PyUnicode_AsUnicode(_PyObject_CAST(op)))
      |                                             ^
C:/Program Files/MSYS2/mingw64/include/python3.9/cpython/unicodeobject.h:283:21: note: in expansion of macro 'PyUnicode_AS_UNICODE'
  283 |     ((const char *)(PyUnicode_AS_UNICODE(op)))
      |                     ^~~~~~~~~~~~~~~~~~~~
D:/Sources/gemrb/gemrb/plugins/GUIScript/PythonConversions.cpp:129:15: note: in expansion of macro 'PyUnicode_AS_DATA'
  129 |    wrap.str = PyUnicode_AS_DATA(obj);
      |               ^~~~~~~~~~~~~~~~~
C:/Program Files/MSYS2/mingw64/include/python3.9/cpython/unicodeobject.h:580:45: note: declared here
  580 | Py_DEPRECATED(3.3) PyAPI_FUNC(Py_UNICODE *) PyUnicode_AsUnicode(
      |                                             ^~~~~~~~~~~~~~~~~~~
```
[
According to the docs](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUnicode), this will be removed by v3.10 or v3.12.

bradallred:
this probably shouldn't stop us from making Python 3 the default now.

PR #1443, opened 2021-08-20
Py3 defaults

lynxlynxlynx:
Seeing if it builds.

~@bradallred we don't _brew_ anything for python2 on macs - is v3 also available by default?~ Looks like it.

#1425 is a blocker.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,7 @@ skip_commits:
   message: /\[ci skip\]/
 only_commits:
   files:
+    - CMakeLists.txt
     - gemrb/**/*
     - cmake/**/*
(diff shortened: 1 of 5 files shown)


bradallred:
hmmm, it doesnt come installed as part of the system so I dont know where it is coming from.

bradallred:
oh, maybe its part of the developer tools

lynxlynxlynx:
Probably just preinstalled on the virtual machine.

lynxlynxlynx:
This is now ready, but breaks the mac build, because it ships currently too new python. Windows is broken as before and doesn't get to the plugin.

Issue #1462, opened 2021-08-25
Leveling up doesn't cause a portrait refresh


bradallred:
This should be an easy one for somebody. It appears to be a simple case of a missing `core->SetEventFlag(EF_PORTRAIT);` somewhere. That somewhere being wherever experience is being added to a PC.

#### Steps to reproduce
1. open console and cheat enough xp to level up (`sps(IE_XP, 999999)`)
2. observe the '+' icon does not appear on the portrait
3. you can drag the portrait to another slot to swap them forcing a refresh
4. see that you now have the '+' icon

#### Expected behavior
When a character levels up the portrait refresh bit is set

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

PR #1480, opened 2021-09-09
Fixes issue where portrait would not refresh on level up.

TimHyers:
## Description

This should fix issue #1462 where portraits were not automatically refreshing when a character leveled up.

## Checklist


lynxlynxlynx:
Thanks!

rsantellan:
This happen to me on latest master.

image

lynxlynxlynx:
Works for me and I don't see anything wrong in the screenshot — you can level up the mage.

rsantellan:
> Works for me and I don't see anything wrong in the screenshot — you can level up the mage.

There is no + on the portrait on the sidebar. Is that ok ?

lynxlynxlynx:
Ah, good catch. It looks like a regression in the icon part though, unrelated to this.

PR #1481, opened 2021-09-16
Give windows their own variable context

bradallred:
1. dont needlessly pollute globals. Removes the need to "unset" variables that are only used while the window is active
2. prevent 2 windows controls from conflicting with each other by using the same variable name
3. easily allows implementing of "canceling" something instead of committing it (eg game settings)

## Checklist


bradallred:
I decided to push this up as a PR so that I might get some help identifying and fixing the remaining breakage. The fixes should be pretty straight forward and have largely been either adding parameters to callbacks, or changing `GemRB.GetVar`/`GemRB.SetVar` to the `Window` equivalents.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -54,14 +53,13 @@
...
 total_income = 0
-if GameCheck.IsIWD2():
-	ItemButtonCount = 6
-else:
-	ItemButtonCount = 4
-RepModTable = None
-SpellTable = None
 BarteringPC = 0
 
+ItemButtonCount = 4 if GameCheck.IsIWD2() else 6
```
lynxlynxlynx, line 55: you got it the other way around

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -54,14 +53,13 @@
...
+CureTable = GemRB.LoadTable("speldesc", True) # additional info not supported by core (PST)
+RepModTable = GemRB.LoadTable ("repmodst", True) # all but PST; FIXME: does IWD2 sufffer a 10x price increase?
```
lynxlynxlynx, line 57: well yes, that's why it was ifdefed

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -121,12 +118,9 @@ def CloseStoreWindow ():
...
 	
-	if GemRB.GetVar ("Inventory"):
-		GemRB.SetVar ("Inventory", 0)
-		GUIINV.OpenInventoryWindow ()
-	else:
+	if GemRB.GetView("WIN_INV") == None:
 		GemRB.GamePause (0, 3)
-		GUICommonWindows.SetSelectionChangeHandler( None )
+	else:
 		GUICommonWindows.CloseTopWindow()
```
lynxlynxlynx, line 124: isn't bag use broken now, since both windows are top windows, so the inventory got closed on store entry and we're not reopening it? So even the view fetch itself would fail here?
Also, since the store window is really two windows, doesn't the other one still need closing?

gemrb/GUIScripts/GUISTORE.py
bradallred: right, but why is the data seemingly wrong? IESDP makes it seem like IWD2 data behaves identically to the others. If we really want to ignore the file seems like we ought to override it with an empty one, at least then it is mod-able.

lynxlynxlynx: like the commit says, it's a leftover from the previous game they started with (bg2). IWDs don't have any reputation system at all. But yeah, a special 2da override would also work, though since it has to be an override, it's not easily moddable except manually.

Issue #1250, opened 2021-05-21
Shops: add red overlay for unsellable items 


lynxlynxlynx:
Merchants can be picky and we should be more obvious about it. #426 The original uses a red tint:
https://imgur.com/Bkzar4H

All the code is in GUISTORE.py and the buttons should just get an overlay like for unusable items.

rsantellan:
After some tinkering I have:
![image](https://user-images.githubusercontent.com/1820598/131235629-b230067d-d175-4cb7-80c7-528f415b1602.png)
That is better than use to be:
![image](https://user-images.githubusercontent.com/1820598/131235645-f678d0ae-ff98-4a33-8df6-f66035fa6b88.png)
But not really like should be:
![image](https://user-images.githubusercontent.com/1820598/131235650-8cb09d7e-f9c5-4a9b-8ba1-b3811e46112d.png)
The problem is that the item do not have colors, so is more than a layout to it.
I was looking and is a function to change colors:
https://github.com/gemrb/gemrb/blob/adeda7953862c6258889a9e961d09ca004b43e76/gemrb/core/Video/Video.cpp#L338
It this a correct way to aproach the change of color of the item to a more sepia one ?
Going down the rabbit hole.....

lynxlynxlynx:
That would clearly require different blending, so yeah, by not going the red overlay route, you've made the problem harder.

rsantellan:
@bradallred  one question about this:
https://github.com/gemrb/gemrb/blob/adeda7953862c6258889a9e961d09ca004b43e76/gemrb/core/Video/Video.cpp#L348
It was supposed to pass an extra flag to show the sepia and gray scale ? Why the todo is there ?

bradallred:
the TODO is there because we don't handle `BlitFlags::GREY` or `BlitFlags::SEPIA` for drawing primitives, Shouldn't be relevant tho, I cant see why you would need to.

bradallred:
`Button.SetOverlay` is terribly named and isn't what you want. You want `Button.EnableBorder` (which is also terribly named). Honestly, the 2 should be combined somehow. Really all we need is a set of rectangles with colors which would be much simpler and a lot more flexible.

rsantellan:
@bradallred the problem is that the item maintain the color, so I was thinking on taking out the color of the item and that with the EnableBorder should accomplish it. But perhaps I'm over-complicating a simple issue.

bradallred:
[...]

PR #1482, opened 2021-09-17
 Shops: add red overlay for unsellable items #1250 

rsantellan:
## Description
Add a darker overlay for unsellable items.

## Checklist


gemrb/GUIScripts/GUISTORE.py
```diff
@@ -1738,6 +1738,11 @@ def SetupItems (pc, Slot, Button, Label, i, storetype, idx, steal=0):
 				#containers are always clickable
 				Button.SetState (IE_GUI_BUTTON_ENABLED)
 			else:
+				color = {'r': 61, 'g': 47, 'b': 24, 'a': 100}
+				Button.SetBorder (0, color, 1,1)
+				Button.SetBorder (1, color, 1,1)
+				Button.SetBorder (2, color, 1,1)
+				Button.SetBorder (3, color, 1,1)
```
lynxlynxlynx, line 1745: Why are you setting all the borders — isn't one enough?

bradallred:
we should squish this into one commit

gemrb/GUIScripts/GUISTORE.py
rsantellan: I thought that you have to put all the borders to work, I should try with only one. Should it be the number 0 or one specific ?

lynxlynxlynx: Use the same one we use in the inventory, which would be 2. Just in case the other uses get triggered.

rsantellan: Change it to leave only index 2.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -1738,6 +1738,8 @@ def SetupItems (pc, Slot, Button, Label, i, storetype, idx, steal=0):
 				#containers are always clickable
 				Button.SetState (IE_GUI_BUTTON_ENABLED)
 			else:
+				color = {'r': 61, 'g': 47, 'b': 24, 'a': 100}
+				Button.SetBorder (2, color, 1,1)
```
lynxlynxlynx, line 1742: One other thing. Does it still work properly (no overlay) when you scroll the item list or change to a character that can use the item?

rsantellan: Is only applied on the right side, on the item that you own. Is the ones that they will not buy, when the character can't use them is already implemented on the red. Scrolling work good. Don't know how to test the other part, also I mess up my local repo and after update latest master can't talk to anybody that segfault. #1484

lynxlynxlynx: ok, then it must reset correctly (elsewhere), since you say scrolling works fine. I was worried the overlay would persist on other items that are sellable, since you only explicitly enable it.

lynxlynxlynx:
Thanks!

Issue #1486, opened 2021-09-26
Crash when using TTFs/demo 


MarcelHB:
#### Bug description
Right now, I can't start the demo anywhere as it ends up in a memory corruption when trying to read from a TTF file:
```
[ResourceManager/ERROR]: Couldn't find 'svtriobj.ids'.
hread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffb887e13c8 in GemRB::TTFFontManager::read (stream=<optimized out>,
    offset=36200,
    buffer=0x1f1c72432a0 "\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒", '▒' <repeats 16 times>, "▒▒▒▒▒▒▒▒▒▒▒▒", count=164)
    at D:/Sources/gemrb/gemrb/plugins/TTFImporter/TTFFontManager.cpp:56
56              dstream->Seek((int)offset, GEM_STREAM_START);
(gdb) bt
#0  0x00007ffb887e13c8 in GemRB::TTFFontManager::read (
    stream=<optimized out>, offset=36200,
    buffer=0x1f1c72432a0 "\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒\r▒", '▒' <repeats 16 times>, "▒▒▒▒▒▒▒▒▒▒▒▒", count=164)
    at D:/Sources/gemrb/gemrb/plugins/TTFImporter/TTFFontManager.cpp:56
#1  0x00007ffb88727a37 in ?? ()
   from C:\Program Files\MSYS2\mingw64\bin\libfreetype-6.dll
...
#6  0x00007ffb887e1ff3 in GemRB::TTFFont::GetGlyph (this=0x1f1bdc713e0,
    chr=<optimized out>)
[...]

PR #1488, opened 2021-09-27
TTFManager: saving the file data stream for later

MarcelHB:
A simple fix suggestion for #1486.

On Windows, I can now see that many glyphs are missing, a very few on Linux. But this is a different story I suppose.

## Checklist

### DIFF
--- a/gemrb/plugins/TTFImporter/TTFFontManager.cpp
+++ b/gemrb/plugins/TTFImporter/TTFFontManager.cpp
@@ -55,8 +55,10 @@ unsigned long TTFFontManager::read(FT_Stream		stream,
 
 void TTFFontManager::close( FT_Stream stream )
 {
-	if (stream)
-		free(stream);
+	if (stream) {
+		delete static_cast<DataStream*>(stream->descriptor.pointer);
+		delete stream;
+	}
 }
 
 TTFFontManager::~TTFFontManager(void)
@@ -81,7 +83,7 @@ bool TTFFontManager::Import(DataStream* stream)
 	if (stream) {
 		FT_Error error;
 
-		ftStream = (FT_Stream)calloc(sizeof(*ftStream), 1);
+		ftStream = new std::remove_pointer<FT_Stream>::type{};
 		ftStream->read = read;
 		ftStream->close = close;
 		ftStream->descriptor.pointer = stream;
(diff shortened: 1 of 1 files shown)


bradallred:
we cant change that `free` to `delete`, that is undefined behavior. we need to match `free` with `calloc` or change to `new`

MarcelHB:
Good idea, time for some C-to-C++ beauty.

bradallred:
doesn't `~TTFFontManager` need to call `close()` then? looks like this potentially leaks if you were to destroy `TTFFontManager` without closing it first.

MarcelHB:
Hm, the resource so far is effectively owned by FT2 that calls static `close` at some point over being given as a function pointer (line 88), I guess somewhere in `FT_Done_Face` that is in the d'tor. But unless we have succesfully created an FT2 face (line 98), we haven't moved away the pointer (line 106). So unless FT2 has taken over, the `ImporterBase` d'tor is doing the job instead.

bradallred:
I meant the non static `TTFFontManager::Close`, which looks like it calls the other along with `FT_Done_Face`. We already call `Close()` at the top of `TTFFontManager::Import` so I dont see why calling it in the destructor as well would be an issue.

MarcelHB:
In `TTFFontManager::Close` it's the `FT_Stream` stream, not the `DataStream` stream. And since `FT_Stream` is `null` at least in the first call on top of `Import`, nothing is happening there, but in the destructor.

I'll check with Valgrind.

bradallred:
then we have more problems. if an object is allocating memory, that memory needs to be freed when the object goes out of scope.

Calling `Close()` in the destructor (or at any other point in the lifecycle) _should not_ cause a problem, Im not making a statement about how things are, I'm suggesting how they should be.

It's also an unpleasant smell if we are leaving memory to be cleaned up at some undetermined period of time beyond the objects lifecycle. I don't even know why we have a static `close()`, the `FT_Stream` is a member of the object, not class. seems like we should just not use `FT_Stream_CloseFunc`

MarcelHB:
Yeah, I'm with you and suspect that there is some not-so-good lifetime management biting us here, also because I now spot gaps in the font use, as if not all glyphs have been extracted.

For now, I just haven't found time to go deeper how exactly a font manager, a font and the Freetype resources are spending their lifetime, or how they are supposed to under the current setup.

bradallred:
ok, I can just merge this for now and we can circle back. I dont have time right now either.

Issue #1378, opened 2021-07-24
Simplify Map::SortQueues


lynxlynxlynx:
`cstdlib` provides `qsort`, so surely we don't need to reimplement it.

https://www.cplusplus.com/reference/cstdlib/qsort/

bradallred:
we can probably just use `std::sort` too. AFIK there is no reason to prefer `qsort` specifically

rsantellan:
Looking at the code:
 https://github.com/gemrb/gemrb/blob/4e0fcacb97b7f39156cc91aa1145ad0400227c39/gemrb/core/Map.cpp#L2848
The ordering condition is always the **Y** position ?

rsantellan:
> Looking at the code:
> 
> https://github.com/gemrb/gemrb/blob/4e0fcacb97b7f39156cc91aa1145ad0400227c39/gemrb/core/Map.cpp#L2848
> 
> The ordering condition is always the **Y** position ?

And what will be a good test to do to the function ?

lynxlynxlynx:
You can print the global IDs of the Scriptables in the final queue and abort immediately, then compare the order before the actual refactoring and afterwards.

PR #1493, opened 2021-10-05
Change to std::sort

rsantellan:
## Description
This merge is for simplify #1378

## Checklist

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -2825,7 +2825,7 @@ void Map::GenerateQueues()
 void Map::SortQueues() const
 {
 	struct {
-         bool operator()(Actor* a, Actor* b) const { return a->Pos.y < b->Pos.y; }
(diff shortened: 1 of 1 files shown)


gemrb/core/Map.cpp
```diff
@@ -2824,40 +2824,15 @@ void Map::GenerateQueues()
 //the original qsort implementation was flawed
 void Map::SortQueues() const
 {
+	struct {
+         bool operator()(Actor* a, Actor* b) const { return a->Pos.y < b->Pos.y; }
```
lynxlynxlynx, line 2828: Both parameters should be const.

lynxlynxlynx:
Thanks, it looks ok. What sort of testing did you do?

rsantellan:
@lynxlynxlynx I did 
```
		for (int j=0;j<Qcount[q];j++) {
			Actor* actor = baseline[j];
			StringBuffer text = StringBuffer();
			text.append(std::to_string(actor->Pos.y));
			Log(MESSAGE, "Map", text);
		}
```
On master and in this branch to test it was the same.

Issue #1420, opened 2021-08-08
 BG1/BG2: Subtitles appear even with the setting turned off


Laulajatar:
#### Bug description
Even with the subtitle option checked off, the texts from party members saying their idle or selection dialogue gets dumped into the chat log.

#### Steps to reproduce
1. Disable subtitles
2. Stand around make a coffee
3. Check chat log

#### Expected behavior
When subtitles are turned off, they shouldn't appear in the log.

#### Screenshots
![Screenshot_2021-08-08_14-16-11](https://user-images.githubusercontent.com/68446748/128631677-d4714e3a-a3c3-4983-bf32-8487c488434a.png)
![Screenshot_2021-08-08_14-16-19](https://user-images.githubusercontent.com/68446748/128631680-fbe738cb-c2ba-42a0-8015-979067951b49.png)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.0
- [x] 0.8.8
(Just didn't test 0.9.0)

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Ok, so this is the "Subtitles" key (out of three possible), which we do check in DisplayStringCore. I guess it should effectively toggle DS_CONSOLE.

Issue #757, opened 2020-05-25
bg2: journal text has the wrong color 


lynxlynxlynx:
For the journal journal section, titles should be green:
https://i2.wp.com/lparchive.org/Baldurs-Gate-2-and-Throne-of-Bhaal/Update%2060/26-BG2ToBch8023.jpg

Something like #003d00. Blocked by #308

bradallred:
I checked the original and they are the red color we use now. I don't know where this green comes from. I wonder if it is dependent on the entry somehow..

lynxlynxlynx:
No, that's the quest log.

PR #1502, opened 2021-12-06
Change if subtitles option is disable will not show the pc text on the chat log

rsantellan:
## Description
This will fix the option of enabling and disabling subtitles on the chat log. Issue #1420

## Checklist

### DIFF
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -435,6 +435,9 @@ void DisplayStringCore(Scriptable* const Sender, int Strref, int flags)
 	unsigned int channel = SFX_CHAN_DIALOG;
 
 	Log(MESSAGE, "GameScript", "Displaying string on: %s", Sender->GetScriptName() );
+	// Check if subtitles are not enabled
+	ieDword charactersubtitles = 0;
+	core->GetDictionary()->Lookup("Subtitles", charactersubtitles);
 	if (flags & DS_CONST) {
 		if (Sender->Type!=ST_ACTOR) {
 			Log(ERROR, "GameScript", "Verbal constant not supported for non actors!");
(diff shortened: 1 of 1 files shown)


gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -485,6 +485,11 @@ void DisplayStringCore(Scriptable* const Sender, int Strref, int flags)
 			strlcpy(Sound, sb.Sound, sizeof(Sound));
 		}
 		if (sb.text) {
+			ieDword charactersubtitles = 0;
+			core->GetDictionary()->Lookup("Subtitles", charactersubtitles);
+			if (!charactersubtitles) {
+				flags &= ~DS_CONSOLE;
+			}
```
lynxlynxlynx, line 492: It has been looked up already around line 462, so why not just toggle it there? Did you find some other use than what was mentioned in the bug that requires us to always disable?

rsantellan: It never enters the if https://github.com/gemrb/gemrb/blob/b94b3069cb09fd4aa49ddc1526c177131432c12a/gemrb/core/GameScript/GSUtils.cpp#L438
So I put it when it was showing the text.

lynxlynxlynx: What are you testing with? Both idle sounds and selection sounds are verbal constants.

rsantellan: I have followed the steps:
Steps to reproduce

    Disable subtitles
    Stand around make a coffee
    Check chat log

lynxlynxlynx: Checked the code and I see, it depends on what kind of actor is running it. Still, there's no need to look up the configuration twice. And you can do the actual check in the same block that disables it for PST.

rsantellan: Change to put it on the upper if of PST.

lynxlynxlynx: You can move it up further, so the lookup at 462 isn't needed.

rsantellan: @lynxlynxlynx move it above the if for be called only once.

lynxlynxlynx:
Thanks!

rsantellan:
> Thanks!

Thank you!

lynxlynxlynx:
If you want to help with something else, #757 is even simpler.

PR #1504, opened 2021-12-13
Added color depending if it is journal or quests.

rsantellan:
## Description
Fix colors of issue #757

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg2/GUIJRNL.py
+++ b/gemrb/GUIScripts/bg2/GUIJRNL.py
@@ -148,6 +148,9 @@ def UpdateLogWindow (JournalWindow):
 		Label.SetText (45202)
 		list.sort(key=SortByDate)
 
+	Color = "800000"
+	if Section == 4:
+		Color = "003d00"
 	for i in range(len(list)):
 		je = list[i]
 		hours = je['GameTime'] // 4500
@@ -162,7 +165,7 @@ def UpdateLogWindow (JournalWindow):
 		# each journal entry consists of the title and description
 		# but the game displays the entry date between the two
 		je2 = GemRB.GetString(je['Text']).split("\n",1)
-		JournalTitle = "\n[color=800000]" + je2[0] + "[/color]\n"
+		JournalTitle = "\n[color=" + Color + "]" + je2[0] + "[/color]\n"
 		if len(je2) == 1:
 			# broken entry, bg1 style (no title)
 			JournalText = ""
--- a/gemrb/GUIScripts/bg2/GUIJRNL.py
+++ b/gemrb/GUIScripts/bg2/GUIJRNL.py
@@ -149,6 +149,7 @@ def UpdateLogWindow (JournalWindow):
 		list.sort(key=SortByDate)
 
 	Color = "800000"
+	# Section with number 4 is defined on the Journal Button.
 	if Section == 4:
 		Color = "003d00"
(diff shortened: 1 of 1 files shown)


gemrb/GUIScripts/bg2/GUIJRNL.py
```diff
@@ -148,6 +148,9 @@ def UpdateLogWindow (JournalWindow):
 		Label.SetText (45202)
 		list.sort(key=SortByDate)
 
+	Color = "800000"
+	if Section == 4:
```
Mingun, line 152: It will be valuable to add a comment what is `4` magic constant here

rsantellan: @Mingun is defined on the same file: https://github.com/gemrb/gemrb/blob/33b898f32d5957ddb044552159a176d3d0a55ba1/gemrb/GUIScripts/bg2/GUIJRNL.py#L71 
I think it will be redundant but I can put a comment.

Mingun: - Because there are more that 70 lines between that lines it is better to add a comment
- In the future code could be refactored and both lines may turn out to be in different files and that link would be lost

Maybe even better to add a named constant and use it

rsantellan: @Mingun a few points.

- 70 lines is that not long.
- If the file is split on the future the constant should be on the file or in another place ? 
- Should I move all **Sections** to constants ? Section var is already a **Global** one.

Let me know which is the best way to go.

Regards

lynxlynxlynx: Journals are different among all the games, so there's not much opportunity to share more code. A comment never hurt though.

rsantellan: Added a comment, don't think is a great one. Really don't know how to make it better.

Mingun: Just
```python
if Section == 4: # Journal
```
would be enough, but that is also good

lynxlynxlynx: In general, it's more important to explain the why, not the what, which is often easily read from the code. Here there's not much different. I'd go with "only the personal journal section used a different title color", so readers know other section types have been checked already and the code is complete.

rsantellan: @lynxlynxlynx @Mingun I change the comment, I like to put it above not on the side because the if can longer over the time.

lynxlynxlynx: Eh, that's full of spelling errors, let's just use mine.

gemrb/GUIScripts/bg2/GUIJRNL.py
```diff
@@ -148,6 +148,10 @@ def UpdateLogWindow (JournalWindow):
 		Label.SetText (45202)
 		list.sort(key=SortByDate)
 
+	Color = "800000"
+	# The color is change for been used on the Journal section.
```
lynxlynxlynx, line 152:
```suggestion
	# only the personal journal section used a different title color
```

lynxlynxlynx:
Thanks!

PR #1523, opened 2022-01-19
UTF 16 everywhere

bradallred:
This is a PR to get the ball rolling on #1516 

it's not well tested and i'm sure there are bugs. the branch is pushed to the main repo so anybody can checkout out and help get it finished. Mainly it's just the necessary changes to use the `String` class thought the core.

@fizzet maybe this interests you

PR #1529, opened 2022-01-28
Missing CString header for MSYS2/GCC

MarcelHB:
The only thing that prevents #1523 from building for me.

## Checklist


lynxlynxlynx:
Thanks!

Issue #1527, opened 2022-01-28
%ls formatting needs to be rewritten with libfmt


bradallred:
this is needed to make more progress on #1516

since `String` will be UTF-16 we need to use the `WideToChar` formatter so we can convert to whatever system encoding the logging system uses.

we should probably also take this opportunity to remove `print`.

effectively this means that the logging system needs to be refactored to use libfmt

lynxlynxlynx:
I can look at this if you want. It'd be good if you addressed the drawing regressions in the meantime, since that's completely your area & design.

bradallred:
thanks, I've assigned this to you. I am aware of the drawing issues, but it's not a simple fix and will require some of the [roadmap](https://github.com/gemrb/gemrb/wiki/Graphic-Rendering-Roadmap) to be implemented.

lynxlynxlynx:
We've accrued quite some good work, so I'd like to make a new release within the next few months. Meaning we should try not to get into anything too invasive on master.

lynxlynxlynx:
I'll go function by function, leaving some scaffolding (back to char) until all are converted, so the work can be split and easier to look at. I've started with `error`, since it has fewer users and is a bit easier to regex into submission. 

Checking what would happen in variadic function form: we'd create 20 instances, which sounds like a lot, but I don't know if va_list is eventually any better. Also the linker doesn't like it in variadic function form, as if the GEM_EXPORT attribute wasn't passed along. Any preferences?

bradallred:
I’m not near a computer to find the docs, but libfmt has an “arg list” type you can use to pack all the arguments then pass it to format. We don’t explicitly need to go the variadic argument route.

lynxlynxlynx:
Turns out it's practically unavoidable, as fmt needs type info at compile time.
WideToChar can become quite noisy — I guess we can't add a formatter directly to String easily?

Here's error for any comments:
http://sprunge.us/VMxbVC?diff

bradallred:
> WideToChar can become quite noisy — I guess we can't add a formatter directly to String easily?

I'm not sure what you mean by noisy.

A formatter directly to `String` should be nice, but I couldn't figure out a way to do it since libfmt provides its own formatter for `std::wstring` and `std::u16string` both and I dont know how to make the compiler select our version.

PR #1534, opened 2022-01-30
Use fmt for logging

lynxlynxlynx:
#1527

There's only one notable difference that I noticed — fmt doesn't can't handle displaying null pointers, but besides the audio driver (workaround included), I've not hit that anywhere.


bradallred:
> fmt doesn't can't handle displaying null pointers

this certainly makes sense for an automatic (`{}`) format since it would dereference the pointer, but I would think you could specify something like `{:p}` to have it print at least the address.

lynxlynxlynx:
We'll just have to be a bit more careful, since I guess p would always print the address, which we never need.

bradallred:
if you're talking about cstrings, then we just keep on weeding them out over time :)

lynxlynxlynx:
It was from openal, so not within our powers. :)

lynxlynxlynx:
Eh, should have merged sooner. Will push manually once I resolve.

PR #1538, opened 2022-02-01
Fixing the logging header for Windows

MarcelHB:
Having `GEM_EXPORT` with the now-template definitions misdefines some linking expectations, resulting in:

```
D:/Sources/gemrb/gemrb/core/Logging/Logging.h:45:17: error: function 'void GemRB::error(const char*, const char*, ARGS&& ...)' definition is marked dllimport
   45 | GEM_EXPORT void error(const char* owner, const char* format, ARGS&&... args)
      |                 ^~~~~
D:/Sources/gemrb/gemrb/core/Logging/Logging.h:54:17: error: function 'void GemRB::Log(GemRB::log_level, const char*, const char*, ARGS&& ...)' definition is marked dllimport
   54 | GEM_EXPORT void Log(log_level level, const char* owner, const char* message, ARGS&&... args)
      |                 ^~~
```

## Checklist

### DIFF
--- a/gemrb/core/Logging/Logging.h
+++ b/gemrb/core/Logging/Logging.h
@@ -42,7 +42,7 @@ GEM_EXPORT void LogMsg(Logger::LogMessage&& msg);
 
 /// Log an error and exit.
 template<typename... ARGS> [[noreturn]]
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #1542, opened 2022-02-03
use FixedSizeString for ieVariable

bradallred:
this PR supersedes #1401

- use `FixedSizeString` for ieVariable
- remove strnlwrcpy
- remove strnuprcpy
- remove strnspccpy
- a bunch of changes required to achieve the above 4

### DIFF
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -2491,7 +2491,7 @@ void GameScript::ExecuteAction(Scriptable* Sender, Action* aC)
 
 Trigger* GenerateTrigger(char* String)
 {
-	StringToLower(String);
+	StringToLower(String, String + strlen(String));
 	ScriptDebugLog(ID_TRIGGERS, "Compiling: {}", String);
 
 	int negate = 0;
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -1467,7 +1467,7 @@ void InitializeIEScript()
 		else {
 			idtargets[i]=poi->Function;
 		}
-		ObjectIDSTableNames[i] = MakeLowerCaseResRef(idsname);
+		ObjectIDSTableNames[i] = ResRef::MakeLowerCase(idsname);
 	}
 	MaxObjectNesting = atoi( objNameTable->QueryField( 1 ) );
 	if (MaxObjectNesting<0 || MaxObjectNesting>MAX_NESTING) {
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -1946,8 +1946,8 @@ static Trigger* ReadTrigger(DataStream* stream)
 			&tR->int1Parameter, &tR->int2Parameter, tR->string0Parameter,
 			tR->string1Parameter );
 	}
-	strlwr(tR->string0Parameter);
-	strlwr(tR->string1Parameter);
+	StringToLower(tR->string0Parameter);
+	StringToLower(tR->string1Parameter);
 	tR->triggerID &= 0x3fff;
 	stream->ReadLine( line, 1024 );
 	tR->objectParameter = DecodeObject( line );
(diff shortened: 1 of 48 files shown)


bradallred:
I was hoping this would fix the windows build too, but it looks like there are other issue there now

gemrb/plugins/GAMImporter/GAMImporter.cpp
```diff
@@ -755,16 +755,14 @@ int GAMImporter::PutPlaneLocations(DataStream *stream, Game *game) const
...
 	Variables::iterator pos=NULL;
 	const char *name;
 	ieDword value;
 
 	for (unsigned int i=0;i<KillVarsCount;i++) {
 		//global variables are locals for game, that's why the local/global confusion
 		pos=game->kaputz->GetNextAssoc( pos, name, value);
-		strnspccpy(tmpname, name, 32, core->HasFeature(GF_NO_NEW_VARIABLES));
-		stream->WriteVariable(tmpname);
+		stream->WriteVariable(ieVariable(name));
```
lynxlynxlynx, line 765: Looks like this doesn't preserve the current behaviour. Ignoring the space removal is likely fine, since I guess you checked that we do the same already when reading them, so the runtime vars don't have spaces.

But the uppercasing was deemed important for PST, so we should leave that intact until proven redundant.

gemrb/plugins/TLKImporter/TLKImporter.cpp
```diff
@@ -55,7 +55,7 @@ TLKImporter::TLKImporter(void)
 	for(int i=0;i<gtcount;i++) {
 		ieVariable key;
 
-		strnuprcpy(key, tm->GetRowName(i), sizeof(ieVariable)-1 );
+		strnuprcpy(key.begin(), tm->GetRowName(i), sizeof(ieVariable)-1 );
```
lynxlynxlynx: Is there some reason why all these strnlwrcpy/strnuprcpy/strlcpy calls weren't converted to regular assignment + case change (where applicable)?

bradallred: they are later on; those functions don't even exist anymore. I was merely trying to limit changes for easier commit digestion.

gemrb/core/GameScript/GameScript.cpp
```diff
@@ -2491,7 +2491,7 @@ void GameScript::ExecuteAction(Scriptable* Sender, Action* aC)
 
 Trigger* GenerateTrigger(char* String)
 {
-	StringToLower(String);
+	StringToLower(String, String + strlen(String));
```
lynxlynxlynx: strnlen would be safer, I already see Sonar complaining, but it would be even better to just convert GenerateTrigger to take a std::string.

gemrb/plugins/GAMImporter/GAMImporter.cpp
```diff
@@ -786,13 +786,14 @@ int GAMImporter::PutVariables(DataStream *stream, const Game *game) const
 		if (core->HasFeature(GF_NO_NEW_VARIABLES)) {
 			/* This is one anomaly that must have a space injected (PST crashes otherwise). */
 			if (strcmp("dictionary_githzerai_hjacknir", name) == 0) {
-				strncpy(tmpname.begin(), "DICTIONARY_GITHZERAI_ HJACKNIR", sizeof(ieVariable));
+				// FIXME: this should be moved to where the variable is written to disk
```
lynxlynxlynx, line 789: this is the place, it happens a few lines lower.

gemrb/plugins/GAMImporter/GAMImporter.cpp
```diff
@@ -762,7 +762,7 @@ int GAMImporter::PutKillVars(DataStream *stream, const Game *game) const
 	for (unsigned int i=0;i<KillVarsCount;i++) {
 		//global variables are locals for game, that's why the local/global confusion
 		pos=game->kaputz->GetNextAssoc( pos, name, value);
-		stream->WriteVariable(ieVariable(name));
+		stream->WriteVariableUC(ieVariable(name));
```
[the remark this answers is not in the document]
bradallred: according to the comment above the function definition, this is only called in PST

lynxlynxlynx:
The windows build looks like it still suffers from the RNG template, while the rest of the spam is about deleted operators. So neither are related to this branch, which is good. :)

bradallred:
the problem might be better solved by defining the move ctor+assignment as default.

gemrb/core/GameScript/GameScript.cpp
bradallred: yeah, my thoughts were to kick this toward stringifying GameScript at some point

lynxlynxlynx:
We already had a copy ctor, but sadly adding the move assignment operator didn't appease msvc.

lynxlynxlynx:
I'm away for a few days, so it's hard to review or test, but looking back, before the latest pushes, there was only a redundant or poorly worded fixme left.

PR #1543, opened 2022-02-03
Faster GLSL uniforms lookup

MarcelHB:
I found some oddly bad looking, and unnecessary slow code that's on the hot path when repopulating GLSL uniform variables.

## Checklist


bradallred:
thanks!

PR #1552, opened 2022-02-09
CMake: centralizing the GEM_BUILD_DLL control

MarcelHB:
Some recent changes made linker errors appear in some more plugins on Windows/GCC (probably also MSVC when we have made it until there), like:
```
ld.exe: CMakeFiles/BIFImporter.dir/objects.a(BIFImporter.cpp.obj): in function `GemRB::IndexedArchive::~IndexedArchive()':
D:/Sources/gemrb/gemrb/core/IndexedArchive.h:30: undefined reference to `__imp__ZTVN5GemRB14IndexedArchiveE'
ld.exe: CMakeFiles/BIFImporter.dir/objects.a(BIFImporter.cpp.obj): in function `GemRB::IndexedArchive::IndexedArchive()':
D:/Sources/gemrb/gemrb/core/IndexedArchive.h:30: undefined reference to `__imp__ZTVN5GemRB14IndexedArchiveE
```
Not sure what exactly the trigger was, but `GEM_EXPORT` is not set in the plugins then, hiding these functions. Historically, these case were fixable by putting `ADD_DEFINITIONS(-DGEM_BUILD_DLL)` into _some_ of the respective plugin CMake files, but this is bad since it ignores setting for static/dynamic linking, and we don't need any partially-static-partially-dynamic stuff for any component as well.

I'd like to move it into the root CMake file, respecting the user's setting as far as supported.

## Checklist


lynxlynxlynx:
Thanks!

Issue #1555, opened 2022-02-10
BG: Cannot enter any UI from in-game 


MarcelHB:
When in game, I cannot open any UI such as the inventory or maps, _both_ from the menu bar to the left and hotkey. Other games look good so far.

There is no Python error around. Just affirmation:
```
[KeyMap]: Looking up key: 105(i)
[KeyMap]: RunFunction(GUICommon::ResolveKey)
```

![bg_ui](https://user-images.githubusercontent.com/238558/153469355-603fc431-1510-487e-9e96-1eb3c0fabbdc.gif)

PR #1556, opened 2022-02-11
CString: fixing RTrim 

MarcelHB:
The implementation has problems: First index is oob, and what if the string length is > 127. See #1555

## Checklist


bradallred:
thanks!

Issue #1536, opened 2022-01-31
Major drawing corruption & texture errors


MarcelHB:
#### Bug description

After walking around, two or three areas in IWD2, I noticed flickering (N)PCs and then a lot of rectangles of corrupted screen content. 

Console is full of:
```
[SDLVideo/ERROR]: Invalid texture
```
Starting with Valgrind, the only thing I noticed that looks somewhat related is:
```
==11796== Invalid read of size 4
==11796==    at 0x7045F27: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x7132EA7: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x71337A8: PyTraceBack_Print (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x711F7A1: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x7120255: _PyErr_Display (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x7120499: PyErr_Display (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x712DE7A: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x7045097: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x6FADD8C: _PyEval_EvalFrameDefault (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x6FA4DD2: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x7120634: ??? (in /usr/lib/x86_64-linux-gnu/libpython3.9.so.1.0)
==11796==    by 0x6E9E164: bool GemRB::CallPython<int, &(int GemRB::noop<int>(_object*))>(_object*, _object*, int*) (PythonCallbacks.h:50)
==11796==  Address 0x24fe9020 is 15,552 bytes inside a block of size 16,424 free'd
==11796==    at 0x484621F: free (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #1558, opened 2022-02-11
SDL sprite copy fix

MarcelHB:
The default copy ctor is bad for us.

See #1536

## Checklist


bradallred:
thanks again!

PR #1566, opened 2022-02-18
MSVC branch fixes

MarcelHB:
Notable changes:

- Dropping `GEM_BUILD_DLL` as its actual idea was not clear.
- Introducing `GEM_BUILDING_CORE` to make clear when core classes are needed to be exported, and when imported.
- Removes `GEM_EXPORT` from template classes, as MSVC actually expects import linking otherwise.
- Introduces `GEM_PLUGIN_EXPORT` as always-export attribute for things we need both at building core DLL _and_ plugin DLLs.
- Moves plugin schema headers into `gemrb/includes/Plugins`, since it's header-only and limited to the plugin users.
- Workaround in `IniSpawn` as MSVC complains about ambiguity otherwise


bradallred:
Neat, we'll give it a crack!

Issue #1501, opened 2021-11-12
Migrate rendering to SDL Shader API 


bradallred:
I put together a [graphics rendering roadmap](https://github.com/gemrb/gemrb/wiki/Graphic-Rendering-Roadmap) a while ago detailing how we can solve some of the ongoing issues/performance/complexity with rendering.

Since then SDL has been [awarded a grant](https://www.patreon.com/posts/new-project-top-58563886) from Epic games to develop shader capabilities for `SDL_Renderer`. This is _huge_ news for us.

This would truly enable us to fix all the issues I mentioned in the roadmap without having to force OpenGL on people. We wouldn't have any direct dependency on OpenGL for these benefits, nor would we rely on hacks to control the active shader.

lynxlynxlynx:
That sounds great and I guess it will take a while to get the SDL side sorted out. Maybe a thing for 2023, we'll see.

bradallred:
Ryan has added more information to the SDL [discussion board](https://discourse.libsdl.org/t/future-3d-api/33485)

it looks like it will be a new API entirely. It may not change my roadmap plans much in the end. Time will tell.

lynxlynxlynx:
Btw, I'm not sure how useful it is for anything (besides learning that they use hq4x for scaling), but EEs also ship with a bunch of shader source files.

bradallred:
that could be potentially very useful. lots of our blending is known to be incorrect (magic missile off the top of my head) and shaders can provide clues at the very least.

do you have a link or something?

lynxlynxlynx:
It's plain GLSL files, so if you open the game with NearInfinity, it will all be nicely categorized.

MarcelHB:
I noticed that an RPi 400, with specs that should outnumber old Pentium platforms, really suffers from framedrops in some situations, like when flying over a map.

Profiling shows that `GLSLProgram::SetUniformValue` consumes a significant amount of time, also since its called millions of times. Looking at the current way, the switching between two shaders within one logical draw pass, that requires resetting the uniforms all the times, leaves me at one idea: Certrain things should be grouped for drawing, like glyphs or tiles, so that there is no need to call SDL and the driver with that administrative overhead all the time. (*)

(*) I have no tools right now for deep insights by profiling the GPU on the RPi. It's just that I looked a bit deeper at Vulkan over the last year and it tells the user: really think of how you put your drawable elements together, and that is sometimes orthogonal to what OpenGL 2 allows the user to do.

[...]

Issue #943, opened 2020-09-23
[subviews] GL: write a shader for dithering


lynxlynxlynx:
> right, as discussed in #441 we can't do dithering on SDL2 as it would be astronomically expensive to do without a proper shader. BG:EE doesn't dither either FWIW, just 0-100% opacity. It's been this way for ages now.

_Originally posted by @bradallred in https://github.com/gemrb/gemrb/issues/883#issuecomment-665524517_

bradallred:
this is a reasonably close approximation to master AFICT. I'm not sure if the result is different on non retina displays or if antialiasing is a factor.

lynxlynxlynx:
Yeah, looks fine, thanks.

PR #1587, opened 2022-03-06
Shader refactoring

MarcelHB:
See discussion start in https://github.com/gemrb/gemrb/issues/1501#issuecomment-1060012688


bradallred:
your PR seems to be missing BlitF.glsl

MarcelHB:
My bad, here it is.

bradallred:
looks to me like the problem is `v_color`, which is just an alias for `gl_Color` and that is probably getting set to black as par of another operation. Fog of war simply does a rectangular fill for any tiles that are completely obscured for example.

MarcelHB:
When I drop `v_color` from the shader, that does not help anything wrt to this issue. From a fast scan, this value is set by `SDL_SetTextureColorMod` and its alpha counterpart, and this is always refreshed by us.

bradallred:
We can assume the texture sample is correct or else you would have failure in all circumstances.
So the only way I can see to get black here is if `v_color` (`gl_Color`) is black. You can test this by doing `SDL_SetRenderDrawColor(255, 255, 255, 255)` at the top of `BeginCustomRendering`. You can also ditch all the `SDL_RenderFlush` calls.

bradallred:
the other obvious thing to check is the alpha. if you've left `u_stencil` set, but `s_stencil` isn't then that will probably result in black. You also aren't sampling from the correct color channels using `u_channel`

MarcelHB:
`u_stencil` is always reset if the request's flag doesn't match the known state, or I'm missing some case. The channel isn't correct, yeah, but also not the missing link unfortunately.

bradallred:
I finally got around to checking this out to poke it and yeah, the problem is the `SDL_RenderFlush` remove that and it works other than the stenciling which is hopefully only broken due to the aforementioned color channel issue.

MarcelHB:
I haven't gone much deeper into checking stenciling, also this `glMultiTexCoord` needs some double-checking.

Hm, but to me, this sounds like a bad idea to skip `SDL_RenderFlush`.

bradallred:
> Hm, but to me, this sounds like a bad idea to skip SDL_RenderFlush.

I dont understand this. We cant flush the renderer or SDL will reset things we are depending on. Why would you _want_ to flush it? it just harms performance and gains us nothing now that we have taken over the shader.

MarcelHB:
Let's say SDL was instructed drawing primitives right before, like lines, that are not subject to the texture shader program but one some [more primitive one](https://github.com/libsdl-org/SDL/blob/120c76c84bbce4c1bfed4e9eb74e10678bd83120/src/render/opengl/SDL_shaders_gl.c#L247). When asking for a flush, everything potentially not subject to the texture shader program should be done then.

We then do all the `gl` injection and go continue by `SDL_` calls. If we haven't flushed and there is a draw command for lines in the queue, SDL will eventually (when flushing/executing at a later point) load its line shader program, draw, then switch to the texture shader program, and draw. I'm not sure at this point what will be happening to the uniforms/multi tex samplers that we have set. Some research says it's probably not UB, i. e. we can expect uniforms and samplers to survive, only if the shaders have compatible uniform/sampler layouts.

MarcelHB:
Btw. you really tried dropping the flush, not reinstating the copy? Because doing the first option doesn't even help me, just the second action.

bradallred:
SDL does batching, switching programs implicitly flushes the queue (IIRC from reading the source some time ago). SDL render functions don't call OpenGL directly; they set state inside of SDL render commands and its not until these commands are dispatched that OpenGL is called. As you noted now that we have usurped SDL's `SDL_RenderCopy` shader so from its perspective the book keeping is correct. At least as long as we refrain from calling things like `glColor` or the blending functions directly. All we are (should be) doing is setting uniforms.

> Some research says it's probably not UB, i. e. we can expect uniforms and samplers to survive fully, only if the shaders have compatible uniform/sampler layouts.

yes, but what of the uniform SDL is attempting to set (`tex0`)? I don't know that it matters, just something to remember. I think as long as its the same type of uniform declared in the same order (eg first) than it will work and if it didn't then things wouldn't work at all.

> And yes, since it's possibly a not-so-good impact on performance, this is what I discussed wrt grouping drawings into epochs that are guaranteed to be free of anything that even potentially could interfere with previous SDL drawstates.

SDL already does this and the amount of work to do anything similar is huge. There is much lower hanging fruit, and a lot of it would be required to do that anyway. There is no way you can do what you suggest while keeping palette rendering in software for example. The min we have to reupload a texture to the GPU we lose any ground gained.

> Btw. you really tried dropping the flush, not reinstating the copy? Because doing the first option doesn't even help me, just the second action.

I left the copy commented out and removed the flush. thats it. Of course I loaded a different save, so it's entirely possible the draw order is such that it just kicked the can down the road.

bradallred:
I'm also using the latest release of SDL2 FWIW

bradallred:
Also, to be clear, if we _do_ need to change the OpenGl state beyond setting our uniforms then, yes, we do need `SDL_RenderFlush`.

MarcelHB:
There are actually two shaders in play `SHADER_RGB` and `SHADER_RGBA`, and not switching properly between them, as SDL does depending on the texture format, caused this issue. So now I'm playing with two of them. I try to fix the stenciling now and then evaluate how much benefit is left then.

MarcelHB:
On SDL2 v2.0.10+ everything looks good now. On previous versions, there is only one shader to take care of, and while I made it compile and run, there are visual errors around.

Then again, I'm willing to ignore this for now since RPi OS has now reached Deb 11 with v2.0.14, in contrast to Deb 10 with v2.0.9, and since nobody else complained about RPis for now ...

lynxlynxlynx:
I'm at 2.0.10, so setting that as a minimum is fine by me. :) It's only there for the OpenGL backend anyway and we already warn people not to use 2.0.12.

MarcelHB:
Feel free to test and review.

The only thing I left open for now: do we really want dithering in this mode? Is there somebody insisting on it for better visual quality or something?

> EDIT: what sorts of speedups do you see?

The really bad areas have increased about 50% maybe, but we are talking about an average from 5fps to 8fps.

bradallred:
re dithering, see #943. Are you asking because it's a bottleneck?

> The really bad areas have increased by about 50% maybe, but we are talking about an average from 5fps to 8fps.

Is this with or without `SDL_RenderFlush()`? Also would be interesting for you to try with a release build of SDL2's master branch. There has been a lot of work done to optimize their OpenGL backend since the last release.

> we already warn people not to use 2.0.12

versions later than 2.0.12 should be fine tho

gemrb/plugins/SDLVideo/Shaders/BlitRGB.glsl
```diff
@@ -0,0 +1,31 @@
...
+uniform sampler2D s_stencil;
+uniform int u_greyMode;
+uniform int u_channel;
+uniform int u_stencil;
+
+void main() {
+	vec4 color = texture2D(s_sprite, v_texCoord);
+	gl_FragColor = color;
+	gl_FragColor.a = 1.0;
+	gl_FragColor *= v_color;
```
bradallred, line 16: shouldn't we be multiplying by `v_color` in all cases? the way its written now this will be undone for `u_greyMode`s

I also wonder if there is a noticeable performance difference if we had an `else` instead of setting `gl_FragColor` multiple times.

bradallred:
You might be able to get quite a bit more performance by eliminating the double blit from the `BLIT_STENCIL_MASK` branch. You already eliminated the reason for it, so it should be a simple matter of changing the flags passed to `RenderCopyShadedGL` and removing `std::static_pointer_cast<SDLTextureVideoBuffer>(scratchBuffer)->Clear(drect);` and the block at the bottom:

```cpp
if (flags & (BlitFlags::ALPHA_MOD | BlitFlags::HALFTRANS)) {
	Uint8 alpha = SDL_ALPHA_OPAQUE;
	if (flags & BlitFlags::ALPHA_MOD) {
		alpha = tint->a;
	}
	
	if (flags & BlitFlags::HALFTRANS) {
		alpha /= 2;
	}
	SDL_SetTextureAlphaMod(ScratchBuffer(), alpha);
}
SDL_SetRenderTarget(renderer, CurrentRenderBuffer());
SDL_SetTextureBlendMode(ScratchBuffer(), SDL_BLENDMODE_BLEND);
ret = SDL_RenderCopy(renderer, ScratchBuffer(), &drect, &drect);
```

by remove, I mean `#ifdef` it out of the `USE_OPENGL_BACKEND` build

MarcelHB:
> re dithering, see #943. Are you asking because it's a bottleneck?

No, it's just that I personally would never chose dithering over smooth transparency, so I was just wondering if we maintain this for visual backward compatibility or somebody really wants to see it.

> Is this with or without `SDL_RenderFlush()`? Also would be interesting for you to try with a release build of SDL2's master branch. There has been a lot of work done to optimize their OpenGL backend since the last release.

> by remove, I mean `#ifdef` it out of the `USE_OPENGL_BACKEND` build

I'll have a look at both of your suggesions later or tomorrow.

gemrb/plugins/SDLVideo/Shaders/BlitRGB.glsl
MarcelHB: > shouldn't we be multiplying by `v_color` in all cases? the way its written now this will be undone for `u_greyMode`s

I copied that from the original `SHADER_RGB` code https://github.com/libsdl-org/SDL/blob/1868c5b5210636530d6945c0ec16b11e3506248d/src/render/opengl/SDL_shaders_gl.c#L266-L270 and suspect that we need to keep it unless we want do dome some RE to also maintain color channels aligned with SDL's draw state. Otherwise, I'll probably be failing with something like the black areas as initially.

Partially, yes. But I've hardcoded the alpha to `1.0` there, too. But see below ...

> I also wonder if there is a noticeable performance difference if we had an `else` instead of setting `gl_FragColor` multiple times.

Lecture always tells you -- I cannot verify this -- that if-branching in shaders does not lead to selected execution but always ends up rolling out every code line into the shader cores with discard masks. Anyway, I guess this tiny little shader is even too trivial for small platforms, and all the profiling tools I'm aware of only work with OpenGL v3.1+

MarcelHB:
Some interesting numbers that can be seen when disabling the FPS limiter on high-performance hardware: SDL v2.0.20 vs. current master is about 10% improvement for whatever reason with our OpenGL customizations.

This PR r/n is about 100% more FPS in some scenes. With that suggestion of Brad, it's about 120% but I spot blending issues even on the non-stencil path, so that needed to be checked first.

Anyway, I added dithering with no notable drawback. That made me spot some effects of imprecision w. r. t. stencil mask coordinates or mixing effects that I still need to investigate.

MarcelHB:
So here's something interesting:

Screenshot_2022-03-19_15-16-18

blue is where there is non-zero stencil value, like from the wall polygons. red means stencil flag is set and we apply dithering/transparency to those fragements.

This works good where it's supposed to be working, like the PC to the right. But you see there is blue and red areas on the PCs animations that even move deterministically forth and back with their standby animations, but it varies while moving the camera.

Not sure how these non-zero channel values up there, and the uncovered red ones are bad (2nd PC to the right, helmet pixels around the 1st PC). Looks as if there drawing-dirt on the stencil texture or something that may need to be cleared at some point. Will investigate later or tomorrow.

bradallred:
it looks like the same stencil is being used for all the sprites. Possibly because it's being bound in `SetStencilBuffer` instead of where we render the sprite that uses it, but also does seem like any time we draw with a stencil we should flush the renderer after.

MarcelHB:
I got it. When trying to find a way to get the right stencil tex coords, I tried `v_texCoord` just for fun and it worked suprisingly good for reasons I cannot explain, but who knows how this fits into exploiting SDL so let's try. It has its trouble in some other cases as seen above.

So the natural thing of course is supplying UV textures for the stencil map but since we are bound to the single vertex stream of SDL, we need to put in a matrix that transforms the render texture coordinates into the OpenGL space (0,1) of the stencil. So this works and is stable.

Yet it's not fully pixel-perfect, sometimes one may notice a flicking or a little shifting when moving the viewport. I guess it's the some precision issue of the interpolation. Feel free to try it out now, because it may be largely unnoticeable as well, depending on the screen scaling, map, polygon pattern and so on.

bradallred:
Correct, the stencils need to have coordinates independent of the sprites they are stenciling. Otherwise we would have to have a stencil for every animation frame and things like `AreaAnimation` couldn't share a singular viewport stencil.

I'm not an OpenGL guru, but can't we just supply the vertices when we bind the stencil texture using the same method that SDL uses for `SDL_RenderCopy`?

MarcelHB:
> I'm not an OpenGL guru, but can't we just supply the vertices when we bind the stencil texture using the same method that SDL uses for `SDL_RenderCopy`?

Hm, it is using some very old FFP compatible stuff that needs to be enabled manually via client state control calls. One could maybe ship them in one of the other channels of the same structure (like sec. color array), but then again one matrix just offloads a little bit of calc to the GPU without any further state mgmt considerations. 

Performance-wise it's probably the same, just less to think about.

bradallred:
For reference, this is how SDL is setting texture geometry

```cpp
data->glEnable(data->textype);
data->glEnableClientState(GL_TEXTURE_COORD_ARRAY);
data->glActiveTextureARB(GL_TEXTURE0_ARB);
data->glBindTexture(textype, texturedata->texture);
...
data->glVertexPointer(2, GL_FLOAT, sizeof(float) * 5, verts + 0);
data->glColorPointer(4, GL_UNSIGNED_BYTE, sizeof(float) * 5, verts + 2);
data->glTexCoordPointer(2, GL_FLOAT, sizeof(float) * 5, verts + 3);
...
data->glDrawArrays(op, 0, (GLsizei) count);
```

you can combine this with `glClientActiveTexture(GL_TEXTURE1)` and then use `gl_MultiTexCoord1` I assume

vertices appear to be populated as such:
```cpp
float xy[8];
const int xy_stride = 2 * sizeof (float);
float uv[8];
const int uv_stride = 2 * sizeof (float);
const int num_vertices = 4;
const int indices[6] = {0, 1, 2, 0, 2, 3};
const int num_indices = 6;
const int size_indices = 4;
float minu, minv, maxu, maxv;
float minx, miny, maxx, maxy;

minu = (float) (real_srcrect.x) / (float) texture->w;
minv = (float) (real_srcrect.y) / (float) texture->h;
maxu = (float) (real_srcrect.x + real_srcrect.w) / (float) texture->w;
maxv = (float) (real_srcrect.y + real_srcrect.h) / (float) texture->h;

minx = real_dstrect.x;
miny = real_dstrect.y;
maxx = real_dstrect.x + real_dstrect.w;
maxy = real_dstrect.y + real_dstrect.h;

uv[0] = minu;
uv[1] = minv;
uv[2] = maxu;
uv[3] = minv;
uv[4] = maxu;
uv[5] = maxv;
uv[6] = minu;
uv[7] = maxv;

xy[0] = minx;
xy[1] = miny;
xy[2] = maxx;
xy[3] = miny;
xy[4] = maxx;
xy[5] = maxy;
xy[6] = minx;
xy[7] = maxy;

retval = QueueCmdGeometry(renderer, texture,
                xy, xy_stride, &texture->color, 0 /* color_stride */, uv, uv_stride,
                num_vertices,
                indices, num_indices, size_indices,
                renderer->scale.x, renderer->scale.y);
```

with the OpenGL implementation of `QueueCmdGeometry` being:

```
static int
GL_QueueGeometry(SDL_Renderer *renderer, SDL_RenderCommand *cmd, SDL_Texture *texture,
        const float *xy, int xy_stride, const SDL_Color *color, int color_stride, const float *uv, int uv_stride,
        int num_vertices, const void *indices, int num_indices, int size_indices,
        float scale_x, float scale_y)
{
    GL_TextureData *texturedata = NULL;
    int i;
    int count = indices ? num_indices : num_vertices;
    GLfloat *verts;
    size_t sz = 2 * sizeof(GLfloat) + 4 * sizeof(Uint8) + (texture ? 2 : 0) * sizeof(GLfloat);

    verts = (GLfloat *) SDL_AllocateRenderVertices(renderer, count * sz, 0, &cmd->data.draw.first);
    if (!verts) {
        return -1;
    }

    if (texture) {
        texturedata = (GL_TextureData *) texture->driverdata;
    }

    cmd->data.draw.count = count;
    size_indices = indices ? size_indices : 0;

    for (i = 0; i < count; i++) {
        int j;
        float *xy_;
        if (size_indices == 4) {
            j = ((const Uint32 *)indices)[i];
        } else if (size_indices == 2) {
            j = ((const Uint16 *)indices)[i];
        } else if (size_indices == 1) {
            j = ((const Uint8 *)indices)[i];
        } else {
            j = i;
        }

        xy_ = (float *)((char*)xy + j * xy_stride);

        *(verts++) = xy_[0] * scale_x;
        *(verts++) = xy_[1] * scale_y;

        /* Not really a float, but it is still 4 bytes and will be cast to the
           right type in the graphics driver. */
        SDL_memcpy(verts, ((char*)color + j * color_stride), sizeof(*color));
        ++verts;

        if (texture) {
            float *uv_ = (float *)((char*)uv + j * uv_stride);
            *(verts++) = uv_[0] * texturedata->texw;
            *(verts++) = uv_[1] * texturedata->texh;
        }
    }
    return 0;
}
```

I hope there is something helpful to glean from that.

bradallred:
> When that stencil needs to respect both being in front of and behind a cover, and maybe just in some cases because visually sometimes it's good. When doing the stuff like in the previous paragraph, I'm not sure if I'm still missing something or the stencil is blown up?

Stencils are never behind something; they simply hide the pixels of the sprite they are rendered with according to the channel uniform. If a sprite is both behind one wall and in front of another we generate an appropriate stencil containing only pixels from the blocking walls.

MarcelHB:
Goddamit, that's a problem in at least SDL v2.0.14. After ruling out geometry edge cases, bad clear params and stencil cache problems, I tried compiling again with v2.0.20/master and it's stable.

I'm somewhat convinced so since I put TNO into a position for a save game where he oscillates between full cover and correct cover every few frames, so I can reliably reproduce this. I'll try bisecting SDL2 for finding the reason and maybe a workaround unless it's too expensive.

MarcelHB:
Story is that SDL before v.18 hits a different path if a flip bit is set. If set, the OpenGL backend applies some old-school matrix-stacking over texture space coordinates, not viewport space.

In contrast, with v.18+ this calculation is moved to the render frontend via CPU so we end up getting the expected vertices in every case.

Solution: I think of how to fill that matrix correctly depending over some `#ifdef` and flag settings.

bradallred:
IMO, we can restrict `USE_OPENGL_BACKEND` to SDL 2.18+

MarcelHB:
I'll give it a try first because that version is still too new for my Linux distros here.

MarcelHB:
Looks good now, I don't find any further edge cases across the SDL versions with only minimal intrusion to do for this.

So the last thing that would be cool is to have the double blitting step eliminated since that changes to performance to almost-no-impact compared to the untouched render path. But since the easy way of reordering the macros causes errors, I will check what this is about the next session.

bradallred:
> So the last thing that would be cool is to have the double blitting step eliminated since that changes the performance to almost-no-impact compared to the untouched render path. But since the easy way of reordering the macros causes visual errors

I'm not seeing an issue when I apply this patch to remove it

```diff
diff --git a/gemrb/plugins/SDLVideo/SDL20Video.cpp b/gemrb/plugins/SDLVideo/SDL20Video.cpp
index 7958b143a..b62425120 100644
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -401,21 +401,21 @@ void SDL20VideoDriver::BlitSpriteNativeClipped(SDL_Texture* texSprite, const Reg
 		// 3. blend texture to scratchpad
 		// 4. copy scratchpad segment to screen
 
-		std::static_pointer_cast<SDLTextureVideoBuffer>(scratchBuffer)->Clear(drect); // sets the render target to the scratch buffer
-
 		SDL_Rect stencilRect = drect;
 		stencilRect.x -= stencilBuffer->Origin().x;
 		stencilRect.y -= stencilBuffer->Origin().y;
 
 #if USE_OPENGL_BACKEND
-		RenderCopyShadedGL(texSprite, &srect, &drect, flags & ~(BlitFlags::ALPHA_MOD|BlitFlags::HALFTRANS), tint, &stencilRect);
+		RenderCopyShadedGL(texSprite, &srect, &drect, flags, tint, &stencilRect);
+		SDL_RenderFlush(renderer);
 #else
+		std::static_pointer_cast<SDLTextureVideoBuffer>(scratchBuffer)->Clear(drect); // sets the render target to the scratch buffer
+
 		SDL_Texture* stencilTex = CurrentStencilBuffer();
 		SDL_SetTextureBlendMode(stencilTex, stencilAlphaBlender);
 		RenderCopyShaded(texSprite, &srect, &drect, flags & ~(BlitFlags::ALPHA_MOD|BlitFlags::HALFTRANS), tint);
 		// alpha masking only
 		SDL_RenderCopy(renderer, stencilTex, &stencilRect, &drect);
-#endif
 
 		if (flags & (BlitFlags::ALPHA_MOD | BlitFlags::HALFTRANS)) {
 			Uint8 alpha = SDL_ALPHA_OPAQUE;
@@ -431,6 +431,7 @@ void SDL20VideoDriver::BlitSpriteNativeClipped(SDL_Texture* texSprite, const Reg
 		SDL_SetRenderTarget(renderer, CurrentRenderBuffer());
 		SDL_SetTextureBlendMode(ScratchBuffer(), SDL_BLENDMODE_BLEND);
 		ret = SDL_RenderCopy(renderer, ScratchBuffer(), &drect, &drect);
+#endif
 	} else {
 		UpdateRenderTarget();
 #if USE_OPENGL_BACKEND
```

MarcelHB:
Thanks for looking at it. I guess we are good to go then. :slightly_smiling_face:

MarcelHB:
Just some numbers I took yesterday with SDL v2.0.14, `-O3`, some IWD2 still scene with stenciling and no FPS cap:

* master + SDL-only: 450fps
* master + GL: 160ps
* this PR with double blitting: 380fps
* this PR now: 420fps

bradallred:
> master + SDL-only: 450fps

note, that this is probably not using OpenGL, unless you force it, so it's not a relatable data point for comparison.

lynxlynxlynx:
That'd be a bit depressing — all this work and software rendering is still faster.

MarcelHB:
So, some RPi benchmarks with good news. First, `opengl` is default SDL backend there, too.

* SDL with `software`: stable around 20fps but also ugly
* SDL with `opengl`: up to 35fps but dropping down to 10-15fps when stenciling in play
* SDL with `opengl` + our master injections: even worse, more dropping < 10fps
* This PR: Dropping is no worse than 25fps.

That means you can now already play on this platform again, given the latest Raspian OS, at an acceptable level without heavy stuttering.

bradallred:
> SDL with opengl + our master injections: even worse, more dropping < 10fps

I don't know what this means.

This opens up the next avenue for efficiency (software palettes) so 30 fps is easily within reach.

MarcelHB:
> I don't know what this means.

I refer to the current master's `USE_OPENGL_BACKEND` code path. From all of the combinations I tested on the Raspbi, that one suffers from the worst frame drops.

MarcelHB:
I'd propose to defer that hardware paletting into a later run. Thinking a bit around here, I see the `#ifdef`ing grow to an uncomfortable amount that should be maybe justify a split into distinct renderer, while not trying to rebuild the old OGL thing.

lynxlynxlynx:
So ... is there a reason this hasn't been merged yet?

bradallred:
I'm going to decompose this into separate commits when I get a min

MarcelHB:
Aside from the few lines around `Video.h` changes maybe, I did not see much room for splitting changes without the risk of leaving a bad state in the rev history.

bradallred:
merged most of this in #1607

PR #1607, opened 2022-05-04
Unified shaders

bradallred:
First, sorry for the delay. War has unpredictable consequences and pet projects fall by the wayside.

## Description
This is a breakdown of #1587 into smaller chunks for better regression finding.

This doesn't contain any of the uniform caching optimizations from the previous patch. I mistakenly thought we would be able to ask SDL to create a renderer with an OpenGL 3.x context so we could use bitwise operators and just pass render flags directly to the shader rather than though a series of uniforms. However, those changes should also be a separate commit, if indeed we want them. Measurements on my Mac showed _no_ performance improvement for cashing uniforms. I understand this may not apply to other OpenGL backends such as the RPI. If they are in fact a worthwhile optimization we can add them back in with comments for applicable platforms and gains etc.

@MarcelHB would you mind testing this out and make sure it works for you before I merge it into master?


MarcelHB:
I'll have look at it around the weekend.

gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -415,21 +375,100 @@ int SDL20VideoDriver::RenderCopyShaded(SDL_Texture* texture, const SDL_Rect* src
...
+			stencilTexY = -dstrect->y;
+		}
+
+#if !SDL_VERSION_ATLEAST(2, 0, 18)
+		// In versions earler, SDL uses a different vertex setup in case
+		// of flipping: (-w/2, -h/2) to (w/2, h/2) that are transformed by
+		// the OpenGL backend via matrices.
+		if (flags & BlitFlags::MIRRORX) {
+			stencilTexX = dstrect->w / 2.0f + stencilRect->x;
+			stencilTexY = dstrect->h / 2.0f + stencilRect->y;
```
MarcelHB, line 445: `stencilRect`  is no pointer.

MarcelHB (commented):
Aside from the minor fix requirement above, it builds and performs as good as I what I've seen in the other PR.

bradallred:
good catch. thanks! I'll merge this and close the other.

Issue #1574, opened 2022-02-23
Replace snprintf with fmt::format_to


bradallred:
for example: `key.SNPrintF("xPos%d", count);` can be rewritten as `fmt::format_to(key.begin(), "xPos{}", count);`

the goal of course is to use the same simpler formatting string style everywhere. the first parameter to `format_to` simply needs to be something that qualifies as an output iterator (`operator++`, and `operator*`)

PR #1614, opened 2022-05-23
use libfmt for formatting strings

bradallred:
this is for #1574


gemrb/core/Strings/CString.h
```diff
@@ -122,12 +122,14 @@ class FixedSizeString {
...
-	int SNPrintF(const char* format, ...) noexcept {
-		va_list args;
-		va_start(args, format);
-		int ret = vsnprintf(str, sizeof(str), format, args);
-		va_end(args);
-		return ret;
+	template<typename ...ARGS>
+	bool Format(const char* format, ARGS&& ...args) noexcept {
+		auto ret = fmt::format_to_n(begin(), LEN + 1, format, std::forward<ARGS>(args)...);
+		if (ret.size > LEN) {
```
lynxlynxlynx, line 128: should we also be doing it for 0?

lynxlynxlynx: From what you changed, yeah. It's safe to leave it as-is and if we ever need to detect useless calls somewhere, that can be added.

lynxlynxlynx:
So far LGTM, feel free to merge if you're done.

Issue #137, opened 2019-04-15
Externalize message colors


lynxlynxlynx:
iwd2 for example uses gold for xp. So we need new entries in the colors.2da table of each game's unhardcoded/ folder and then use them in core (DisplayMessage). ~See `GameData::GetColor` on how to fetch the new colors from the table. The colors would replace the old DMC_BG2XPGREEN (...) defines. (done)~

PST:
>     Dialog text color #8CA27B (currently #70FF70)
>     Response option color #FF2C21 (currently #FF0000)
>     Response number color #84C7D6 (currently #FF0000, and wrong format)
>     Dialog creature name color #B59E73 (currently #B9B63F)
>     Dialog PC name color #84C7D6 (currently #A2A0A4)
>     Dialog chosen response text #9CC7D6 (currently #8080FF)
>     gained item/xp color #C6FFC6 (currently #BCEFBC)
Journal messages. Load/Save hints. Save messages ... look at where all the ~DMC_~ GUIColors enum constants are used.

bradallred:
this will be related to #233 
would be nice to just do them both at the same time.

lynxlynxlynx:
For #233 there is one tricky aspect to this — the color for the actor name is computed, so it's different for the various interlocutors. We'll have to pregenerate a palette with many options and then limit the choice to those.

bradallred:
I'm not sure what you mean. I was thinking a single grey-scale palette that is then combined with a tint color. There is no reason the tint color cant be dynamically computed.

lynxlynxlynx:
That avoids the speed problem? Ok then. :)

bradallred:
These colors have been externalized in BG:EE and can be found in BGEE.lua. I think our colors are a bit off for most, but I'm not sure and I've already updated a few.

rsantellan:
Searching for the DMC_ string I found

```
grep 'DMC_' * -nr | awk -F 'DMC_' '{print "DMC_"$2}' | awk -F',' '{print $1}' | awk -F')' '{print $1}' | awk -F' ' '{print $1}' | sed 's/^[ \t]*//' | sort | uniq -c
     33 DMC_ //IGNORE THIS ONE
    102 DMC_BG2XPGREEN
      4 DMC_DIALOG
      4 DMC_DIALOGPARTY
     12 DMC_GOLD
     84 DMC_LIGHTGREY
     30 DMC_RED
    210 DMC_WHITE

```
There are more colors or only this ones ?

[...]

PR #1615, opened 2022-05-24
137 vector change

rsantellan:
Remove unused functions.
Make function private.
Copy colors to all colors.2da
Update to use a ENUM and map to the colors in the table.
I create this because the old one has a lot of commit of a merge.

## Description
Externalize the colors of the message to table. #137

## Checklist


gemrb/core/DisplayMessage.cpp
```diff
@@ -144,7 +146,7 @@ Color DisplayMessage::GetSpeakerColor(String& name, const Scriptable *&speaker)
 		case ST_TRIGGER:
 		case ST_PROXIMITY:
 		case ST_TRAVEL:
-			name = core->GetString( speaker->DialogName );
+			name = core->GetString(speaker->DialogName );
```
lynxlynxlynx, line 149: this is an unrelated change, so it shouldn't be here.

gemrb/core/DisplayMessage.cpp
```diff
@@ -187,30 +190,65 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
 	}
 }
 
+std::vector<Color> DisplayMessage::GetAllColors() const
+{
+	std::vector<Color> auxiliaryColors;
+	if (auxiliaryColors.empty()) {
```
lynxlynxlynx, line 196: this is always true

gemrb/core/DisplayMessage.cpp
```diff
@@ -187,30 +190,65 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
...
+		AutoTable colorTable = gamedata->LoadTable("colors", true);
+		assert(colorTable);
+		for (TableMgr::index_t r = 0; r < colorTable->GetRowCount(); r++) {
+			ieDword c = colorTable->QueryFieldUnsigned<ieDword>(r, 0);
+			auxiliaryColors.push_back(Color(c));
+		}
+	}
+	return auxiliaryColors;
+}
```
lynxlynxlynx, line 205: I disagree with this approach; why store all the colors again? We have them in GameData already. Like I explained before, we just need their names. And then also the new `DisplayMessage::GetColor` can be removed.

gemrb/core/DisplayMessage.h
```diff
@@ -33,16 +33,20 @@
...
+/** GUIColors options */
+enum class GUIColors {
+	DIALOG = 12,
+	DIALOGPARTY = 13,
+	GOLD = 11,
+	RED = 8,
+	XPCHANGE = 10,
+	LIGHTGREY = 9,
+	WHITE = 7
+};
```
lynxlynxlynx, line 49: please order this and add a comment why it doesn't start at 0

gemrb/core/DisplayMessage.cpp
rsantellan: The problem is that on https://github.com/gemrb/gemrb/blob/5834c9ada3b08f2c77604351f20e8a82b4b76bed/gemrb/core/GameData.cpp#L640 the colors created is ` std::map<std::string, Color> colors;` so to map my enum I pass it as vector Color. 
The problem with this approach is that I have 13 colors defined on colors.2da so I have to map the positions to the enum.
I don't like this at all, for me will be better to map the enum to the names so there is no mistake there.

gemrb/core/DisplayMessage.h
```diff
@@ -37,7 +37,7 @@
 
 namespace GemRB {
 
-/** GUIColors options */
+/** GUIColors options. This starts at 7 because is the row in the colors.2da that corresponds to white color.*/
```
lynxlynxlynx, line 40: A better comment would be that the first six colors from colors.2da are not indexed here because they are only used on the python side.

gemrb/core/DisplayMessage.h
```diff
@@ -39,13 +39,13 @@ namespace GemRB {
...
-	GOLD = 11,
+	WHITE = 7,
 	RED = 8,
-	XPCHANGE = 10,
 	LIGHTGREY = 9,
-	WHITE = 7
+	XPCHANGE = 10,
+	GOLD = 11,
+	DIALOG = 12,
+	DIALOGPARTY = 13
```
lynxlynxlynx, line 48: Now that they're in order, you don't need to manually assign values to all of them. WHITE, as the first, is enough.

gemrb/core/DisplayMessage.cpp
```diff
@@ -190,28 +188,26 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
...
-	if (size < colortInt) {
-		return ColorRed;
+	
+	int colorInt = static_cast<int>(color) - 1;
+	if (static_cast<int>(GUIColorNames.size()) < colorInt) {
+		// It never should happend this, but just for been defensive.
+		colorInt = 0;
 	}
-	return colors.at(colortInt);
+	return gamedata->GetColor(GUIColorNames[colorInt]);
```
lynxlynxlynx, line 210: Let's use `std::map<GUIColors, std::string>` then for GUIColorNames, so this function can be simplified or removed altogether. Because it's a map, it also means you don't have to check for non-existing keys, as new entries would get created automatically and eventually the default color from the table would be used.

gemrb/core/DisplayMessage.cpp
```diff
@@ -187,30 +188,63 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
 	}
 }
 
+std::vector<std::string> DisplayMessage::GetAllColors() const
+{
+	std::vector<std::string> auxiliaryColors;
+	AutoTable colorTable = gamedata->LoadTable("colors", true);
+	assert(colorTable);
+	for (TableMgr::index_t r = 0; r < colorTable->GetRowCount(); r++) {
```
miscco, line 196: This is missing a call to `reserve` which will lead to a whole bunch of allocations
```suggestion
        auxiliaryColors.reserve(colorTable->GetRowCount());
	for (TableMgr::index_t r = 0; r < colorTable->GetRowCount(); r++) {
```

gemrb/core/DisplayMessage.cpp
```diff
@@ -187,30 +188,63 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
...
+		auxiliaryColors.push_back(colorTable->GetRowName(r));
+	}
+	return auxiliaryColors;
+}
+
+Color DisplayMessage::GetColor(GUIColors color) const
+{
+
```
miscco, line 204:
```suggestion
```

gemrb/core/DisplayMessage.cpp
```diff
@@ -187,30 +188,63 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
 	}
 }
 
+std::vector<std::string> DisplayMessage::GetAllColors() const
+{
+	std::vector<std::string> auxiliaryColors;
```
miscco, line 193: Arent those names static? A view type like `std::string_view` would be much better here

lynxlynxlynx: I don't see any benefit, but there's no reason for the function to work with a temporary either.

gemrb/core/DisplayMessage.cpp
rsantellan: @lynxlynxlynx this will not be the same as the function **gamedata->GetColor** at the end ?

lynxlynxlynx: No, the gamedata version will still be the main one to get the precached Color values by color (row) name.
When I say that this function here can probably go away, I'm thinking of just replacing calls to it by `gamedata->GetColor(GUIColorNames[color])` (color being one of the GUIColors values).

rsantellan: Sorry I don't understand this. Now I have:

- A **std::vector<std::string>** that has all the string names on the **colors.2da table.**
- A **Enum** that has the row where the color is set.
- Using the **Enum** I use that position on the vector to retrieve the color name to use on **gamedata->GetColor**

And the proposed change will work something like this:
If I use a **std::map<GUIColors, std::string>** I will have to also have a way to convert the **Enum** to a position on the string, so I have to keep the **std::vector<std::string>** and also use the map, and then I will not need  **gamedata->GetColor** because I have the color there, but if not in the map I will call **gamedata->GetColor** ?

This is what I'm thinking, I'm not fully comprehend where the change is going or how it will work.

lynxlynxlynx: > I will have to also have a way to convert the Enum to a position on the string
What string? The enum still holds the row numbers and the std::map will map these enum values to the row names. Exactly like the vector, except that the indices are properly typed (enum vs general int), so there's no need to convert the index. Or care about missing indexes (keys).

gemrb/core/DisplayMessage.cpp
```diff
@@ -188,27 +188,24 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
 	}
 }
 
-std::vector<std::string> DisplayMessage::GetAllColors() const
+std::map<int, std::string> DisplayMessage::GetAllColors() const
```
lynxlynxlynx, line 191: Why not the map I proposed? There's not only no need for the full range of integer keys, but more bugs are possible that way.

rsantellan:
Closed in favor of  #1625

Issue #1616, opened 2022-05-28
 Static assertion error in strtounsigned


MarcelHB:
I tried to compile the latest revision with MSYS2/GCC v12.1. It fails because of the following static assertion:

```
[  1%] Building CXX object gemrb/core/CMakeFiles/gemrb_core.dir/Interface.cpp.obj
In file included from D:/Sources/gemrb/gemrb/core/Audio.h:26,
                 from D:/Sources/gemrb/gemrb/core/Interface.h:32,
                 from D:/Sources/gemrb/gemrb/core/Interface.cpp:21:
D:/Sources/gemrb/gemrb/includes/globals.h: In instantiation of 'T GemRB::strtounsigned(const char*, char**, int) [with T = long long unsigned int]':
D:/Sources/gemrb/gemrb/core/TableMgr.h:72:30:   required from 'RET_T GemRB::TableMgr::QueryFieldUnsigned(const ROW_T&, const COL_T&) const [with RET_T = long long unsigned int; ROW_T = long long unsigned int; COL_T = int]'
D:/Sources/gemrb/gemrb/core/Interface.cpp:553:80:   required from here
D:/Sources/gemrb/gemrb/includes/globals.h:220:33: error: static assertion failed: Type is too big for conversion.
  220 |         static_assert(sizeof(T) <= sizeof(long), "Type is too big for conversion.");
      |                       ~~~~~~~~~~^~~~~~~~~~~~~~~
D:/Sources/gemrb/gemrb/includes/globals.h:220:33: note: the comparison reduces to '(8 <= 4)'

```

MarcelHB:
https://github.com/gemrb/gemrb/blob/b344079ac5979962a9b6b02fc6207a73549cdce2/gemrb/core/Interface.cpp#L553
Do we see a good reason for `size_t`? The table values should be something fixed like `uint32_t` (or 64), I presume?

https://github.com/gemrb/gemrb/blob/b344079ac5979962a9b6b02fc6207a73549cdce2/gemrb/includes/globals.h#L220
Similarly, maybe replace `long` by something static as well to avoid platform confusion?

bradallred:
`uint16_t` is good enough for this particular call. There are probably others of different sizes.

> Similarly, maybe replace long by something static as well to avoid platform confusion?

I dont care if we nix that assert at this point. It was there so I could detect all the callers that were using inappropriate types at the time it was written. However, you shouldn't just change the `long` to something else since `strtol` explicitly returns a `long`.

PR #1617, opened 2022-05-28
Replace some uses of size_t

MarcelHB:
## Description
See #1616, there were two errors and I noticed a bunch of stuff around it that could be type-wired and cleaned with `uint32_t` at the end.

## Checklist

### DIFF
--- a/gemrb/plugins/2DAImporter/2DAImporter.cpp
+++ b/gemrb/plugins/2DAImporter/2DAImporter.cpp
@@ -102,7 +102,9 @@ bool p2DAImporter::Open(DataStream* str)
 			row++;
 		}
 	}
+
 	delete str;
+	assert(rows.size() < std::numeric_limits<index_t>::max());
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


gemrb/plugins/2DAImporter/2DAImporter.cpp
```diff
@@ -102,7 +102,9 @@ bool p2DAImporter::Open(DataStream* str)
 			row++;
 		}
 	}
+
 	delete str;
+	assert(rows.size() < std::numeric_limits<index_t>::max());
```
lynxlynxlynx, line 107: I don't think we need to be this paranoid. You'd need specially crafted data and tons of free memory to be able to fail here.

MarcelHB: Force of habit when it comes to _all input is evil_ :shrug:

bradallred (approved):
this all looks good to me

MarcelHB:
No merge button for me 🤷

PR #1619, opened 2022-05-29
SaveGameIterator: fix of buffer overflow

MarcelHB:
Fixes the last of the known crashes mentioned in #1616.

## Description
Cannot apply `strftime` like this anymore.

## Checklist


lynxlynxlynx:
Thanks!

Issue #1620, opened 2022-05-29
Regression: all ARE files are extracted again on loading 


MarcelHB:
#### Bug description

When loading a game, GemRB again starts to extract _all_ ARE files that are contained. This was improved with #1337

#### Expected behavior
Only the one we need right now.

#### Screenshots
From late chapter in BG1:
```
[SAVImporter]: Decompressing ar0002.are
[SAVImporter]: Decompressing ar0004.are
[SAVImporter]: Decompressing ar0005.are
[SAVImporter]: Decompressing ar0010.are
[SAVImporter]: Decompressing ar0011.are
[SAVImporter]: Decompressing ar0012.are
[SAVImporter]: Decompressing ar0100.are
[SAVImporter]: Decompressing ar0103.are
[SAVImporter]: Decompressing ar0104.are
[SAVImporter]: Decompressing ar0105.are
[SAVImporter]: Decompressing ar0106.are
[SAVImporter]: Decompressing ar0107.are
...
```

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

MarcelHB:
https://github.com/gemrb/gemrb/blob/a7df1085fb145f9a23b5377ca16d3852a581ac01/gemrb/plugins/SAVImporter/SAVImporter.cpp#L43
`fnlen` reads 11 and reserves 11 chars and `pos` of the extension is 6. So the check of `.length() - 4` fails since I assume that we'd get 10 for a string like `ar1234.are`.

MarcelHB:
Regression by e56e48972e20f9455962de9001e2d048075999a8. Since Brad touched the math there, this rules out any reading errors, and the fix is simply changing 4 into 5 at subtraction.

PR #1623, opened 2022-05-29
SAVImporter: fixes broken lazy ARE lookup

MarcelHB:
See #1620 for details.

## Checklist


bradallred:
can't we just decrement `fnlen` by one instead?

MarcelHB:
I was flipping a coin between doing that at the expense of an additional stream-skipping for that one byte we'd read less then, and what I did here.

PR #1625, opened 2022-06-03
137 vector to map change

rsantellan:
Remove unused functions.
Make function private.
Copy colors to all colors.2da
Update to use a ENUM and map to the colors in the table.
I create this because the old one has a lot of commit of a merge.

## Description
Externalize the colors of the message to table. #137

## Checklist


gemrb/core/DisplayMessage.h
```diff
@@ -77,9 +77,9 @@ class GEM_EXPORT DisplayMessage
 	/** displays a string in the textarea */
 	void DisplayString(String text, const Color &color, Scriptable *target) const;
 	Color GetColor(const GUIColors color) const;
-	std::map<int, std::string> GetAllColors() const;
+	std::map<GUIColors, std::string> GetAllColors() const;
 private:
```
lynxlynxlynx, line 81: this "private" line can be removed, as the previous block is already like that.

gemrb/core/DisplayMessage.cpp
```diff
@@ -188,20 +188,22 @@ void DisplayMessage::DisplayString(String text, const Color &color, Scriptable *
...
-std::map<int, std::string> DisplayMessage::GetAllColors() const
+std::map<GUIColors, std::string> DisplayMessage::GetAllColors() const
 {
-	std::map<int, std::string> auxiliaryColors;
+	std::map<GUIColors, std::string> auxiliaryColors;
 	AutoTable colorTable = gamedata->LoadTable("colors", true);
 	assert(colorTable);
 	for (TableMgr::index_t r = 0; r < colorTable->GetRowCount(); r++) {
-		auxiliaryColors[static_cast<int>(r)] = colorTable->GetRowName(r);
+		if (static_cast<int>(r) >= static_cast<int>(GUIColors::WHITE) && static_cast<int>(r) <= static_cast<int>(GUIColors::DIALOGPARTY)) {
```
lynxlynxlynx, line 197: You can cast to one of the types already used, avoiding the need for so many casts.
But also this is a bit fragile:
- you can move these bounds conditions to the loop definition
- it would be cleaner if you now added two more entries to the enum, to mark the start and the end, so these bounds here won't need to be updated every time a new color is added. I suggest FIRST_COLOR and LAST_COLOR

gemrb/core/DisplayMessage.cpp
```diff
@@ -193,10 +193,11 @@ std::map<GUIColors, std::string> DisplayMessage::GetAllColors() const
 	std::map<GUIColors, std::string> auxiliaryColors;
 	AutoTable colorTable = gamedata->LoadTable("colors", true);
 	assert(colorTable);
-	for (TableMgr::index_t r = 0; r < colorTable->GetRowCount(); r++) {
-		if (static_cast<int>(r) >= static_cast<int>(GUIColors::WHITE) && static_cast<int>(r) <= static_cast<int>(GUIColors::DIALOGPARTY)) {
-			auxiliaryColors[static_cast<GUIColors>(r)] = colorTable->GetRowName(r);
-		}
+	TableMgr::index_t index = static_cast<int>(GUIColors::FIRST_COLOR) + 1;
```
lynxlynxlynx, line 196: If you bump FIRST_COLOR to 7, then you could directly cast both index and finish to TableMgr::index_t. Or not cast them at all and have them be of GUIColors type (not sure if that would work below), so you'd only have to have one cast to index_t when looking up the row name.

gemrb/core/DisplayMessage.cpp
```diff
@@ -193,10 +193,11 @@ std::map<GUIColors, std::string> DisplayMessage::GetAllColors() const
...
+	TableMgr::index_t finish = static_cast<int>(GUIColors::LAST_COLOR);
+	while (index < finish) {
+		auxiliaryColors[static_cast<GUIColors>(index)] = colorTable->GetRowName(index);
+		index ++;
```
lynxlynxlynx, line 200:
```suggestion
		index++;
```

gemrb/core/DisplayMessage.cpp
rsantellan: > If you bump FIRST_COLOR to 7, then you could directly cast both index and finish to TableMgr::index_t. Or not cast them at all and have them be of GUIColors type (not sure if that would work below), so you'd only have to have one cast to index_t when looking up the row name.

The problem to bump it is that it will mess up the colors numbers, I tried using them without casting and I can't do operation on them. I could store the FIRST_COLOR and don't do any operation, but seams like a waste of memory.

lynxlynxlynx: enums can have duplicate values without issues, you just have to explicitly specify them, so the +1 rule doesn't take over.

gemrb/core/DisplayMessage.cpp
```diff
@@ -193,11 +193,11 @@ std::map<GUIColors, std::string> DisplayMessage::GetAllColors() const
 	std::map<GUIColors, std::string> auxiliaryColors;
 	AutoTable colorTable = gamedata->LoadTable("colors", true);
 	assert(colorTable);
-	TableMgr::index_t index = static_cast<int>(GUIColors::FIRST_COLOR) + 1;
+	TableMgr::index_t index = static_cast<int>(GUIColors::FIRST_COLOR);
```
lynxlynxlynx, line 196: Cast directly to index_t, now there's still another implicit cast going on.

lynxlynxlynx:
Looks good to me now, thanks. :)

PR #1627, opened 2022-06-05
Fix calls to Scriptable::As through null pointers.

gsz:
Calling a non-static member function on a null pointer is undefined
behavior. Added non-null checks at affected call sites.

## Description
There are multiple places in the code where Scriptable::As is called on a pointer that can be null. This can cause crashes on some platforms, which I have experienced (I can add a separate bug report with details if it's necessary).

From this discussion (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99074) it seems that versions of GCC from 8 up to 11 are affected.

## Checklist

### DIFF
--- a/gemrb/core/Core.cpp
+++ b/gemrb/core/Core.cpp
@@ -272,7 +272,7 @@ bool WithinPersonalRange(const Scriptable *actor, const Point &dest, int distanc
 int EARelation(const Scriptable* Owner, const Actor* target)
 {
 	ieDword eao = EA_ENEMY;
-	const Actor* actor = Owner->As<const Actor>();
+	const Actor* actor = Scriptable::As<const Actor>(Owner);
 	if (actor) {
 		eao = actor->GetStat(IE_EA);
 	}
--- a/gemrb/core/DialogHandler.cpp
+++ b/gemrb/core/DialogHandler.cpp
@@ -196,7 +196,7 @@ void DialogHandler::EndDialog(bool try_to_break)
 	}
 
 	Actor *tmp = GetSpeaker();
-	Actor *target = GetTarget()->As<Actor>();
+	Actor *target = Scriptable::As<Actor>(GetTarget());
 	speakerID = 0;
 	targetID = 0;
 	originalTargetID = 0;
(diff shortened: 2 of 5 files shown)


bradallred:
we already have a static version of this for cases when it may be null e.g. `const Actor* actor = Owner->As<const Actor>();` -> `const Actor* actor =-Scriptable::As<const Actor>(Owner);` which should return `nullptr` appropriately.

We really don't have to worry about UB here anyway. If you look at the implementation of `Scriptable::As` you will see all it does is call the static method and pass `this`. No UB results from passing a null `this` to a static method.

gsz:
Like I said, this caused a crash for me (using gcc 9.4.0 on Ubuntu 20.04). Segfault while doing dynamic cast, same as in the example described in the linked GCC bug report. I believe what happens is that the compiler infers that the pointer is not null (operating under the assumption that it hasn't been fed invalid code) and then optimizes away non-null check in dynamic cast implementation.

bradallred:
Yeah, obviously if its crashing or causing noisy analyzer/sanitizer output we want to address that. Indeed your theory seems likely albeit surprising since the instance method `As` is implemented in terms of the static method which cannot optimize that check away, but the compiler is indeed free to produce separate implementations (which it must be doing here).

 I'm not opposed to these changes, just wondering if we have a preference for using `Scriptable::As<>` for constancy.

lynxlynxlynx:
Yes, that would be cleaner.

gsz:
Here's a version calling Scriptable::As.

As for how a crash can happen, the compiler is inlining the calls to As.

gemrb/core/Core.cpp
```diff
@@ -272,7 +272,7 @@ bool WithinPersonalRange(const Scriptable *actor, const Point &dest, int distanc
 int EARelation(const Scriptable* Owner, const Actor* target)
 {
 	ieDword eao = EA_ENEMY;
-	const Actor* actor = Owner->As<const Actor>();
+	const Actor* actor = Scriptable::As<Actor>(Owner);
```
lynxlynxlynx, line 275: Should be `const Actor` still, so we won't get analyzer warnings from casting away the const.

gsz: Oops.

lynxlynxlynx:
Thanks!

PR #1630, opened 2022-06-06
Fix starting number of the ENUM.

rsantellan:
## Description
Fix bad color.

## Checklist


lynxlynxlynx:
Thanks for the quick response. 👍🏾

Issue #1516, opened 2022-01-13
UTF-16 everywhere


bradallred:
We have several bugs and complexities that are arising from the fact that we have a mix of encodings thought the engine. The only sane thing to do is to migrate to UTF-16 thought and translate to/from at the importer boundary.

This is a much larger undertaking than it will appear on the surface due to the nature of the BAM fonts. A character in a BAM font is looked up with a frame+cycle combination (1 byte each, hence the need for UTF-16). Currently, the font is constructed to support the TLK encoding so in order for the fonts to continue to work, the font `GlyphAtlasPage` objects will have to become encoding aware such that we can look up using a UTF-16 code point and map back to the encoding the BAM font expects.

Alternatively, we could remap the glyph encoding at initialization time, but our TTF fonts rely on being able to produce glyphs dynamically so I don't think it's a good idea. However, the TTF fonts should already be in UTF-16 to work, so perhaps we can work around that issue.

bradallred:
@fizzet tagging you because this may be of interest to you and you seem to have the requisite skills and knowledge.

fizzet:
If only I had the time... :stuck_out_tongue_closed_eyes:

bradallred:
I don't expect any of us to have the time in the short term :)

bradallred:
I created a new [wiki page](https://github.com/gemrb/gemrb/wiki/Text-Encoding-Documentation) outlining the plan

TheCycoONE:
Incidentally the old 16-bit unicode encoding predating surrogates is called UCS-2.

TheCycoONE:
Also not sure which event you are talking about but SDL_TextInputEvent is defined to always return UTF-8.

MarcelHB:
Any idea how to continue best here? I have a bunch of umlaut display errors and console warnings so I'm tempted to give it some progress, yet I don't want to cross any ongoing effort or prerequisites.

bradallred:
Please, go ahead :)

I had looked at using libfmt for unicode conversion, but it's not part of the public API, so I guess we just stick with iconv (or codecvt).

The easiest thing to get going would be to start with adding iconv to `StringFromEncodedData` and `formatter<WideToChar>`. We also have to examine the GUIScript code to make sure everything is being properly converted. I think most of the time we just copy the string without a care for the encoding.

PR #1634, opened 2022-06-28
String explosion

bradallred:
This could use more testing/review, and hopefully we have some console commands to help there.

This is some bug fixes/enhancements to some of our string algorithms as well cutting over some C string manipulation to use those algorithms instead so we can be better positioned to finish #1516 at some point.

### DIFF
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -1603,18 +1603,14 @@ static int ParseIntParam(const char *&src, const char *&str)
 	}
 	if (*str=='*') { //there may be an IDS table
 		str++;
-		char idsTabName[9];
-		char *cur = idsTabName;
-		const char *end = idsTabName + 9 - 1;
-		while (*str != ',' && *str != ')') {
-			// limit IDS file length to 8 characters
-			 if (cur != end) {
-				*cur = *str;
-				++cur;
-			}
+		ResRef idsTabName;
+		char *cur = idsTabName.begin();
+		const char *end = idsTabName.bufend();
+		while (cur != end && *str != ',' && *str != ')') {
+			*cur = *str;
+			++cur;
 			++str;
 		}
-		*cur = 0;
 
 		if (idsTabName[0]) {
 			return GetIdsValue(src, idsTabName);
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -1807,8 +1807,6 @@ GameScript::~GameScript(void)
 
 Script* GameScript::CacheScript(const ResRef& resRef, bool AIScript)
 {
-	char line[10];
-
 	SClass_ID type = AIScript ? IE_BS_CLASS_ID : IE_BCS_CLASS_ID;
 
 	Script *newScript = (Script *) BcsCache.GetResource(resRef);
@@ -1821,8 +1819,10 @@ Script* GameScript::CacheScript(const ResRef& resRef, bool AIScript)
 	if (!stream) {
 		return NULL;
 	}
-	stream->ReadLine( line, 10 );
-	if (strncmp( line, "SC", 2 ) != 0) {
+	
+	std::string line;
+	stream->ReadLine(line, 10);
+	if (line.compare(0, 2, "SC") != 0) {
 		Log(WARNING, "GameScript", "Not a Compiled Script file");
 		delete stream;
 		return nullptr;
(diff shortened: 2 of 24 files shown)


gemrb/core/InterfaceConfig.cpp
```diff
@@ -39,28 +39,17 @@ InterfaceConfig::~InterfaceConfig() noexcept
...
-	const char* value = NULL;
-	if (key) {
-		// lowercase the key so that the key is not case sensitive
-		char* keyCopy = strdup(key);
-		for (char* c = keyCopy; *c != '\0'; ++c) *c = tolower(*c);
-		if (configVars->get(keyCopy)) {
-			value = configVars->get(keyCopy)->c_str();
-		}
-		free(keyCopy);
+	if (configVars->get(key)) {
```
lynxlynxlynx, line 49: This check is redundant, the function can be simplified to the one return.

lynxlynxlynx: Or better yet, should check `key` as StringMap's `get` doesn't check itself.

gemrb/core/GameScript/GameScript.cpp
```diff
@@ -1821,8 +1819,10 @@ Script* GameScript::CacheScript(const ResRef& resRef, bool AIScript)
 	if (!stream) {
 		return NULL;
 	}
-	stream->ReadLine( line, 10 );
-	if (strncmp( line, "SC", 2 ) != 0) {
+	
+	std::string line;
+	stream->ReadLine(line, 10);
+	if (line.compare(0, std::string::npos, "SC") != 0) {
```
lynxlynxlynx, line 1825: Why not just use operator== in such cases?

gemrb/core/Streams/DataStream.cpp
```diff
@@ -123,35 +123,28 @@ strret_t DataStream::ReadRegion(Region& r)
...
 	//Windows: \r\n
 	//Old Mac: \r
 	//otherOS: \n
-	while (i < ( maxlen - 1 )) {
+	while (Pos < size && i < (maxlen - 1)) {
 		char ch;
-		Read(&ch, 1);
+		i += Read(&ch, 1);
 		if (ch == '\n')
 			break;
```
lynxlynxlynx, line 142: Not sure if it matters, but this case now returns a value 1 greater than before.

bradallred: yeah, it should be fine. I think we only check it for `Error` and use `string::length()` for the actual data examination. It should be returning the number of bytes read which will always be longer.

gemrb/plugins/2DAImporter/2DAImporter.cpp
```diff
@@ -25,81 +25,55 @@
...
 	return stricmp(str.c_str(), key.c_str()) == 0;
 }
 
-p2DAImporter::p2DAImporter() noexcept
-{
-	colNames.reserve(10);
-	rowNames.reserve(10);
-	rows.reserve(10);
-}
-
```
lynxlynxlynx, line 41: Why did you remove these? They were added to reduce the amount of reallocations and helped significantly when you looked at a heap profile.

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -1603,18 +1603,14 @@ static int ParseIntParam(const char *&src, const char *&str)
...
-				++cur;
-			}
+		ResRef idsTabName;
+		char *cur = idsTabName.begin();
+		const char *end = idsTabName.bufend();
+		while (cur != end && *str != ',' && *str != ')') {
+			*cur = *str;
+			++cur;
 			++str;
 		}
```
lynxlynxlynx, line 1613: This can be further simplified by using `FindNotOf` instead of the loop.

bradallred: `FindNotOf` wont work on a c string tho. Needs further refactoring to get there.

gemrb/core/GameScript/GameScript.cpp
bradallred: ah, yes, an oversight on my part certainly.

gemrb/plugins/2DAImporter/2DAImporter.cpp
bradallred: they've been added to `p2DAImporter::Open`

gemrb/core/Interface.cpp
```diff
@@ -3616,16 +3613,18 @@ bool Interface::ResolveRandomItem(CREItem *itm) const
...
+		if (p != ResRef::npos) {
+			NewItem[p] = '\0';
+			diceSides = strtosigned<int>(NewItem.begin() + p + 1, nullptr, 10);
 		} else {
 			diceSides = 1;
 		}
-		int diceThrows = strtosigned<int>(NewItem, &endptr, 10);
+		
+		char *endptr;
+		int diceThrows = strtosigned<int>(NewItem.begin(), &endptr, 10);
```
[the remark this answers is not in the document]
bradallred: I think I've done what you asked.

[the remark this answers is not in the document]
lynxlynxlynx: It looks fine, thanks. It did made me look at the function more deeply and we do have at least one problem, but it's not related to these changes: it looks like we're always generating just 1 gp and the bolt example would create gold instead. I'll take a look once this is merged.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -2023,19 +2023,15 @@ static void InitActorTables()
...
-			char **classnames = (char **) calloc(tmpbits, sizeof(char *));
-			classnames[0] = strtok(strdup(classname.c_str()), "_");
-			while (numfound<tmpbits && (classnames[numfound] = strdup(strtok(NULL, "_")))) {
-				numfound++;
-			}
-			numfound = 0;
+			auto classnames = Explode(classname, '_', tmpbits);
 			bool foundwarrior = false;
+			//we have to account for dual-swap in the multiclass field
+			auto numfound = classnames.size() + 1;
```
lynxlynxlynx, line 2033: No, this should remain 0, or it will break the iteration below.

lynxlynxlynx:
LGTM now. :)

Issue #1636, opened 2022-07-08
Multiple unresolved externals in GUIScript.dll using MSVC


jabaldwin:
I'm attempting to build GemRB from source following [the guide on the website](https://gemrb.org/MSVC-build.html#get-gemrb-sources-and-build-them). I encounter several "unresolved external" linking errors within GUIScript.cpp, which is odd since that file doesn't seem to have been updated since the v0.9.0 release. Any help is appreciated! 

Build output attached:
[gemrb_build_log.txt](https://github.com/gemrb/gemrb/files/9075071/gemrb_build_log.txt)

MarcelHB:
> I encounter several "unresolved external" linking errors within GUIScript.cpp, which is odd since that file doesn't seem to have been updated since the v0.9.0 release. Any help is appreciated!

`GUIScript.cpp` has undergone [a lot of changes](https://github.com/gemrb/gemrb/compare/v0.9.0...master#diff-96ea9d8971dee67edce84478072a33a6730e52ef01fb174cfa0b7dda99cfacfe) since v0.9.0.

You can maybe execute a `git status` to check what revision you are really compiling on.

jabaldwin:
Sorry, I must have read the history of GUIScript.cpp wrong, my mistake. I have tried compiling on the most recent commit to master and several slightly older commits, all with the same issues. Is there a particular commit that should be more "stable" re: being compile-able?

MarcelHB:
The most recent commit rev [looks good](https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/44121819) from what we see in MSVC 2019, and I wasn't seeing any linker problem locally either.

But I have found something suspicious, could you maybe give this modification a try (`ScriptEngine.h` line 57)?
```diff
diff --git a/gemrb/core/ScriptEngine.h b/gemrb/core/ScriptEngine.h
index 86e85c6c7..69f0daab4 100644
--- a/gemrb/core/ScriptEngine.h
+++ b/gemrb/core/ScriptEngine.h
@@ -54,7 +54,7 @@ public:
 };

 template <class T>
-class GEM_EXPORT ScriptingRef : public ScriptingRefBase {
+class ScriptingRef : public ScriptingRefBase {
 private:
        T* ref;
 public:
```

jabaldwin:
That change got the build to succeed, thank you!

Now I'm having a problem immediately on running the app:
`The code execution cannot proceed because iconv-2.dll was not found.`

[...]

PR #1637, opened 2022-07-09
Fixes for Windows

MarcelHB:
## Description
Two things addressed:

* The initial problem of #1636 
* Errors for unused variables on uses of `ResolveFilePath` on Windows & GCC.

## Checklist


lynxlynxlynx:
Thanks!

PR #1638, opened 2022-07-10
Sdl ambient

Northfear:
## Description
I had a working ambient sounds support for SDLAudio plugin for quite some time, so I thought that I might as well do a PR with it.
Also had to move music stopping code a bit to make sure that music is stopped after movie sound buffer is queued (movie sound stops otherwise with SDLAudio).

## Checklist

- [*] Commit messages are descriptive and explain the rationale for changes
- [*] I used the same coding style as the surrounding code 
- [*] I have tested the proposed changes
- [*] I extended the documentation, if necessary

### DIFF
--- a/gemrb/plugins/SDLAudio/SDLAudio.cpp
+++ b/gemrb/plugins/SDLAudio/SDLAudio.cpp
@@ -506,7 +506,7 @@ int SDLAudio::SetupNewStream(int x, int y, int z,
 
 tick_t SDLAudio::QueueAmbient(int stream, const ResRef& sound)
 {
-	if (stream == 0 || stream > AMBIENT_CHANNELS + 1) {
+	if (stream <= 0 || stream > AMBIENT_CHANNELS) {
 		return -1;
 	}
 
--- a/gemrb/plugins/SDLAudio/SDLAudio.cpp
+++ b/gemrb/plugins/SDLAudio/SDLAudio.cpp
@@ -31,15 +31,14 @@
 
 using namespace GemRB;
 
-static void SetChannelPosition(const Point &listenerPos, const Point &p, int channel)
+static void SetChannelPosition(const Point &listenerPos, const Point &p, int channel, float rolloff)
 {
 	int16_t angle = AngleFromPoints(listenerPos, p) * 180 / M_PI - 90;
 	if (angle < 0) {
 		angle += 360;
 	}
-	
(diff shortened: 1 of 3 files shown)


gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -322,22 +325,35 @@ Holder<SoundHandle> SDLAudio::Play(StringView ResRef, unsigned int channel,
 		*length = time_length;
 	}
 
-	Mix_VolumeChunk(chunk, MIX_MAX_VOLUME * (GetVolume(channel) * volume / 10000.0f));
+	Mix_VolumeChunk(chunk, MIX_MAX_VOLUME * GetVolume(channel) / 100);
```
lynxlynxlynx, line 328: This looks like a mistake, now it's only taking the preset relative volume into account, not the user settings.

gemrb/plugins/SDLAudio/SDLAudio.cpp
```diff
@@ -31,15 +31,14 @@
...
 {
 	int16_t angle = AngleFromPoints(listenerPos, p) * 180 / M_PI - 90;
 	if (angle < 0) {
 		angle += 360;
 	}
-	
-	constexpr float AUDIO_DISTANCE_ROLLOFF_MOD (1.3 * 1.3); // squared to save a sqrt
-	uint8_t distance = std::min<uint8_t>((SquaredDistance(listenerPos, p) / AUDIO_DISTANCE_ROLLOFF_MOD), 255u);
+
+	uint8_t distance = std::min(static_cast<int32_t>(std::sqrt(SquaredDistance(listenerPos, p)) / rolloff), 255);
```
lynxlynxlynx, line 41: `std::sqrt(SquaredDistance(X)) == Distance(X)` to a close enough degree.
So you can just use `Distance(listenerPos, p) / rolloff` now that you've started removing the squaring.

gemrb/plugins/SDLAudio/SDLAudio.cpp
Northfear: It's taking preset relative volume into account while mixing the audio chunk.
User setting is taken into account on the line 335 for the channel itself
`Mix_Volume(chan, MIX_MAX_VOLUME * volume / 100);`

PR #1652, opened 2022-07-25
- VCPKG using manifest file

DudeMcDude:
## Description
- Fixed libiconv usage - newer version changed dll name
- Added manifest file for VCPKG, pinned versions to current latest (in case stuff like the above happens again)

## Checklist

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -39,6 +39,7 @@ install:
 
     pushd c:\tools\vcpkg
     git pull
+    c:\tools\vcpkg\bootstrap-vcpkg.bat
     popd
     #vcpkg remove sdl2:x86-windows --recurse
     # (lib)vlc is missing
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
All the commented stuff should be removed instead. To recache, try commenting out the vcpkg line under `cache:` (we will reinstate it later). And thanks for putting the manifest in the right place. :)

lynxlynxlynx:
It got killed for taking over an hour. There is a var that could help with priming the cache, but it wouldn't help here, so I think we have to temporarily avoid the main build. So I suggest you uncomment the APPVEYOR_SAVE_CACHE_ON_ERROR line and just `exit 1` (or whatever the powershell version is) in before_build. The build will fail, but the cache will be primed, so then we can remove the hack and the whole thing should build in time.

DudeMcDude:
Oh, it looks like caching doesn't work for PRs (sensibly).  
https://www.appveyor.com/docs/build-cache/#saveupdate-cache-before-build-finishes

So I suppose we'll try it out in a branch?

lynxlynxlynx:
I'll just merge and then you can retry after that fails.

lynxlynxlynx:
Oh and please add -q to the pull invocation, so it won't spam the log so much.

lynxlynxlynx:
Ufff, that was close. But we have the primed caches now. :)

PR #1653, opened 2022-07-25
Rest of VCPKG PR

DudeMcDude:
## Description
Continuing #1652 

## Checklist

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -39,6 +39,7 @@ install:
 
     pushd c:\tools\vcpkg
     git pull
+    c:\tools\vcpkg\bootstrap-vcpkg.bat
     popd
     #vcpkg remove sdl2:x86-windows --recurse
     # (lib)vlc is missing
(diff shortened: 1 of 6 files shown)


lynxlynxlynx:
Thanks!

DudeMcDude:
I forgot to comment APPVEYOR_SAVE_CACHE_ON_ERROR...

Also for some reason it looks like it stopped using the cache on the commit right after that one. Wonder why that is...

lynxlynxlynx:
That's odd indeed. It used the cache in the subsequent PR, but never in the main branch.
One thing we didn't pin was the vcpkg version itself. Let's always grab the same release tag / commit, so there won't be any surprises.

lynxlynxlynx:
Bleh, another cycle. It used the cache in 6434b6cda89fc2, but merging your PRs again skipped it — 447cdc2644a.
It looks like the cancelled builds cause it: https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/44275180
We use rolling builds not to be too much of a burden.

It's now clear how to fix this — indeed we can't keep APPVEYOR_SAVE_CACHE_ON_ERROR enabled all the time.

PR #1658, opened 2022-07-25
Formations ctor crash 

DudeMcDude:
Would otherwise crash on debug build, at least when running IWD2.

## Description
Was doing reserve instead of resize.

## Checklist


lynxlynxlynx:
Thanks! I'll wait with the merge, since I don't want to cancel the current Appveyor build. :) The cache is back.

PR #1659, opened 2022-07-26
VCPKG manifest added to CMakelists.txt for win

DudeMcDude:
## Description
To simplify installation instructions for us MSVC peasants :)

## Checklist

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -73,7 +73,7 @@ before_build:
     $replace = "GemRBPath=."
     (Get-Content $file) -replace $find, $replace | Set-Content $file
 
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Ok, that worked. So please rebase on current master and force push. Or if that's a problem for you, I can just cherry-pick the first commit.

Also, why would anyone want to override VCPKG_MANIFEST_DIR instead of using the provided one?

DudeMcDude:
> 



> Ok, that worked. So please rebase on current master and force push. Or if that's a problem for you, I can just cherry-pick the first commit.
> 
> Also, why would anyone want to override VCPKG_MANIFEST_DIR instead of using the provided one?

Cherry picking would be easiest :)
At least I found out how to debug it in case it goes wrong...

PR #1664, opened 2022-07-30
BIFImporter: wrong delete operator

MarcelHB:
## Description
There is a mismatching `delete`-operator in use. Interestingly, GCC started to complain about this only today.

## Checklist


lynxlynxlynx:
Thanks!

Issue #1663, opened 2022-07-29
Keys in bags are not visible to scripts


rsantellan:
#### Bug description
I'm on Umar hills and just have all the pieces of the Symbol of Amaunator, but the door to going down is not opening.
It says **Locked door**

```
[ResourceManager]: Found 'CHMB3g1.bam' in 'chitin.key'.
[ResourceManager]: Searching for 'ARM_04c'...
[ResourceManager]: Found 'ARM_04c.wav' in 'chitin.key'.
[ResourceManager]: Found 'CEFF4g18.bam' in 'chitin.key'.
[ResourceManager]: Found 'formatio.2da' in 'GemRB Unhardcoded data'.
[ResourceManager]: Searching for 'error02'...
[ResourceManager/WARNING]: Couldn't find 'error02'... Tried error02.wav error02.acm error02.wav error02.ogg 
[ResourceManager]: Searching for 'AMB_D20'...
[ResourceManager]: Found 'AMB_D20.wav' in 'chitin.key'.
[ResourceManager]: Searching for 'ARM_03e'...
[ResourceManager]: Found 'ARM_03e.wav' in 'chitin.key'.
[ResourceManager]: Found 'CIFB3g12.bam' in 'chitin.key'.
^C[ResourceManager]: Found 'GUIOPT.chu' in 'chitin.key'.
[ResourceManager]: Searching for 'GUIERR2'...
[ResourceManager]: Found 'GUIERR2.mos' in 'Override'.
[ResourceManager]: Found 'GUIERRB.bam' in 'chitin.key'.
[ResourceManager]: Searching for 'GAM_03'...
[ResourceManager]: Found 'GAM_03.wav' in 'chitin.key'.
```

#### Steps to reproduce
1. On umar hills
2. Get Symbol of Amauntur
3. Try to open de door

#### Expected behavior
Door open

#### Savegame
[000000061-Cant open door.tar.gz](https://github.com/gemrb/gemrb/files/9215281/000000061-Cant.open.door.tar.gz)

GemRB version (check as many as you know apply)
- [X] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [X] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
[...]

PR #1669, opened 2022-08-04
Keys in bags are not visible to scripts 

rsantellan:
## Description
This is for fixing this bug #1663

Change logic in Highlightable::TryUnlock to search on the Bags also.
The remove logic was also changed so if not removed from the inventory then it will search in all the bags.

## Checklist

### DIFF
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -411,12 +411,7 @@ bool RemoveItemOfStore(const ResRef& storename, const ResRef& itemname)
 		Log(ERROR, "GameScript", "Store cannot be opened!");
 		return false;
 	}
-
-	//RemoveItem doesn't use trigger, and hopefully this will always run on bags (with no triggers)
-	unsigned int idx = store->FindItem(itemname, false);
-	if (idx == (unsigned int) -1) return false;
-	STOItem *si = store->GetItem(idx, false);
-	store->RemoveItem(si);
+	store->RemoveItemByName(itemname);
 	//store changed, save it
 	gamedata->SaveStore(store);
 	return true;
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -404,7 +404,7 @@ bool HasItemCore(const Inventory *inventory, const ResRef& itemname, ieDword fla
 	return false;
 }
 
-bool RemoveItemOfStore(const ResRef& storename, const ResRef& itemname)
+bool RemoveStoreItem(const ResRef& storename, const ResRef& itemname)
 {
 	Store* store = gamedata->GetStore(storename);
 	if (!store) {
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -404,6 +404,24 @@ bool HasItemCore(const Inventory *inventory, const ResRef& itemname, ieDword fla
 	return false;
 }
 
+bool RemoveItemOfStore(const ResRef& storename, const ResRef& itemname)
+{
+	Store* store = gamedata->GetStore(storename);
+	if (!store) {
+		Log(ERROR, "GameScript", "Store cannot be opened!");
+		return false;
+	}
+
+	//RemoveItem doesn't use trigger, and hopefully this will always run on bags (with no triggers)
+	unsigned int idx = store->FindItem(itemname, false);
+	if (idx == (unsigned int) -1) return false;
+	STOItem *si = store->GetItem(idx, false);
+	store->RemoveItem(si);
+	//store changed, save it
+	gamedata->SaveStore(store);
+	return true;
+}
+
 //finds and takes an item from a container in the given inventory
 static bool GetItemContainer(CREItem &itemslot2, const Inventory *inventory, const ResRef& itemname, int count)
 {
--- a/gemrb/core/Scriptable/Scriptable.cpp
+++ b/gemrb/core/Scriptable/Scriptable.cpp
@@ -1820,14 +1820,13 @@ bool Highlightable::TryUnlock(Actor *actor, bool removekey) const {
 		for (int idx = 0; idx < game->GetPartySize(false); idx++) {
 			Actor *pc = game->FindPC(idx + 1);
 			if (!pc) continue;
-
-			if (pc->inventory.HasItem(KeyResRef,0)) {
+			if (HasItemCore(&pc->inventory, KeyResRef, 0)) {
 				haskey = pc;
 				break;
 			}
 		}
 	// actor is not in party, check only actor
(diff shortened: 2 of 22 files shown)


gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -404,6 +404,24 @@ bool HasItemCore(const Inventory *inventory, const ResRef& itemname, ieDword fla
...
+
+	//RemoveItem doesn't use trigger, and hopefully this will always run on bags (with no triggers)
+	unsigned int idx = store->FindItem(itemname, false);
+	if (idx == (unsigned int) -1) return false;
+	STOItem *si = store->GetItem(idx, false);
+	store->RemoveItem(si);
+	//store changed, save it
+	gamedata->SaveStore(store);
+	return true;
+}
```
lynxlynxlynx, line 423: Did you plan to use this in some other function as well now? Otherwise it doesn't make much sense to be here.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -1837,7 +1836,24 @@ bool Highlightable::TryUnlock(Actor *actor, bool removekey) const {
...
+				//maybe we could speed this up if we mark bag items with a flags bit
+				const CREItem *itemslot = haskey->inventory.GetSlotItem(i);
+				if (!itemslot)
+					continue;
+				const Item *itemStore = gamedata->GetItem(itemslot->ItemResRef);
+				if (!itemStore)
+					continue;
+				if (core->CanUseItemType(SLOT_BAG, itemStore)) {
+					//the store is the same as the item's name
+					RemoveItemOfStore(itemslot->ItemResRef, KeyResRef);
```
lynxlynxlynx, line 1852: Is it important if it fails? I guess not.

gemrb/core/GameScript/GSUtils.cpp
rsantellan: Do I should move it to Scriptable.cpp ? My first thought was doing everything on this class but I have to pass a lot of objects so I did the minimum  here to remove from store and put the rest of the logic on **Scriptable.cpp**

gemrb/core/Scriptable/Scriptable.cpp
rsantellan: You could have multiple bags and it should look it in all of them.

gemrb/core/GameScript/GSUtils.cpp
lynxlynxlynx: First check if it can be reused, since it looks familiar. Eg. by StoreGetItemCore or something else like that. That will impact the best place to have it. A static method of Store is also an option.

rsantellan: Is very similar to StoreGetItemCore the difference is the check of the amount in stock. I could convert the two of them in one with an extra parameter to be call to just remove or take in account the amount in stock.

lynxlynxlynx: You'd need more, since it modifies the item by reference. So let's skip this then. Move it to Store, remove the comments (useless) and be consistent with the spacing.

rsantellan:
@lynxlynxlynx this have a lot of commit because I did a rebase because the file **gemrb/core/GameScript/GSUtils.cpp** was changed on master. I can cherry-pick to another MR is you want to be clean.

lynxlynxlynx:
I think you have to rebase everywhere, also to your own branch, if you want to avoid this from happening. It's not normal. So when you update your feature branch from your master, use rebase there as well.

gemrb/core/Store.cpp
```diff
@@ -339,6 +339,14 @@ void Store::RemoveItem(const STOItem *itm)
 	}
 }
 
+void Store::RemoveItemByName(const ResRef &itemname)
+{
+	unsigned int idx = this->FindItem(itemname, false);
+	if (idx == (unsigned int) -1) return;
+	STOItem *si = this->GetItem(idx, false);
```
lynxlynxlynx, line 346: should be const

gemrb/core/Store.cpp
```diff
@@ -339,6 +339,14 @@ void Store::RemoveItem(const STOItem *itm)
 	}
 }
 
+void Store::RemoveItemByName(const ResRef &itemname)
```
lynxlynxlynx, line 342: Please use camelCase names for vars

gemrb/core/Store.cpp
```diff
@@ -339,6 +339,14 @@ void Store::RemoveItem(const STOItem *itm)
...
+{
+	unsigned int idx = this->FindItem(itemname, false);
+	if (idx == (unsigned int) -1) return;
+	STOItem *si = this->GetItem(idx, false);
+	this->RemoveItem(si);
```
lynxlynxlynx, line 347: all these `this` are redundant

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -404,6 +404,19 @@ bool HasItemCore(const Inventory *inventory, const ResRef& itemname, ieDword fla
 	return false;
 }
 
+bool RemoveItemOfStore(const ResRef& storename, const ResRef& itemname)
```
lynxlynxlynx, line 407: I suggest the simpler RemoveStoreItem instead.

rsantellan:
> I think you have to rebase everywhere, also to your own branch, if you want to avoid this from happening. It's not normal. So when you update your feature branch from your master, use rebase there as well.

I did that and didn't work, this time using the github gui for fetching and merging and didn't work also. I will keep looking how to fix this, because I don't see anything bad on it (I use it in other repos) but this is the only place that I do it from a fork. I will make the changes and when is ready I make another PR with only my commits.

gemrb/core/Store.cpp
rsantellan: But **RemoveItem** is not a const it will not work. I have to define RemoveItem as const

lynxlynxlynx: Why? It's just about this STOItem that you don't modify and RemoveItem also doesn't.

lynxlynxlynx:
This merge commit says otherwise:
https://github.com/gemrb/gemrb/pull/1669/commits/be8d2bbe36ab6184dd65cd85126ed90df8bbf0d3

Merging by default doesn't rebase, fast-forward, but creates a merge commit.

rsantellan:
@lynxlynxlynx I update with the changes, I will follow this on next one: https://stackoverflow.com/questions/40413071/after-rebasing-my-github-fork-commits-from-others-are-in-my-pull-request# 
I will cherry pick the changes if they are good and create a clean PR.

lynxlynxlynx:
You dealt with the last change requests nicely, but it's too noisy to review as a whole. So please go ahead.

rsantellan:
> You dealt with the last change requests nicely, but it's too noisy to review as a whole. So please go ahead.

I create this one #1677

Issue #1671, opened 2022-08-05
enable opengl for windows builds


lynxlynxlynx:
Looking at AppVeyor logs confirms we're not building with any GL. Luckily there's a handy package in VCPKG:
https://github.com/microsoft/vcpkg/tree/master/ports/opengl

So for most users it should just be about installing that and passing the OPENGL_BACKEND flag to cmake. I'm not sure if we can use the same approach for AppVeyor, because on closer inspection it so far only supports Windows 8+, while the bots run on Windows Server:
https://github.com/microsoft/vcpkg/blob/master/ports/opengl/portfile.cmake#L4
https://www.appveyor.com/docs/windows-images-software/#operating-system

Since we use the 2019 image, we should be fine:
https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version

But would we be cutting off users on 7 or older?

MarcelHB:
The check above is testing for the SDK, not Windows, that just follows the same version style. But even newer Windows SDKs can be used for applications on [older Windows, such as 7](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/). Of course this may fail on using stuff that is exclusive to new versions.

PR #1672, opened 2022-08-05
windows: fetch opengl, try building it on appveyor

lynxlynxlynx:
Blind try for #1671

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -74,7 +74,11 @@ before_build:
     $replace = "GemRBPath=."
     (Get-Content $file) -replace $find, $replace | Set-Content $file
 
-    cmake -G "Visual Studio 16 2019" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DINSOURCEBUILD=1 -DWIN32_USE_STDIO=1 -DCMAKE_INSTALL_PREFIX="$env:archive_source" -DCMAKE_TOOLCHAIN_FILE=C:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET="x86-windows" -A Win32 .
+    $env:gl = "None"
+    if ($env:APPVEYOR_REPO_TAG_NAME -Or $env:APPVEYOR_REPO_COMMIT_MESSAGE -like '*buildgl*') {
+      $env:gl = "OpenGL"
+    }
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Oh wow, look at all the choices :)
```
[SDL 2 Driver/DEBUG]: Render driver count: 5
[SDL 2 Driver/DEBUG]: Driver 4 named software
[SDL 2 Driver/DEBUG]: Driver 3 named opengles2
[SDL 2 Driver/DEBUG]: Driver 2 named opengl
[SDL 2 Driver/DEBUG]: Driver 1 named direct3d11
[SDL 2 Driver/DEBUG]: Driver 0 named direct3d
```

MarcelHB:
Hate to disappoint you, but this is a compile-time information: https://github.com/zielmicha/SDL2/blob/373232d63fffdca74fffefab97c5e455576fbb91/src/render/SDL_render.c#L47-L79

MarcelHB:
Hate to disappoint you, but this is a compile-time composition: https://github.com/libsdl-org/SDL/blob/17b329d199899da4bc328632f852e3c96f98c665/src/render/SDL_render.c#L92 + https://github.com/libsdl-org/SDL/blob/17b329d199899da4bc328632f852e3c96f98c665/src/render/SDL_render.c#L661

lynxlynxlynx:
No matter, we are looking for the opengl one anyway. -1 doesn't cut it in this case.

lynxlynxlynx:
Hmm, now it fails with a permission error before even reaching the new code (or the log is buffered and we just don't see). @DudeMcDude can you try building with this PR applied and enabling opengl?

bradallred:
> or the log is buffered and we just don't see

yeah, it _is_ on another thread, so that can happen, but if the process is exiting without crashing all pending messages should have been written and flushed...

DudeMcDude:
> Hmm, now it fails with a permission error before even reaching the new code (or the log is buffered and we just don't see). @DudeMcDude can you try building with this PR applied and enabling opengl?

opengles matches "opengl" pattern :)
Other than that seems to work (launched demo at least after fixing that).

lynxlynxlynx:
Ah, good catch, thanks!

lynxlynxlynx:
Didn't help for the test, so I guess it's just a limitation of the VM. It would be bad to disable the test, but we can compromise: build opengl only for tags and not test, while the rest remain using software rendering and do run the test. Would also mean we'd have pretty much identical builds around release time for both video drivers.

lynxlynxlynx:
Hmm, so it doesn't like something else — the last build was without opengl and the same testing regimen.

lynxlynxlynx:
Ah, I see it now, we still force opengl, doh.

lynxlynxlynx:
So now it's confirmed that SDL's solution (at least in 2.0.22) isn't enough or the renderer is bad:
```
[SDL 2 Driver]: Creating display: opengl
[SDL 2 GL Driver/FATAL]: OpenGL backend must be used instead of direct3d
[SDL20Video/DEBUG]: Renderer: direct3d
```
Looking at the code it means it failed to create the renderer internally and then fell back to -1 behaviour:
https://github.com/libsdl-org/SDL/blob/main/src/render/SDL_render.c#L972
Which explains why the bot didn't like when we forced opengl. So it's all clear. :)

lynxlynxlynx:
fix gl env var sharing — we shouldn't be testing in gl mode

PR #1677, opened 2022-08-08
1663 Keys in bags are not visible to scripts

rsantellan:
## Description
This is for fixing this bug #1663

Change logic in Highlightable::TryUnlock to search on the Bags also.
The remove logic was also changed so if not removed from the inventory then it will search in all the bags.

## Checklist

### DIFF
--- a/gemrb/core/Scriptable/Scriptable.cpp
+++ b/gemrb/core/Scriptable/Scriptable.cpp
@@ -1820,14 +1820,13 @@ bool Highlightable::TryUnlock(Actor *actor, bool removekey) const {
 		for (int idx = 0; idx < game->GetPartySize(false); idx++) {
 			Actor *pc = game->FindPC(idx + 1);
 			if (!pc) continue;
-
-			if (pc->inventory.HasItem(KeyResRef,0)) {
+			if (HasItemCore(&pc->inventory, KeyResRef, 0)) {
 				haskey = pc;
 				break;
 			}
 		}
 	// actor is not in party, check only actor
-	} else if (actor->inventory.HasItem(KeyResRef, 0)) {
+	} else if (HasItemCore(&actor->inventory, KeyResRef, 0)) {
 		haskey = actor;
 	}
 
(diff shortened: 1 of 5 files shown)


gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -1837,7 +1836,24 @@ bool Highlightable::TryUnlock(Actor *actor, bool removekey) const {
...
+			int i= haskey->inventory.GetSlotCount();
+			while (i--) {
+				//maybe we could speed this up if we mark bag items with a flags bit
+				const CREItem *itemslot = haskey->inventory.GetSlotItem(i);
+				if (!itemslot)
+					continue;
+				const Item *itemStore = gamedata->GetItem(itemslot->ItemResRef);
+				if (!itemStore)
+					continue;
+				if (core->CanUseItemType(SLOT_BAG, itemStore)) {
```
lynxlynxlynx, line 1850: This function does a lot of things and its use here is not so obvious. Do you need more than the `int ret = (slotmatrix[item->ItemType]&slottype);` check?

rsantellan: I only need to now if is a SLOT_BAG, how can I get the slotmatrix ?

lynxlynxlynx: You'll then need to add an accessor to Interface. Maybe `CheckItemType(Item*, int slotType)`. I see it can then be used elsewhere as well.

rsantellan: Added you method in Interface.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -1845,9 +1845,7 @@ bool Highlightable::TryUnlock(Actor *actor, bool removekey) const {
 				if (!itemslot)
 					continue;
 				const Item *itemStore = gamedata->GetItem(itemslot->ItemResRef);
-				if (!itemStore)
```
lynxlynxlynx: Someone has to have this responsibility. Now the code would crash on a null itemStore.

gemrb/core/Interface.cpp
```diff
@@ -3177,6 +3177,11 @@ int Interface::CanUseItemType(int slottype, const Item *item, const Actor *actor
 	return ret;
 }
 
+int Interface::CheckItemType(const Item *item, int slottype) const
+{
+	return (slotmatrix[item->ItemType]&slottype);
```
lynxlynxlynx, line 3182:
```suggestion
	return slotmatrix[item->ItemType] & slottype;
```

lynxlynxlynx: and camelCase ...

lynxlynxlynx:
Thanks!

PR #1680, opened 2022-08-10
appveyor: more debugging why it's not using the provided GamePath

lynxlynxlynx:
grrr

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -109,15 +109,13 @@ after_build:
     }
 test_script:
 - ps: |-
-    if ($env:gl -ne "OpenGL") {
-      cd $env:archive_source
-      cp GemRB.cfg.sample GemRB.cfg.bots
-      Add-Content -Path GemRB.cfg.bots -Encoding ascii -Value "GamePath=minimal"
-      Add-Content -Path GemRB.cfg.bots -Encoding ascii -Value "GameType=test"
+    cd $env:archive_source
+    cp GemRB.cfg.sample GemRB.cfg.bots
+    Add-Content -Path GemRB.cfg.bots -Encoding ascii -Value "GamePath=minimal"
+    Add-Content -Path GemRB.cfg.bots -Encoding ascii -Value "GameType=test"
 
-      ./gemrb.exe -q -c GemRB.cfg.bots
-      rm GemRB.cfg.bots
-    }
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Ok, so it's really not a new appending problem.

@DudeMcDude does running the minimal test work for you? It should just start up and exit immediately.

You can try with your existing working config by appending `GamePath=minimal ` and `GameType=test` to it, then running it.

DudeMcDude:
It brings up the demo instead. Are you sure I should specify GamePath=minimal and not an actual path?

lynxlynxlynx:
That's the path in our source tree. We fall back to the demo if nothing is found and the fact that we don't find the minimal test any more is the regression. Can you step through Interface.cpp around line 910 to see why it doesn't work in practice any more? Maybe the string changes need a fix in the macro etc.

DudeMcDude:
The working directory is the build path, whereas the chitin files are in the GemRBPath (which is the install path). It tries to look up minimal relative to the working directory, but it looks up the demo files relative to GemRBPath (eventually).
So either the minimal folder needs to be copied to the build path, or the minimal test should fall back to GemRBPath as well.

lynxlynxlynx:
I can see how that could be broken for a build-only test, but this one is ran on the installed files in a build where the build dir is the source dir (-DINSOURCEBUILD=1), so where both the exe and "minimal" are in the same folder. And we run the test from it.

At least that's how I read the recipe, I'll readd a file listing just so it's clearer.

lynxlynxlynx:
Bleh, this build ignored caches so it will likely fail. Anyway, we set `GemRBPath=.` earlier in the build and that code hasn't been touched at all. So I'd be surprised if GemRBPath didn't point to the CWD when resolving the minimal dataset.

DudeMcDude:
Ok, so should I just set the path to where it minimal actually is?

lynxlynxlynx:
For your usecase, if you want, yes. But it should already be fine for appveyor. That's the conundrum.

lynxlynxlynx:
Not sure why it's throwing VCPKG_MANIFEST_DIR errors though.

lynxlynxlynx:
Hah, found the reason.

This results in a two-byte representation, so of course we don't see either var as set:
```
echo GamePath=minimal >> GemRB.cfg.bots
echo GameType=test >> GemRB.cfg.bots
```
image

Issue #1678, opened 2022-08-09
BG2: Not pausing entering menu 


rsantellan:
#### Bug description
On the original entering the menu or the inventory will pause the game.
That was lost recently.

#### Steps to reproduce
1. Go to a place with space
2. Click on the other side of the map
3. Enter menu
4. See error, you will arrive to the other side

#### Expected behavior
It should pause the game

#### Screenshots

GemRB version (check as many as you know apply)
- [X] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [X] SDL2 without `USE_OPENGL_BACKEND`

bradallred:
`CreateTopWinLoader` already takes a pause parameter so it's quite easy to do.

rsantellan:
> `CreateTopWinLoader` already takes a pause parameter so it's quite easy to do.

I created the PR, I can only test it on BG2, don't know how the other behave.

lynxlynxlynx:
bg1 famously didn't pause in the inventory, bg2, at least in tob, not in the map window, so this looks fine.

PR #1683, opened 2022-08-11
Pausing on accessing the menu

rsantellan:
Added pause when opening:

- Options
- Inventory
- Character Record
- Priest Spells
- Mage Spells
- Sorcerer Spells
- Journal

This is only applying to BG2. Missing what the others do.

## Description
Fixing bug #1678 

## Checklist


lynxlynxlynx:
Thanks!

PR #1688, opened 2022-08-20
Stats overflow crashes fix

Arc0re:
## Description

Using the linux build (Ubuntu 22.04.1 LTS x86_64, package from the default repository), I had multiple crashes during combat, which I tracked down to be caused by an overflow: the resist_stat of RESISTPOISON was read as -1, which is then cast to a signed integer, causing a massive overflow in the Actor::stats_t maximum_values table when checking.

I don't know neither the games or the data file formats enough to figure out what causes this wrong read, but it seems IDSImporter::GetValue(StringView txt) should maybe be changed to not return an int anymore (that would mean changing the GemRB::IDSImporter::Pair struct, but I don't know the implications to that).
It also seems the RESISTPOISON is the only one having a wrong value, when I logged them all.

Maybe this is a bug in the original game, no idea.

## Checklist

- [] I extended the documentation, if necessary

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -4277,6 +4277,7 @@ ieDword Interface::TranslateStat(const char *stat_name)
 	ieDword stat = sym->GetValue(StringView(stat_name));
 	if (stat==(ieDword) ~0) {
 		Log(WARNING, "Core", "Cannot translate symbol: {}", stat_name);
+		stat = 0; // dont let the overflow happen
 	}
 	return stat;
 }
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -7088,10 +7088,15 @@ void Actor::ModifyDamage(Scriptable *hitter, int &damage, int &resisted, int dam
 				damage -= resisted;
 			} else {
 				int resistance = (signed)GetSafeStat(it->second.resist_stat);
-				// avoid buggy data
-				if ((unsigned)abs(resistance) > maximum_values[it->second.resist_stat]) {
-					resistance = 0;
-					Log(DEBUG, "ModifyDamage", "Ignoring bad damage resistance value ({}).", resistance);
+				// sanity check
+				if (it->second.resist_stat <= MAX_STATS) {
+					// avoid buggy data
+					if ((unsigned)abs(resistance) > maximum_values[it->second.resist_stat]) {
+						resistance = 0;
+						Log(DEBUG, "ModifyDamage", "Ignoring bad damage resistance value ({}).", resistance);
+					}
+				} else {
+					Log(ERROR, "resist_stat overflow: ", "{}", it->second.resist_stat);
 				}
 				resisted += (int) (damage * resistance/100.0);
 				damage -= resisted;


lynxlynxlynx:
Thanks, yet should fix it at the source, not patch it over. Were you trying with our latest code? 
`Interface::ReadDamageTypeTable` reads and translates the stat as expected for me (74 == IE_RESISTPOISON).

Arc0re:
> Thanks, yet we should fix it at the source, not patch it over. Were you trying with our latest code? `Interface::ReadDamageTypeTable` reads and translates the stat as expected for me (74 == IE_RESISTPOISON).

I was only 3 or 4 commits behind when I wrote the code, didn't try with the most recent one yes, I shall give it a try. I agree this seems to be a data problem, but those data files are supposed to be coming from the original games (I'm using the Gog version), could it be a bug in the original data?

lynxlynxlynx:
Part of the mapping is in our data (dmgtypes.2da) and part is in the original stats.ids (stat name to value). There's no bug in this data or we'd notice long ago.

For some time after 0.9.0 we did have problems though, due to an insufficient Explode method. RESISTPOISON would have been broken then as well, since if you look at the ids file, it has trailing whitespace after the name. But instead of speculating, it's really easier if you just step through to see why it's being read as -1 for you.

lynxlynxlynx:
I tried the latest appimage, set poison resistance on someone and cast cloudkill on everyone. No problems.

lynxlynxlynx:
It would be also interesting to know what compiler you used.

bradallred:
I don't really understand the use of `abs` here. cant we just do:
```diff
diff --git a/gemrb/core/Scriptable/Actor.cpp b/gemrb/core/Scriptable/Actor.cpp
index d09cd063a..7f03ec4b7 100644
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -7081,9 +7081,9 @@ void Actor::ModifyDamage(Scriptable *hitter, int &damage, int &resisted, int dam
 				}
 				damage -= resisted;
 			} else {
-				int resistance = (signed)GetSafeStat(it->second.resist_stat);
+				auto resistance = GetSafeStat(it->second.resist_stat);
 				// avoid buggy data
-				if ((unsigned)abs(resistance) > maximum_values[it->second.resist_stat]) {
+				if (resistance > maximum_values[it->second.resist_stat]) {
 					resistance = 0;
 					Log(DEBUG, "ModifyDamage", "Ignoring bad damage resistance value ({}).", resistance);
 				}
```

bradallred:
now I wonder why `stat_t` is unsigned if it allows negative values. something for later tho.

lynxlynxlynx:
It's just the way these are stored in CREs, not much we can do about it.

lynxlynxlynx:
@Arc0re will you have time to dig into this or shall we just close it? It works both for me and from the official build ...

Arc0re:
Hello, sorry for the late answer - can't seem to get notifications or something from Github...
The compiler I used is: gcc version 11.2.0 (Ubuntu 11.2.0-19ubuntu1) , with the cmake build configured as:
cmake .. -DSDL_BACKEND=SDL2 -DOPENGL_BACKEND=None
The options I picked aren't random: they seem to be the exact replica of the built available in the default Ubuntu repository.

Arc0re:
I'm going to try the latest build and let you know.

Arc0re:
Okay so after a fresh build, I don't have any more crashes related to the resist stat, but I hit https://github.com/gemrb/gemrb/commit/dd2ae3a705c89f2e1feecdd240809c310938b074 ("Projectile", "Travelled over full distance..."). Maybe something is just wrong with my data files and this isn't worth looking over indeed

lynxlynxlynx:
No, I'm also on gog, but good to know. If you have a reproducer for this projectile assert or at least the full message, that would be helpful.

Arc0re:
> No, I'm also on gog, but good to know. If you have a reproducer for this projectile assert or at least the full message, that would be helpful.

[ResourceManager]: Searching for 'TRA_22'...
[ResourceManager/WARNING]: Couldn't find 'TRA_22'... Tried TRA_22.wav TRA_22.ogg TRA_22.acm TRA_22.wav 
[ResourceManager]: Found 'magicmis.pro' in 'GemRB Unhardcoded data'.
[ResourceManager]: Searching for 'TRA_02'...
[ResourceManager]: Found 'TRA_02.wav' in 'chitin.key'.
**[Projectile/FATAL]: Travelled over full distance (1.2592592592592593 = 34 / 27)! Origin: (3326, 2174), Destination: (3353, 2168), Pos: (3360, 2175)**
AL lib: (EE) alc_cleanup: 1 device not closed

Here is the last lines of the log. To trigger this error, I just got into the fight outside the Friendly Arm Inn. I hope this is useful

lynxlynxlynx:
Thanks, that's good. So magic missiles triggered it. I just tried and had no problems, but everyone was static, let's see what happens if the target moves ... Do you do anything special in the fight?

Arc0re:
> Thanks, that's good. So magic missiles triggered it. I just tried and had no problems, but everyone was static, let's see what happens if the target moves ... Do you do anything special in the fight?
> 
> EDIT: nope, no luck. Moved after the projectile was launched, tried with a moving target, all fine. Just approaching the guy on the stairs.

Well I'll have to step-by-step this one when I have the time, thanks for your time!

lynxlynxlynx:
Maybe I need to try at longer range; the check looks a bit suspicious anyway, since it's comparing a missile that can have a curved path to a path along a line. I think we'll just snap it to 1.0 in these rare cases and be done with it.

PR #1701, opened 2022-10-02
Fixes for SDL v2.24 on Windows

MarcelHB:
## Description
After a bunch of updates, I was unable to build GemRB on MSYS2 anymore. It appears that there has been some [major changes](https://github.com/libsdl-org/SDL/pull/5727) around SDL2's cmake setup with v2.24.

What happens now: linker receives the instruction to take `SDL2main` into linking of the plugins, that then expect to be linked against whatever satifies `SDL_main`. But since the executables for that are supposed to be elsewhere, we don't want that for the plugins.

Unfortunately I haven't found a better way to switch the behavior before finding SDL2, so I removed that token. At some later point, I'll be checking Linux and static linking and may come up with another PR.

## Checklist


lynxlynxlynx:
Is this finished?

MarcelHB:
Yes, open for comments and alternative ideas.

lynxlynxlynx:
It looks fine to me and I can't think of anything nicer. The replacement likely affect the global var, but that's inconsequential.

lynxlynxlynx:
Thanks!

PR #1705, opened 2022-11-06
Consolidate Geometry into Geometry.h/cpp

bradallred:
Mainly this is to add generic Circle/Ellipse plotting and to use these functions in the myriad of places we are creating such shapes.

A couple of things worth mentioning:

1. lights can now be elliptical and the Projectile/ScriptedAnimation code has been updated to use the previously unused `LightY` variable.
2. the circles created for PathFinding are slightly different at larger sizes, but I suspect this is closer to the originals based on existing comments and the fact that the originals certainly would have used something faster than brute force to calculate a circle.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,15 @@ ELSE(DEFINED CMAKE_BUILD_TYPE)
 	SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
 ENDIF(DEFINED CMAKE_BUILD_TYPE)
 
+if(CMAKE_BUILD_TYPE MATCHES "Rel.*")
+	include(CheckIPOSupported)
+	check_ipo_supported(RESULT supported)
+
+	if(supported)
+		set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE CACHE BOOL)
+	endif()
+endif()
+
 # VCPKG manifest file: platforms/windows/vcpkg.json
 # must be declared before PROJECT()
 IF(WIN32)
--- a/gemrb/core/Geometry.h
+++ b/gemrb/core/Geometry.h
@@ -25,6 +25,8 @@
 
 #include "Region.h"
 
+#include <vector>
+
 namespace GemRB {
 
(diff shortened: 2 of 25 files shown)


CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ if(COMMAND cmake_policy)
 endif(COMMAND cmake_policy)
 
 set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```
lynxlynxlynx, line 8: This should be tested for first:
https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html#module:CheckIPOSupported

LTO in general increases build times, can expose bugs in our code and debuggers ... It's probably better to have this as an off-by-default option, so it's easy to enable for those who want it.

gemrb/core/Geometry.h
```diff
@@ -51,6 +53,29 @@ GEM_EXPORT bool intersectSegments(const Point& a, const Point& b, const Point& c
...
+  3____⋱⋮⋰___2
+  5    ⋰⋮⋱   4
+   5 ⋰  ⋮  ⋱4
+	  7  6
+ */
+GEM_EXPORT std::vector<Point> PlotCircle(const Point& origin, uint16_t radius, uint8_t octants = 0xff) noexcept;
+
+/*
+ return a vector of Points composing an ellipse bounded by rect
+ p1 and p2 are points relative to the center of rect
```
lynxlynxlynx, line 73: The explanation would better fit as a comment in the implementation, since the function doesn't take any p1/p2 params and so the text doesn't make much sense.

gemrb/core/Light.cpp
```diff
@@ -0,0 +1,82 @@
...
+		
+		const Point& q3 = points[i + 2];
+		const Point& q4 = points[i + 3];
+		assert(q3.y == q4.y);
+		
+		int yint = (q1.y / float(radii.y)) * intensity;
+		for (int x = q1.x; x >= 0; --x) {
+			int xint = (x / float(radii.x)) * intensity;
+			// by the power of Pythagurus
+			int dist = sqrt(xint * xint + yint * yint);
```
lynxlynxlynx, line 67: You can use `std::hypot` here.

bradallred:
Looks like the build is failing in master; the failure here is the same.

lynxlynxlynx:
Yeah, got it covered, just finishing some other work.

bradallred:
the windows build failure is with Python, so im just gonna go ahead and merge

PR #1708, opened 2022-11-10
A number of spell immunity fixes

Tomsod:
Hello again!  Here are some fixes for the spell protection effects, which didn't work properly for quite a while now.  I'll elaborate on the six commit points:

1. "Magic attack" spells such as Pierce Magic now correctly ignore most spell protections they're supposed to remove.  (`check_type()`, `bool pierce`)

Spells with the secondary type 4, which include spell protection strippers (and also Breach), ignore some protective effects, although not all.  [Here's a handy chart.](https://docs.google.com/spreadsheets/d/1y7r2Z3FfZk74NdAVHuZuiNmWh14y1tAytNrsFMSY7zc)  As you can see, they go through decrementing level immunity (Spell Turning), school immunity (Spell Immunity), decrementing level bounce (Spell Turning), and spell trap, although not generic level immunity (Spell Thrust (technically level 4) and Globe of Invulnerability, and see also note on Breach) or secondary type immunity (Spell Shield).  I haven't actually tested some of the more obscure effects, though, so I'm not 100% sure I got everything right, though.  Also the chart mentions e.g. Spell Trap absorbing lower-level type 4 spells (but only some of them!), which I'm not even sure how to implement properly.

2. Spell protections now make an exception for self-cast spells.   (`check_type()`, `bool self`)

Spells cast on oneself, including bounced spells, are even more exempt from protections, probably for convenience reasons.  See [this wiki article](https://baldursgate.fandom.com/wiki/Spell_Turning#Gameplay) about bounced spells and immunities.  I think they only can be blocked by source immunity (e.g. BG2Fixpack's implementation of Sunfire gives the caster 3 second immunity to itself).

3. Implemented projectile immunity.  The old code didn't work properly.  (`check_type()`, `fx_projectile_immunity_ref`)

Neither the damage nor added effects were blocked, now they do.  I remember seeing some nonfunctional code when I first wrote this fix, but now I can't find it!

4. Spell Deflection and similar limited level immunity effects are now correctly removed after reaching their limit.  (`DecreaseParam1OfEffect()`)

Just like Spell Trap, (Minor) Spell Deflection and Turning create an effect per each level, and an absorbed spell should decrease all of them at once.  I also made the decrease function return false if the effects are already spend (they don't disappear immediately, which caused them to overperform, even in IE).

5. Fixed effects that allow MR not checking saving throws.  (`check_resistance()`)

A typo: if MR is allowed, unconditional `return` was issued, skipping the saving throw code.

6. Spells that dispel other spell effects now work correctly.  (`RemoveLevelEffects()`)

Another typo: source ref was compared with `removed` variable which wasn't initialized before the comparison.

### DIFF
--- a/gemrb/core/EffectQueue.cpp
+++ b/gemrb/core/EffectQueue.cpp
@@ -1031,11 +1031,11 @@ static int check_resistance(Actor* actor, Effect* fx)
 
 	//opcode immunity
 	// TODO: research, maybe the whole check_resistance should be skipped on caster != actor (selfapplication)
-	if (caster != actor && actor->fxqueue.HasEffectWithParam(fx_opcode_immunity_ref, fx->Opcode)) {
+	if (actor->fxqueue.HasEffectWithParam(fx_opcode_immunity_ref, fx->Opcode)) {
 		Log(MESSAGE, "EffectQueue", "{} is immune to effect: {}", fmt::WideToChar{actor->GetName()}, globals.Opcodes[fx->Opcode].Name);
 		return FX_NOT_APPLIED;
 	}
-	if (caster != actor && actor->fxqueue.HasEffectWithParam(fx_opcode_immunity2_ref, fx->Opcode)) {
+	if (actor->fxqueue.HasEffectWithParam(fx_opcode_immunity2_ref, fx->Opcode)) {
 		Log(MESSAGE, "EffectQueue", "{} is immune2 to effect: {}", fmt::WideToChar{actor->GetName()}, globals.Opcodes[fx->Opcode].Name);
 		// totlm's spin166 should be wholly blocked by spwi210, but only blocks its third effect, so make it fatal
 		return FX_ABORT;
--- a/gemrb/core/EffectQueue.cpp
+++ b/gemrb/core/EffectQueue.cpp
@@ -1041,10 +1041,13 @@ static int check_resistance(Actor* actor, Effect* fx)
 		return FX_ABORT;
 	}
 
-	//not resistable (but check saves - for chromatic orb instakill)
-	// bg2 sequencer trigger spells have bad resistance set, so ignore them
-	if (fx->Resistance == FX_CAN_RESIST_CAN_DISPEL && signed(fx->Opcode) != EffectQueue::ResolveEffect(fx_activate_spell_sequencer_ref)) {
-		return check_magic_res(actor, fx, caster);
+	// check magic resistance if applicable
+	// (note that no MR roll does not preclude saving throws -- see e.g. chromatic orb instakill)
+	if (fx->Resistance == FX_CAN_RESIST_CAN_DISPEL && check_magic_res(actor, fx, caster) == FX_NOT_APPLIED) {
+		// bg2 sequencer trigger spells have bad resistance set, so ignore them
+		if (signed(fx->Opcode) != EffectQueue::ResolveEffect(fx_activate_spell_sequencer_ref)) {
+			return FX_NOT_APPLIED;
+		}
 	}
 
 	if (globals.pstflags && (actor->GetSafeStat(IE_STATE_ID) & STATE_ANTIMAGIC)) {
--- a/gemrb/core/EffectQueue.cpp
+++ b/gemrb/core/EffectQueue.cpp
@@ -856,9 +856,9 @@ static int check_type(Actor *actor, const Effect& fx)
 
 	//decrementing immunity checks
 	//decrementing level immunity
-	if (fx.Power && fx.Resistance != FX_NO_RESIST_BYPASS_BOUNCE && !self && !pierce) {
-		efx = const_cast<Effect*>(actor->fxqueue.HasEffectWithParam(fx_level_immunity_dec_ref, fx.Power));
-		if (efx && DecreaseEffect(efx)) {
+	if (fx.Power && fx.Resistance != FX_NO_RESIST_BYPASS_BOUNCE && !self && !pierce
+	    && actor->fxqueue.HasEffectWithParam(fx_level_immunity_dec_ref, fx.Power)) {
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Always a pleasure when you pop in. :)

The chart and I guess testing were made against the EEs, not vanilla, so this requires more thought. 
Concrete test cases would be great, without requiring the whole table, of course. Eg. for projectile immunity aurstaf is a good item one can test against (spawn as many needed to fill the inventory, then you will be able to pick up the next one spawned on the ground).

And it's likely to be wrong in parts, so please split the commit into sensible chunks, so it's easier to reason about and bisect back to, if needed.

Tomsod:
No, I'm pretty sure the chart is also mostly correct for vanilla BG2, perhaps except for the mentioned spell trap weirdness.  It's just the most visually convenient source that I have, but mage chess was studied and discussed a lot even before EE.

I've tested projectile immunity with Protection From Normal Missiles, actually, but it working on anything beyond simple arrows/bolts/bullets depends on game version and installed mods, so it's probably not reliable for testing on-hit effects.

As for the separate commits... *sigh* thought of it myself, but a bit too late. I'll do it tomorrow.  You're still cool with force-pushing, right?

lynxlynxlynx:
Yeah, force pushing to PR branches is fine. 
For the test cases it would be good to have one for each change in a sense extra justification for it, but also good to recheck for if the code changes later.

Tomsod:
Okay, here you go.  I also noticed another change here -- opcode immunity not working on a self-cast broke Durlag's Goblet for me, so it's probably shouldn't be there.

gemrb/core/EffectQueue.cpp
```diff
@@ -1031,11 +1031,11 @@ static int check_resistance(Actor* actor, Effect* fx)
 
 	//opcode immunity
 	// TODO: research, maybe the whole check_resistance should be skipped on caster != actor (selfapplication)
-	if (caster != actor && actor->fxqueue.HasEffectWithParam(fx_opcode_immunity_ref, fx->Opcode)) {
+	if (actor->fxqueue.HasEffectWithParam(fx_opcode_immunity_ref, fx->Opcode)) {
```
lynxlynxlynx, line 1034: The check was added to fix the statues from Watcher's keep level 1 from moving immediately, before the plot related scripting animates them. Does that still work fine?

gemrb/core/EffectQueue.cpp
```diff
@@ -1585,13 +1585,14 @@ void EffectQueue::RemoveAllNonPermanentEffects()
...
 {
+	bool found_one = false;
 	ResRef removed;
 	for (auto& fx : effects) {
 		if (fx.Power > level) {
 			continue;
 		}
 
-		if (removed != fx.SourceRef) {
+		if (found_one && removed != fx.SourceRef) {
```
lynxlynxlynx, line 1595: I think it would be cleaner to just check `!removed.IsEmpty()` instead of adding another variable.

gemrb/core/EffectQueue.cpp
```diff
@@ -1736,33 +1730,31 @@ const Effect *EffectQueue::HasEffectWithParamPair(EffectRef &effect_reference, i
...
-		if( value>amount) {
-			value -= amount;
-			amount = 0;
-		} else {
-			amount -= value;
-			value = 0;
-		}
-		fx.Parameter1 = value;
-		if (value) {
-			return;
```
lynxlynxlynx, line 1746: The removal of this short-circuit affects other users as well, but I don't think it makes a difference. They're chaos shield, both stoneskins and mirror image, none of which one can stack AFAIK.

I suggest you add a comment, so later down the line nobody "optimizes" the loop with a break when `found` gets set to true — it intentionally affects all effects of the desired opcode. You can just delete the existing one at the top of the function and do it there.

lynxlynxlynx:
Thanks! It also reminded me that I don't know if we handle (bg2-only) feedback properly. That's mschool.2da and [msectype.2da](https://gibberlings3.github.io/iesdp/files/2da/2da_bgee/msectype.htm) and I don't see us referencing the tables or strings anywhere, even though they're present.

gemrb/core/EffectQueue.cpp
Tomsod: I agree.  I think when I wrote this I was still unclear how your custom string types worked.  Some of this code is several years old, actually, I just never submitted it for some reason.  Anyway, it's the last commit, so I'll just amend it.

gemrb/core/EffectQueue.cpp
Tomsod: I plan to visit the Keep shortly in my current game, I'll check then.

gemrb/core/EffectQueue.cpp
Tomsod: Can do!

gemrb/core/EffectQueue.cpp
Tomsod: Okay, the statues are still and circleless, and (two of them) only attack after picking up the holy book.  Everything seems OK?  Two statues close to the entrance are doing their idle arms-waving animation, though.

lynxlynxlynx: Do you mean the ogre and archer? The spawns are level dependent, but I couldn't get anyone to wave on master.

Tomsod: Ogre and golem, actually.  And by waving I mean they're just wobbling them a little.

Tomsod: Oh, yes, I see, they're indeed resisting HOLDRING's paralysis (opcode 109).  (The other statues don't.)  But they're not moving anyway, so maybe it's OK?

Tomsod: Got it, this is just some mod adding paralyze immunity to all golems (and ogres I guess?)  Should be okay in vanilla.

lynxlynxlynx:
Thanks!

Issue #1709, opened 2022-11-12
BG2: Some spells occasionally not shown in the spellcast bar 


Tomsod:
I occasionally (as recently as last month with git) had a problem with some prepared spells not being shown in the spellcasting menu.  After some digging, it turned out something would change the case of prepared spells' resrefs (usually to lowercase), while known spells remained as usual (usually all caps), which caused Spellbook.py, which compared these resrefs case-sensitively, to react poorly.  Now, I never figured out the cause or even how to reliably reproduce it (and I don't even have a save right now, sorry), but I at least was able to fix the consequences by rewriting the spellbook code so it would ignore case.  I can provide a pull request if you want, but really, finding out what mangles spells' case would also be prudent.

Here's the code just in case:
```diff
diff --git a/gemrb/GUIScripts/Spellbook.py b/gemrb/GUIScripts/Spellbook.py
index e834f9674..b51ac3e77 100644
--- a/gemrb/GUIScripts/Spellbook.py
+++ b/gemrb/GUIScripts/Spellbook.py
@@ -51,16 +51,17 @@ def GetUsableMemorizedSpells(actor, BookType):
 			if not Spell0["Flags"]:
 				# depleted, so skip
 				continue
-			if Spell0["SpellResRef"] in spellResRefs:
+			resref = Spell0["SpellResRef"].lower()
+			if resref in spellResRefs:
 				# add another one, so we can get the count more cheaply later
-				spellResRefs.append (Spell0["SpellResRef"])
+				spellResRefs.append (resref)
 				continue
-			spellResRefs.append (Spell0["SpellResRef"])
-			Spell = GemRB.GetSpell(Spell0["SpellResRef"])
+			spellResRefs.append (resref)
+			Spell = GemRB.GetSpell(resref)
 			Spell['BookType'] = BookType # just another sorting key
 			Spell['SpellIndex'] = GemRB.GetSpelldataIndex (actor, Spell["SpellResRef"], 1<<BookType) # crucial!
 			if Spell['SpellIndex'] == -1:
-				GemRB.Log (LOG_ERROR, "GetUsableMemorizedSpells", "Memorized spell not found! " + Spell["SpellResRef"] + " of type " + 1<<BookType)
+				GemRB.Log (LOG_ERROR, "GetUsableMemorizedSpells", "Memorized spell not found! " + Spell["SpellResRef"] + " of type " + str(1<<BookType))
 			Spell['SpellIndex'] += 1000 * 1<<BookType
[...]

PR #1710, opened 2022-11-13
Spellbook: compare resrefs case-insensitively

Tomsod:
Normally it would be fine, but an inexplicable bug occasionally converted memorized spell resrefs to lowercase, at which point they would be misparsed and would dissappear from the action bar.

Also, fix an unrelated typo (string + int addition attempt).

This is related to #1709.


lynxlynxlynx:
Thanks!

PR #1717, opened 2022-11-15
BG2: fix scrolls cast on other actors instead of protagonist

Tomsod:
The old code for `Spell:Cast` effect (which scrolls use) created a `ForceSpell()` action with the target's script name (unless casting on self), but the protagonist has the script name "none" (there was even a comment about that!), so usually the spell would be cast on some other actor with the same name instead.  Really, this extends to any creature with a non-unique script name.

Anyway, now the code uses `GenerateActionDirect()` instead, which takes an actor and directly substitutes its actor ID, and this solves the problem.

### DIFF
--- a/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
+++ b/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
@@ -4660,14 +4660,8 @@ int fx_cast_spell (Scriptable* Owner, Actor* target, Effect* fx)
 	if (fx->Parameter2 == 0 || target->Type == ST_CONTAINER) {
 		// no deplete, no interrupt, caster or provided level
 		// ForceSpell doesn't have a RES variant, so there's more work
-		std::string tmp;
-		if (target == Owner) {
-			// charname has no scriptname ...
-			tmp = fmt::format("ForceSpell(Myself,{})", ResolveSpellNumber(fx->Resource));
-		} else {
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Ah, so you mean if someone else used the effect targeting the protagonist, it wouldn't find it.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
```diff
@@ -4661,13 +4661,13 @@ int fx_cast_spell (Scriptable* Owner, Actor* target, Effect* fx)
...
 		std::string tmp;
+		tmp = fmt::format("ForceSpell({},{})", target == Owner ? "Myself" : "[-1]", ResolveSpellNumber(fx->Resource));
+		Action *forceSpellAction;
 		if (target == Owner) {
-			// charname has no scriptname ...
-			tmp = fmt::format("ForceSpell(Myself,{})", ResolveSpellNumber(fx->Resource));
+			forceSpellAction = GenerateAction(std::move(tmp));
 		} else {
-			tmp = fmt::format("ForceSpell(\"{}\",{})", target->GetScriptName(), ResolveSpellNumber(fx->Resource));
+			forceSpellAction = GenerateActionDirect(std::move(tmp), target);
```
lynxlynxlynx, line 4669: This will work for the Myself case as well, so the ternary and branching can be removed.

Tomsod: Oh, all right.  I assumed there was a reason to special-case it.

lynxlynxlynx:
Thanks!

PR #1721, opened 2022-11-16
Fix actors dropping 1 gold on death instead of their IE_GOLD

Tomsod:
Gold was unfortunately deleted _before_ it could be dropped.  A simple typo, it's only not 0 gold because of `SanitizeItem()`.


lynxlynxlynx:
thanks!

PR #1722, opened 2022-11-16
BG2: Allow ammo for self-firing launchers and fix their bonuses

Tomsod:
This is a bit controversial, but in the original game the few self-firing ranged weapons with launcher bits set (e.g. Gesen shortbow) could do both with or without ammo.  Some view this as bug, but see for instance [this Beamdog thread](https://forums.beamdog.com/discussion/28124/7413-core-feature-request-the-gesen-bow) where they ultimately decided that it doesn't need fixing.  Personally I'd like this feature, so I submit this PR.

Also, this commit incidentally fixes said weapons so that their thac0 and damage bonuses wouldn't double when firing without ammo (they were counted both as launchers and ammo).

### DIFF
--- a/gemrb/core/Inventory.cpp
+++ b/gemrb/core/Inventory.cpp
@@ -930,7 +930,7 @@ bool Inventory::EquipItem(ieDword slot)
 	
 	int armorLevel = itm->AnimationType[0] - '1';
 	int weaponslot;
-	ieDword equip;
+	ieDword equip = IW_NO_EQUIPPED;
 	switch (effect) {
 	case SLOT_EFFECT_FIST:
 		SetEquippedSlot(IW_NO_EQUIPPED, 0);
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Ah, I missed this part when refactoring (adding this caching in Inventory). The last chunk looks fine, but the rest give me pause — do throwing weapons still work fine?

Tomsod:
Ordinary throwing weapons?  Sure, I've just checked the Boomerang Dagger, it has expected range, thac0, and damage range.  The code only affects them if they have launcher bit set (and ammo can be found), otherwise it defaults to the old behavior.
If you want to be super nitpicky, I've noticed that it's sometimes necessary to un- and re-equip an ammo-optional launcher to switch it to the ammoless mode, as just removing all applicable ammo turns it into a fist, I think.  But it's minor enough that I don't care.

lynxlynxlynx:
Ok, thanks!

PR #1723, opened 2022-11-16
Clang v15 fixes

MarcelHB:
## Description

There are two problems when trying to compile GemRB on Clang v15 here:

* An effectively unsued variable `month` in `Calendar.cpp`
* Non-UTF8 russian comments in `ACMReader`

I think we can safely drop both these.

## Checklist


lynxlynxlynx:
Thanks!

PR #1724, opened 2022-11-17
Actor: fix displaying consecutive saving throws

Tomsod:
A simple typo, the code was supposed to block repetitive messages from similar effects, but due to using && instead of ||, it often ended up blocking many subsequent unrelated saving throw reports.

So basically a boolean logic derp.  Without the fix, I would only see ST messages once every few minutes if at all.


lynxlynxlynx:
Nice catch.

PR #1725, opened 2022-11-17
Actor: fix weapon attacks polluting aura

Tomsod:
Prevent actions that don't require aura themselves (e.g. weapon attacks) from setting it on a cooldown.  Allows combining attacks with spells.

So basically, as per the original game, if you can cast a spell then shoot an arrow, you should also be able to shoot an arrow then cast a spell (or attack then drink a potion etc.)  Current master forbids the latter, seems to be unintentional though.


lynxlynxlynx:
Yes, this was unintentional, thanks. :)

Issue #1404, opened 2021-08-03
META: Ensure save game compatibility for all games


lynxlynxlynx:
Usually early-game saves made in gemrb load fine in the originals, while later we often cause crashes. Historically it has been for a number of reasons including not writing everything back, not writing something specific back (like expecting some unused field to be 0), case differences, new variables (pst) and sometimes even the order of the sections (#359 is an example). And sometimes the bad state is only visible ingame, where the actionbar might be foobared, search map corrupted and so on.

Instructions with two approaches are on the site: https://gemrb.org/Dev-docs.html#save-game-compatibility

The first goal is to check what our current status is with saves right after character generation:
- [ ] bg1
- [ ] bg2:soa
- [ ] bg2:tob
- [ ] iwd1
- [ ] iwd1:how
- [ ] iwd2
- [ ] pst

While the real goal is to be able to load any game. A good test case would be loading an end-game save:
- [ ] bg1
- [ ] bg2:soa
- [ ] bg2:tob
- [ ] iwd1
- [ ] iwd1:how
- [ ] iwd2 (as far as we support it)
- [ ] pst

Open new bugs for actual items, this one is to track overall progress.

burner1024:
PST new game save doesn't load.

lynxlynxlynx:
iwd2 will need the new spell focus feat effect moved lower, replacing some noop in the original. Now we crash on load, since the opcode number is too high.

MarcelHB:
> PST new game save doesn't load.

Yeah, this one damn variable leaks into the savegame again.

MarcelHB:
Reprodcibly making original BG2 hang: https://github.com/gemrb/gemrb/issues/2125

lynxlynxlynx:
With our new test setup (from #2440) this became easier. By default the GAM and WMP are compared via resaving late original saves. The GAM drags in CRE, so the coverage is very good. Ideally the SAV would get unpacked as well (just a limitation of external tools), to get ARE, but the main stuff is there. For bg1 only ~APR and HP~ minor spellbook differences remain.

lynxlynxlynx:
Ok, so I went through the resave differences in bg1, bg2, iwd1, iwd2, how, bg2ee and pst. Remaining differences:
- bg1: none
- bg2: current area, some effects fields (like FirstApply — tricky, since we do run the effects at least once), gesen bow equipped header (#2486)
- iwd1: current area [the save is from plain iwd1, tested in iwd mode on a how install]
[...]

PR #1731, opened 2022-11-19
Data stream cleanup

bradallred:
## Description
Some minor enhancements to DataStream for reading data along with cleanup in the importers to favor things like ReadRegion/ReadSize/ReadPoint and removal of needless temporaries

I've tested loading+saving+loading in BG2. Will do some more testing in other games soon.

## Checklist

### DIFF
--- a/gemrb/plugins/AREImporter/AREImporter.cpp
+++ b/gemrb/plugins/AREImporter/AREImporter.cpp
@@ -628,7 +628,7 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 		ieWord Type, VertexCount;
 		ieDword FirstVertex, Cursor, Flags;
 		ieWord TrapDetDiff, TrapRemDiff, Trapped, TrapDetected;
-		ieWord LaunchX, LaunchY;
+		Point launchP;
 		Point pos;
 		Point talkPos;
 		ieVariable Name, Entrance;
@@ -642,7 +642,6 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 		str->ReadVariable(Name);
 		str->ReadWord(Type);
 		Region bbox;
-		ieWord tmp;
 		str->ReadRegion(bbox, true);
 		str->ReadWord(VertexCount);
 		str->ReadDword(FirstVertex);
@@ -658,16 +657,14 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 		str->ReadWord(TrapRemDiff);
 		str->ReadWord(Trapped);
 		str->ReadWord(TrapDetected);
-		str->ReadWord(LaunchX);
-		str->ReadWord(LaunchY);
+		str->ReadPoint(launchP);
 		str->ReadResRef( KeyResRef );
 		str->ReadResRef(script0);
 		/* ARE 9.1: 4B per position after that. */
 		if (16 == map->version) {
 			str->ReadPoint(pos); // OverridePoint in NI
 			if (pos.IsZero()) {
-				str->ReadScalar(pos.x); // AlternatePoint in NI
-				str->ReadScalar(pos.y);
+				str->ReadPoint(pos); // AlternatePoint in NI
 			} else {
 				str->Seek(8, GEM_CURRENT_POS);
 			}
@@ -696,10 +693,9 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 		if (VertexCount <= 1) {
 			// this is exactly the same as bbox.origin
 			if (VertexCount == 1) {
-				str->ReadWord(tmp);
-				assert(tmp == bbox.x);
-				str->ReadWord(tmp);
-				assert(tmp == bbox.y);
+				Point pos;
+				str->ReadPoint(pos);
+				assert(pos == bbox.origin);
 			}
 
 			if (bbox.size.IsInvalid()) {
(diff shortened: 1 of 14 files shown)


gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -1448,9 +1436,8 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 		str->ReadWord(TrapSize);
 		str->ReadWord(ProID);
 		str->ReadDword(Ticks);  //actually, delaycount/repetitioncount
-		str->ReadWord(X);
-		str->ReadWord(Y);
-		str->ReadWord(Z);
+		str->ReadPoint(point);
+		str->Seek(2, GEM_CURRENT_POS); // unknown/unused 'Z'
```
lynxlynxlynx, line 1440: this is likely Projectile::ZPos

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -1698,21 +1696,17 @@ int AREImporter::PutHeader(DataStream *stream, const Map *map) const
...
-	stream->WriteFilling(8); // westref
-	stream->WriteDword(tmpDword);
-	stream->WriteFilling(8); // southref
-	stream->WriteDword(tmpDword);
-	stream->WriteFilling(8); // eastref
-	stream->WriteDword(tmpDword);
+	stream->WriteFilling(12); // northref
+	stream->WriteFilling(12); // westref
+	stream->WriteFilling(12); // southref
+	stream->WriteFilling(12); // eastref
```
lynxlynxlynx, line 1702: We should instead read all these 8 fields and write them back here. They're used in the originals.

gemrb/plugins/CREImporter/CREImporter.cpp
```diff
@@ -710,6 +693,8 @@ void CREImporter::ReadChrHeader(Actor *act)
 		//fallthrough
 	case CREVersion::GemRB:
 		for (int i = 0; i < QSPCount; i++) {
+			// FIXME: why is the data larger than required?
```
lynxlynxlynx, line 696: It just is, we can't do anything about it except change our underlying type and waste a little bit of memory.

gemrb/plugins/CREImporter/CREImporter.cpp
```diff
@@ -1299,85 +1243,50 @@ Effect *CREImporter::GetEffect()
...
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_SETTRAPS]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_LORE]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_LOCKPICKING]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_STEALTH]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_TRAPS]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_PICKPOCKET]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_FATIGUE]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_INTOXICATION]);
+	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_LUCK]);
+	str->ReadScalar<Actor::stat_t, ieByteSigned>(act->BaseStats[IE_TRACKING]);
```
lynxlynxlynx, line 1287: luck should be signed; tracking wasn't, but this is a good change.

bradallred: you mean both tracking and luck should be signed?

lynxlynxlynx: You inverted the previous way they were treated, hence the comment. It doesn't really matter for tracking.

bradallred: I can see that, but you said it is a good change, so I was asking for clarity on what that was to mean.

I'll change them around to match the original code. If we should change signedness of anything we can address that later.

gemrb/plugins/CREImporter/CREImporter.cpp
bradallred: It has nothing to do with the `act->PCStats->QSlots[i+3]` line? Makes it feel like we should be reading 4 separate bytes into these holes.

gemrb/plugins/AREImporter/AREImporter.cpp
bradallred: yeah, thats something to keep in mind for #1404.

lynxlynxlynx: Exactly. We don't need them for travel, since the world map has all the needed info as well, but it would lower the amount of differences in saves and perhaps allows for simplification of our world map adjacency logic.

gemrb/plugins/CREImporter/CREImporter.cpp
lynxlynxlynx: If it wasn't for the fallthrough, this would only apply to our own unused format and we could just change it. But iwd2 runs this code as well.

gemrb/plugins/AREImporter/AREImporter.cpp
bradallred: yeah, I had the same inkling, but the field is private, and we seem to restrain it to these preset values:
https://github.com/gemrb/gemrb/blob/138886d702265939d96b5f6b4ba5a5c8215d5915/gemrb/core/Projectile.h#L47-L52

so its something out of scope to fix here IMO

gemrb/plugins/CREImporter/CREImporter.cpp
```diff
@@ -710,6 +693,9 @@ void CREImporter::ReadChrHeader(Actor *act)
 		//fallthrough
 	case CREVersion::GemRB:
 		for (int i = 0; i < QSPCount; i++) {
+			// FIXME: we are reading 4 bytes then suspiciously discarding 3 of them
+			// I wonder if those the bytes are supposed to fill in the 3 byte holes we are leaving in QSlots
```
lynxlynxlynx, line 697: Like I said, there's nothing suspicious here, it's just the way the data is stored. The QSlot gap (index offset) is for the hardcoded set of slots.

PR #1733, opened 2022-11-23
BG2: Make haste APR bonus work properly by moving it from FXOpcodes to Actor

Tomsod:
All forms of haste computed attacks per round way incorrectly, especially combined with other APR modifiers.

> It used to be set in the effect queue, but that doesn't really work as, for one, the fixed APR of most launchers would override it (items come after spells following a reload).  Also the effect sets all those states which are meant to be used elsewhere, so now there's a block late in actor stat refresh code that manually adjusts APR based on those states.

Belm and Kundane didn't work properly with improved haste for the same reason.
It's also after wspatck handler code now, which helps, although because it sets base it would only be wrong by one tick anyway.  (Incidentally, setting base also explains how that code respects the limit of 5 APR despite not calling `ClampStat()`.)
Also, isn't `IE_ATTACKNUMBERDOUBLE` misnamed?  It's for weak haste, which does not double anything.

> Further, type 0 and type 2 haste are supposed to round the APR to integers, which they didn't before.  And, by being applied manually and last, haste can now exceed the limit of 5 APR, which it should.

[It's all in IESDP.](https://gibberlings3.github.io/iesdp/opcodes/bgee.htm#op16)  Normal haste rounds down, weak haste rounds up.  Normal and improved haste can break the limit, weak can't because of math.  I suppose if anything else messed with the APR stat later that tick it could get clamped again, but evidently nothing does.  I have got code that enforces a variable limit through PCF, but I ended up not needing it.

### DIFF
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -2780,6 +2780,24 @@ void Actor::RefreshEffects(bool first, const stats_t& previous)
 		if (!(BaseStats[IE_STATE_ID] & STATE_DEAD)) pcf_hitpoint(this, 0, BaseStats[IE_HITPOINTS]);
 	}
 
+	// apply haste bonuses to attacks per round, possibly exceeding the limit of 5
+	// this must be done after ALL other modifiers, including skill effects in RefreshPCStats() above
+	if (GetStat(IE_STATE_ID) & STATE_HASTED) {
+		int apr = GetStat(IE_NUMBEROFATTACKS);
+		switch ((signed) GetStat(IE_IMPROVEDHASTE)) {
+		case 1: // improved
+			apr += apr; // x2
+			break;
+		case 0: // normal
+			apr += 2 - apr % 2; // +1, round down
+			break;
+		case -1: // weak
+			if (apr % 2) apr++; // round up
+			break;
+		}
+		Modified[IE_NUMBEROFATTACKS] = apr;
+	}
+
 	for (int i=0; i < MAX_STATS; ++i) {
 		if (first || Modified[i]!=previous[i]) {
 			PostChangeFunctionType f = post_change_functions[i];
@@ -2986,7 +3004,6 @@ void Actor::RefreshPCStats() {
 			// the adjustment results in a base of 2-5 (2+[0-3]) and the modified stat degrades to 4+(4-[2-5]) = 8-[2-5] = 3-6
 			// instead of 4+[0-3] = 4-7
 			// For a master ranger at level 14, the difference ends up as 2 (1 apr).
-			// FIXME: but this isn't universally true or improved haste couldn't double the total apr! For the above case, we're half apr off.
 			ieDword warriorLevel = GetWarriorLevel();
 			if (warriorLevel) {
 				int mod = Modified[IE_NUMBEROFATTACKS] - BaseStats[IE_NUMBEROFATTACKS];
--- a/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
+++ b/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
@@ -1575,6 +1575,10 @@ int fx_set_hasted_state (Scriptable* /*Owner*/, Actor* target, Effect* fx)
 	// print("fx_set_hasted_state(%2d): Type: %d", fx->Opcode, fx->Parameter2);
 	target->fxqueue.RemoveAllEffects(fx_set_slow_state_ref);
 	target->fxqueue.RemoveAllEffectsWithParam( fx_display_portrait_icon_ref, PI_SLOWED );
+	int old_type = -2, new_type = -2; // lowest priority by default
+	if (target->GetStat(IE_STATE_ID) & STATE_HASTED) {
+		old_type = (signed) target->GetStat(IE_IMPROVEDHASTE); // we cramp all three types into this stat now
+	}
 	if (fx->TimingMode==FX_DURATION_INSTANT_PERMANENT) {
 		BASE_STATE_CURE( STATE_SLOWED );
 		BASE_STATE_SET( STATE_HASTED );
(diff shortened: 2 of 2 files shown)


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -2780,6 +2780,20 @@ void Actor::RefreshEffects(bool first, const stats_t& previous)
...
+	if (GetStat(IE_STATE_ID) & STATE_HASTED) {
+		int apr = GetStat(IE_NUMBEROFATTACKS);
+		if (GetStat(IE_ATTACKNUMBERDOUBLE)) { // weak haste
+			if (apr % 2) apr++; // round up
+		} else if (GetStat(IE_IMPROVEDHASTE)) {
+			apr += apr; // x2
+		} else { // normal haste
+			apr += 2 - apr % 2; // +1, round down
+		}
+		Modified[IE_NUMBEROFATTACKS] = apr; // do NOT call SetStat or it will be clamped
```
lynxlynxlynx, line 2794: Working on Modified here is fine, but the comment shouldn't be needed. Just increase the limit in the maximum_* array.

lynxlynxlynx:
Are ranged weapons supposed to be able to be boosted? I guess it's the difference between the original "fixed" mode and tobex's final mode.

Yes, IE_ATTACKNUMBERDOUBLE is misnamed, not misused, but it's a waste of a stat either way. Slight speedup to keep it, but that's all.

And wow, boots of speed indeed affect APR! A tob-starting fighter is the easiest way to check.

If on a reload item effects are queued later, this means other things will now be broken. Boots of speed would disable the APR doubling of improved haste.

gemrb/core/Scriptable/Actor.cpp
Tomsod: Then _anything_ will be able to exceed 5 APR, which is not the original game behavior.  E.g. level 13 fighter with two stars dual-wielding Belm and Kundane would have 1 + 0.5 + 2 + 1 + 1 = 5.5 APR, but the original game clamps them to 5.  But haste, and haste alone, can exceed that to 6 or 10.  I've commented above that wspatck bonuses are properly clamped by gemrb, which is correct behavior.  But haste is, and must be, an exception.

lynxlynxlynx: Ok, then I suggest you just remove the comment.

Tomsod:
There's some inconsistency whether ranged APR is supposed to be affected by wspatck (I've read it isn't in EE? I don't have EE), but here's a proof fixed APR weapons are affected by haste: Energy Blades (`ENEBLADE.ITM`).  It has APR set to 4.5 plus improved haste for a total of 9.  It's also a dart.

That stat is not a waste of stat with my PR, as it governs the haste APR bonus.  I highly suspect it does so too in the original game.

Item effects were shifted on reload for as long as I remember, is that not intended?  It's annoying sometimes (e.g. DUHM and giant belts or dex gauntlets), but not critical.

Boots are problematic, yes.  Perhaps don't have them unset `IE_IMPROVEDHASTE` and check that first?  But what of normal haste then?

gemrb/core/Scriptable/Actor.cpp
Tomsod: I can do that, if you don't think it's helpful.  No big deal.

lynxlynxlynx: To avoid confusion, yeah.

lynxlynxlynx:
I don't see anything special in CREImporter that would break the ordering on load/save. However, I can easily imagine that only persists for a little bit, until we apply equipping effects when adding the inventory. Minor nuisance, I wouldn't touch it.

I suggest you convert IE_IMPROVEDHASTE to a bitfield. That way states can be anded together by effects, separately checked for, and IE_ATTACKNUMBERDOUBLE can be repurposed (very valuable property).

Tomsod:
While the idea is very tempting, doesn't it have the potential to break mods/scripts that use `CheckStat()` on these?

lynxlynxlynx:
I doubt any do, especially the second one, but I've asked the community to weight in. The base check people are doing is definitely with states, but let's see, maybe SCS cares.

Tomsod:
I'll be waiting then.  The backup solution is to give weak haste the lowest priority -- only have it set `IE_ATTACKNUMBERDOUBLE` if `STATE_HASTE` was not set before, otherwise nothing.  Other types of haste would reset it then.  And if `IE_IMPROVEDHASTE` will only be touched by type 1 haste, it won't be overridden.  This would preserve the values of both stats.

lynxlynxlynx:
No response, but I've checked SCS. It's using the improved haste stat only by checking if it's not zero:
https://github.com/Gibberlings3/SwordCoastStratagems/search?q=IMPROVEDHASTE
The other one is not referenced, so we're safe with simplifying.

Tomsod:
But if I make `IE_IMPROVEDHASTE` into a bitfield responsible for both type 0 (normal) and type 1 (improved) haste, then those scripts will think normal haste is improved, which is slightly wrong.  Are you sure you want that free stat that much?

A safer solution would be to flip `IE_ATTACKNUMBERDOUBLE`, so that normal haste (stat set to 1) would have priority over weak (default stat value 0), and improved would alone set `IE_IMPROVEDHASTE` that would have top priority.  No new free stat that way, but scripts that check for improved haste will work 100% correctly.  And I'm ready to believe no one ever checks for weak haste, so it should be safe for the other stat, too.

Alternatively, for SCS compatibility in particular, it checks if the stat is _greater_ than zero.  Then if improved is 1, normal is 0, and weak is -1, the check would still work correctly and we can cramp it all in one stat.

lynxlynxlynx:
Yeah, but if you set to a fixed amount, effects will still override each other. What I had in mind was that normal haste doesn't touch it, improved BORs (let's say) 1 and weak BORs 2. Normal mode can be detected by checking if the stat is 0 and STATE_HASTED is on, so the same `else` as before.

I just double-checked and the effect bonuses are indeed not cumulative. Casting improved haste doubled the rate without the boots. So you're also right that improved haste should be checked for first.

Tomsod:
I got the idea, but then scripts will think weak haste is improved, which is not perfect.  Probably tolerable, as the former is rare, but what I meant for the (1, 0, -1) variant is roughly this pseudocode:
```c
old = -2;
if (state & STATE_HASTED)
    old = stat(IMPROVEDHASTE);
state |= STATE_HASTED;
switch (haste_type) {
    case 0:
    new = 0;
    break;
    case 1:
    new = 1;
    break;
    case 2:
    new = -1;
    break;
}
if (new > old)
  stat(IMPROVEDHASTE) = new;
```
That way they will override each other in the proper order 1 > 0 > 2.

lynxlynxlynx:
You're right, better limit the special logic to one place, no need to pollute the trigger code as well.

Tomsod:
I'm doing that last thing I posted then?

lynxlynxlynx:
Yes please.

Tomsod:
Here, the priorities seem to work fine now.

By the way, it's now fairly easy to make the -2 weapon speed bonus not stackable.  Is it supposed to be?

lynxlynxlynx:
Thanks! And yeah, pretty sure the speed bonus should be cumulative.

Issue #326, opened 2019-10-11
[bug] BG2: Issues with Con bonuses


caillean7:
Split from #171

**Description**

- [ ] Con bonus HP does not immediately update after a level up, often leaving the PC wounded

- [ ] (supertricky) related: Con bonus should (?) raise HP as well as max HP

PR #1741, opened 2022-11-26
BG: modify current HP following Con changes

Tomsod:
> In the original BG1&2, temporary increases in constitution also increased current HP equal to max HP bonus.  In EE, these bonuses also cause equal amount of damage on expiring.  This commit implements the EE behavior, projecting any max HP change from constitution to current HP.

ATM GemRB does not adjust current HP at all, which makes e.g. DUHM significantly worse defensively.

Personally I feel the EE way is more fair (especially if you want to add early Con equipment), that said, the original game behavior is literally one comparison operator change away, if you'd like that more.  Or it could be controlled by those EE flags whenever you implement them.

### DIFF
--- a/gemrb/GUIScripts/LUCommon.py
+++ b/gemrb/GUIScripts/LUCommon.py
@@ -458,8 +458,6 @@ def SetupHP (pc, Level=None, LevelDiff=None):
 
 	#update our hp values
 	GemRB.SetPlayerStat (pc, IE_MAXHITPOINTS, CurrentHP+OldHP)
-	# HACK: account also for the new constitution bonus for the current hitpoints
-	GemRB.SetPlayerStat (pc, IE_HITPOINTS, GemRB.GetPlayerStat (pc, IE_HITPOINTS, 1)+CurrentHP+5)
 	return
 
 def ApplyFeats(MyChar):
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
This isn't penalizing actors twice — directly and then indirectly from the decreased max hp?

Tomsod:
No, max HP decrease by itself would only cut current HP down to the new maximum, but the former already will be at or below the latter because the new code decreased it correspondingly.

Oh yeah, I forgot I complained about this already.  Yes, this PR fixes #326.  Just checked, PCs do not become wounded after a level-up anymore.

In fact, there's apparently some unrelated code that even heals a few HP on a level-up, which might have been another attempt to fix this.  I don't know where it is, though.

lynxlynxlynx:
It's at gemrb/GUIScripts/LUCommon.py:462 and can hopefully be removed now.

Tomsod:
Oh, that's where five healed hit points come from.  Kind of an odd number!  Yes, that code should be removed.  Do I add this to the PR?

lynxlynxlynx:
Please, I won't be near a computer for at least 5 hours more.

lynxlynxlynx:
Thanks!

PR #1744, opened 2022-11-29
SPLImporter: handle short icon names

### LINKED ISSUE #1743: BG2 (modded): spellbook can't handle 7-character spell icon names
So, aTweaks has a new Gate icon as part of "PnP Fiends" component, I believe.  It's installed as `fl#905c.bam` for the in-spellbook version.  But this code mishandles the filename as it expects it to be 8 characters long:
https://github.com/gemrb/gemrb/blob/0fe43af76537e80dfc137315e4efa8fbde84a7e1/gemrb/plugins/SPLImporter/SPLImporter.cpp#L124-L128
And the engine ends up with `fl#905cc.bam` which doesn't actually exist.  Presumably it worked for the original game's engine, otherwise the mod's authors would have noticed, so just adding/replacing the 8th character is evidently wrong.  What was this code supposed to do, again?  Replacing the last character (7th or 8th or whatever) works for me, but this particular spell name already ends with 'c', so maybe it should just check for that, instead.

Tomsod:
> aTweaks adds a seven-character-long-filename spellbook icon, and the old code mangled it by assuming all filenames are 8 characters long.

Fixes #1743.  Thanks to @bradallred for the neater code.


lynxlynxlynx:
Thanks!

PR #1745, opened 2022-11-29
Enable static-linking on GCC/Clang for Windows

MarcelHB:
So far, Windows was excluded for static linking. Concerning GCC & Clang, there is no reason for that (any more) while MSVC remains untested and thus disabled.

Why would one need that? `gprof` won't work with dynamically loaded libraries. Also, AMD µProf is virtually broken for non-MSVC on Windows, so this is a fallback.

There's a little whitespace/tab inconsistency cleanup in accordance to editorconfig settings.

## Checklist


lynxlynxlynx:
Looks fine, thanks!

Issue #1692, opened 2022-09-07
Going to open a door with the full party make it drop fps considerably 


rsantellan:
#### Bug description
On the bridge district (Where I notice) trying to open a door brings down the fps. It's only when the full party is selected.
I believe this is related to #1441
#### Steps to reproduce
1. Go to Bridge District
2. Click on a door with the full party
3. Move to the other side and click again

#### Expected behavior
It not should drop the fps like that

#### Screenshots

https://user-images.githubusercontent.com/1820598/188778821-070c1bb5-49ea-4d2c-a260-2922323938a3.mp4

GemRB version (check as many as you know apply)
- [X] master as of this issue
- [x] 0.9.1
- [x] 0.9.0
- [x] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [X] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Probably _just_ the general performance problem / trade-off, unless you can demonstrate this worked better in 0.8.8 or some later version.

We'll probably need to add new threads to speed the search (eg. the direction loop in FindPath looks parallelizable), but that wouldn't help on all hardware. And since pathing is critical, we can't just wait until it's done, doing other stuff like processing mouse movement in the meanwhile.

rsantellan:
@lynxlynxlynx I tried on all the tags beginning on the v0.8.8 and in all of them happen the same. Depending where you are the impact is more notable

bradallred:
the profiler shows we are spending quite a bit of time in `Actor::NewPath` and half of that time is spent in `Map::NormalizeDeltas`. Not sure how to improve that.

bradallred:
Things still hitch, but those optimizations gave me ~10% improvement (hitches are 10% shorter). We could squeeze another ~14% if we could eliminate ConvertCoordToTile from the pathfinding by having a flavor of `GetBlocked` that takes a tile coord. we'd probably have to cache the tile coord for each actor pos too.

Any better fix is going to have to actually change some logic (or offloading to another thread as mentioned)

bradallred:
I'm not sure it is related, but this looks like a bug to me:
https://github.com/gemrb/gemrb/blob/ab71947f35bdea7e4c862bace1b9779ac6847a58/gemrb/core/Map.cpp#L2575-L2594

[...]

PR #1749, opened 2022-12-01
Faster vector normalization in Map code

MarcelHB:
I was investigating why there is an extreme slowdown in some cases of changing areas that do not look particularly difficult at a first glance.

Anway, the profilers indicate another interesting thing: When spending a lot of time waiting for map changes, `Map::NormalizeDeltas` has an insane amount of calls, and `sqrt` even consumes some systime.

Looking at this, there one obvious thing: One can do that with only one instead of two `sqrt` operations. And doing a bit of micro-benchmarking fun, it turns out that using something called [Quake Inverse Fast Sqrt](https://en.wikipedia.org/wiki/Fast_inverse_square_root) is even faster, if you can live with tiny little errors that don't really matter here (doing `ceil` and such). What was even mind-blowing: This optimization is not just faster by 20% in `-O2` but almost 2000% (Clang)/3000% (gcc) faster under `-O3`.

deltas
(that's the whole `NormalizeDeltas` body code with old / new code)

There is still some noticeable I/O concern, I guess, and we should maybe look for ways to largely reduce the # of calls that end up in `NormalizeDeltas` in this situation but changing areas has really improved a little in Release mode, like relative 25-30% on that function counter.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,11 @@ IF(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
 	if (NOT APPLE AND NOT UNSAFE_PLUGIN)
 		string(APPEND CMAKE_MODULE_LINKER_FLAGS " -Wl,--no-undefined")
 	endif (NOT APPLE AND NOT UNSAFE_PLUGIN)
+
+	# Fast math helps us covering some things that need a little more rework soon
+	string(APPEND CMAKE_CXX_FLAGS " -ffast-math")
+ELSEIF(MSVC)
+	string(APPEND CMAKE_CXX_FLAGS " /fp:fast")
 ENDIF()
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/")
--- a/gemrb/core/Map.h
+++ b/gemrb/core/Map.h
@@ -457,13 +457,13 @@ private:
 
 	MainAmbients dayAmbients;
 	MainAmbients nightAmbients;
-	
+
 	friend class AREImporter;
 
 public:
 	Map(TileMap *tm, TileProps tileProps, Holder<Sprite2D> sm);
 	~Map(void) override;
-	static void NormalizeDeltas(double &dx, double &dy, const double &factor = 1);
+	static void NormalizeDeltas(double &dx, double &dy, double factor = 1);
 	static Point ConvertCoordToTile(const Point&);
 	static Point ConvertCoordFromTile(const Point&);
 
--- a/gemrb/core/PathFinder.cpp
+++ b/gemrb/core/PathFinder.cpp
@@ -453,10 +453,9 @@ PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, u
 	return nullptr;
 }
 
-void Map::NormalizeDeltas(double &dx, double &dy, const double &factor)
+void Map::NormalizeDeltas(double &dx, double &dy, double factor)
 {
 	constexpr double STEP_RADIUS = 2.0;
-	constexpr double STEP_RADIUS_SQ = STEP_RADIUS * STEP_RADIUS;
 
 	double ySign = std::copysign(1.0, dy);
 	double xSign = std::copysign(1.0, dx);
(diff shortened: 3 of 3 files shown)


MarcelHB:
As a disappointing observation afterwards, gcc uses something else on `std::sqrt` but only with `-ffast-math`. For x86, you'll get the even better/faster `RSQRTSS` instruction only with `-O3 -funsafe-math-optimizations -mrecip -ffinite-math-only -fno-trapping-math`.

So this solution looks as if it was even a saner performance choice than dealing with every platform and compiler individually.

bradallred:
yeah, we noted `Map::NormalizeDeltas` in #1692 and made some improvements in 92fdee2. I actually meant to also commit a change to do something like enable `-funsafe-math-optimizations`, or `-fno-math-errno` which would actually eliminate the `sqrt` call entirely in an optimized build since modern CPUs have an instruction for it. It would be interesting to benchmark an O3 build with those options vs this.

I have no doubt that this is much much faster than `std::sqrt`, however, the real issue with path finding is that it is recalculating entire paths very frequently, when all it really needs to do is calculate a much smaller path to splice into an existing path. This happens mainly when PCs bump into each other, which is why you see it when changing areas a lot.

Anyhow, i'm not _opposed_ to having a `sqrt` approximation, but im skeptical it resolves #1692, tho I'd be happy to be wrong  :)

If we are going to include this id rather just put it in geometry.h and access its use for other places we use `sqrt`.

bradallred:
> As a disappointing observation, gcc uses something else on `std::sqrt` but only with `-ffast-math`. For x86, you'll get the even better/faster `RSQRTSS` instruction only with `-O3 -funsafe-math-optimizations -mrecip -ffinite-math-only -fno-trapping-math`.
> 
> So this solution looks as if it was even a saner performance choice than dealing with every platform and compiler individually.

So you apparently replied while I was typing my response, but `-fno-math-errno` should be all that is required for the optimizer to use the CPU instruction for sqrt. The reason it cannot do so without such is because `sqrt` can set `errno`, but we never check this so we dont care. the only reason I never added the option is because I was going to figure out how to get MSVC to also optimize it away and it just fell out of my mind.

lynxlynxlynx:
A quick look at the log suggests bumping is not much of an issue on area move. Maybe if you're far, but there's a slowdown even if you're close to the map edge. We should probably redo the edge grace period/size thing and avoid much of the issue altogether.

MarcelHB:
Ok, so I suggest the following:

- We don't use a lot of FP math, and I don't see any pressure on the profiler except for this function, so we don't generally need aggressive optimization and it's management.
- We can enable `-ffast-math` on GNU/Clang though since this flag is universally usable there.
- For MSVC, Godbolt says that `/fp:fast` does the same kind of job but even more aggressively.
- I remove that approx. sqrt but modify the function to still use only one `std::sqrt`, since looking at two of them is confusing when expecting vector normalization.
- I still haven't looked at the door thing and whether they have common ancestors, but the overall issue -- to reduce the number of Map math in general -- is handled in a 2nd step.

bradallred:
yeah, maybe I'm unfairly assuming bumping is the issue; I really haven't dug into this past profiling and testing. However, since you can get bogged down by creating a narrow corridor and sending groups though in opposing directions it seemed like a likely suspect.

One thing that should quickly help (if it's not already being done) is to keep the formation intact (not necessarily the literal formation) when the destination is an object/travel. It seems like all the actors are getting the object/travel location as their destination causing issues when they get near. If we created a special formation(s) that grouped around a point, or for the map edge, created a line along the edge it seems like that would help a lot.

It might be helpful for anyone unaware looking to test these issues, you can edit the search map in-game by activating the search map overlay and sampling a tile with right click and painting with left. You can quickly setup any sort of test you want.

> I still haven't looked at the door thing and whether they have common ancestors

they are related which is why I mention it (And you should probably reference it in the commit)

lynxlynxlynx:
Doors and I guess attacking are the only actions where we move the whole party. For containers, moving and talking, we use the first selected pc. I guess because there are potential skill checks involved. For doors, we do try preserving the formation. So at best we could be smarter about formation placement, prerotating them better in narrow spaces. Not sure if that would be universally faster, since there are some unfortunate formation and searchmap combos.

We do that for border travel as well, but there are better solutions available for that.

MarcelHB:
Interesting, #1692 is probably what I was looking at all the time. Except I can reproduce the hanging in my case/save game only if the party actually changes the map, not by just clicking the door from anywhere at a distance. Anyway, if that's related to formation, or some kind of narrow corridor path then it's probably different geometry.

MarcelHB:
So for now this is the low-hanging patchwork we discussed.

lynxlynxlynx:
What I was talking about for the edges is taking another look at Map.cpp's LastGoCloser / CT_GO_CLOSER. In the end Game::EveryoneNearPoint gets called with a single point, so everyone's position is compared to that, not to their closest viable point (projection to the edge).

Issue #1759, opened 2022-12-05
Formation issues


burner1024:
#### Bug description
1. Formations are not geometrically perfect in the first place.
2. They get distorted further when rotating.
3. Presumed destination get "pre-bumped" by actors and in particular, the same selected actors.

#### Steps to reproduce
1. Select a few actors, try various formations and rotate.

#### Expected behavior
- Formations actually matching their geometry
- No distortion on rotation.
- Selected characters should try to bump busy spots. Or, at least ignore positions of currently selected characters when calculating destinations.

#### Screenshots
This is not a proper hexagon.
![Captura desde 2022-12-06 02-23-05](https://user-images.githubusercontent.com/1279852/205724943-d5a1fb36-2d40-4326-8272-984ecad06efa.png)

These T's are gravitating towards south (other formations also get distorted similarly).
![Captura desde 2022-12-06 02-24-50](https://user-images.githubusercontent.com/1279852/205725367-e38d50ae-2240-46fa-a094-a4d8fa88f7ef.png) ![Captura desde 2022-12-06 02-24-57](https://user-images.githubusercontent.com/1279852/205725372-38626d65-16d9-4175-8760-7d11805147ec.png)

T turned into mush by currently selected characters
![Captura desde 2022-12-06 02-28-14](https://user-images.githubusercontent.com/1279852/205725758-88e1b3d7-769f-4370-9ab3-e0500919261b.png)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
1. Likely just erroneously taking the isometric perspective into account.
2. Probably the same thing.
3. What do you mean? We do take the search map into account, yes.

burner1024:
3. I mean I expect to see a T.

bradallred:
re 3 we purposely deviate from the original to show where the actor will be directed to. Maybe this could be enhanced to ignore actors tho.

burner1024:
Yes, ideally this would ignore all bumpable actors. Destinations jump all round when rotating in a crowded area, screenshots don't show that very well, but I'm struggling with recording a video unfortunately.

[...]

PR #1761, opened 2022-12-07
Fix formations generator

burner1024:
## Description
make_formations.py is outdated and doesn't match the corresponding 2da's, see #1759.

Hope you don't mind conversion to python 3.
Files are generated with 20 positions.
Files generated with updated script match current 2das exactly, with exception of GATHER formation. I couldn't figure out the algo used to calculate the current one, if there even was any. But first 6 positions match exactly, and the rest shouldn't matter according to [this](https://github.com/gemrb/gemrb/issues/1759#issuecomment-1338761973).

This is the new gather, for reference.
6-new

## Checklist


admin/make_formation.py
```diff
@@ -314,6 +315,15 @@ def generate_wedge():
...
 	print()
 
+def generate_none():
+	print("NONE", end=' ')
+	end = " "
+	for i in range(num_coords):
+		if i == (num_coords - 1):
+			end = ""
+		print("0 0", end=end)
+	print()
```
lynxlynxlynx, line 325: Why not just use something like `print("NONE ", "0 0 " * (num_coords - 1),  "0 0\n")`

admin/make_formation.py
```diff
@@ -77,34 +77,31 @@ def generate_t():
...
-				yloc = 24 * i // 3
+		if i == 0:
+			xloc = 0 # centre
+			yloc = -36 # front
+		elif i == 1:
+			xloc = 48 # right
+			yloc = -24
+		elif i == 2:
+			xloc = -48 # left
+			yloc = -24
```
lynxlynxlynx, line 95: this block is redundant

lynxlynxlynx:
Thanks! And yes, a python3 port is welcome. 
It all started with you complaining about formation shape, but these descriptions and code all look like they preserve the existing state. Is the shape fix coming later, did I misunderstand something?

admin/make_formation.py
burner1024: That doesn't generate the same line, but I fixed it.

admin/make_formation.py
burner1024: All 1-3 are at negative y, and all the rest are at positive.

burner1024:
Yes, first I wanted to get the exact state, then make changes.
I think formations are mostly good, except the pentacle one.

lynxlynxlynx:
The same files are generated here as well. Please also fix generate_formation_files.sh though, it should be using unhardcoded instead of override. And ideally it would inspect $0 to find out where it is and derive the path prefix from there.

burner1024:
Fixed shell script, and improved pentacle.
Comparison:
5-2-old
5-new

bradallred:
you should also commit the generated formato.2da files

burner1024:
Converted to draft for now, will revisit the shapes once there's a working projection fix.

burner1024:
Ok, I made Protect and Gather match bg2 more closely, screenshots attached.

Do you want me to review other formations?

1o
1g
2o
2g

lynxlynxlynx:
Thanks and please do.

burner1024:
OK, I adjusted the rest, mostly the difference was gemrb variant placing characters closer together than originals.
<details> 
  <summary>Screenshots, originals and updated GemRB</summary>

wedge_o
wedge_g
tria_o
tria_g
t_o
t_g
s_o
s_g
rank_o
rank_g
line_o
line_g
3b2_o
3b2_g
2b3_o
2b3_g

</details>

lynxlynxlynx:
Thanks!

PR #1766, opened 2022-12-17
Fix typo in readme sports => supports

Binnette:
Just a typo in readme


lynxlynxlynx:
Thanks, but that's not a fix, it's just the use of the transitive form of "to sport".

Issue #1770, opened 2023-01-07
Demo & Near Infinity not working well


MarcelHB:
#### Bug description

When I'm opening the demo's `chitin.key` in the newest [Near Infinity](https://github.com/Argent77/NearInfinity/releases/tag/v2.3-20221108), I select `Unknown Game`. Trying something else for example  loops/hangs.

I see two issues:

1. Not all resources are listed, for example the new BAM + PVRZ visuals are missing.
2. Of the other resources, most of them don't work and are rejected with some inconclusive error dialog (highlighted red): TIS, WED, ARE und also MUS.

So things to have a look at:

* Are the resources not complying to the standards?
* Can this be fixed somehow by us? With the help of tools? (The error messages in stdout log may be helpful)
* Are these bugs/misinterpretations at NI, so that one can maybe file bug reports?

#### Steps to reproduce
1. Open demo in NearInfinity
2. Try to find and open the demo's resources.

#### Screenshots
![2023-01-07_18-25](https://user-images.githubusercontent.com/238558/211163325-70411104-4415-4faa-a193-05332b84d3b5.png)
![Screenshot_2023-01-07_18-35-18](https://user-images.githubusercontent.com/238558/211163336-4c11b0a2-54ef-43d7-8ff8-045cf819f6ba.png)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

lynxlynxlynx:
Best game mode is bg2:tob.

Not all resources are listed, since chitin.key is not kept up to date. Had to do it manually, which is too much of a chore — we need to script it. I think this is the only bit we can handle on our end,

Recent versions of NI are worse for this, for me.

ARE doesn't load due to the song header iirc, I've opened a bug about it.

MarcelHB:
Good to know, thanks, I was also looking for a script somewhere, but then this sounds like a good idea -- I just helped myself by throwing the stuff into some other game's `override/`.

lynxlynxlynx:
Locally I symlink data to override. Having to rebiff after every change would be quite a chore, so let's try to avoid it. If that means moving everything to override, let's do it.

PR #1773, opened 2023-01-09
Demo & Near Infinity

MarcelHB:
## Description
See #1770. Python script to rebuild `chitin.key` and moved `demo/data` to `demo/override` for plug and play (with NI). And a littly by-fix for not making `KEYImporter` crash when there is no BIFF.

## Checklist


lynxlynxlynx:
Thanks, LGTM!

Issue #1760, opened 2022-12-06
Pretty fog of war


burner1024:
#### Bug description
Original BG2 makes fog soft when 3D acceleration is enabled. Without 3D, it's ugly and jagged, just like current GemRB one.
Not a big issue per se, but it does stand out for someone used to the original.

#### Steps to reproduce
Just look at fog.

#### Expected behavior
Pretty fog.

#### Screenshots
![Captura desde 2022-12-06 07-14-10](https://user-images.githubusercontent.com/1279852/205775092-38aeb8c2-bfb9-4f9c-8e22-f29474d3f935.png)
![Captura desde 2022-12-06 07-05-09](https://user-images.githubusercontent.com/1279852/205775174-80c6fbe0-28a6-4d65-8293-0534be8da530.png)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [x] SDL2 without `USE_OPENGL_BACKEND`

bradallred:
are there just a different set of sprites? or is it applying an effect?

bradallred:
yeah, but that could also just be a 32bit sprite with a proper alpha channel. The "ugly" sprites are just color keyed so have either complete opacity, or transparency with no in between.

bradallred:
<img width="245" alt="image" src="https://user-images.githubusercontent.com/1126744/206484767-0c6ebeb7-2b3d-411e-9a64-763731c28f87.png">

It looks like it's probably 32bit sprites judging from the settings options. Just gotta locate them and choose them. I wonder what other sprites are supposed to be 32bit.

MarcelHB:
I can see smooth fog with 3D Acceleration enabled, regardless of that texture setting. I personally never did because these games always had other glitches for me then.

Not sure if any of my Win tools support APIs that old but I can later try to find out what BG2 was trying to do in this case.

bradallred:
If there isn't a sprite, then I would suppose a simple Gaussian blur. Looking at the glsl from the EE editions shows they indeed are using a blur for something.

MarcelHB:
It's actually quite simple: For a full cover, a quad primitive with some dark greay and alpha = 0.5 is blended over the background:

[...]

PR #1776, opened 2023-01-13
Smooth fog edges

MarcelHB:
## Description
See #1760. Turns out interpolating some vertex colors alone isn't sufficient, we will need to paint cells in a way to cover the edges to look good. Here we are now:

With modern SDLv2 (notice the machine in the lower left, and the room in the upper right):

Fallback:

## Checklist

### DIFF
--- a/gemrb/core/Map.h
+++ b/gemrb/core/Map.h
@@ -25,6 +25,7 @@
 #include "globals.h"
 
 #include "Bitmap.h"
+#include "FogRenderer.h"
 #include "Interface.h"
 #include "MapReverb.h"
 #include "Scriptable/Scriptable.h"
@@ -484,7 +485,7 @@ public:
 	/* transfers all ever visible piles (loose items) to the specified position */
 	void MoveVisibleGroundPiles(const Point &Pos);
 
-	void DrawMap(const Region& viewport, uint32_t debugFlags);
+	void DrawMap(const Region& viewport, FogRenderer& fogRenderer, uint32_t debugFlags);
 	void PlayAreaSong(int SongType, bool restart = true, bool hard = false) const;
 	void AddAnimation(AreaAnimation anim);
 	aniIterator GetFirstAnimation() { return animations.begin(); }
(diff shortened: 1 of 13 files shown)


lynxlynxlynx:
Looks nice, better than the original! Makes it harder to cheese ranged stuff as well.
We can add a config check for that ini key or wait and see if anyone complains.
There's so much code now that it feels right to create a new class for fog stuff instead of burdening the too big Map further.

MarcelHB:
Ok, I'll think of something.

bradallred:
yeah, it'll contain those globals too.

One thing I see: can't we add an enum value for 0 to avoid all the casting?

MarcelHB:
> One thing I see: can't we add an enum value for 0 to avoid all the casting?

Yeah, no objection.

MarcelHB:
Feel free to look again.

gemrb/core/Map.h
```diff
@@ -458,6 +460,24 @@ class GEM_EXPORT Map : public Scriptable {
...
+		SE = S|E // 12
+	};
+
+	// Size of Fog-Of-War shadow tile (and bitmap)
+	static constexpr int CELL_SIZE = 32;
+
+	// Buffers of const allocation size
+	static std::vector<float> fogVertices;
+	static std::vector<Color> fogColors;
+
```
lynxlynxlynx, line 480: These were all moved to the renderer, so can be removed here and in the cpp (the two vectors).

MarcelHB: Oh yeah, I missed cleaning up that section with the lines above as well.

bradallred:
I'm curious if this works if you were to force the software renderer on the non OpenGL flavor.

MarcelHB:
Suprisingly, it suffers from other problems in SW rendering:
software_shadows

bradallred:
That’s ok. The new SDL2 based one won’t. It will use the masking code from our SDL1 driver. Looks like the fog works as long as we can tap into SDL_RenderGeometey. On Jan 14, 2023, at 12:08, Marcel ***@***.***> wrote:﻿
Suprisingly, it suffers from other problems in SW rendering:

MarcelHB:
Hold on, I sill see something to fix after refactoring the code .

lynxlynxlynx:
LGTM now, thanks.

PR #1781, opened 2023-02-10
Added support for embedded portable handhelds

kloptops:
## Description
This PR is the changes required to allow GemRB to run on embedded portable handheld devices running linux.

The changes I have made are for the following purposes:
- These handhelds have limited hardware and as such require gles2 for rendering. The current shaders do not work, I have added a flag to disable GLSL shaders and this allows the game to run well. (-DUSE_NO_GLSL)
- The game is run with a companion program that listens to the handhelds gamepad and translates it into keyboard/mouse events, this allows more fine grained control and better portability between devices/platforms. GemRB currently listens for the same events and is doubling up events, i have added an option to disable the sdl controller api inside of gemrb. (-DSDL_DISABLE_CONTROLLER_API)
- The last option is to fix resolution independence on devices with higher resolution screens. These devices do not run with an x windows server so the current fullscreen/windowed code doesn't scale up the screen resolution. (-DSDL_RESOLUTION_INDEPENDANCE="ON")

Thank you so much for this excellent engine, we have a very active community and many people are excited to use this to play such classic games.

## Checklist


lynxlynxlynx:
Thanks! Some cleanup will be needed, but first I want to check this odd GLSL stuff — did you actually load a game and saw it render well? I'm not sure you get much benefit from enabling a crippled GLES compared to the software rendering.

kloptops:
Okay you're right, the GLSL stuff can be removed. I had issues with it when I was first getting the engine to run on the hardware. I ended up disabling the shader stuff then never tried it again.

Here is bg1 running on the hardware.


IMG_6420

MarcelHB:
Can you elaborate what kind of devices you have been testing GemRB here? Like the SoC of that ... GameBoy? :wink: For future references and compatibility.

kloptops:
So we have been testing it on the [Anbernic RG351V](https://anbernic.com/products/anbernic-new-rg351v) (Rockchip RK3326), [RG353V](https://anbernic.com/products/rg353v-rg353vs) (Rockchip RK3566), and [RG552](https://anbernic.com/products/anbernic-rg552) (Rockchip RK3399) on the [AmberElec](https://amberelec.org)/[ArkOS](https://github.com/christianhaitian/arkos)/[UnofficialOS](https://github.com/RetroGFX/UnofficialOS) operating systems.

We use a system called [PortMaster](https://github.com/christianhaitian/PortMaster) that allows us to package up the games in a way that they can run on all the above oses/hardware.

CMakeLists.txt
```diff
@@ -392,6 +395,13 @@ IF(NOT OPENGL_BACKEND STREQUAL "None")
 		ENDIF()
 	ENDIF()
 
+	IF(SDL_DISABLE_CONTROLLER_API)
+		ADD_DEFINITIONS("-DSDL_DISABLE_CONTROLLER_API")
+	ENDIF()
+
+	IF(SDL_RESOLUTION_INDEPENDANCE)
+		ADD_DEFINITIONS("-DSDL_RESOLUTION_INDEPENDANCE")
+	ENDIF()
```
lynxlynxlynx, line 404: You can add these to `cmake/cmake_config.h.in` instead, the cmakedefine stuff can simplify it.

CMakeLists.txt
```diff
@@ -97,6 +97,9 @@ OPTION(USE_FREETYPE "Enable FreeType support" ON)
 OPTION(USE_PNG "Enable LibPNG support" ON)
 OPTION(USE_VORBIS "Enable Vorbis support" ON)
 
+OPTION(SDL_DISABLE_CONTROLLER_API "Don't use the sdl controller apis." OFF)
```
lynxlynxlynx, line 100:
```suggestion
OPTION(SDL_DISABLE_CONTROLLER_API "Don't use the SDL controller APIs." OFF)
```

lynxlynxlynx:
You can then add AmberElec / ArkOS /UnofficialOS to the exotic OSes list in the README as well.

gemrb/core/GUI/Window.cpp
```diff
@@ -558,7 +558,9 @@ bool Window::DispatchEvent(const Event& event)
 			DispatchTouchUp(target, event.touch, event.mod);
 			break;
 		default:
-			assert(false); // others should be handled above
+			// assert(false); // others should be handled above
+			// NO! DO NOTHING!!!!
```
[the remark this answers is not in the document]
kloptops: Yes this kept getting triggered when when a button was pressed on our controller, but it was not assigned to any keyboard/mouse functions.

It's no issue if it is ifdef'd, what would be the recommended define?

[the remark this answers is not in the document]
lynxlynxlynx: SDL_DISABLE_CONTROLLER_API?

CMakeLists.txt
Mingun: Maybe also expand description so people known when they could use those settings? Like "Don't use the SDL controller APIs. Used for ..."

Also, maybe invert the setting to `USE_SDL_CONTROLLER_API` to make it look like `USE_PNG`, `USE_VORBIS` above?

gemrb/core/GUI/Window.cpp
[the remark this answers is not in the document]
kloptops: Yup, fair call.

CMakeLists.txt
kloptops: Yeah no worries. Let me fix that.

gemrb/plugins/SDLVideo/SDLVideo.cpp
```diff
@@ -207,6 +207,10 @@ int SDLVideoDriver::ProcessEvent(const SDL_Event & event)
 				}
 			}
 			break;
+#ifndef USE_SDL_CONTROLLER_API
+		default:
+			break;
+#else
```
Mingun, line 213: It will be logical to change the code to
```c++
#ifdef USE_SDL_CONTROLLER_API
...
#else
		default:
			break;
#endif
```
just for consistency

kloptops: the default is superfluous, would you rather i leave it off?

lynxlynxlynx: No, but you can move it to the end of the switch. It's superfluous only to the compiler.

gemrb/plugins/SDLVideo/SDLVideo.cpp
```diff
@@ -228,6 +228,9 @@ int SDLVideoDriver::ProcessEvent(const SDL_Event & event)
 				EvntManager->DispatchEvent(std::move(e));
 			}
 			break;
+#else
+		default:
+			break;
 #endif
```
lynxlynxlynx, line 234: both cases, whether the define is set or not, need the default.

kloptops: The default case wasn’t there originally, so I figured you only wanted it when USE_SDL_CONTROLLER_API is undefined. 

But no worries, I’ll have to do it tomorrow.

lynxlynxlynx: It's just MISRA/MITRE/CERT defensive programming best practice. It clearly shows that we indeed thought about all the possible cases the value can take.

lynxlynxlynx:
Thanks!

PR #1782, opened 2023-02-27
Fix typo found in demo dialogue.

desecaw13:
The Riddler says "and old grey guy" which does not make sense to me. So I have changed "and" to "an", which does make sense to me.

## Checklist


lynxlynxlynx:
Thanks and very well done, not many dare to edit binary files. :)

PR #1784, opened 2023-03-11
Allow enums to be Iterable

bradallred:
I noticed we do this a few times which requires a lot of manual casting. This adds a convenient method for iterating an enum that takes the burden of casting onto itself and does so safely/efficiently by using the underlying type.

## Checklist

### DIFF
--- a/gemrb/core/EnumIndex.h
+++ b/gemrb/core/EnumIndex.h
@@ -1,0 +1,133 @@
+/* GemRB - Infinity Engine Emulator
+ * Copyright (C) 2023 The GemRB Project
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ *
+ */
+#ifndef EnumIndex_h
+#define EnumIndex_h
+
+#include "EnumFlags.h"
+
+#include <array>
+#include <bitset>
+#include <cassert>
+#include <initializer_list>
+#include <type_traits>
+
+namespace GemRB {
+
+template <typename ENUM>
+using under_t = typename std::underlying_type<ENUM>::type;
+
+template <typename ENUM, typename ARG = under_t<ENUM>>
+// constexpr // FIXME: make this constexpr in c++14, assert cannot be used in constexper before then
+ENUM EnumIndex(ARG val) noexcept
+{
+	static_assert(std::is_same<under_t<ENUM>, ARG>::value, "Will not implicitly convert to EnumIndex");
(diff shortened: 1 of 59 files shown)


lynxlynxlynx:
Now you can remove the new begin/end members in GUIColors and HCStrings.

bradallred:
I did think about it, but that lends itself to non obvious bugs when somebody wants to add something to them. It's also a bit more ergonomic to do `for(const EnumType e : EnumIterator<EnumType>())` as opposed to `for(const EnumType e : EnumIterator<EnumType, Begin, End>())`.

lynxlynxlynx:
Yeah, fetching the first element ourselves would already be better. There are not many users and half of them already pass all the data.

gemrb/core/WorldMap.h
```diff
@@ -94,8 +96,8 @@ class GEM_EXPORT WMPAreaEntry {
 	ieStrRef LocCaptionName = ieStrRef::INVALID;
 	ieStrRef LocTooltipName = ieStrRef::INVALID;
 	ResRef LoadScreenResRef;
-	ieDword AreaLinksIndex[4]{};
-	ieDword AreaLinksCount[4]{};
+	EnumArray<WMPDirection, int> AreaLinksIndex;
+	EnumArray<WMPDirection, int> AreaLinksCount;
```
lynxlynxlynx, line 100: Why the change of type, doesn't it just create more work for us in the importer?

gemrb/core/EnumIndex.h
```diff
@@ -0,0 +1,133 @@
...
+	static_assert(std::is_unsigned<under_t<ENUM>>::value, "EnumIndex must be unsigned");
+
+	under_t<ENUM> val;
+public:
+	explicit EnumIterator(const ENUM& e) : val(UnderType(e)) {}
+	EnumIterator() : EnumIterator(BEGIN) {}
+	EnumIterator operator++() {
+		++val;
+		return *this;
+	}
```
lynxlynxlynx, line 56: Is this good enough? Specifically for sparse enums like `Direction` in FogRenderer.

bradallred: C++ doesn't have good enough reflection to constrain or increment any other way. We just have to know what we are doing.

We don't use `EnumIterator` with `FogRenderer::Direction`, but if we did we should iterate over "invalid" Directions such as `N|S`. The "invalid" directions would still exist inside an `EnumArray` tho, you can see this with `BAM_FLAGS` and we just fill those invalid gaps with `BlitFlags::NONE`. These slot are just unused filler.

gemrb/core/WorldMap.h
bradallred: a type change had to occur somewhere and the compiler complained about `GUIScript` first, which uses `int` :)

I don't know what is preferable, AFICT it doesn't matter which type we use (`int` vs `ieDword`).

lynxlynxlynx: I'd prefer to keep matching the "spec", which uses ieDword.

gemrb/core/WorldMap.cpp
```diff
@@ -444,9 +439,9 @@ void WorldMap::SetEncounterArea(const ResRef& area, const WMPAreaLink *link) {
 	AddAreaLink(std::move(ldest));
 	AddAreaLink(std::move(lsrc));
 
-	for (i = 0; i < 4; ++i) {
-		ae.AreaLinksCount[i] = 2;
-		ae.AreaLinksIndex[i] = static_cast<ieDword>(idx);
+	for (WMPDirection dir : EnumIterator<WMPDirection>()) {
+		ae.AreaLinksCount[dir] = 2;
+		ae.AreaLinksIndex[dir] = static_cast<int>(idx);
```
lynxlynxlynx, line 444: One more here.

PR #1788, opened 2023-04-05
GUI script refactor

bradallred:
## Description
This PR supersedes #1481. It is still WIP, and has much more work to do. I also haven't tested anything since rebasing on master. I'm pretty sure stores are currently broken entirely as I was working though updating GUIStore the last time I was working on this. 

I wanted to do something much less disruptive, so I took a more surgical approach on this go around so hopefully there is better backwards compatibility with scripts that haven't been updated.

## Checklist


gemrb/GUIScripts/GUICommonWindows.py
```diff
@@ -824,52 +824,6 @@ def ActionQSongRightPressed (which):
...
-	if TopIndex > Max - 10:
-		if Max>10:
-			if TopIndex > Max:
-				TopIndex = Max - 10
-		else:
-			TopIndex = 0
-	GemRB.SetVar ("TopIndex", TopIndex)
-	UpdateActionsWindow ()
-	return
-
```
lynxlynxlynx, line 872: This is odd for a first commit — what replaced them?

bradallred: AFICT, nothing is calling these in master

lynxlynxlynx: Like the rest, the callers are generated (SetActionIcon); these two do exactly as documented.

bradallred: TIL, thanks.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -1994,6 +1994,74 @@ static PyObject* GemRB_Control_SetVarAssoc(PyObject* self, PyObject* args)
...
 
+PyDoc_STRVAR( GemRB_Window_GetVar__doc,
+"===== Window_GetVar =====\n\
+\n\
+**Prototype:** GWindow.GetVar (var)\n\
+\n\
+**Metaclass Prototype:** GetVar (var)\n\
+\n\
+**Description:** Get the value of a window variable.\n\
+Variables are set as a result of calling SetVarAssoc on a GControl.\n\
```
lynxlynxlynx, line 1996: I suggest you either mention it here or add both Window_SetVar and SetVarAssoc to a "See also" section. Like before core may change the values as well.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -1994,6 +1994,74 @@ static PyObject* GemRB_Control_SetVarAssoc(PyObject* self, PyObject* args)
...
+
+PyDoc_STRVAR( GemRB_Window_SetVar__doc,
+"===== Window_SetVar =====\n\
+\n\
+**Prototype:** GWindow.SetVar (var, val)\n\
+\n\
+**Metaclass Prototype:** SetVar (var, val)\n\
+\n\
+**Description:** Set the value of a window variable.\n\
+Variables are set as a result of calling SetVarAssoc on a GControl.\n\
```
lynxlynxlynx, line 2027: Same as above and I guess SetValue as well.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -188,7 +187,6 @@ def ChangeStoreView(func):
 	Store = GemRB.GetStore ()
 	#based on shop type, these buttons will change
 	store_buttons = Store['StoreButtons']
-	BarteringPC = GemRB.GameGetFirstSelectedPC ()
```
lynxlynxlynx, line 191: Did this update the actor before, if you changed the selection while in store? Now we explicitly only work with the initiator.

gemrb/GUIScripts/GUICommonWindows.py
```diff
@@ -1360,11 +1361,11 @@ def ret (btn = None):
...
 			if optwin:
 				rtgbtn = optwin.GetControl(0) # return to game button
-			if optwin and rtgbtn: # not in PST or IWD2
-				rtgbtn.SetState(IE_GUI_BUTTON_UNPRESSED)
-			if btn:
-				btn.SetState(IE_GUI_BUTTON_SELECTED)
-				GemRB.SetVar ("OPTBTN", btn.ID)
+				if btn:
+					btn.SetState(IE_GUI_BUTTON_SELECTED)
+					optwin.SetVar("OPTBTN", btn.ID) # cant use btn.ID because it is "too large to convert to C long"
```
lynxlynxlynx, line 1357: This code and comment combo doesn't make sense.

gemrb/GUIScripts/GUISTORE.py
```diff
@@ -48,23 +43,15 @@
...
-RepModTable = None
-SpellTable = None
-PreviousPC = 0
-MaxAmount = 0
-BuySum = 0
-SellSum = 0
 
+ItemButtonCount = 6 if GameCheck.IsIWD2() else 4
+CureTable = GemRB.LoadTable("speldesc", True) # additional info not supported by core (PST)
+RepModTable = GemRB.LoadTable ("repmodst", True) # all but PST; FIXME: does IWD2 sufffer a 10x price increase?
```
lynxlynxlynx, line 52: Yes it does, see comments on the previous PR.

gemrb/core/DialogHandler.cpp
```diff
@@ -163,8 +158,20 @@ bool DialogHandler::InitDialog(Scriptable* spk, Scriptable* tgt, const ResRef& d
 		Log(DEBUG, "DialogHandler", "Could not find a proper state");
 		return false;
 	}
+	
+	Actor* dialogPC = GetSpeaker();
+	if (dialogPC && dialogPC->InParty) {
+		core->GetDictionary()->SetAt("DIALOG_PC", dialogPC->InParty);
+	} else if (tgt->Type == ST_ACTOR) {
+		core->GetDictionary()->SetAt("DIALOG_PC", static_cast<Actor*>(tgt)->InParty);
+	}
```
lynxlynxlynx, line 167: This is too volatile, the var basically shouldn't be used for anything else but the portrait icon. So a name like DIALOG_STARTER or similar would be more fitting.

gemrb/GUIScripts/GUIOPT.py
```diff
@@ -111,15 +111,22 @@ def InitOptionsWindow (Window):
...
-	if not GemRB.SaveConfig():
-		print("ARGH, could not write config to disk!!")
-
 ###################################################
 
-def CloseVideoOptionsWindow ():
-	GemRB.GetView("SUB_WIN", 0).Close()
-	TrySavingConfiguration()
+def SaveSettings(Window, settings):
+	# volume sliders are multiplers, so apply thay before commition
```
lynxlynxlynx, line 117: This comment can be improved. :)

I'm sceptical this will work, but let's see once the job is done.

lynxlynxlynx:
Looking at this one, maybe a third of commits are redundant, since they already went in with the pst portraits PR. The first one needs to be dropped. The first new thing is b83cab3, but later there are more things that have already been merged. Rebasing would take care of that, but there would probably be conflicts, since many of the changes in the other PR were later amended. 

So it's probably easiest to branch off fresh master first, then just cherry-pick in the work that is new.

lynxlynxlynx:
I cherry-picked one change to master, but most are already inside. What remains in here is everything related to the new window var context that's indeed still a WIP.

lynxlynxlynx:
Alright, I've extracted the paperdoll related changes and integrated them into master, fixing as needed. It wasn't really tied to the rest of the work.

What remains in the diff:
- window var context stuff on the C++ side that could still be manually lifted and finished
- on the python side this manifests mostly through `GemRB.GameGetSelectedPCSingle` replacement, which is more easily redone, since it's just a search & replace.
- GUIOPT seems like the only real user (for the cancellation mechanic)

So if you want to pick this up again at some point, starting from master and copying chunks from the overall diff would be massively easier than trying to resolve the merge conflicts.

Issue #1746, opened 2022-11-29
What is missing for VSync, or higher FPS in general? 


MarcelHB:
#### Description

Unless one modifies the code, the fps is capped to 30fps. It would be nice to have an option to sync the framerate to the refresh rate of the display, and in general this raises a question: What's actually preventing us from implementing that so far?

So the benefits are for example: mouse cursor isn't slowing down (in windowed mode), scrolling a text area or flying over the map can be a lot smoother.

Problems I've observed so with some quick tests (without _any_ capping):
- Moving around the map by pushing the cursor to the borders is way too fast.
- ...

Is there anything else that I missed and that needs to be addressed first w. r. t. higher FPS? Scripts, actions, scheduling, ...?

lynxlynxlynx:
See discussion here https://github.com/gemrb/gemrb/issues/144#issuecomment-585403875, we need to at least implement that max ini key.

MarcelHB:
Of course, that's what I had in mind for this kind of enhancement.

burner1024:
> just FYI, even in the originals boosting the frame rate had other consequences. generally everything runs faster and its really not a great experience. We would have to do quite a bit to make game logic and animations run from a clock rather than happen every "tick" of the game loop

I beg to differ, for me the original 30 FPS is not a great experience.

And unless I misunderstand, it's actually the other way around. I've tried GemRB with 90 fps (btw, `NOFPSLIMIT` seems to do nothing, so modified cap in `Video.h`), and the only differences I see with 30 fps version is cursor move screen speed and pendulum swing speed. I would say, twice as fast, although one would expect 3 times.

lynxlynxlynx:
Yeah, people like to run at higher AI framerates. I don't understand the appeal, but it has been requested often enough.

AFAIK the originals had no drawing-related FPS settings, hence the confusion.

burner1024:
It certainly would be nice to have 2 different settings, one for game speed, and another one for drawing interface.
But still, I imagine the first step would be to tie the whole game process except the interface into one timer/clock/setting.

MarcelHB:
Sure, this topic probably cannot be addressed without having two decoupled clocks, which we then can also try to parameterize.

Issue #144, opened 2019-05-01
[feature] Customize AI Frame Rate


cygarniczka:
BG, BG2, IWD, and IWD2 (not sure about PST) allows for advance option customization which one of the being is setting FPS.
In all four games the ini setting responsible for AI Frame Rate is **Maximum Frame Rate** 

AI Frame allows for change in game dynamics. IMO the game best work when set between 38-43. Game feels a lot smother than hardlocked 30.

Description from the BGConfig.exe

> AI Frame Rate controls the overall speed of the game.  Baldur's Gate II is designed to run with an AI Frame Rate at 30.  Increasing this option on a fast machine will increase the overall speed of the game.  Reducing this option on a slow machine may increase playability by reducing frame-rate slowdown.  WARNING: This option must be equal on all players' machines during a multiplayer game.

lynxlynxlynx:
From another user's testing, I believe this also changes the display FPS (currently hardcoded to ~30 in the video driver). The ai frame rate is hardcoded as AI_UPDATE_TIME, which is definitely 15, so my guess would be it was set to half of that config value.

Maybe a script using [SmallWait](https://gibberlings3.github.io/iesdp/scripting/actions/bg2actions.htm#83) could be made for testing this in the original. One actor with a block that does SmallWait(1) DisplayMessage(...) repeated half the config value times, while another actor with just Wait(1) DisplayMessage(...), to signal a stop after 1 second. Both triggered at the same time, so first a hotkey needs to set a var ... it can get complicated fast.

burner1024:
The other user is probably me.
I don't quite understand, what's needed to test?

There's actual frame rate - displayed fps, and there's `Maximum Frame Rate`, which is a cap on game fps, or "game speed". With CTRL-D, they are displayed:
![Captura de pantalla de 2020-02-13 01-49-13](https://user-images.githubusercontent.com/1279852/74368107-8b793180-4e05-11ea-906b-36f8b4119a12.png)
Game speed is capped by MFR, MFR itself is hard capped at 90, displayed FPS is capped by both game speed and monitor FPS limit:
![Captura de pantalla de 2020-02-13 02-18-30](https://user-images.githubusercontent.com/1279852/74369391-e2800600-4e07-11ea-8550-ed259f06f7ca.png).
![Captura de pantalla de 2020-02-13 02-21-08](https://user-images.githubusercontent.com/1279852/74369368-d8f69e00-4e07-11ea-9f20-f2af5d05b2bd.png)

[...]

Issue #1771, opened 2023-01-07
Video Driver Enhancements


bradallred:
This is something I've been working on. Its not ready to show off yet (much of it is unimplemented still), but knowledge of this might be helpful for ongoing BGEE work, in particular the changes for BAMImporter, Font, and the addition of Atlases.

I'll fill this in with more details eventually, but for now here is an almost complete diagram:
![refactor](https://user-images.githubusercontent.com/1126744/211166529-4968b313-aefd-4ec7-999c-055c77cf23d2.png)

Class removals:
TTFFont
TTFFontManager
BAMFont
BAMFontManager
SpriteSheet
GlyphAtlasPage
GlyphIndexEntry
SDLSurfaceSprite2D
SDLTextureSprite2D

Anything for SDL1, SDL20XXX stuff moves to SDLVideo

Class/Interface Additions:
TTFImporter
PaletteManager
BufferAtlas
PixelBuffer
StreamingVideoBuffer
SDLSoftwareRenderer
SDLOpenGLRenderer

## Regarding SDL 1.2 and OpenGL

There is one annoying hitch: maintaining a standard SDL_Renderer backend becomes more annoying and its already of questionable value due to its current limitations. My proposal then, is this: nix the non-OpenGL capabilities for the current SDL2 renderer. This will simplify the code and bugs/maintenance for that renderer a ton. We then create a second SDL2 renderer that use software rendering and promote it as our new reference implementation so we can just wipe SDL 1.2 entirely (the software rendering bits from it will move to the SDL2 software renderer).

This abstraction sets things up nicely for ports to other rendering backends. Adding a new rendering backend would be able to just add the relevant rendering bits and compile them directly into the SDL video driver and set the relevant setting in the cfg instead of the ifdef soup and hacks we are currently running with. 

Likewise this helps with non SDL ports (SMFL, console, whatever) since the video driver now only cares about creating buffers, drawing geometry, and rendering textures. Any port that wants/needs to rely on software rendering can use the non SDL specific software rendering functions we have (including RLE capabilities). Hardware based rendering is reduced to implementing two relatively simple interfaces.

bradallred:
@MarcelHB does any of this seem helpful to what you are doing?

MarcelHB:
[...]

PR #1789, opened 2023-04-10
VSync, FPS capping, Game state update

MarcelHB:
## Description

I wanted to give it a try and like to invite you for testing. Since it was suspiciously easy for now:

* `CapFPS` is a new tech setting that allows the user to switch between VSync (on SDL2), some custom frame rate limit and no limit.
* I did a simple approach for handling game state and drawing: Assuming refresh rates are much more frequent than game updates. So don't touch the game state for a while and just draw in the meantime.
* Half the value of `Maximum Frame Rate` is now `ai_update_time`, and that's the frequency of the game state's update.
* There is some annoying workaround for SDL1 to work like it did so far, since it appears to lack VSync handling.
* Some logic to prevent scrolling around the map linearly to the fps, thus becoming too fast and while preventing the current relationship.

Benefits:
* On my 144Hz display, the cursor no longer feels like being dragged through water on windowed mode, since SDL gives me 144 fps.
* Moving around the map is now smoother.
* Not sure about scrollbars and texts, but some things appear to look smoother (journal).

See #144, #1746 for previous discussion.

## Checklist

### DIFF
--- a/gemrb/GemRB.cfg.noinstall.sample
+++ b/gemrb/GemRB.cfg.noinstall.sample
@@ -75,6 +75,9 @@ Bpp=32
 #Fullscreen [Boolean]
 Fullscreen=0
 
+# Cap FPS, 0 = request VSync (SDL2 only, default), -1 = no limit, 30+ = cap value
+#CapFPS=0
+
 # Use sprited fog of war [Boolean]. For nostalgia. By
 # default it looks more like accelerated FoW in BG2.
 
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -507,7 +507,6 @@ bool Interface::ReadGameTimeTable()
 		return false;
 	}
 
-	Time.ai_update_time = AI_UPDATE_TIME;
 	Time.round_sec = table->QueryFieldUnsigned<unsigned int>("ROUND_SECONDS", "DURATION");
 	Time.turn_sec = table->QueryFieldUnsigned<unsigned int>("TURN_SECONDS", "DURATION");
 	Time.round_size = Time.round_sec * Time.ai_update_time;
@@ -596,10 +595,25 @@ void Interface::DisableMusicPlaylist(size_t SongType)
 /** this is the main loop */
 void Interface::Main()
 {
-	ieDword speed = 10;
+	int speed = 10;
 
 	vars->Lookup("Mouse Scroll Speed", speed);
-	SetMouseScrollSpeed((int) speed);
+	SetMouseScrollSpeed(speed);
+
+	// We had 36 at 30fps originally, so more fps
+	// should not speed this up unless adjusted
+	auto scrollSpeed = [=](int fps) {
+		return std::max<int>(1080.0 * (speed / 36.0) / fps, 1);
+	};
+
+	if (config.CapFPS == 0) {
+		int refreshRate = video->GetDisplayRefreshRate();
+		if (refreshRate > 0) {
+			SetMouseScrollSpeed(scrollSpeed(refreshRate));
+		}
+	} else if (config.CapFPS >= 30) {
+		SetMouseScrollSpeed(scrollSpeed(config.CapFPS));
+	}
 
 	const Font* fps = GetTextFont();
 	// TODO: if we ever want to support dynamic resolution changes this will break
@@ -612,6 +626,8 @@ void Interface::Main()
 	tick_t frame = 0;
 	tick_t time = GetMilliseconds();
 	tick_t timebase = time;
+	tick_t lastGameUpdate = 0;
+
 	double frames = 0.0;
 
 	do {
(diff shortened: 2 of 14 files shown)


bradallred:
> There is some annoying workaround for SDL1 

Good thing we don't care about SDL 1

gemrb/core/Interface.cpp
```diff
@@ -599,7 +598,21 @@ void Interface::Main()
 	ieDword speed = 10;
 
 	vars->Lookup("Mouse Scroll Speed", speed);
-	SetMouseScrollSpeed((int) speed);
+	SetMouseScrollSpeed(static_cast<int>(speed));
```
bradallred, line 601: just use change `ieDword speed` -> `int speed`. no need to cast.

gemrb/core/Interface.cpp
```diff
@@ -599,7 +598,21 @@ void Interface::Main()
...
+
+	// 1080 comes from originally 30 fps -> scroll speed 36
+	auto scrollSpeed = [](int fps) {
+		return static_cast<int>(std::min(1080.0 / fps, 1.0));
```
bradallred, line 605: same here. just use `std::min<int>` and pass `1`.

gemrb/GemRB.cfg.noinstall.sample
```diff
@@ -75,6 +75,9 @@ Bpp=32
 #Fullscreen [Boolean]
 Fullscreen=0
 
+# Cap FPS, 0 = request VSync (SDL2 only, default), -1 = no limit, 30+ = cap value
+#CapFPS=30
```
lynxlynxlynx, line 79: It's indeed 0 by default, so the commented-out value should reflect that.

gemrb/core/Interface.cpp
```diff
@@ -612,15 +625,21 @@ void Interface::Main()
...
-		for (auto it = timers.begin(); it != timers.end();) {
-			if (it->IsRunning()) {
-				it->Update(time);
-				++it;
-			} else {
-				it = timers.erase(it);
+		bool doGameStateUpdate = time - lastGameUpdate >= 1000.0 / Time.ai_update_time;
+
+		if (doGameStateUpdate) {
+			for (auto it = timers.begin(); it != timers.end();) {
```
bradallred, line 636: I'm pretty sure these timers are realtime, so they don't need to be limited to game ticks.

gemrb/core/Video/Video.cpp
```diff
@@ -133,14 +133,19 @@ void Video::SetStencilBuffer(const VideoBufferPtr& stencil)
 	stencilBuffer = stencil;
 }
 
-int Video::SwapBuffers(unsigned int fpscap)
+int Video::SwapBuffers(int fpscap)
```
bradallred, line 136: I'm not sure what the motivation is for signing this. 0 can be unlimited.

bradallred (changes requested):

lynxlynxlynx:
Yeah, looks fine, thanks! Besides the comment, only the manpage update is missing.

Btw, the original actually ran game updates at the FPS setting. It used double ticks to deal with haste, which we've managed to avoid so far, since it would also mean other complications. Quote from the forum:
>  The main engine loop ticks along at the set FPS value, normally 30tps. I will call this an "engine tick".
>   Creature AI ticks at random mutiples of the engine ticks based on the object's internal ID. Normal creatures tick every other engine tick, Hastened creatures tick every engine tick, Slowed creatures tick every 4 engine ticks.
>   The persistent effect list is updated every AI tick as long as the game is unpaused. 15 of these updates constitute a period. Note how AI ticks can occur at 15tps, 30tps, or 7.5tps given the state of the creature.
>   "world time" ticks every other engine tick while the game is unpaused; spells use world time for duration.

gemrb/core/Interface.cpp
MarcelHB: I'll better check this thoroughly.

gemrb/core/Video/Video.cpp
MarcelHB: I feed it `config.CapFPS` with `-1` (no limit), `0` (VSync) and anything else as a real cap value.

gemrb/core/Interface.cpp
lynxlynxlynx: It's used by the GUI via SetTimer. Doesn't look like there'd be much difference either way.

gemrb/core/Video/Video.cpp
bradallred: maybe I am missing something, but this doesn't seem like it can change dynamically, and we are passing `vsync` to `CreateDriverDisplay` so it seems a bit awkward. Maybe it would be better to just set this up in `CreateDriverDisplay` and not have a parameter here at all.

gemrb/core/Interface.cpp
MarcelHB: PyDoc best describes their use as "time interval in ticks". I can remove the check but I have no idea how this is meant to be understood.

bradallred: It would make a difference for some animation types (eg color anims), since you are limiting them to 30fps. Certainly it would make a difference to `ScrollView` since it uses `PointAnimation` to auto scroll.

bradallred: re `GemRB_SetTimer`, that documentation is out of date. these timers used to be ran by `GlobalTimer`.

MarcelHB:
Rebase and cleanup.

If we really don't care about SDL1 anymore, we should chop that code next :wink:

bradallred:
it will be gone after #1771 is merged

burner1024:
Walking speed definitely doesn't match, and I think animations speed in general too.

burner1024:
Should a new issue be opened for that?

lynxlynxlynx:
Yes please.

Issue #1790, opened 2023-04-10
 Some IWD: HoW savegames fail to load under Release mode


MarcelHB:
#### Bug description
A bunch of my IWD HoW savegames fails to load with following stack trace, but **only** with `CMAKE_BUILD_TYPE=RelWithDebugInfo` (thus I think on `Release`, too) (GCC v12.2.0), so I was unable to see any deeper insights over `Debug` so far:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
GemRB::WEDImporter::GetWallGroups (this=0x146c8b00) at D:/Sources/gemrb/gemrb/plugins/WEDImporter/WEDImporter.cpp:393
393                             auto wp = polygonTable[polyIndex];
(gdb) bt
#0  GemRB::WEDImporter::GetWallGroups (this=0x146c8b00) at D:/Sources/gemrb/gemrb/plugins/WEDImporter/WEDImporter.cpp:393
#1  0x00007ffd4b282da5 in GemRB::AREImporter::GetMap (this=<optimized out>, resRef=..., day_or_night=<optimized out>) at D:/Sources/gemrb/gemrb/plugins/AREImporter/AREImporter.cpp:1578
#2  0x00007ffd3427213c in GemRB::Game::LoadMap (this=this@entry=0x12ee8eb0, resRef=..., loadscreen=loadscreen@entry=true) at D:/Sources/gemrb/gemrb/core/Game.cpp:847
#3  0x00007ffd342723a9 in GemRB::Game::GetMap (this=this@entry=0x12ee8eb0, areaname=..., change=change@entry=true) at D:/Sources/gemrb/gemrb/core/Game.cpp:685
#4  0x00007ffd34229a3c in GemRB::GameControl::ChangeMap (this=this@entry=0x14199f80, pc=0x130c5540, forced=forced@entry=true) at D:/Sources/gemrb/gemrb/core/GUI/GameControl.cpp:2532
#5  0x00007ffd342ebab8 in GemRB::Interface::HandleFlags (this=this@entry=0xd7240) at D:/Sources/gemrb/gemrb/core/Interface.cpp:479
#6  0x00007ffd342f3308 in GemRB::Interface::Main (this=0xd7240) at D:/Sources/gemrb/gemrb/core/Interface.cpp:629
#7  0x00007ff6f7658e20 in main (argc=3, argv=0xd44b0) at D:/Sources/gemrb/platforms/windows/GemRB.cpp:53
(gdb) p polyIndex
[...]

PR #1791, opened 2023-04-10
Fast-math fix for GCC/Clang

MarcelHB:
See #1790.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,7 +257,7 @@ IF(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
 	endif (NOT APPLE AND NOT UNSAFE_PLUGIN)
 
 	# Fast math helps us covering some things that need a little more rework soon
-	string(APPEND CMAKE_CXX_FLAGS " -ffast-math")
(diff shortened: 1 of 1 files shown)


bradallred:
is there not a fix required for MSVC as well?

MarcelHB:
Thought so too but the [docs](https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior) don't show any more sophisticated control.

Best guess is dropping `/fp:fast` from reading what it says but I have no way to check MSVC for either being affected so far or being fixed by some similar measure.

bradallred:
good enough until someone complains or can test it.

lynxlynxlynx:
Actually that's a good point. A user did just complain yesterday. He gets crashes sometimes when scrolling text, also always at the end of chargen. So it could be something like this.

MarcelHB:
Is the user using some AppVeyor nightly, or using MSVC in general? Because v0.9.1 did not have that back then.

lynxlynxlynx:
Latest nightly, but I've asked him to retest with a build from before that change.

Issue #1797, opened 2023-05-24
Tooltip delay setting doesn't work


burner1024:
#### Bug description

Tooltip delay setting doesn't work (doesn't change the delay), console throws errors:
```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "/opt/gemrb/share/gemrb/GUIScripts/GUIOPT.py", line 304, in DisplayHelpTooltipDelay
[Python/ERROR]:     GemRB.SetTooltipDelay (GemRB.GetVar ("Tooltips") * TOOLTIP_DELAY_FACTOR/10)
[Python/ERROR]: TypeError: 'float' object cannot be interpreted as an integer
[Python/ERROR]: Traceback (most recent call last):
```

#### Steps to reproduce
1. Go to tooltip delay setting
2. Change it.

#### Expected behavior
Working setting.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

burner1024:
There's another bug with this: maximum on that setting is supposed to disable tooltips, which it doesn't. Not sure where to fix this.
![Captura desde 2023-05-25 11-21-42](https://github.com/gemrb/gemrb/assets/1279852/100dbd47-d269-4cfc-8f6e-87ed38720218)

lynxlynxlynx:
Oh, there's more. The easiest way would be to just detect the highest value and then set the delay to something enormous.

burner1024:
I don't fully understand how settings work.
If this can be done in the same `DisplayHelpTooltipDelay` function, I can send a PR.

lynxlynxlynx:
Yes, exactly there, it's the same call you already modified.

PR #1804, opened 2023-05-25
(effectively) disable tooltips on max delay setting, like description says

burner1024:
## Description
Set delay tooltip to a very high value on max setting ("disable" it).
Ref #1797

## Checklist


lynxlynxlynx:
Thanks!

Issue #1810, opened 2023-05-26
Invisibility pulsating too strong 


burner1024:
#### Bug description
Invisibility transparency goes 0-100%.

#### Steps to reproduce
1. Cast invisibilty on any party member.
2. See how they are pulsating.

#### Expected behavior
Should be something like 0-50%.

#### Screenshots
Original
![0](https://github.com/gemrb/gemrb/assets/1279852/f977d00f-15dd-46ec-8789-977623fa6dab)

Gemrb
![2](https://github.com/gemrb/gemrb/assets/1279852/863366c9-e028-468f-960c-c254b7e7fdd5)

(Avatars and colors are all different too, don't know why).

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
More like 50-90% from that gif. And slower.

burner1024:
Definitely not 50%. Maybe gif doesn't represent it well.
Hard to catch, since pulsating doesn't stop with pause. Here's the lowest I could screen.
![Captura desde 2023-05-27 01-00-24](https://github.com/gemrb/gemrb/assets/1279852/05d3b475-3c7c-4915-91d3-2cf7f72aa2bb)
I think it touches 0.

And here's max I could cap:
![Captura desde 2023-05-27 01-05-28](https://github.com/gemrb/gemrb/assets/1279852/c58ef777-8610-44f7-9304-d1c96f75f73a)

lynxlynxlynx:
If that's 0, then you're talking about opacity, not transparency/invisibility.

burner1024:
Yes.

lynxlynxlynx:
You can play with this by setting the 32 to a higher number at the end of fx_set_invisible_state. 128 might do it.

burner1024:
I put 128, seems closer.
And I think you're right about speed, I changed the interval to 8 and I think it's better.
![0](https://github.com/gemrb/gemrb/assets/1279852/05309e51-cbff-4778-9b6d-d0b94fc3f98b)
![1](https://github.com/gemrb/gemrb/assets/1279852/2ad45d7a-7eca-4da3-a755-b062cfb01ac8)

[...]

Issue #1098, opened 2020-12-12
Brightness setting not restored when restarting game


FrElvire:
#### Bug description
When setting Brightness in game all is right but when you restart the game, brightness is not restored even if brightness slider is at the right place.

Tried to reproduce with contrast setting, but the slider does nothing....

Tested with IWD1 and BG2

For information : game files (read only) and save/cache/gem-icewind.ini (read/write) are not at the same place. gem-icewind.ini is modified by setting brightness in game.

GemRB version: master

lynxlynxlynx:
Yeah, I don't think contrast is even implemented. Comparing the same run in the original and in gemrb makes it obvious we / SDL doesn't treat the values the same.

bradallred:
it probably behaves differently for SDL 2

lynxlynxlynx:
It's even more complicated. The SDL_SetWindowBrightness function adjusts gamma of the whole screen. On Windows it also resets the system's active color profile (ICC) and undoes some other display options like those available on the NVIDIA Control Panel. So it's not user friendly in that regard.

Plus it might not even work properly. The docs:
> Many platforms will refuse to set the display brightness in modern times. You are better off using a shader to adjust gamma during rendering, or something similar.

bradallred:
we do have a shader so we could incorporate it there

burner1024:
Yes, contrast setting does nothing, and brightness currently goes something like from 0 to 150%, while in original it's maybe 75% to 125%.
See [here](https://github.com/gemrb/gemrb/pull/1761#issuecomment-1354060397) for comparison screenshots. Unfortunately, GemRB looks noticeably darker, and raising brightness does not result in the same look as the original.

burner1024:
Right... I re-checked, and with original set to minimal contrast colors look the same, so it's indeed contrast that makes the difference.

burner1024:
Reading issue description again, the problem described there looks similar to #1762, the setting gets reset.

That's separate from other issues:
- Brightness has different limits in gemrb compated to original game.
- Brightness control only works in full screen mode.
- Contrast is not implemented at all.

PR #1812, opened 2023-05-27
set lower limit for invisibility opacity, decrease pulse interval

burner1024:
## Description
Fix invisibility opacity limits and pulse interval (#1810)

Original

https://github.com/gemrb/gemrb/assets/1279852/1eddec3e-87d8-4b23-b595-f2f171ae0061

New GemRB

https://github.com/gemrb/gemrb/assets/1279852/5a2f4f1e-477f-4fd4-b42f-94a27c6fc7b3

I'm struggling to match original exactly, because of [brightness/contrast differences](#1098), as well as avatar/color differences.
However, it's closer than current master.

## Checklist

### DIFF
--- a/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
+++ b/gemrb/plugins/FXOpcodes/FXOpcodes.cpp
@@ -1789,16 +1789,16 @@ int fx_set_invisible_state (Scriptable* /*Owner*/, Actor* target, Effect* fx)
 	}
 	ieDword Trans = fx->Parameter4;
 	if (fx->Parameter3) {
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks. If you add spaces around the three operators it will be even better, matching the main style.

burner1024:
Sure, added.

Aren't there some autoformatters, though? I try to use those as much as I can, can't be bothered with style discussions and corrections.

lynxlynxlynx:
Yes, we have an issue open for that as well. It basically boils down to some refinement and git blame not being able to ignore refs without passing parameters out of the box (last I checked).

Issue #1374, opened 2021-07-23
Replace HashMap with std::unordered_map


bradallred:
now that we use c++11 that actually has a more or less proper hash map supplied by the std library

I don't expect there to be any benefit other than it being familiar and more idiomatic, but I dont expect it to be painful to do either.

bradallred:
@MarcelHB Thanks for doing the dirty work! There are still custom hash maps being used by `Variables` and `Cache` if you happen to be interested in addressing those as well.

MarcelHB:
Hm, those aren't simply a another map implementation but involve active reference counting. So instead of trashing `Cache` and `Variables`, I'd go for removing the data tree management and leave the interface as-is to the required degree, including the RC, and think of a better ownership of the data.

bradallred:
I wasn't suggesting junking them, just that they both use a `MyAssoc` that is effectively a hashmap. Reference counting can be done via `shared_ptr`, if it is in fact needed. I'd expect things that are cached to be const, so shared ownership there would be silly. `Variables` don't use a refcount, so thats much more straight forward.

PR #1813, opened 2023-06-17
Dropping custom map implementations

MarcelHB:
## Description
See the idea of #1374. I don't see any impact, or don't know if there is any good place to look for one, but compilers' libs ship bucketed hashing as well.

All the `Logging/Logging.h` additions are the result of some unfortunate inclusion chain that broke after removing `HashMap.h`.

## Checklist


lynxlynxlynx:
Nice, thanks!

PR #1814, opened 2023-06-18
Formalize DISABLE_WERROR as a CMake option

majcosta:
## Description
Just a QoL thing for people using ccmake or IDEs such as VScode's cmake extension. Options show up on the list of cache variables to be turned on and off.

Still defaults to disabled, so behavior is unchanged

## Checklist


lynxlynxlynx:
Sure, why not. Thanks!

MarcelHB:
Good idea, since GCC 13 is spilling plenty of false positives in all versions of libfmt right now.

Issue #1815, opened 2023-06-22
 iwd1 autopause options issue


gregariousmonk:
Version 0.9.0 installed via Lunux Mint software manager.
Icewind Dale Complete (GOG) (how)

Autopause options are visible on the autopause options panel, but the options for "enemy spotted", "trap found", "", and "center" are not selectable (ie no highlight appears when clicking on those options (all others can be selected normally.

I traced the issue to GUIopts.py, there is a check for those options, but it only checks for iwd2.  I added "or IWD1" and the options are now selectable.  I have not yet tested if the selected options work correctly or not.  This code appears to be the same in 0.9.1.

GemRB version (check as many as you know apply)
- [ ] master as of this issue
- [ ] 0.9.1
- [ x] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ x] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Thanks, please submit a pull request, if you can.

gregariousmonk:
I have attempted to do as you have requested, but honestly it's been a while since I worked on anything collaborative and my git skills are rudimentary.  If I find other bugs I can fix I'll figure out how this works properly.

PR #1816, opened 2023-06-23
Update GUIOPT.py

gregariousmonk:
## Description
iwd1 autopause options issue #1815

Added gamecheck for IWD1 to the autopause options as per inline comment. The options for "enemy sighted", "trap found", "spell cast", and "center" are present in IWD1 but do not work correctly without the added check, as described below..  

The option highlight does not appear on the 4 indicated options prior to adding the fix and the option behaviours do not change when clicked on; after adding the fix the option highlights appear as for the other options and the options can be set/unset normally.

## Checklist


lynxlynxlynx:
Thanks, I've cleaned it up and committed.

PR #1818, opened 2023-06-25
Cleanup of Variables, Caches

MarcelHB:
## Description
See part two of #1374. Need to review this myself again tomorrow since it's non-trivial size. Anyway, rationale:

* `Cache` has been internally refactored but active RC has been left intact since I did not want to start following these use-cases.
* `LRUCache` has been refactored since it was built on top of `Variables` and its use was rather awkward (user, i. e. audio drivers managing eviction from the outside).
* `Variables` has been replaced, mostly in favor of `unordered_map` with various degrees of facade by its holder.

Also, these structures are now owner in every case.

## Checklist


gemrb/core/Cache.h
```diff
@@ -29,65 +31,94 @@ namespace GemRB {
...
+			return nullptr;
+		}
+
+		template<typename ... ARGS>
+		std::pair<V*, bool> SetAt(const K& key, ARGS && ... args) {
+			auto insertion =
+				map.emplace(
+					std::piecewise_construct,
+					std::forward_as_tuple(key),
+					std::forward_as_tuple(std::forward<ARGS>(args)...)
```
lynxlynxlynx, line 68: These are defined in <tuple>, so it might be prudent to include it directly.

gemrb/core/GameData.cpp
```diff
@@ -268,23 +235,22 @@ Effect* GameData::GetEffect(const ResRef &resname)
...
+	auto newEffect = em->GetEffect();
+	if (newEffect == nullptr) {
 		return nullptr;
 	}
 
-	EffectCache.SetAt(resname, (void *) effect);
-	return new Effect(*effect);
+	EffectCache.SetAt(resname, std::move(*newEffect));
+
+	auto effectCopy = new Effect(*newEffect);
```
lynxlynxlynx, line 245: Wouldn't it be safer/more future-proof to create the copy before the std::move?

gemrb/core/Cache.h
MarcelHB: Yes, won't do any harm.

gemrb/core/GameData.cpp
MarcelHB: Good catch

gemrb/plugins/OpenALAudio/OpenALAudio.h
```diff
@@ -90,6 +90,27 @@ struct AudioStream {
 struct CacheEntry {
 	ALuint Buffer;
 	tick_t Length;
+
+	CacheEntry(ALuint buffer, tick_t length) : Buffer(buffer), Length(length) {}
+	CacheEntry(const CacheEntry&) = delete;
+	CacheEntry& operator=(const CacheEntry&) = delete;
```
MarcelHB, line 96: Reminder to add proper move methods in order to prevent `Buffer` from hanging around more than once (also SDLAudio) in case of some implicit movement.

gemrb/core/SrcMgr.cpp
```diff
@@ -30,27 +30,16 @@ namespace GemRB {
...
+	auto lookup = srcs.find(resource);
+	if (lookup != srcs.cend()) {
+		return &lookup->second;
 	}
 
-	src = new SrcVector(resource);
-	srcs.push_back(src);
-	SrcCache.SetAt(resource, (void *) src);
-	return src;
+	return &srcs.emplace(resource, resource).first->second;
```
lynxlynxlynx, line 42: emplace_back? I'm surprised it compiles.

gemrb/core/LRUCache.h
```diff
@@ -19,46 +19,177 @@
...
-	VarEntry* tail = nullptr;
-
-	void removeFromList(VarEntry* e);
-};
+	public:
+		using key_t = std::string;
+		using public_key_t = StringView;
+
+	private:
+		/* Queue items are place into a doubly-linked list, the front element is the LRU. */
```
lynxlynxlynx, line 42:
```suggestion
		/* Queue items are placed into a doubly-linked list, the front element is the LRU. */
```

gemrb/core/SrcMgr.cpp
MarcelHB: I was looking at `srcs` and missed any benefit in having both the list `srcs` and a `SrcCache` here. So, no, there is no `emplace_back` anymore now.

gemrb/core/Interface.cpp
```diff
@@ -2166,9 +2140,10 @@ void Interface::HandleGUIBehaviour(GameControl* gc)
 				guiscript->RunFunction( "GUIWORLD", "NextDialogState" );
 
 			// the last node of a dialog can have a new-dialog action! don't interfere in that case
-			ieDword newvar = 0; vars->Lookup("DialogChoose", newvar);
+			ieDword newvar = GetVariable("DialogChoose", 0);
+
 			if (var == (ieDword) -1 || newvar != (ieDword) -1) {
-				vars->SetAt("DialogChoose", (ieDword) -3);
+				vars["DialogChoose"] = 3;
```
lynxlynxlynx, line 2146: -3

gemrb/core/Interface.cpp
```diff
@@ -3422,9 +3437,12 @@ bool Interface::ReadItemTable(const ResRef& TableName, const char *Prefix)
 bool Interface::ReadRandomItems()
 {
 	ieDword difflev=0; //rt norm or rt fury
-	vars->Lookup("Nightmare Mode", difflev);
+	auto lookup = vars.find("Nightmare Mode");
+	if (lookup != vars.cend()) {
+		difflev = lookup->second;
+	}
```
lynxlynxlynx, line 3443: How about something like `ieDword difflev = GetVariable("Nightmare Mode", 0)` instead?

gemrb/core/Interface.cpp
```diff
@@ -2349,9 +2353,15 @@ int Interface::PlayMovie(const ResRef& movieRef)
...
+	auto lookup = vars.find("Dialog Movie Subtitles");
+	if (lookup != vars.cend()) {
+		subtitles = lookup->second;
+	}
 	if (!subtitles) {
-		vars->Lookup("Display Subtitles", subtitles);
+		lookup = vars.find("Dialog Subtitles");
+		if (lookup != vars.cend()) {
+			subtitles = lookup->second;
+		}
```
lynxlynxlynx, line 2364:
```
ieDword subtitles = GetVariable("Dialog Movie Subtitles", 1);
if (!subtitles) subtitles = GetVariable("Dialog Subtitles", 0)
```

gemrb/core/Interface.cpp
```diff
@@ -2218,10 +2193,14 @@ void Interface::AskAndExit()
 {
 	// if askExit is 1 then we are trying to quit a second time and should instantly do so
 	ieDword askExit = 0;
-	vars->Lookup("AskAndExit", askExit);
+	auto lookup = vars.find("AskAndExit");
+	if (lookup != vars.cend()) {
+		askExit = lookup->second;
+	}
```
lynxlynxlynx, line 2199: `ieDword askExit = GetVariable("AskAndExit", 0);`

gemrb/core/GUI/GameControl.cpp
```diff
@@ -683,8 +680,13 @@ bool GameControl::OnKeyPress(const KeyboardEvent& Key, unsigned short mod)
 				case GEM_LEFT:
 				case GEM_RIGHT:
 					{
+						auto& vars = core->GetDictionary();
 						ieDword keyScrollSpd = 64;
-						core->GetDictionary()->Lookup("Keyboard Scroll Speed", keyScrollSpd);
+						auto lookup = vars.find("Keyboard Scroll Speed");
+						if (lookup != vars.cend()) {
+							keyScrollSpd = lookup->second;
+						}
```
lynxlynxlynx, line 688: `ieDword keyScrollSpd = core->GetVariable("Keyboard Scroll Speed", 64);`

gemrb/core/Interface.cpp
```diff
@@ -4309,32 +4329,36 @@ void Interface::GetResRefFrom2DA(const ResRef& resref, ResRef& resource1, ResRef
 	}
 }
 
-std::vector<ieDword>* Interface::GetListFrom2DAInternal(const ResRef& resref)
+std::vector<ieDword> Interface::GetListFrom2DAInternal(const ResRef& resref)
 {
 	// FIXME: this is a new/free mismatch inside Variables
 	// luckily, ieDword is trivially destructable and we use the default global allocator
```
lynxlynxlynx, line 4335: This comment can now be removed as well. :)

gemrb/core/Interface.cpp
MarcelHB: Nothing, probably just an oversight since _GetVariable_ was introduced half-way after writing enough of this code :)

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -5400,15 +5416,19 @@ bool Actor::CheckOnDeath()
...
-		ieDword start = 0;
-		vars->Lookup(varname, start);
-		vars->SetAt(varname, start + 1, nocreate);
+
+		auto lookup = vars.find(varname);
+		if (lookup != vars.cend()) {
+			lookup->second += 1;
+		} else if (!nocreate) {
+			vars[varname] = 1;
+		}
```
lynxlynxlynx, line 5431: We now have 4 blocks like this in the same file, so it's ripe for a static function.

gemrb/core/Interface.h
```diff
@@ -805,7 +810,10 @@ class GEM_EXPORT Interface
 	}
 	inline void ResetActionBar()
 	{
-		vars->SetAt("ActionLevel", 0, false);
+		auto lookup = vars.find("ActionLevel");
+		if (lookup != vars.cend()) {
+			lookup->second = 0;
+		}
```
lynxlynxlynx, line 816: This can safely be unconditional. Even if the var doesn't exist at some early stage yet, we'll create it soon enough and it doesn't do any harm to have it set potentially too early.

gemrb/core/GUI/Control.cpp
```diff
@@ -191,15 +198,13 @@ void Control::BindDictVariable(const varname_t& var, value_t val, ValueRange val
...
 	if (GetDictOp() == BitOp::SET) {
 		// SET implies the dictionary value should always mirror Value
 		UpdateDictValue();
 	} else {
-		value_t dictVal = INVALID_VALUE;
-		if (core->GetDictionary()->Lookup(VarName, dictVal)) {
-			UpdateState(VarName, dictVal);
-		}
+		int value = core->GetVariable(VarName.c_str(), INVALID_VALUE);
+		UpdateState(VarName, value);
```
lynxlynxlynx, line 207: Before, this didn't get called if the dict item didn't exist. It would be fine if UpdateDictValue gets called before, but that doesn't look likely. I suggest we just check for INVALID_VALUE first.

lynxlynxlynx:
Very nice btw, finally we are done with the clutch of having to use temporary variables for lookups. Turns out testing the presence in the dicts wasn't that useful.

MarcelHB:
Thanks for the 1st round of reviewing, everything is updated and the code now builds on GitHub. I'll soon do another trip myself since I want to attempt to patterns of recurring stuff such as types or helper methods that can maybe optimized even more.

lynxlynxlynx:
The most verbose stuff is the conditional setting, but adding a wrapper and then using a bunch of lambdas everywhere wouldn't be an improvement.

MarcelHB:
I was thinking about these explicit `std::unoredered_map<RR, dWord, Hasher>` or `LoadInitialVariables`. Also the Mac build is still complaining.

lynxlynxlynx:
The mac/windows error looks trivial though, it's just one line.

MarcelHB:
So I found some room for improvement of my primary concerns (we already types for map + hasher for `ResRef`), found one wrong key, removed the `Variables.cpp` leftover into `Interface.cpp` (it uses `core`, so ...) and both Mac + Windows builds are hopefully addressed by now.

MarcelHB:
AppVeyor compiles but starting fails, but it looks like it does so before this PR already.

lynxlynxlynx:
Yeah, something again changed on the host or some other bullshit we didn't manage to freeze in place, since it started failing out of the blue.

bradallred:
This is really great. Overall, much cleaner and it sounds like you are aware of the few areas that aren't quite there yet.

I'm assuming since you were able to remove `Variables` that there was some separation of its uses that removed the need for storing a "variant" type?

MarcelHB:
Somebody obviously wanted to build code once for three different use cases (pointers, ints, strings) but there was no polymorphic use, so yes, there was absolutely no reason to consider variant-ic code here.

bradallred:
Were we just not using that functionality, or were the uses really just distinct for one another so we just use different maps now?

MarcelHB:
They were really distinct, all users were just calling something like `SetType( GEM_VARIABLES_POINTER )` at init. time and that's it.

NinjaSpectre:
Slightly off topic, but concerning AppVeyor I believe an escape ` is missing to nullify the regex any character . thus changing
https://github.com/gemrb/gemrb/blob/af5921fa88509cc10bd1a671f1fe346319b06d3c/.appveyor.yml#L75
to the below should hopefully be the fix for the minor annoyance.
```PowerShell
$find = '#GemRBPath=`.'

lynxlynxlynx:
The regex dot matches the literal dot too, so that wouldn't change anything. Besides, the error is in a different path, filled in by cmake, not appveyor.

NinjaSpectre:
The regex currently matches twice in GemRB.cfg.sample.in which I believe causes the problem.

lynxlynxlynx:
Still that shouldn't matter, since the good entry is below the other. But let's try.

lynxlynxlynx:
That helped — it's not like sed, but only does the first replacement it can. Must be some powershell change and a version bump on the host started showing it.

Back on track, the release testing should find regressions if any, since Variables is so important for scripting. @MarcelHB @bradallred any other thoughts on this PR? Otherwise just merge it. :)

bradallred:
I didn't see anything concerning. I do wonder if anybody tested SDLAudio? If you and @MarcelHB feel good about it, I say merge it.

bradallred:
Tho, let's be sure to get Sonar and Coverity to check things out afterward.

MarcelHB:
Give it a go.

lynxlynxlynx:
Thanks!

lynxlynxlynx:
The only major complaint from Sonar is this: https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=BUG&id=gemrb_gemrb&open=AYkHiyebfkQwUWT-saJM&tab=code

MarcelHB:
Sigh, ... why? Ok, the standard says so but my mental broke, again.

lynxlynxlynx:
It's suggesting a move instead, since V is a template param of the class.

MarcelHB:
Yes, I was reading footnotes in docs, will do later/at weekend.

Issue #1825, opened 2023-06-28
Pathfinding in narrow corridors is suboptimal 


graag:
#### Bug description
The pathfinding in narrow spaces, with other party members nearby, finds paths that are very unoptimal. It is overly eager to walk around actors compared to the original engine (BG2). This leads to situations where some party members take a different path and walk into enemies / traps.

Part of the problem is that the "bounding circles" used to determine if a map node is blocked are to big which is a one line fix. However it can be improved using a simple heuristic which I'll detail in a PR.

#### Steps to reproduce
1. Go to a narrow corridor with nearby intersections
2. Walk around and observe selected paths for party members

#### Expected behavior
Pathfinding should be more in line with how it works in the original engine.

#### Screenshots

* Pathfinding in GemRB master: https://youtu.be/Fe4MDv6oEV4
* Pathfinding in the original engine: https://youtu.be/PTxUTKtNeE8
* Simple pathfinding fix: https://youtu.be/Ud7Sto3us3c
* Shortcomings of the oneline fix: https://youtu.be/oo62qzbdOq4
* Improved pathfinding with heuristic: https://youtu.be/kiZ8ZN9e0mY

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [x] 0.9.1
- [x] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`
- [x] Linux Appimage - whatever driver it is compiled with

Issue #1750, opened 2022-12-03
Speed up edge map travel


lynxlynxlynx:
>  What I was talking about for the edges is taking another look at Map.cpp's LastGoCloser / CT_GO_CLOSER. In the end Game::EveryoneNearPoint gets called with a single point, so everyone's position is compared to that, not to their closest viable point (projection to the edge).

_Originally posted by @lynxlynxlynx in https://github.com/gemrb/gemrb/issues/1749#issuecomment-1334562533_

lynxlynxlynx:
This is noticeable now not so much due to stuttering (almost none compared to doors), but sometimes pcs even get rerouted to the (0,0) corner somehow.

Issue #1015, opened 2020-10-24
Pathfinding not considering diagonals - problem in AR9106 (HoW) / ar1008


fizzet:
#### Bug description
There's a cupboard container at the north end of the house which appears to be unreachable at the moment. It looks like the actors need too much space.

lynxlynxlynx:
For reference:
![image](https://user-images.githubusercontent.com/121515/97088007-58db3600-162e-11eb-8048-cfe460a200b3.png)

The problem then becomes pretty obvious. If you teleport into the island of walkability, you can get close enough, so it's that diagonal corner transition that is blocking things. I even tried with just blocking one searchmap square and it doesn't help — you could walk right up to the border before as well.

@kmfrick do I see correctly that this is because the pathfinder considers only the four cardinal directions when considering adjacency?

kmfrick:
> For reference:
> ![image](https://user-images.githubusercontent.com/121515/97088007-58db3600-162e-11eb-8048-cfe460a200b3.png)
> 
> The problem then becomes pretty obvious. If you teleport into the island of walkability, you can get close enough, so it's that diagonal corner transition that is blocking things. I even tried with just blocking one searchmap square and it doesn't help — you could walk right up to the border before as well.
> 
> @kmfrick do I see correctly that this is because the pathfinder considers only the four cardinal directions when considering adjacency?

You're right. I'll have a look in chat, notes and old issues to see if there's a specific reason for this. If there isn't one, I'll adjust it to consider diagonal directions too and test for performance before submitting a PR.

lynxlynxlynx:
Thanks. From memory it just seemed to be enough and since it saves a few cycles, we left it like so.

kmfrick:
It's more complicated than that. It's fixed only if we check diagonal directions AND we use `GetBlockedNavmap` instead of `GetBlockedInRadius` to check if the adjacent node is blocked when expanding. I'm afraid this might cause some quantum tunneling, I'll go back to the thieves' maze and test there. Also it does seem weird to me that a single diagonal corner is enough to warrant walkability, but hey, there _is_ a container there.

lynxlynxlynx:
Just noting another test case is recorded in #1100.

MarcelHB:
I tried the obvious thing:
```diff
--- a/gemrb/core/PathFinder.cpp
+++ b/gemrb/core/PathFinder.cpp
@@ -46,11 +46,11 @@

 namespace GemRB {

[...]

PR #1826, opened 2023-06-28
Sanitize party pathfinding in narrow spaces

graag:
## Description
Current implementation of circle drawing uses a "tweaked" version of the Bresenham algorithm: https://funloop.org/post/2021-03-15-bresenham-circle-drawing-algorithm.html. This leads no nicer circles but they tend to be slightly larger then requested radius. In particular for radius 1 we end up with a square where the corners extend quite a bit outside of the intended circle. By going back to the **not** tweaked version we end up with a cross that leads to reduced collisions.
Please take a look at #1825 to compare the GemRB, BG2 and improved behavior.

The fix amounts to this oneliner:
```
diff --git a/gemrb/core/Geometry.cpp b/gemrb/core/Geometry.cpp
index 927c8bfcc..7f0762427 100644
--- a/gemrb/core/Geometry.cpp
+++ b/gemrb/core/Geometry.cpp
@@ -158,7 +158,7 @@ std::vector<Point> PlotCircle(const Point& origin, uint16_t r, uint8_t octants)
        GenOctants(x, y);
        
        while (x < y) {
-               if (fm <= 0) {
+               if (fm < 0) {
                        fm += de;
                } else {
                        fm += dse;
```

Most of the PR tries to improve on the pathfinding by introducing a heuristic that measures if the calculated path is not to long. It works by returning the path length as an additional element of a tuple and comparing it to the euclidean distance of the start and stop points. It ignores short path to minimise bumping and very long paths that are more likely to be longer then the simple geometric distance.

Please take a look at #1825 for an example of improved algorithm at work.

## Checklist

### DIFF
--- a/gemrb/core/Map.h
+++ b/gemrb/core/Map.h
@@ -628,7 +628,7 @@ public:
 	PathListNode* GetLine(const Point &start, const Point &dest, int speed, orient_t Orientation, int flags) const;
 	Path GetLinePath(const Point &start, const Point &dest, int speed, orient_t Orientation, int flags) const;
 	/* Finds the path which leads to near d */
-	PathListNode* FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance = 0, int flags = PF_SIGHT, const Actor *caller = NULL) const;
+	PathListNode* FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance = 0, int flags = PF_SIGHT, const Actor *caller = nullptr) const;
 
 	bool IsVisible(const Point &p) const;
 	bool IsExplored(const Point &p) const;
--- a/gemrb/core/PathFinder.cpp
+++ b/gemrb/core/PathFinder.cpp
@@ -294,6 +294,7 @@ PathListNode *Map::GetLine(const Point &p, int steps, orient_t orient) const
 PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance, int flags, const Actor *caller) const
 {
 	if (core->InDebugMode(ID_PATHFINDER)) Log(DEBUG, "FindPath", "s = {}, d = {}, caller = {}, dist = {}, size = {}", s, d, caller ? MBStringFromString(caller->GetShortName()) : "nullptr", minDistance, size);
+	bool actors_are_blocking = flags & PF_ACTORS_ARE_BLOCKING;
 	
 	// TODO: we could optimize this function further by doing everything in SearchmapPoint and converting at the end
 	NavmapPoint nmptDest = d;
@@ -362,14 +363,15 @@ PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, u
 			if (smptChild.x < 0 ||	smptChild.y < 0 || smptChild.x >= mapSize.w || smptChild.y >= mapSize.h) continue;
 			// Already visited
 			if (isClosed[smptChild.y * mapSize.w + smptChild.x]) continue;
+
+			PathMapFlags childBlockStatus = GetBlocked(nmptChild);
+			bool childBlocked = !(childBlockStatus & (PathMapFlags::PASSABLE | (actors_are_blocking ? PathMapFlags::UNMARKED : PathMapFlags::ACTOR) | PathMapFlags::TRAVEL));
+			if (childBlocked) continue;
+
 			// If there's an actor, check it can be bumped away
 			const Actor* childActor = GetActor(nmptChild, GA_NO_DEAD | GA_NO_UNSCHEDULED);
-			bool childIsUnbumpable = childActor && childActor != caller && (flags & PF_ACTORS_ARE_BLOCKING || !childActor->ValidTarget(GA_ONLY_BUMPABLE));
+			bool childIsUnbumpable = childActor && childActor != caller && (actors_are_blocking || !childActor->ValidTarget(GA_ONLY_BUMPABLE));
 			if (childIsUnbumpable) continue;
-
-			PathMapFlags childBlockStatus = GetBlockedInRadius(nmptChild, size);
-			bool childBlocked = !(childBlockStatus & (PathMapFlags::PASSABLE | PathMapFlags::ACTOR | PathMapFlags::TRAVEL));
-			if (childBlocked) continue;
 
 			// Weighted heuristic. Finds sub-optimal paths but should be quite a bit faster
 			const float HEURISTIC_WEIGHT = 1.5;
@@ -377,7 +379,7 @@ PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, u
 			NavmapPoint nmptParent = parents[smptCurrent2.y * mapSize.w + smptCurrent2.x];
 			unsigned short oldDist = distFromStart[smptChild.y * mapSize.w + smptChild.x];
 			// Theta-star path if there is LOS
-			if (IsWalkableTo(nmptParent, nmptChild, flags & PF_ACTORS_ARE_BLOCKING, caller)) {
+			if (IsWalkableTo(nmptParent, nmptChild, actors_are_blocking, caller)) {
 				SearchmapPoint smptParent = Map::ConvertCoordToTile(nmptParent);
 				unsigned short newDist = distFromStart[smptParent.y * mapSize.w + smptParent.x] + Distance(smptParent, smptChild);
 				if (newDist < oldDist) {
@@ -385,7 +387,7 @@ PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, u
 					distFromStart[smptChild.y * mapSize.w + smptChild.x] = newDist;
 				}
 			// Fall back to A-star path
-			} else if (IsWalkableTo(nmptCurrent, nmptChild, flags & PF_ACTORS_ARE_BLOCKING, caller)) {
+			} else {
 				unsigned short newDist = distFromStart[smptCurrent2.y * mapSize.w + smptCurrent2.x] + Distance(smptCurrent2, smptChild);
 				if (newDist < oldDist) {
 					parents[smptChild.y * mapSize.w + smptChild.x] = nmptCurrent;
--- a/gemrb/core/PathFinder.cpp
+++ b/gemrb/core/PathFinder.cpp
@@ -293,6 +293,7 @@ PathListNode *Map::GetLine(const Point &p, int steps, orient_t orient) const
 // target (the goal must be in sight of the end, if PF_SIGHT is specified)
 PathListNode *Map::FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance, int flags, const Actor *caller) const
 {
+	bool lazyThetaStarFlag = true;
 	if (core->InDebugMode(ID_PATHFINDER)) Log(DEBUG, "FindPath", "s = {}, d = {}, caller = {}, dist = {}, size = {}", s, d, caller ? MBStringFromString(caller->GetShortName()) : "nullptr", minDistance, size);
 	bool actors_are_blocking = flags & PF_ACTORS_ARE_BLOCKING;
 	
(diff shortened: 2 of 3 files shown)


bradallred:
I actually noticed this when I was implementing Geometry.cpp in e1b2d6c which was intended to unify all our circles. I opted to match the pathfinder, since it was the one place where there could have been meaningful regressions. I thought I had opened an issue to question which flavor should ultimately be used, but it seems I did not. On Jun 28, 2023, at 15:58, Konrad Klimaszewski ***@***.***> wrote:﻿Description
Current implementation of circle drawing uses a "tweaked" version of the Bresenham algorithm: https://funloop.org/post/2021-03-15-bresenham-circle-drawing-algorithm.html. This leads no nicer circles but they tend to be slightly larger then requested radius. In particular for radius 1 we end up with a square where the corners extend quite a bit outside of the intended circle. By going back to the not tweaked version we end up with a cross that leads to reduced collisions.
Please take a look at #1825 to compare the GemRB, BG2 and improved behavior.
The fix amounts to this oneliner:
diff --git a/gemrb/core/Geometry.cpp b/gemrb/core/Geometry.cpp
index 927c8bfcc..7f0762427 100644
--- a/gemrb/core/Geometry.cpp
+++ b/gemrb/core/Geometry.cpp
@@ -158,7 +158,7 @@ std::vector<Point> PlotCircle(const Point& origin, uint16_t r, uint8_t octants)
        GenOctants(x, y);
        
        while (x < y) {
-               if (fm <= 0) {
+               if (fm < 0) {
                        fm += de;
                } else {
                        fm += dse;

Most of the PR tries to improve on the pathfinding by introducing a heuristic that measures if the calculated path is not to long. It works by returning the path length as an additional element of a tuple and comparing it to the euclidean distance of the start and stop points. It ignores short path to minimise bumping and very long paths that are more likely to be longer then the simple geometric distance.
Please take a look at #1825 for an example of improved algorithm at work.
Checklist

 Commit messages are descriptive and explain the rationale for changes
 I used the same coding style as the surrounding code 
 I have tested the proposed changes
 I extended the documentation, if necessary
 The proposed change builds also on our build bots (check after submission)




You can view, comment on, or merge this pull request online at:
  https://github.com/gemrb/gemrb/pull/1826

Commit Summary

  5bb902d  Sanitize party pathfinding in narrow spaces


File Changes  (4 files)

  
    M
    gemrb/core/Geometry.cpp
    (2)
  
  
    M
    gemrb/core/Map.h
    (4)
  
  
    M
    gemrb/core/PathFinder.cpp
    (22)
  
  
    M
    gemrb/core/Scriptable/Scriptable.cpp
    (19)
  


Patch Links:

  https://github.com/gemrb/gemrb/pull/1826.patch
  https://github.com/gemrb/gemrb/pull/1826.diff

lynxlynxlynx:
Thanks for the detailed report and fixes. I don't remember if anyone REd the actual shapes used by the originals (we still have a TODO in the code), but thinner is more pragmatic choice either way, so I have not problems with that.

Since only one caller cares about the distance, this tuple approach is quite verbosely ugly. I'd rather see another optional pointer param for anyone that wants to get the distance set.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -2181,11 +2183,20 @@ void Movable::WalkTo(const Point &Des, int distance)
 		return;
 	}
 
+	Point smp_pos = Point(Pos.x/16, Pos.y/12);
+	Point smp_des = Point(Des.x/16, Des.y/12);
+	int hdist = int(Distance(smp_pos, smp_des) * 1.3);
```
lynxlynxlynx, line 2188: You can just use: `Distance(ConvertCoordToTile(Pos), ConvertCoordToTile(Des))`

bradallred:
I think we need to do a bit of testing with the search map overlay and compare to the originals. Shrinking the circle is going to allow actors to stand quite close together IIRC.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -2181,11 +2183,20 @@ void Movable::WalkTo(const Point &Des, int distance)
...
 	if (BlocksSearchMap()) area->ClearSearchMapFor(this);
-	PathListNode* newPath = area->FindPath(Pos, Des, circleSize, distance, PF_SIGHT | PF_ACTORS_ARE_BLOCKING, actor);
-	if (!newPath && actor && actor->ValidTarget(GA_CAN_BUMP)) {
+	PathListNode * newPath;
+	int dist;
+	std::tie(newPath, dist) = area->FindPath(Pos, Des, circleSize, distance, PF_SIGHT | PF_ACTORS_ARE_BLOCKING, actor);
+	// Use a simple heuristic to identify unusually long paths by comparing to the shortest geometrical path -> try pathfinding without actors blocking.
+	// Following the two sides of an rectangle with 1:2 aspect ratio is around 1.3 times longer then the diagional.
+	// Ignore very short paths (below 20) as it would lead to excesive bumping.
+	// Ignore very long paths as they could be complicated and blocking by nearby actors will not matter anyway.
```
lynxlynxlynx, line 2196: Good test cases include the bg1 undercity maze and the ruins in ar3900. Both have narrow walls that can make you cry. And whole party movement in general, since it appears you've tested only with a single actor.

graag:
> Since only one caller cares about the distance, this tuple approach is quite verbosely ugly. I'd rather see another optional pointer param for anyone that wants to get the distance set.

I'm not fond of the tuple approach either. A Pointer param it is then :).

gemrb/core/Scriptable/Scriptable.cpp
lynxlynxlynx: The bg2 starting area is not really a maze-type test, since everything is spaced much better. Still, testing the pathing from the cage area to the Ellesime room is a good way to see how it behaves on long paths and how it responds to changes (eg. closing doors).

You can just teleport to the areas. In GemRB there's a shorthand mta("ar0000") and in the originals you have MoveToArea: https://gibberlings3.github.io/iesdp/appendices/clua/bg1.htm

graag:
> I think we need to do a bit of testing with the search map overlay and compare to the originals. Shrinking the circle is going to allow actors to stand quite close together IIRC. This _might_ not matter, if the pathfinding destination try to put them in the center of a search map square.

Is there anyway this can be automated? I mean by running a script to set the party target to the same position every time instead of trying to click in the same place. Bonus points if this would work in the originals ...

Anyway we could come up with few scenarios:
- moving actor around party members standing in a square (open and narrow space)
- moving party through a narrow corridor
- moving party through a narrow intersecting corridors (like in the area I tested in #1825)
- moving around actors with larger circles - I have no idea where to find them

Anything else?

bradallred:
I don't know that automation is required, the main question I have is if actors are directed to the center of a search path node, or if it is where the mouse pointer is. I know the paths obviously work on the search map, but I don't know if the final step just places them at the click position or not.

Activating the search map overlay would probably help demonstrate the difference between actors allowed proximity.

It might be helpful to know that you can manually setup arbitrary path finding tests using the [debug overlay painting](https://github.com/gemrb/gemrb/commit/2d9df680cb0787ba63f7164472ca65709c4f1380)

lynxlynxlynx:
Yes, that's doable via the ingame scripting. But it's easier to just set up a scenario in the game and share it between the original and gemrb via a save.

I've already mentioned some important test cases, some of the other pathfinder reports have nasty little areas and the kitchen and small loop in the Copper coronet are also a good tests (first for bumping neutrals and more crammed areas, the other for general heuristic balancing).

bradallred:
one thing to consider are the radii used in `Map.cpp`:
https://github.com/gemrb/gemrb/blob/7f26afda7e7b8be9fd800565c4de741e569a4981/gemrb/core/Map.cpp#L198-L201
and
https://github.com/gemrb/gemrb/blob/7f26afda7e7b8be9fd800565c4de741e569a4981/gemrb/core/Map.cpp#L2415-L2424

I haven't dug in, but it seems suspicious that one clamps with a minimum of 1 and the other 2.

I also couldn't find anything regarding why we calculate the radius the way we do, so who knows if that is even correct. Searchmaps like the one in #1015 seem like they could never work with a PC occupying more than one tile, even if we used diagonals.

lynxlynxlynx:
Anyway, the positions of actors near each other would perhaps not change with this, since they already can overlap. Just need to check the diagonals don't allow to bring them too close, since then even the selection circles could intersect.
image

graag:
In the original BG2 I'm not able to manually set a target that overlaps with the circle of another actor. The cursor changes shape to "move" but clicking does not do anything until the target does not overlap.

bg2_manual_target_1
bg2_manual_target_2

graag:
In gemrb with this PR it is possible to set the target manually to get an overlap.

gemrb_manual_target_1
gemrb_manual_target_2

graag:
I've implemented a simple AI script that moves the PC to a specified position.

With this I'm able to force the original into overlaping. I think it's comparable to what I see in this PR. Maybe a bit smaller. Below results from the original:

bg2_closest
bg2_closest_2
bg2_overlap

graag:
Results for the overlap for the gemrb

gemrb_closest
gemrb_closest_2
gemrb_closest_3

graag:
Most of the time both engines do a good job to prevent overlapping ..

BG2:
bg2_overlap_usual

GemRB:
gemrb_overlap_usual

graag:
However they handle differently bumping. BG2 ends up in continuous bumping when I force several PCs to go to a single point. While GemRB spitsout a lot of abandoned target messages.

bg2_bumping

Side note - the abandoned target code path seems to be very slow, there's noticable stutter when 4 PCs are forced by a script into impossible target. It is even visible when manually targeting a position that's not "walkable".

graag:
I'll see if I can make some sense from those radius calculations [bradallred](https://github.com/bradallred) mentioned. Maybe this will decrease the size of the maximal overlaps.

Last side note. The original does a better job at preserving the formation close to walls.
BG2:
bg2_party_target

GemRB:
gemrb_party_target

gemrb/core/Scriptable/Scriptable.cpp
graag: Done. I've also removed the tuple output in favor of the pointer param.

lynxlynxlynx:
Yes, we intentionally take the searchmap into account, so the reticles show closer to what the actual position would be.

gemrb/core/Map.h
```diff
@@ -627,8 +627,8 @@ class GEM_EXPORT Map : public Scriptable {
 	PathListNode* GetLine(const Point &start, int Steps, orient_t Orientation, int flags) const;
 	PathListNode* GetLine(const Point &start, const Point &dest, int speed, orient_t Orientation, int flags) const;
 	Path GetLinePath(const Point &start, const Point &dest, int speed, orient_t Orientation, int flags) const;
-	/* Finds the path which leads to near d */
-	PathListNode* FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance = 0, int flags = PF_SIGHT, const Actor *caller = NULL) const;
+	/* Finds the path which leads to near d. Return path_length if provided. */
+	PathListNode* FindPath(const Point &s, const Point &d, unsigned int size, unsigned int minDistance = 0, int flags = PF_SIGHT, const Actor *caller = NULL, int *path_length = NULL) const;
```
lynxlynxlynx, line 631:
```suggestion
	PathListNode* FindPath(const Point& s, const Point& d, unsigned int size, unsigned int minDistance = 0, int flags = PF_SIGHT, const Actor* caller = nullptr, int* pathLength = nullptr) const;
```
Please use nullptr with new code, same for camel case and symbol placement. We're slowly unifying the styles.

gemrb/core/PathFinder.cpp
```diff
@@ -82,7 +82,8 @@ PathListNode *Map::RunAway(const Point &s, const Point &d, unsigned int size, in
 	}
 	int flags = PF_SIGHT;
 	if (backAway) flags |= PF_BACKAWAY;
-	return FindPath(s, p, size, size, flags, caller);
+	PathListNode *newPath = FindPath(s, p, size, size, flags, caller);
+	return newPath;
```
lynxlynxlynx, line 86: Any particular reason this chunk wasn't just reverted?

bradallred:
> Results for the overlap for the gemrb
> 
> gemrb_closest gemrb_closest_2 gemrb_closest_3

the two actors in the middle image appear to have overlapping occupancy zones. Are they actually in motion in that image?

it wouldn't be a problem from this PR, but maybe we need a new issue.

lynxlynxlynx:
Turns out EEs also have debug overlays for search map debugging. And they draw personal space, which shows that we should keep the current circle algorithm for pathing purposes. It also allows overlaps the same way we do:
image

bradallred:
Wow, that's great! EE's have been of great help to us. Is there a resource to see what other debugging capabilities they have?

bradallred:
One thing I don't understand: if pathing doesn't care about personal space, and from that debug info from EE it appears that overlap is allowed... so what does personal space actually restrict? is it just for bumping? or does it only allow for partial overlap?

lynxlynxlynx:
Only partial overlap is allowed, yes, and EEs still suck at that compared to the originals and us (it still happens actors will walk through each other).

Besides the usual cheats (a few new) listed on IESDP, what were previously ini keys is now in a lua file. The closest to a list of options I could find is in a spoiler for this mod: https://forums.beamdog.com/discussion/64524/mod-reveal-hidden-gameplay-options-for-all-ee-games-patched-to-v2-0-or-higher

graag:
> And they draw personal space, which shows that we should keep the current circle algorithm for pathing purposes.

Sigh and it looked line such a nice fix ;-)

Rendering of search maps in EE is so cool:
bg2_ee_search_map

graag:
The problem in GemRB is that in this particular case pathfinding does not consider the touching corner of PC personal space and the wall as passable while it is true in the original.
gemrb_corner

We can position the character on the both sides it just cannot cross:
gemrb_search_map

lynxlynxlynx:
Yes, because we don't even consider diagonal moves in that case.

graag:
Probably not for this issue. GemRB does not allow to place a move target within actors personal space. This is not the case for EE - only the circles are considered. If the character will move is another issue - as I wrote I was not able to manually place a move target where circles would overlap.

GemRB:
gemrb_pointer

EE:
bd2_ee_pointer

graag:
> Yes, because we don't even consider diagonal moves in that case.

Right. So I've added diagonals - seems to work ...

What's the gotcha? Performance?

lynxlynxlynx:
Yes, we still have performance issues, so doing more calculations is not desirable without major optimizations, maybe even parallelization, since most cases don't actually benefit from also inspecting the diagonals. The discussion was in #1015.

bradallred:
There is probably some fairly low hanging fruit for improving the performance with pathfinding. We seem to completely recalculate paths instead of just splicing sub paths for things like bumping. There were a few other things we identified that I cannot recall.

MarcelHB:
I was thinking: are diagonals really a signficant problem w. r. t. CPU penalty? Under normal circumstances, a diagonal is the same as one step on the X axis, one more on the Y axis. The path finder will end up in a little different scheduling but not in any combinatorial explosion, just hitting this check more often: https://github.com/gemrb/gemrb/blob/99371e6258470bc999935545c96252d25067af5f/gemrb/core/PathFinder.cpp#L363-L364

bradallred:
I think, regardless of performance issues, it ought to be added as an option.

graag:
> Yes, we still have performance issues, so doing more calculations is not desirable without major optimizations, maybe even parallelization, since most cases don't actually benefit from also inspecting the diagonals. The discussion was in #1015.

In the worst case scenarios we will get 4 additional target is passable + line of sight tests for each node (if I understood correctly the current theta* implementation) so it would be on average 2x more costly. However for open spaces by allowing diagonal movement we should end up with less nodes so it could balance out. Of course this is even without any of the proposed optimizations.

Should be easy to add some counters and compare two scenarios. I'll try to run some benchmarks.

graag:
> > Results for the overlap for the gemrb
> > gemrb_closest gemrb_closest_2 gemrb_closest_3
> 
> the two actors in the middle image appear to have overlapping occupancy zones. Are they actually in motion in that image?
> 
> it wouldn't be a problem from this PR, but maybe we need a new issue.

I tried to take the screenshots when the actors stopped - so they should not be moving. The AI script kicked in at some point later and they changed positions.

Probably should be retested with the square personal zones as the consensus is that they should stay.

graag:
I've dumped the original code from this PR. Let's start from scratch :).

1.  6f598641920a29d548cb737b5a12cf4a05e96a81 - fixes a bug where `IsWalkableTo` will consider a LoS as walkable if it crosses a map node with TRAVEL flag, even if there are blocking Actors present
2. 042372aa2c462e905eae272abc137ced34ba8da4 - depending on the actor speed the `GetBlockedInLine` will consider also the start map node when determining blocking, in some cases the starting block will be in an actor personal zone and will be marked as blocked. If we are already at some start node we should be able to leave, I've added a simple check for now.
3. 355a77b8b7f6cc67fd0bc36ba7e59486a6ce3f81 - with two previous commits in place we can implement Theta star as in the original paper https://arxiv.org/pdf/1401.3843.pdf. For the A* fallback the line of sight check is not needed.
3. 927387f6e988777da8fd57b7154269cef178545b - now we are able to implement the Lazy Theta* :-)  http://idm-lab.org/bib/abstracts/papers/aaai10b.pdf

I run some performance benchmarks. A party of four walks from the cage room (BG2 starting area) to the Cambion room.
| counter | master    | theta* with fixes | lazy theta* |
| ------------| ----------- | --- | --- |
| open | 112277.75 | 97051.66 | 81018.66 | 
| neighbour | 247093 | 214120.33 | 179287 |
| walkable | 213278.5 | 176528 | 84029.66 |
| path found | 14.5 | 12 | 10 |

* open - number of times a node is taken from the `open` list
* neighbour - number of times a neighbour has to be considered (is not outside of the map and is not in the `closed` list)
* walkable - number of calls to `IsWalkableTo`
* path found - number of times a path is found and returned

graag:
I did not observe any regressions related to the first three commits. Nothing around travel points and I cannot trigger a case where `GetBlockingInRadius` would block something in addition to what the modified code does. I do not see anything wrong with the Lazy Theta* version - for now I left both version in the code so it would be easier to compare. For sure this would benefit from more extensive testing. 

As for the diagonal search for it is not ready yet. For some reason actors tend to get stuck especially around corners - will have to step this through debugger :cry:.

I think that the `IsWalkableTo` should work faster when done strictly in the search map space. Is there some reason it is done in map coordinates? At the moment depending on the actor speed some search map nodes are queried several times. I guess I'll try to propose an alternate version ...

lynxlynxlynx:
1 &2 are a nice find, I'll cherry-pick them in.

3. only has a small difference, but the change to checking just one block could be problematic. Make sure you're also testing with bigger circle sizes. You can summon bigger things and ctrl-q them to the party for permanence or teleport to an area with some giants or dragons first.

4. Previous attempts at lazy theta* caused some hard to debug problems, so in the end it was left out. Hopefully the new attempt won't run into the same issues.

IsWalkableTo: actually no, I can't think of a reason. But I'm also surprised the same block can be evaluated several times, sicne they're so small.

bradallred:
I can't find the discussion now, but I remember asking the same question about why we use map coordinates. I'm pretty sure there is no technical reason. It probably stems from the original pathfinding code we had and the fact that the actors `Pos`, _does_ need to be in map space. We need to detach that from pathfinding (other than updating it at the end of each step), and I think we are good to go.

lynxlynxlynx:
patch 1 is actually not good, at least not on its own — it breaks travelling between areas when you do it by area border. The worldmap never pops up.

graag:
> patch 1 is actually not good, at least not on its own — it breaks travelling between areas when you do it by area border. The worldmap never pops up.

:sweat: I wa sure I've tested it at some point when coding this ...

graag:
OK. The part 1 does not work because of this: https://github.com/gemrb/gemrb/blob/798e8e2b2b8865ad049f0a06896d758568d7fdb8/gemrb/core/Map.cpp#L2958

I'm not sure that this test is correct. Effectively it only tests for the travel flag and will disregard IMPASSABLE terrain, the IMPASSABLE_DOOR flag or other actors blocking. Do we expect maps that have both TRAVEL and some kind of IMPASSABLE flag set on a node? For the actors it would be a problem when party or an NPC is standing close enough to overlap personal spaces. So it would be enough to get the map flags and test for TRAVEL ...

The whole thing with the TRAVEL flag is some kind of an optimization: af82492
From this I assume there are places that do not have PASSABLE set but have TRAVEL. Is that righ??

graag:
> From this I assume there are places that do not have PASSABLE set but have TRAVEL. Is that right??

If that is the case and we always would like to treat TRAVEL as PASSABLE then we could revert af82492e45c4ce9f1244fac01d8bd6ea383801db and do something like this:

```
PathMapFlags Map::GetBlockedTile(const SearchmapPoint& p) const
{
	PathMapFlags ret = tileProps.QuerySearchMap(p);
	if (bool(ret & PathMapFlags::TRAVEL)) {
		ret |= PathMapFlags::PASSABLE;
	}
	if (bool(ret & (PathMapFlags::DOOR_IMPASSABLE|PathMapFlags::ACTOR))) {
		ret &= ~PathMapFlags::PASSABLE;
	}
	if (bool(ret & PathMapFlags::DOOR_OPAQUE)) {
		ret = PathMapFlags::SIDEWALL;
	}
	return ret;
}
```

Then we know if a node is passable or blocked by an actor / door by looking for the PASSABLE flag and still have access to TRAVEL flag.

lynxlynxlynx:
2958 is correct by definition. Travel is not a bit that's added to other searchmap terrain types, it's a type by itself.
But yeah, it can only be blocked by actors and hypothetically doors.

A related, but not strictly about pathfinding, issue is #1750.

lynxlynxlynx:
Ok, the release is done, so we can play a bit more loose now. :)

lynxlynxlynx:
Started looking into getting this further and things just got wilder and wilder. Wanted to test Brad's idea about prioritizing search in the direction of travel first, but then it looks almost like FindPath is not deterministic; I get significantly different loop counts for almost the same destination (will script it next). But investigating that then showed we sometimes try to move the target instead of the speaker when starting dialog, which is definitely a bug... :)

lynxlynxlynx:
The wrong move target thing is fixed and prioritizing lookup in the direction of the destination has no effect. Which makes sense in hindsight, since we have to check all directions to know which one is optimal.

I resolved the problems with patch 1 now (thanks!) and patch 2 has been in for a while, so we can focus on the last two. :) The main thing is to check we're still acting correctly, since there are several implicit cases.

lynxlynxlynx:
All good with the a* fallback without the walkability test. Not checking the full actor circle for blockage is less clear, I'll have to do more comparative testing. Walking around as a dragon is of course easier if we check one cell, but whether that's correct remains to be determined ...

Turns out it is wrong. We can move on this walkway in ar5500 (Amkhetran), while the original not:
image

Now most actors aren't that big and for them calling `GetBlockedInRadius` is just a roundabout way to call `GetBlocked`, but we can at least avoid some of that needless work.

And the last part of patch 3 should be fine. `GetActor` has to iterate overall all the map, checking the searchmap for each, so it's better to check just the one or few blocks first.

bradallred:
> lookup in the direction of the destination has no effect. Which makes sense in hindsight, since we have to check all directions to know which one is optimal.

No, you don't need to look in all directions. The shortest path between two points is always a line. If such a path exists you already know you don't need to explore other options. That's why pathfinding is generally implemented with backtracking. Typically you use backtracking to find the initial path, then run that path though an algorithm to cut the corners out.

lynxlynxlynx:
We don't do backtracking, but more importantly it would break (in the sense of not giving the best results) any-angle algorithms like theta* if we just went for a rounded guess.

lynxlynxlynx:
Ok, so the last patch got integrated. I left in the option to turn off the lazy algo at runtime, just in case we find some problem too late, but that switch should eventually get removed. Seems to work fine in my testing. 

I didn't do any performance tracking, but just comparing my debug build with 0.9.2 in release mode (appimage), the difference is noticeable. Specifically in pst city areas like outside the mortuary, where there are a lot of wandering actors: after loading there's no noticeable freeze any more.

Issue #1824, opened 2023-06-27
 IWD1 all saving throws for 1st level characters set to 20


gregariousmonk:
#### Bug description
All characters at the start of my most recent game have their saving throws set to 20 for all characters for all saves.

#### Steps to reproduce
Go to 'Record' --> scroll down the right-hand info box to see the saving throws.
All throws for 1st level characters are set to 20.  
The saving throw values appear to be correct after a character levels up.

#### Expected behavior
Saving throws for 1st level characters should be variable, dependent on race/class/ability.
Eg. https://baldursgate.fandom.com/wiki/Saving_throw#Save_progression_by_class_groups

#### Screenshots
![Screenshot from 2023-06-27 20-22-20](https://github.com/gemrb/gemrb/assets/137374871/06dc5957-702a-4407-9ea0-4bd9f9d8590e)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [x] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
They're still from the base creature file, since nothing calls the saving throw code. Does it fix it for you if you add a call to LUCommon.SetupSavingThrows (MyChar) in AcceptPress in CharGen.py?

gregariousmonk:
I've been running the buildbot appimages without altering the code (very often).  
I'll unpack the current one and try that.

gregariousmonk:
Took me a while due to a rash of typos and my editor being set to turn spaces into tabs, but it's working. 

![Screenshot from 2023-06-29 18-48-07](https://github.com/gemrb/gemrb/assets/137374871/071d2914-9b92-4a7c-a525-376ace9ba271)

lynxlynxlynx:
Cool, feel free to submit a PR.

PR #1830, opened 2023-06-29
Update CharGen.py

gregariousmonk:
Fix for #1824 

IWD1 characters do not have saving throws created at chargen.  

Added necessary code to provide the correct saving throws for 1st level characters

## Checklist

### DIFF
--- a/gemrb/GUIScripts/iwd/CharGen.py
+++ b/gemrb/GUIScripts/iwd/CharGen.py
@@ -29,6 +29,7 @@ import GUICommon
 import Spellbook
 import CommonTables
 import LUSkillsSelection
+import LUCommon
 
 CharGenWindow = 0
 CharGenState = 0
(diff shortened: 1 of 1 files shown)


gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -384,7 +385,8 @@ def AcceptPress():
 	GemRB.SetPlayerName (MyChar, GemRB.GetToken ("CHARNAME"), 0)
 	GemRB.SetToken ("CHARNAME","")
 	GemRB.SetPlayerStat (MyChar, IE_XP, CommonTables.ClassSkills.GetValue (ClassName, "STARTXP"))
-
+	LUCommon.SetupSavingThrows (MyChar, Level=None)
```
lynxlynxlynx, line 388: This actually runs? Either way, there's no need for the second parameter, since it has the same default.

gregariousmonk: I'll adjust the code.

Issue #1821, opened 2023-06-27
 IWD1 Prologue - Shattered Blade of Aihonen oddness


gregariousmonk:
#### Bug description
During the siren quest in Easthaven the Shattered Blade of Aihonen description panel has a button marked "Glimglam's Cloak" at the bottom.

#### Steps to reproduce
Do the siren quest in Easthaven and get the Shattered Blade, look at the description panel.

#### Expected behavior
No button marked Glimglam's Cloak.

#### Screenshots
![Screenshot from 2023-06-27 17-37-46](https://github.com/gemrb/gemrb/assets/137374871/4f243043-1800-42b5-b1d0-629c9c3b9d1d)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [x] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
What mods do you have installed?

gregariousmonk:
IWD fixpack and widescreen.

lynxlynxlynx:
It has the item type that's also used for "familiars", but of course doesn't have the string for it. IWD doesn't have familiars, so there are two ways of fixing it:
- just ifdefing out iwd (limiting) gemrb/GUIScripts/InventoryCommon.py:404
- detect familiars better, a few lines earlier. They are not droppable, so checking the IE_INV_ITEM_UNDROPPABLE bit would do

gregariousmonk:
Fixed but only tested here.

![Screenshot from 2023-06-30 00-24-19](https://github.com/gemrb/gemrb/assets/137374871/372a8f34-27f1-461d-a80e-6aefe44a6560)

![Screenshot from 2023-06-30 00-26-31](https://github.com/gemrb/gemrb/assets/137374871/fe86dd58-75ab-4cf5-abcd-48ed40c67f33)

PR #1831, opened 2023-06-30
Update InventoryCommon.py

gregariousmonk:
To address issue #1821

Added a check to DisplayItem to verify if a 'familiar' item is flagged undroppable.

## Checklist


lynxlynxlynx:
Thanks!

PR #1833, opened 2023-06-30
Addressing the Sonar complaint in Cache.h

MarcelHB:
## Description
See https://github.com/gemrb/gemrb/pull/1818#issuecomment-1613298342

## Checklist

### DIFF
--- a/gemrb/core/Cache.h
+++ b/gemrb/core/Cache.h
@@ -42,7 +42,7 @@ class RCCache {
 			V value;
 			int64_t refCount = 1;
 
-			explicit Value(V && value) : value(std::forward<V>(value)) {}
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks. I also noticed what's probably a Variables regression, but I'll have to dig deeper. In various early ToB saves Melissan just spawns and does her Saradush dialog and soon thereafter you're teleported to the pocket plane.

lynxlynxlynx:
Yeah, bisection leads to b4ab5812. I reread it and it looked fine until almost the end — I realized we're not storing and testing  ieVariables, but ResRefs in the maps, so anything longer gets cut off. 🤦🏾

Issue #1834, opened 2023-06-30
[IWD1+HOW GOG] no actor audio 


gregariousmonk:
#### Bug description
The radio button issue for character sounds is resolved, the radio buttons retain their state, but the characters and creatures make no selection/bored/battlecry sounds.

I've also noticed that if I set the Selection/Command sounds radio buttons, they only retain their settings until I exit to the loadscreen.  Upon reloading the game, those character sound settings reset to 'Never'.

The other in-game audio seems to be mostly working as expected (I still get the theme music dropping out on return to loadscreen sometimes).

#### Steps to reproduce
1. Load a game
2. Go to Options-->Sound-->Character Sounds-->set the sounds to "Always"
3. Click on a character or leave them waiting idle or enter battle.
4. No character audio is heard
5. Exit to the Loadscreen and select Load Game
6. Check the Character Sounds settings, they have returned to the "Never' position,

#### Expected behavior
Enabling Selection and Command sounds should create character sounds dependent on the current actions.  The sound settings should be retained from session to session.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Hmm, that's odd. I can reproduce, but when I was testing the fix for the radios, the values did get saved to the ini properly.

gregariousmonk:
I forgot to mention that I don't get any creature battlecries either, combat has only hit and missile sounds and there is no battle music.

gregariousmonk:
The soundsets for the game are present and work when played via the character customization or during chargen.  

They just aren't being played, or are not playing correctly, when characters are selected, given a command, nor battlecries and the banter-ish comments.

lynxlynxlynx:
This will be the same regression as noticed with scripting from the cache refactor. No need to test further.

lynxlynxlynx:
Yep, a different thing, but I found it already.

PR #1839, opened 2023-07-01
Variables: fix of map type

MarcelHB:
## Description
See https://github.com/gemrb/gemrb/pull/1833#issuecomment-1615828748

@lynxlynxlynx you also mention this https://github.com/gemrb/gemrb/issues/1834#issuecomment-1615855313 but I cannot see this fixed by this, so maybe can you check again or elaborate?

## Checklist

### DIFF
--- a/gemrb/includes/ie_types.h
+++ b/gemrb/includes/ie_types.h
@@ -33,6 +33,7 @@
 #include "Strings/StringView.h"
 
 #include <cstdint>
+#include <unordered_map>
 
 namespace GemRB {
 
@@ -107,6 +108,8 @@ enum class ieStrRef : ieDword {
 
 using ieVariable = FixedSizeString<32, strnicmp>;
 using ResRef = FixedSizeString<8, strnicmp>;
+
+using ieVarsMap = std::unordered_map<ieVariable, ieDword, CstrHashCI<ieVariable>>;
 
 template <typename STR>
 inline bool IsStar(const STR& str) {
--- a/gemrb/core/Game.h
+++ b/gemrb/core/Game.h
@@ -222,7 +222,7 @@ using CRRow = int[MAX_CRLEVEL];
 
 class GEM_EXPORT Game : public Scriptable {
 public:
-	using kaputz_t = ResRefMap<ieDword>;
+	using kaputz_t = ieVarsMap;
 
 	Game(void);
 	~Game(void) override;
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -2187,9 +2187,9 @@ Trigger *GenerateTriggerCore(const char *src, const char *str, int trIndex, int 
 
 void SetVariable(Scriptable* Sender, const StringParam& VarName, ieDword value, VarContext context)
 {
-	ResRef key{VarName};
+	ieVariable key{VarName};
 
-	auto SetLocalVariable = [=](ieVarsMap& vars, const ResRef& key, ieDword value) {
+	auto SetLocalVariable = [=](ieVarsMap& vars, const ieVariable& key, ieDword value) {
 		auto lookup = vars.find(key);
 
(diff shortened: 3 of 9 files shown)


lynxlynxlynx:
Ok, much better, but still an extra script block fires in the tob starting area, will have to check in more detail.

lynxlynxlynx:
Ah, you left the keys as ResRefs ...

lynxlynxlynx:
Changing that in CheckVariable and SetVariable fixes my test cases, but please take a systematic look.

MarcelHB:
So I have to make `locals` to use `StringParam` as key (64) since that is what is actually queried in some cases.

The uncomfortable thing: everything always just compiles ...

bradallred:
> So I have to make `locals` to use `StringParam` as key (64) since that is what is actually queried in some cases.
> 
> The uncomfortable thing: everything always just compiles ...

What are the cases? Just because scripting allows 64 characters doesn't mean all uses of strings are valid at that length. `StringParam` is often truncated to `ieVariable` or even `ResRef`.

lynxlynxlynx:
I don't really like it either, the variables dict should hold ieVariables.
You also missed GSUtils lines 2209 and 2265.

MarcelHB:
> > So I have to make `locals` to use `StringParam` as key (64) since that is what is actually queried in some cases.
> > The uncomfortable thing: everything always just compiles ...
> 
> What are the cases? Just because scripting allows 64 characters doesn't mean all uses of strings are valid at that length. `StringParam` is often truncated to `ieVariable` or even `ResRef`.

This `Action::string0Parameter` for example.

lynxlynxlynx:
It's a union, you can use variable0Parameter instead.

gemrb/includes/ie_types.h
```diff
@@ -108,6 +109,11 @@ enum class ieStrRef : ieDword {
 using ieVariable = FixedSizeString<32, strnicmp>;
 using ResRef = FixedSizeString<8, strnicmp>;
 
+using StringParam = FixedSizeString<64, strnicmp>;
+static_assert(std::is_standard_layout<StringParam>::value, "Fixed Size String must be standard layout for use in unions");
```
bradallred, line 113: will you just move this assert to the `union` declaration instead? it applies to everything in the union, I was just lazy and asserted for one because they are all `FixedSizeString`

MarcelHB:
> It's a union, you can use variable0Parameter instead.

Eh, I was wrong w. r. t. use case, I was looking at: https://github.com/gemrb/gemrb/blob/c16e19b9401c069a1ca6296db6526f2f3bb23a1c/gemrb/core/GameScript/GSUtils.cpp#L2188

bradallred:
yeah, I'd wager if we were using `string0Parameter` for an entry in `locals` that was just a mistake and we should use `variable0Parameter` instead. You can always add an `assert` on the length of `string0Parameter` before inserting to catch any possible attempts to use something larger.

lynxlynxlynx:
We weren't, the functions just accept a larger type because of occasional merged variable+context strings that need parsing first.

bradallred:
> We weren't, the functions just accept a larger type because of occasional merged variable+context strings that need parsing first.

@MarcelHB `SetVariable` should be one of these functions. The rest probably are too.

gemrb/includes/ie_types.h
MarcelHB: So you want this line back to `GameScript.h`, and also for both the other types (`ResRef`, `ieVariable`)?

bradallred: you don't need to add the others, they are all the same template and the size doesn't matter for standard layout.

MarcelHB:
Then I assume it's safe to just use `ieVariable` instead of `StringParam` as key? I will change this later today.

lynxlynxlynx:
Yep!

lynxlynxlynx:
Thanks!

PR #1840, opened 2023-07-01
Search map with a walk flag

graag:
## Description

I'll leave #1826 for diagonal pathfinding while here I'll consider a potentially less performance demanding optimization for party movement.

Originally  I considered to store in the search map information that an Actor is moving - this could be used by pathfinding to consider a node as passable as it should be freed in near future. Of course it would have some nasty corner cases like actors moving in opposite directions ;-). I dropped this idea because the search map flags are 8 bit and all possible values are already used.

However looking at EE I noticed that it looks like it stores such info in the search map:
https://github.com/gemrb/gemrb/assets/264425/b17da080-da85-4667-9816-49d5224ed36a

I guess that this is the reason it behaves better then GemRB master:
https://github.com/gemrb/gemrb/assets/264425/3064fa40-af2e-4be1-84cd-8b87f11eb401

So I returned to the idea of storing the info about walking actors in the search map. I tried to switch the map to `unit16_t` but that requires the palettes to be also 16bit for the debug to work, so not for a quick proof of concept ;-). Then I noticed that we actually only test for `PathMapFlags::ACTOR` and not separately for `PathMapFlags::PC` or `PathMapFlags::NPC`. I repurposed `PathMapFlags::PC` as an ACTOR and `PathMapFlags::NPC` as a "walking" flag. The result is extremely hacky, buggy and not optimized :-). But seems to work ...
https://github.com/gemrb/gemrb/assets/264425/731597e1-4b08-489a-a579-e87f2b54343d

Do you think it would be worthwhile to develop this further. If yes what would be the best way to store an additional flag in the search map ?

## Checklist


gemrb/core/GUI/GameControl.cpp
```diff
@@ -295,6 +295,7 @@ void GameControl::CreateMovement(Actor *actor, const Point &p, bool append, bool
 
 	actor->CommandActor(action, !append);
 	actor->Destination = p; // just to force target reticle drawing if paused
+	actor->Moving = true;
```
lynxlynxlynx, line 298: The actor could be held or otherwise immobile, so this is deceptive.

gemrb/core/Scriptable/Scriptable.h
```diff
@@ -247,6 +247,7 @@ class GEM_EXPORT Scriptable {
 	std::list< Action*> actionQueue;
 	Action* CurrentAction = nullptr;
 public:
+	bool Moving = false;
```
lynxlynxlynx, line 250: If anywhere, it should be in Movable.

gemrb/core/PathFinder.h
```diff
@@ -45,7 +45,8 @@ enum class PathMapFlags : uint8_t {
 	DOOR_IMPASSABLE = 32,
 	PC = 64,
 	NPC = 128,
-	ACTOR = (PC | NPC),
+	//ACTOR = (PC | NPC),
+	ACTOR = PC,
```
lynxlynxlynx, line 49: With the current implementation, where we only set the NPC bit and figure out bumpability elsewhere (since it's complicated), there's no need for such a bit. Just set ACTOR to 64 and rename NPC to whatever is sensible. And then remove the two places where it is set.

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -2114,7 +2116,11 @@ void Movable::DoStep(unsigned int walkScale, ieDword time) {
 			if (step->Next) {
 				step = step->Next;
 			} else {
+				// End??
 				ClearPath(true);
+				this->Moving = false;
+				area->ClearSearchMapFor(this);
+				area->BlockSearchMapFor(this);
```
lynxlynxlynx, line 2123: What's the point?

gemrb/core/Map.cpp
```diff
@@ -212,6 +212,44 @@ void TileProps::PaintSearchMap(const Point& Pos, uint16_t blocksize, const PathM
...
+		const Point& p2 = points[i + 1];
+		assert(p1.y == p2.y);
+		assert(p2.x <= p1.x);
+		
+		for (int x = p2.x; x <= p1.x; ++x) {
+			PaintIfPassable(Point(x, p1.y));
+		}
+	}
+}
+
```
lynxlynxlynx, line 252: This is too much code duplication, rather add another param.

gemrb/core/Map.cpp
```diff
@@ -909,8 +947,24 @@ void Map::DoStepForActor(Actor *actor, ieDword time) const
...
+void Map::TravelSearchMapFor(const Movable *actor) const
+{
+	auto flag = PathMapFlags::NPC;
 	tileProps.PaintSearchMap(ConvertCoordToTile(actor->Pos), actor->circleSize, flag);
+
+	std::vector<Actor *> nearActors = GetAllActorsInRadius(actor->Pos, GA_NO_SELF|GA_NO_DEAD|GA_NO_LOS|GA_NO_UNSCHEDULED, MAX_CIRCLE_SIZE*3, actor);
+	for (const Actor *neighbour : nearActors) {
+		if (neighbour->BlocksSearchMap()) {
+			if (!neighbour->Moving) {
```
lynxlynxlynx, line 963: I suggest you reverse their order and merge them.

Issue #1848, opened 2023-07-04
[iwd+how gog] Large shield not displaying on HF fighter paperdoll


gregariousmonk:
#### Bug description
The paperdoll for female human fighter has some kind of issue with the large shield.  Other race class combinations seem to display the shield in the paperdoll.  The shield appears correctly displayed in the game view. The medium and small shields are displayed in the pdoll.

There are frequent warnings and for missing plt/bam files, but I guess those are not critical.

#### Steps to reproduce
1. Open the inventory dialogue
2. Equip a large shield on a human female fighter|paladin|ranger
3. Observe the paperdoll and the missing shield

#### Expected behavior
The paperdoll should reflect the equiped shield

#### Screenshots
[cinnamon-2023-07-03T224434-0230.webm](https://github.com/gemrb/gemrb/assets/137374871/e2f05c8c-ddd2-4d4c-a5d3-d2cced92b7f0)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
> frequent warnings and for missing plt/bam
No, in this case this is very important.

Eg. when equipping on a paladin:
> [ResourceManager]: Searching for 'CHFF4INV'...
> [ResourceManager/WARNING]: Couldn't find 'CHFF4INV'... Tried CHFF4INV.plt  <-- then the bam is taken, as expected
> [ResourceManager]: Searching for 'WPNS1INV'...
> [ResourceManager/WARNING]: Couldn't find 'WPNS1INV'... Tried WPNS1INV.plt   <-- then the bam is taken, as expected
> [ResourceManager]: Searching for 'WPND4INV'...
> [ResourceManager/WARNING]: Couldn't find 'WPND4INV'... Tried WPND4INV.plt  (i only have wpnd1inv)
> [ResourceManager/ERROR]: Couldn't find 'WPND4INV.bam'.
> [GUISCript/WARNING]: BAM/PLT not found for ref: WPND4INV

Should be WPSD4INV. So I guess we misconstruct the filename. All three plain large shields have a correct item and animation type set (D4). I'd start looking if we miscategorize them in Inventory::UpdateShieldAnimation.

[...]

PR #1850, opened 2023-07-05
Update GUIINV.py

gregariousmonk:
## Description
Added code for GUIINV.py to remediate #1848 for shields to display correctly on HF avatars..  There are other BAMs that use the size N but I didn't find any references to them in the iwd+how game files.

## Checklist


lynxlynxlynx:
Are there other valid animations with N? IOW should just the pdoll entries be set to M instead?

gregariousmonk:
There are 33 WPN*.BAM animations but it looks as though they were intended to be used for displaying offhand weapons, which aren't used in iwd+how.  

I was cautious about editing the 2da file in case there were scripting dependencies that I'm not aware of, but changing it and doing limited testing gives the same result as the conditional in the script, ie. the shields display correctly on the pdoll.

You folks have a lot of experience with the code, which method do you think would be better?

lynxlynxlynx:
The size column of pdolls.2da is only ever used in GUIINV and since iwd has its own, it's completely safe to just fix the data that we unhardcoded. Please also do the same in the "iwd" copy.

gregariousmonk:
I'm going to close this pull request and open one without the code change file.

PR #1851, opened 2023-07-05
Edit pdolls.2da for how/iwd

gregariousmonk:
## Description
Remediating #1848 by editing the pdolls.2da files for iwd/how.  The size N listings have been changed to M.

This wrked in testing on my local installation of how.  I am unable to test with iwd only, but it should be the same.

## Checklist


lynxlynxlynx:
Thanks! Consider getting a git GUI or learning a bit more about it, so you won't have to do everything through the browser, which is also quite limiting.

gregariousmonk:
They say that you can learn a great deal from your mistakes.  I have really learned a lot in the last week, lol.

I'm trying to figure out which git client will work best for me, there are a few to choose from (I'm using linux mint).

Issue #274, opened 2019-10-03
[bug] Gerth doesn't buy keys


lynxlynxlynx:
Gerth in ar2100 should buy keys and misc items, he does not. In bg1 and iwd the critical item flag does not prevent selling as in bg2. Start the investigation by looking at the critical/destructible part of Store::AcceptableItemType and add a gameflag for the different behavior (ideally recheck if there is different behaviour first — in bg2, pst, iwd2).

lynxlynxlynx:
Cam says:

> Looking at the item flags at 0x18 for Icewind Dale:
> 
> https://gibberlings3.github.io/iesdp/file_formats/ie_formats/itm_v1.htm#Header_Flags
> 
> Like oBG, bit 0 doesn't affect whether an item can be sold. All items survive effects like disintegration and freezing/stone death irrespective of bit 0. Bit 0 items are handed over if a party member leaves (not that oIWD uses that anywhere) but not if the character is removed via arbitration.
> 
> Bit 11 (BG2's conversable flag) prevents selling, and this includes placement in a bag of holding.

PR #1860, opened 2023-07-12
Getting Gerth to buy keys/misc in iwd+how. (#274)

gregariousmonk:
## Description
First look at enabling Gerth's shop for misc and key items (#274).

Only tested on iwd+how.

[cinnamon-2023-07-11T222144-0230.webm](https://github.com/gemrb/gemrb/assets/137374871/99bfdf10-a8ca-43dd-99f7-82fb67a70b14)

## Checklist


lynxlynxlynx:
This is way too crude, enabling selling too often and ignoring the non-sellable bit.

gemrb/core/Store.cpp
```diff
@@ -135,6 +135,15 @@ int Store::AcceptableItemType(ieDword type, ieDword invflags, bool pc) const
 			ret &= ~IE_STORE_SELL;
 		}
 
+		//...the shop is special (iwd/how/bg1)
+		if ((core->HasFeature(GFFlags::NO_UNDROPPABLE)) || (core->HasFeature(GFFlags::BREAKABLE_WEAPONS))) {
+			if (invflags&IE_INV_ITEM_CONVERSABLE) {
+				ret &= ~IE_STORE_SELL;
+			} else {
+				ret |= IE_STORE_SELL;
```
lynxlynxlynx, line 143: this could potentially enabling the sale to stores that don't allow it.

lynxlynxlynx: also how should IE_INV_ITEM_DESTRUCTIBLE be handled in this case? Now you reenable their sellability.

gregariousmonk: I added more exact gametype detection.

I'm guessing that the major concern here is getting new/modded games, outside the intended ones, caught in the selling factors?

gregariousmonk (commented):
For issue #274

gregariousmonk (commented):
For issue #274

gemrb/core/Store.cpp
```diff
@@ -136,7 +136,7 @@ int Store::AcceptableItemType(ieDword type, ieDword invflags, bool pc) const
 		}
 
 		//...the shop is special (iwd/how/bg1)
-		if ((core->HasFeature(GFFlags::NO_UNDROPPABLE)) || (core->HasFeature(GFFlags::BREAKABLE_WEAPONS))) {
+		if ((core->config.GameType == "how") || (core->config.GameType == "bg1")) {
```
lynxlynxlynx, line 139: No, like I said in the issue, the proper way is to add another game flag.

gemrb/core/Store.cpp
lynxlynxlynx: The gametype checks are there to ensure this code is only used for the games that need it. My comments in this thread are about the code itself — look at the rest of the function, potential execution flow.

gemrb/core/Store.cpp
gregariousmonk: I've added a new game flag and performed a simple test.

[cinnamon-2023-07-13T164020-0230.webm](https://github.com/gemrb/gemrb/assets/137374871/a41e6ec0-c9e9-461d-b2a5-208a61dab0b4)

gemrb/unhardcoded/iwd/gemrb.ini
```diff
@@ -143,6 +143,7 @@ ReverseDoor = 0
 ReverseToHit = 1
 SaveForHalfDamage = 0
 SelectiveMagicRes = 0
+SellableCritsNoConv = 1
```
lynxlynxlynx, line 146: You need to add these also to the other gemrb.ini files, just with 0.

gregariousmonk: Thanks, Jaka, I've pushed the files I missed.

lynxlynxlynx: There's one more, for the demo, just elsewhere in the tree.

gregariousmonk: Fixed and pushed.

Did I miss anything else?

gregariousmonk: Wait, I screwed that one up, sorry.

lynxlynxlynx: The stuff in my first two comments is still not addressed, yeah.

gregariousmonk: I'll sort it out.  Still tracing through the code.

gregariousmonk: Should I make another game flag to enable the selling of items which originally have no bits set for the destructible flag?

lynxlynxlynx: Why, is behaviour different between the games?

gregariousmonk: I can't find where it's being set, but the "critical" items appear to also have the "destructible" bit set to false, the destructible flag doesn't exist in the original iwd/bg1.

If I add a condition for a !"destructible" flag after the condition for critical/not conversable it won't permit the sale.  If I don't add the condition it will permit the sale of indestructible items.

lynxlynxlynx: It's in Interface::SanitizeItem. But again, until the behaviour is proven to be different among games, a game flag makes no sense. There's already a check in this function, I'm just asking what the proper order of the two is, so we can match the originals.

gregariousmonk: Thanks for the leg up, Jaka. I'll go read and break some more code.

gregariousmonk: btw, I agree about the gameflag, that was just frustration talking, sorry.

I have a single test case working with modified interface.cpp code in IWD, I'll unpack bg2 and pst over the weekend and try them out.  

It's been too long since I played pst, too bad it's just for bughunting this time, but it'll be nice to see it again. The GemRB code runs really well, even runs smoothly on a crappy OLPC notebook I use.

gemrb/core/Store.cpp
```diff
@@ -125,27 +125,25 @@ int Store::AcceptableItemType(ieDword type, ieDword invflags, bool pc) const
 		ret &= ~IE_STORE_BUY;
 
 	if (pc && Type < StoreType::BG2CONT) {
-		//don't allow selling of non destructible items
-		if (!(invflags&IE_INV_ITEM_DESTRUCTIBLE )) {
-			ret &= ~IE_STORE_SELL;
-		}
```
lynxlynxlynx, line 131: Err, why did you remove this?

gemrb/core/Store.cpp
```diff
@@ -125,27 +125,25 @@ int Store::AcceptableItemType(ieDword type, ieDword invflags, bool pc) const
...
 
 		//don't allow selling of critical items (they could still be put in bags) ... unless the shop is special
-		if ((invflags&IE_INV_ITEM_CRITICAL) && !(Flags&IE_STORE_BUYCRITS)) {
+		if ((invflags&IE_INV_ITEM_CRITICAL) && (!(Flags&IE_STORE_BUYCRITS))) {
```
lynxlynxlynx, line 130: This doesn't change anything. Also I'm not sure why you're adding so many parentheses everywhere. Two pairs are enough in this line.

gemrb/core/Interface.cpp
```diff
@@ -3563,6 +3563,8 @@ void Interface::SanitizeItem(CREItem *item) const
 	// some slot flags might be affected by the item flags
 	if (!(item->Flags & IE_INV_ITEM_CRITICAL)) {
 		item->Flags |= IE_INV_ITEM_DESTRUCTIBLE;
+	} else if ((core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) && (!(item->Flags & IE_INV_ITEM_CONVERSABLE))) {
+		item->Flags |= IE_INV_ITEM_DESTRUCTIBLE;
```
lynxlynxlynx, line 3567: This is not correct. Both bits may be set, so if the critical one isn't, this else-if would get skipped.

gemrb/unhardcoded/iwd/gemrb.ini
lynxlynxlynx: No need to test pst stores, as they're not really different and hard to find. It's also almost pointless to test with gemrb, the main questions can only be answered by the originals.

gregariousmonk: Gorion's  Scroll (scrl3z) has the critical flag and can be sold to Bentley Mirrorshade in the original (non-EE) bg+totsc.

I'll continue testing...

gemrb/core/Interface.cpp
```diff
@@ -3563,10 +3563,13 @@ void Interface::SanitizeItem(CREItem *item) const
 	// some slot flags might be affected by the item flags
 	if (!(item->Flags & IE_INV_ITEM_CRITICAL)) {
 		item->Flags |= IE_INV_ITEM_DESTRUCTIBLE;
-	} else if ((core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) && (!(item->Flags & IE_INV_ITEM_CONVERSABLE))) {
+	} else if (core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) {
 		item->Flags |= IE_INV_ITEM_DESTRUCTIBLE;
-	}
-
+		}
+	if  (item->Flags & IE_INV_ITEM_CONVERSABLE) {
```
lynxlynxlynx, line 3569: This needs to happen only in iwd, unless you have some evidence items with the conversable bit shouldn't be destructible.

lynxlynxlynx: And to quote the expected behaviour:
> All items survive effects like disintegration and freezing/stone death irrespective of bit 0

IOW no items should have the bit set.

gregariousmonk: I couldn't find any items with bit 11 set in vanilla totsc, so iwd appears to be the only one using it despite the comment in the ITM docs.  I have yet to test/compare with bg2.

From the behaviour of the original games a 'critical' bit should permit selling in both bg1 and iwd/how, but the 'conversable' bit only appears to be used as unsellable in iwd/how.  

Should this be two separate flags as well or stick with one flag?

lynxlynxlynx: Isn't this still the same difference, already documented in the initial report? 

I can't think of any users of bit 11 outside of bg2 (lilarcor) off the top of my head. The chatty sword in iwd you can't interact with and I'm not sure if pst needs it, since it has a whole field for the dialog name.

gemrb/core/Interface.cpp
```diff
@@ -3566,7 +3566,7 @@ void Interface::SanitizeItem(CREItem *item) const
 	} else if (core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) {
 		item->Flags |= IE_INV_ITEM_DESTRUCTIBLE;
 		}
-	if  (item->Flags & IE_INV_ITEM_CONVERSABLE && core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) {
+	if  ((item->Flags & IE_INV_ITEM_CONVERSABLE) && core->HasFeature(GFFlags::SELLABLE_CRITS_NO_CONV)) {
```
lynxlynxlynx, line 3569: useless change: https://en.cppreference.com/w/cpp/language/operator_precedence

lynxlynxlynx:
Ok, took a fresh view:
- the demo is still missing the new key in its ini
- you're still enabling the sale for shops that don't support it (eg. temples), but
- with the changes to Interface.cpp, does Store.cpp even need changing at all?

gregariousmonk:
The change to the test gemrb.ini is in commit aba64f8, I'm not sure why it's not showing up here.

I'll look at the code for interface and store, but I believe you may be correct, the store code should not be needed anylonger since the filtering would happen in interface.  I'll check it this week.

lynxlynxlynx:
Hey, any progress on this?

lynxlynxlynx:
Do you plan to finish this or should we take over?

gregariousmonk:
Hi Jaka, I will finish it.  Apologies for the delay, I've had a large
amount of overtime at work for the past few months and am currently tending
to my wife in hospital in Korea.  I return home in a couple of weeks and
will look at it then.

lynxlynxlynx:
I've completed your work and pushed it; it will be available to everyone with the next release. Thanks!

PR #1866, opened 2023-07-18
Use libfmt for logging

bradallred:
Pretty much title. Just some QoL changes.

Summary:
- changed build option `NOCOLOR` -> `COLOR_LOG`, which is now defaulted to `true` everywhere.
- the generic "string like" `formatter` has been replaced with `format_as` and better restrictions
- we delegate formatting to `fmt::format`

## Checklist


platforms/apple/GemRB.xcodeproj/project.pbxproj
```diff
@@ -9108,7 +9108,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/bash;
-			shellScript = "echo \"$SRCROOT\"\ncmake \"$SRCROOT\"\n";
+			shellScript = "echo \"$SRCROOT\"\ncmake \"$SRCROOT\" -DNOCOLOR:BOOL=0\n";
```
lynxlynxlynx, line 9111: COLOR_LOG?

CMakeLists.txt
```diff
@@ -514,6 +514,10 @@ endif (STATIC_LINK)
 
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR} gemrb/includes gemrb/core platforms/${CMAKE_SYSTEM_NAME})
 
+if (NOT DEFINED COLOR_LOG)
```
lynxlynxlynx, line 517: I suggest you add NOCOLOR as well, for anyone requesting it from before. Another block to handle the transition.

gemrb/includes/win32def.h
```diff
@@ -32,7 +32,6 @@
 #define UNICODE
 #define _UNICODE
 #define NOGDI
-#define NOCOLOR
```
lynxlynxlynx, line 35: There are a few other NOCOLOR mentions in the tree that need to be changed.

CMakeLists.txt
bradallred: Done

gemrb/includes/win32def.h
bradallred: Any others?

CMakeLists.txt
lynxlynxlynx: Should be false though, since you inverted the meaning of the var.

platforms/apple/GemRB.xcodeproj/project.pbxproj
lynxlynxlynx: I suspect you want it inverted here as well.

CMakeLists.txt
bradallred: yeah, I spotted that as soon as I hit enter 🤦🏻‍♂️

lynxlynxlynx:
Ok, two more related things. Now windows is forced to use colors, which makes WIN32_USE_STDIO redundant. But also, there's no logging if you check the appveyor test run.

bradallred:
I don't have Windows to test, but it should be using `COLOR_LOG` like everything else. I'll look at `WIN32_USE_STDIO`, but Windows logger was derived from `StdioLogWriter` before this.

lynxlynxlynx:
What do you mean? We clearly used it to disable colored output; this is the bit this PR removes.

.appveyor.yml
```diff
@@ -83,7 +83,7 @@ before_build:
     if ($env:APPVEYOR_REPO_TAG_NAME -Or $env:APPVEYOR_REPO_COMMIT_MESSAGE -like '*buildgl*') {
       $env:gl = "OpenGL"
     }
-    cmake -G "Visual Studio 16 2019" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENGL_BACKEND="$env:gl" -DINSOURCEBUILD=1 -DWIN32_USE_STDIO=1 -DCMAKE_INSTALL_PREFIX="$env:archive_source" -DCMAKE_TOOLCHAIN_FILE=C:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET="x64-windows" -A x64 -DPython_ADDITIONAL_VERSIONS="3.10" .
+    cmake -G "Visual Studio 16 2019" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENGL_BACKEND="$env:gl" -DINSOURCEBUILD=1 -DCOLOR_LOG=true -DCMAKE_INSTALL_PREFIX="$env:archive_source" -DCMAKE_TOOLCHAIN_FILE=C:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET="x64-windows" -A x64 -DPython_ADDITIONAL_VERSIONS="3.10" .
```
lynxlynxlynx, line 86: Again, should be false.

lynxlynxlynx: You're right, it does. So it's surprising this disabling didn't work. Interesting! Maybe an ordering issue with the cmake files.

bradallred:
arg, the issue with windows is my copy pasta is `_O_RDONLY`

bradallred:
hmmm, that didn't seem to produce any output either

bradallred:
Windows console output is back, but uncolored... the color does work on master, which apparently is hardcoded to use color regardless of options.

bradallred:
@MarcelHB do you have any insight into why color doesn't work here, but does on master?

lynxlynxlynx:
Our ansi colors defined both the background and foreground colors. Maybe we need to be explicit with fmt colors as well? It has fg/bg https://hackingcpp.com/cpp/libs/fmt.html#terminal-styles

bradallred:
hmmmm, it does actually look like the `ghost_white` style is applied, just the styles for the inner text is missing. Does this work as expected on Linux?

MarcelHB:
Color works on Windows. It's just bold-inversion as well.

lynxlynxlynx:
Not on the bot, before:
https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/47575235
after:
https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/47577935

And only noticed just now — WARNING got shortened to WARN.

MarcelHB:
Hm, I also see a change when switching the palette of my terminal: It lacks the background but color is still there (it was black-on-white before), with some highlighting:
terminal

lynxlynxlynx:
Both the linux and windows bot now show one inconsistency: only the opening tag brackets are colored. It's more obvious on the windows one if you select the text.

bradallred:
yes, unfortunately the `fmt::styled` stuff doesn't work as intuitively as I had hoped. It appears each styled item completely resets the style when it is done outputting, rather than keeping a stack of formats inside a context. I suppose the answer is to just make multiple format calls.

lynxlynxlynx:
Now it's all bold, unfortunately.

bradallred:
some reading to consider:
https://github.com/actions/runner/issues/241
https://no-color.org

lynxlynxlynx:
Hmm, now neither build bot has color output.

platforms/windows/GemRB.cpp
```diff
@@ -20,15 +20,24 @@
...
 using namespace GemRB;
 
+Logger::WriterPtr createWin32ConsoleLogger()
+{
+	auto hConsole = GetStdHandle(STD_OUTPUT_HANDLE);
+	DWORD dwMode = 0;
+	GetConsoleMode(hConsole, &dwMode);
+	SetConsoleMode(hConsole, dwMode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+	return createStdioLogWriter();
+}
```
lynxlynxlynx, line 39: Are you sure the console mode will be reverted? It may matter only when ctrl-cing amid a print, but still.
https://learn.microsoft.com/en-us/windows/console/console-modes

bradallred:
> Hmm, now neither build bot has color output.

the linux one cant until the changes to `builder.yml` are merged.

> Are you sure the console mode will be reverted? It may matter only when ctrl-cing amid a print, but still.

I don't know how windows works, does windows even unwind the stack when you do whatever their equivalent of `SIGINT` is?

bradallred:
ok, I think this is about ready, tho perhaps you should test it on linux and ensure the `--color` option works as expected.

If there are no objections to eliminating the build time color option I'm going to squash those two commits before merging.

lynxlynxlynx:
I'll try it tomorrow. It's odd that the builder change is not in effect. One thing that's missing is an update to the manpage.

lynxlynxlynx:
--color doesn't have any effect, which explains why it's not showing on the builder either. Still looking into why.

Overall the look might have regressed on linux, but I can't say, since the bg is black in my main terminal, which I don't access to now. I propose we don't force a black background for the leading tags:
image

lynxlynxlynx:
LoadFromArgs has LogColor instead of LogColors key. Mode 1 looks the best here, but 2 is autodetected.

bradallred:
the style is something easily changed; I didn't bother trying anything but basic colors and retrieving the old colors completely depends on what the terminal palette was set to, so we just need to come up with whatever RGB values we want.

bradallred:
ok, that did fix colors on the linux builder. nice to know that it builds with the branch version of the config.

lynxlynxlynx:
One of the reasons for the NOCOLOR stuff was to support users with fancy terminal emulators using (semi)transparent backgrounds. So it'd be best to just not set a bg color in the tags.

bradallred:
I've removed backgrounds and squashed the `NO_COLOR` bits. We probably do want some better styles for the truecolor capable terminals, but we can adjust those later IMO

lynxlynxlynx:
Looking good!

Issue #1847, opened 2023-07-03
VFS: get rid of most fixed filename buffers 


lynxlynxlynx:
Both Sonar and now also Coverity keep warning about the dangers of potential overflows and while our _MAX_PATH is big enough for practical purposes, all the C char buffers now stand out even more. And it's not really friendly to non-ascii encodings anyway.

Some things may be better as-is due to the host APIs we plug into (unchecked), but at least all the path manipulation can be converted to String and that would remove most users of _MAX_PATH already. 

IIRC @bradallred also mentioned something to this effect somewhere, but it's good to have it tracked either way.

bradallred:
> path manipulation can be converted to String

Ultimately, the APIs for the filesystem operate on `char` buffers that are expected to be encoded in a particular encoding, which depends on OS. We can ofc use `std::string` instead of `char[_MAX_PATH]`. This is for *nix, no idea about other systems, so we might need a `Path` typedef to use the appropriate buffer for other platforms.

Nowadays, `std::filesystem::path` is the ideal way to do this in a cross platform way that handles the encoding and conversion between char widths internally. That isn't available until c++17, however.

Generally, only things that are displayed in the GUI need to be `String` (`wchar_t`), which will eventually mean a subset of UTF-16 (#1516). We've been steering everything else towards UTF-8 for simplicity and its general compatibility with things like ASCII.

MarcelHB:
For Windows (at least Win API) we need wide chars but that max-length test can be rewired. I can look at this for having both platforms.

bradallred:
ok PR #1857 should resolve this, with possibly the exception of Windows code I cannot fix.

bradallred:
@MarcelHB what needs to be done for Windows? That appears to be the only place left where we are using fixed size buffers. We can open a new issue for it if its not something we can quickly address.

MarcelHB:
I'll have a look at it. Just need to find the exact Sonar complaints.

bradallred:
im not sure they would complain about windows code. The offenders are in PluginLoader.cpp and VFS.cpp. I'm not sure they all need to change, but VFS might have a better way of working. Its also totally fine to say "goo enough" and close this :)

lynxlynxlynx:
Just git grep for _MAX_PATH; the only other user is in Vita-specific code. Sonar just complains that fixed-size buffers are prone to overruns and bugs/exploits like that.

PR #1867, opened 2023-07-18
Refactoring of some `_MAX_PATH` code sections

MarcelHB:
## Description
See #1847.

You still need buffers in some way for Win API but I guess Sonar and Coverty don't panic anymore.

## Checklist

### DIFF
--- a/gemrb/core/System/VFS.cpp
+++ b/gemrb/core/System/VFS.cpp
@@ -55,66 +55,68 @@
 
 #ifdef WIN32
 
-#include <tchar.h>
+#include <array>
 
 using namespace GemRB;
 
-struct DIR {
-	TCHAR path[_MAX_PATH];
-	bool is_first;
-	struct _finddata_t c_file;
-	intptr_t hFile;
-};
-
 struct dirent {
-	char d_name[_MAX_PATH];
+	dirent() : buffer(_MAX_PATH, L'\0'), d_name(const_cast<char*>(buffer.data())) {}
+
+	std::string buffer;
+	const char *d_name;
 };
 
+struct DIR {
+	DIR() : path(), is_first(true), hFile(0) {}
+	~DIR() {
+		_findclose(hFile);
+	}
 
-// buffer which readdir returns
-static dirent de;
+	std::wstring path;
+	bool is_first;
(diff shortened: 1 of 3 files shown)


platforms/vita/dirent.h
```diff
@@ -26,50 +26,54 @@
...
-inline dirent* readdir(DIR *dirp)
+inline dirent* readdir(DIR *dir)
 {
 	static dirent de;
 	//vitasdk kind of skips current directory entry..
-	if (dirp->is_first) {
-		dirp->is_first = 0;
+	if (dir->is_first) {
+		dir->is_first = false;
 		strncpy(de.d_name, ".", 1);
```
bradallred, line 62: shouldn't these `strncpy`s just be an assignment now?

MarcelHB: I did not want to touch that a lot (yet I noticed my renaming introduced a mistake ...).

Ok but I've found a ctor of std::string that ought to supersede the 2nd `strncpy`.

MarcelHB: Soft-tested my changes on godbolt

gemrb/core/System/VFS.cpp
```diff
@@ -55,66 +55,68 @@
...
+	// consider $PATH\*.*\0 for _wfindfirst
+	if (strlen(filename) > _MAX_PATH - 5) {
+		return nullptr;
+	}
 
-	mbstowcs(t_filename, filename, _MAX_PATH - 1);
-	StringCbPrintf(dirp->path, _MAX_PATH * sizeof(TCHAR), TEXT("%s%s*.*"), t_filename, TEXT("\\"));
+	std::wstring buffer(_MAX_PATH, L'\0');
+	mbstowcs(const_cast<wchar_t*>(buffer.data()), filename, buffer.length() - 1);
+	dir->path = std::move(fmt::format(L"{}\\*.*", buffer.c_str()));
```
bradallred, line 92: `std::move` is redundant here; function returns are already rvalues.

MarcelHB: Right

lynxlynxlynx:
Thanks!

PR #1868, opened 2023-07-19
GUIAnimation: try making msvc happy

lynxlynxlynx:
It's the only one not liking the inherited constructor, so checking if a template param will help.

### DIFF
--- a/gemrb/core/GUI/GUIAnimation.h
+++ b/gemrb/core/GUI/GUIAnimation.h
@@ -29,7 +29,7 @@
 namespace GemRB {
 
 template <class T>
-class GEM_EXPORT GUIAnimation {
+class GUIAnimation {
 protected:
 	tick_t begintime;
 	T current;
@@ -100,7 +100,7 @@ extern ColorCycle GlobalColorCycle;
 
 // This is supposed to be a fast optionally infinitely repeating transition
 // between 2 colors. We will create a global instance to syncronize many elements with the same animation.
-class GEM_EXPORT ColorAnimation : public GUIAnimation<Color> {
+class ColorAnimation : public GUIAnimation<Color> {
 public:
 	Color begin;
 	Color end;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Didn't help. @MarcelHB what version of MSVC do you have available? 19 fixed a related bug early on, but the builder already uses that.

Reading the specs and errata, I don't see anything wrong: https://en.cppreference.com/w/cpp/language/using_declaration

bradallred:
seems like its the same thing I encountered with `CstrCmp`. We have `GEM_EXPORT` on a template (`GUIAnimation`). try removing that?

lynxlynxlynx:
Alas that didn't help, but I guess it's a sensible change on its own.

bradallred:
I dont understand what MSVC is complaining about, but this seems like it ought to work:
```diff
diff --git a/gemrb/core/GUI/GUIAnimation.h b/gemrb/core/GUI/GUIAnimation.h
index ffa89675f..41c4fa248 100644
--- a/gemrb/core/GUI/GUIAnimation.h
+++ b/gemrb/core/GUI/GUIAnimation.h
@@ -109,7 +109,7 @@ public:
 	int timeOffset = 0;
 
 public:
-	using GUIAnimation::GUIAnimation;
+	ColorAnimation() = default;
 
 	ColorAnimation(const Color& begin, const Color& end, bool repeat)
 	: GUIAnimation(), begin(begin), end(end), repeat(repeat), cycle(7) {
```

lynxlynxlynx:
Sure, but that's pretty much what we had before.

MarcelHB:
I don't have MSVC under normal circumstances, I always play around with snippets on godbolt.org.

MarcelHB:
> Sure, but that's pretty much what we had before.

I can replay the issue, and that fixes it.

lynxlynxlynx:
Ok, I'll partially revert then. 🤷🏾‍♂️

Issue #340, opened 2019-10-11
[bug] Bad non-ASCII character handling in creature names


caillean7:
Split from #171 

**Description**
I play BG in Russian, so there's a lot of weiяd characters. I think they weren't displayed correctly in the console debug messages, so instead of PC names there were empty strings and I had to guess which is which when reading dumps.

caillean7:
Something I noticed recently was that I couldn't use umlauts in the name for my PC when creating the character. When typing a name containing umlauts, I can't confirm the name. Not sure if this is due to the English version of the game, but it would be nice if you could use umlauts, anyway.

lynxlynxlynx:
This is a wider issue. #47 tackled part of it, but both our current and subviews code has plenty of todos for full i18n support.

I doubt the English versions allowed umlauts though.

caillean7:
Probably not. They have the US keyboard layout on the original though, so I couldn't test that.

MarcelHB:
In PST, I frequently hear beeps on Windows, mostly after loading. Why that: Some bytes of the names are apparently in an encoding that mismatches what the Windows command line expects, thus printing BEL (0x7) to the console and making a sound.

caillean7:
@MarcelHB : Out of interest - I gather from your screenshots you're playing in German - could you use umlauts in character names, if you wanted?

MarcelHB:
It mostly works, yet there are incompatibilities between savegames when it comes to umlauts, and the inputs always appends garbage when typing such glyphs, i.e. I press `ü` and get `ü\xC3\xBC` - but I can delete the garbage and everything is good again.

caillean7:
Ah, so you can use them at least, if you don't care about save game compatibility. Again, would be cool if you could regardless of the language you play in. There are languages containing umlauts, like Finnish, and there's no game version available, so folks play in English, and it's nice if you can at least name your character however you want.

MarcelHB:
It's a not so trivial thing here, as in theory, these things must be considered:

1. The game has assets packed in some encoding, i.e. ISO-8859-1 afaik for German.
2. The operating system has some quirks to respect - and we're multiplatform here - for example file names. Windows has UTF-16 over WinAPI, Linux uses byte streams.
3. The user provides input in whatever encoding, like PC names, save game names, map notes.
4. The save games contain data of 1. and 3.  and let's say, you now open them on another native game or GemRB platform.

caillean7:
[...]

PR #1872, opened 2023-07-29
UTF-16 <> Python

MarcelHB:
## Description

So this fixes all � and error messages I'm aware of, mostly around the message window, shops and sound file paths.

Please have a look around.

## Checklist


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3185,7 +3185,7 @@ bool Actor::GetSavingThrow(ieDword type, int modifier, const Effect *fx)
 		if (core->HasFeedback(FT_COMBAT) && (lastSave.prevType != type || lastSave.prevRoll != ret)) {
 			// "Save Vs Death" in all games except pst: "Save Vs. Death:"
 			String msg = core->GetString(DisplayMessage::GetStringReference(HCStrings(ieDword(HCStrings::SaveSpell) + type)));
-			msg += L" " + fmt::to_wstring(ret);
+			msg += u" " + fmt::format(u"{}", ret);
```
bradallred, line 3188: there are probably a number of places where we can simply collapse the `+` into the `format` call. eg `msg += fmt::format(u" {}", ret)`.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7011,7 +7011,7 @@ void Actor::PerformAttack(ieDword gameTime)
 		} else {
 			hitMiss = core->GetString(DisplayMessage::GetStringReference(HCStrings::Miss));
 		}
-		String rollLog = fmt::format(L"{} {} {} {} = {} : {}", leftRight, roll, (rollMod >= 0) ? L"+" : L"-", abs(rollMod), roll + rollMod, hitMiss);
+		String rollLog = fmt::format(u"{} {} {} {} = {} : {}", leftRight, roll, (rollMod >= 0) ? u"+" : u"-", abs(rollMod), roll + rollMod, hitMiss);
```
bradallred, line 7014: you don't need to address this unless you want, but `format` can certainly take the burden of determining the sign.

gemrb/core/Strings/String.h
```diff
@@ -31,15 +31,15 @@
 #include "Format.h"
 #include <fmt/xchar.h>
 
-#define WHITESPACE_STRING_W L"\n\t\r "
+#define WHITESPACE_STRING_W u"\n\t\r "
```
bradallred, line 34: I'm debating on wether we should change the `W` suffix now. I guess its clear enough it doesn't matter assuming we no longer use `wchar_t` anywhere.

gemrb/core/Strings/StringConversion.cpp
```diff
@@ -26,77 +26,65 @@
 
 namespace GemRB {
 
+static bool IsBigEndian() {
```
bradallred, line 29: we already have this function somewhere... not able to check where ATM.

gemrb/core/Strings/StringConversion.cpp
```diff
@@ -26,77 +26,65 @@
...
-				ch = (string[++i] << 8) + currentChr;
-			}
-			dbString.push_back(ch);
-		} else {
-			dbString.push_back(currentChr);
-		}
-		++dbLen;
+	if (!string) return u"";
+
+	char* in = reinterpret_cast<char*>(const_cast<ieByte*>(string));
```
bradallred, line 40: if we change the incoming data, then we ought to change the signature. Is this just because of a bad `iconv` decision to not use `const`? If thats the case then this is fine.

gemrb/core/GUI/TextSystem/GemMarkup.cpp
```diff
@@ -164,9 +202,14 @@ GemMarkupParser::ParseMarkupStringIntoContainer(const String& text, TextContaine
 				}
 				break;
 			case INT:
-				if (*it == L']') {
+				if (*it == u']') {
+					std::wstring_convert<std::codecvt_utf16<wchar_t, 0x10ffff, std::little_endian>, wchar_t> conv;
+					std::wstring wToken = conv.from_bytes(
```
bradallred, line 207: seems like we should be able to do this without `std::wstring`, no?

MarcelHB: What's your idea? I did not found anything related to `std::u16string` yet.

gemrb/core/Scriptable/Actor.cpp
MarcelHB: Yeah, I'll look again.

gemrb/core/Strings/String.h
MarcelHB: Sure, why not.

gemrb/core/Scriptable/Actor.cpp
MarcelHB: I'll skip that 😉

gemrb/core/Strings/StringConversion.cpp
MarcelHB: It's inside of the `DataStream` but in there. Can look for a better position for both if you like.

gemrb/core/Strings/StringConversion.cpp
MarcelHB: Hm, looks like this can be dropped indeed, probably just left over of trying to understand what iconv takes and gives.

MarcelHB: Interesting, I was looking at the docs that say `const char**`: https://www.gnu.org/savannah-checkouts/gnu/libiconv/documentation/libiconv-1.17/iconv.3.html

But my lib complains on both Windows and Linux:
```
C:/MSYS2/mingw64/include/iconv.h:82:15: note: candidate function not viable: no known conversion
      from 'const char **' to 'char **' for 2nd argument
extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_...
```

MarcelHB: Great, some context and history: https://stackoverflow.com/questions/45938990/why-does-the-iconv-function-need-a-non-const-inbuffer

gemrb/core/Strings/String.h
MarcelHB: Wait, please check the nature of `WHITESPACE_STRING_VIEW` three lines below.

gemrb/plugins/GUIScript/PythonConversions.h
```diff
@@ -236,6 +236,7 @@ PyObject* PyString_FromResRef(const ResRef& resRef);
 PyObject* PyString_FromString(const char* s);
 PyObject* PyString_FromStringView(StringView sv);
 PyObject* PyString_FromStringObj(const std::string&);
+PyObject* PyString_FromSystemStringObj(const std::string&);
```
bradallred, line 239: what is `PyString_FromStringObj(const std::string&)` used for now? I can only think of `path_t` strings, but that would fall into "system encoding".

gemrb/core/Interface.h
```diff
@@ -732,7 +732,7 @@ extern GEM_EXPORT Interface * core;
 
 template<typename T>
 void SetTokenAsString(const ieVariable& key, T&& newValue) {
-	core->GetTokenDictionary()[key] = fmt::format(L"{}", std::forward<T>(newValue));
+	core->GetTokenDictionary()[key] = fmt::format(u"{}", std::forward<T>(newValue));
```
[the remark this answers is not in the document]
MarcelHB: No, this does not work, `to_string` won't deduce the type we need and the result fails because `std::string` vs `std::u16string`.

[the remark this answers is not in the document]
bradallred: ah, I didnt realize that `format` had a variable return type. Makes sense when you think about it.

gemrb/plugins/GUIScript/PythonConversions.h
MarcelHB: Apparent remaining users of this method deal with 2DA/IDS files.

bradallred: Then those shouldn't be using `return PyUnicode_Decode(s.c_str(), s.length(), core->TLKEncoding.encoding.c_str(), "strict");`, but probably just utf-8. In fact, nothing should be using TLK encoding in GUIScript.

We should probably just combine these and add a parameter for the encoding.

MarcelHB: For my German IWD2, there is a `CHARSTR.2DA` file with umlaute encoded in ISO-8559-1. So not sure what else is out there.

bradallred: I'm surprised to hear there is a 2da that uses localized text, but whatever decoding needs to happen should happen elsewhere (like the data importer) and TLK encoding is still probably only relevant to TLKs. I have to assume it's possible for TLK encoding to differ from that of random data files.

We can of course address this later.

gemrb/plugins/GUIScript/PythonConversions.cpp
```diff
@@ -137,15 +137,23 @@ PyObject* PyString_FromStringObj(const std::string& s)
...
-	std::string mbstr = MBStringFromString(s);
-	return PyString_FromStringObj(mbstr);
+	return PyUnicode_Decode(reinterpret_cast<const char*>(s.c_str()), s.length() * sizeof(char16_t), "UTF-16", "strict");
 }
 
 String PyString_AsStringObj(PyObject* obj)
 {
-	auto wrap = PyStringWrapper(obj, core->TLKEncoding.encoding.c_str());
-	return StringFromCString(wrap.CString());
+	auto utf16Object = PyUnicode_AsUTF16String(obj);
```
bradallred, line 150: seems like it would be cleaner to just take the native (utf-8) data and pass it thought our `StringFromUtf8` function.

MarcelHB: By native, you mean what Python uses internally? Isn't that somewhat platform/build/version-dependent?

bradallred:
no, not really, you have to go out of your way to find a Python 3 implementation that doesn't use utf8 internally.

bradallred:
This is probably of interest to us: https://peps.python.org/pep-0393/

gemrb/plugins/GUIScript/PythonConversions.h
lynxlynxlynx: Btw, that table is currently unused, so it may not even matter (I don't have it here to check and IESDP doesn't either). IDS files should all be fine.

gemrb/core/SaveGameIterator.h
```diff
@@ -29,7 +29,7 @@
 
 namespace GemRB {
 
-#define SAVEGAME_DIRECTORY_MATCHER "%d - %[A-Za-z0-9- _+*#%&|()=!?':;]"
+#define SAVEGAME_DIRECTORY_MATCHER "%d - %[^\n]"
```
lynxlynxlynx, line 32: They don't get included in the a-zA-Z ranges? Too bad sscanf doesn't support :alpha:, but we should block at least  / too. Also \\ to be safe.

lynxlynxlynx:
Does this also fix #340 ?

MarcelHB:
> Does this also fix #340 ?

No, I'll look what I can do.

MarcelHB:
> Does this also fix #340 ?

Ok, it does work without any further considerations, it just noticed it failing on MSYS terminal that has some apparent issues around encodings not related to here, but dumping on Linux looks good.

Another point to mention, just to avoid confusion: I don't want to address savegame things here w. r. t. user inputs and the saving and loading them in whatever encoding while keeping things compatible in both directions.

Now I'll just look at this Python PEP.

lynxlynxlynx:
So you're saying name input works fine, but you haven't checked if it's shown properly in-game? From memory the only other things that get saved are custom biographies and pst map notes (which IIRC we already handled before).

MarcelHB:
I see that it isn't shown properly, at least when starting a new game from GemRB, and then loading it.

bradallred:
yeah, we are relying on `wcstombs` which is going to behave differently everywhere. We should probably drop `MBStringFromString` and just write things with utf-8 or maybe the TLK encoding for compatibility with originals.

MarcelHB:
> So you're saying name input works fine, but you haven't checked if it's shown properly in-game? From memory the only other things that get saved are custom biographies and pst map notes (which IIRC we already handled before).

The most ugly thing is the voice set folder names in IWD2.

Other than that, folder names aside, is there any reason to bother with UTF-8 explicitly at all now? Should it work like whatever-it-is-TLK encoding or UTF-16? There is some users left for `MBStringFromString` and it looks like some of them could be replaced for good.

bradallred:
> is there any reason to bother with UTF-8 explicitly at all now?

SDL at least does things explicitly in UTF-8. There may be other reasons at the importer level; I imagine our data files are all utf-8. Thats why I was saying you cannot rely on the TLK encoding to determine what the data encoding is.

> There is some users left for MBStringFromString and it looks like some of them could be replaced for good.

Do it.

lynxlynxlynx:
All our data files are in ASCII. Checked with `find gemrb/override/ gemrb/unhardcoded/ | xargs file -i | grep text/plain | grep -v charset=us-ascii`, which returned nothing. We ship two UTF-8 licensing files in the demo, but we don't read those.

MarcelHB:
* We have to stick to UTF-8 for Python's `PyRun_String`.
* `stdout` isn't fully clear to me whether people agree more to sticking to UTF-8 for the best hope and compatibility, or whether to extract this from the env vars (I only see UTF-8 on my distro as well) and to do as required.
* And yes, SDL text events.

The `MBStringFromString`s related to save game packing is what I'll leave out for now.

bradallred:
I would assume you can/should use the `locale` for `stdout`. you can change the GemRB locale to whatever you need.

bradallred (commented):
I don't get it. What is `TextScreen.py` calling then?

bradallred:
right, but the cost of doing the conversion with python is greater since you must allocate a buffer, convert into that buffer, then copy the result into the string. If we already have a means of converting from any of the possible encodings to UTF-16 it would be less code and quicker to just toss the raw data plus encoding at it since it would be just allocate the string, then convert in place.

lynxlynxlynx:
The removal of TextArea_SetChapterText is indeed bad, since it's used. We define another calling variant on the python side and use that, hence it doesn't come up in a simple grep.

MarcelHB:
Damn, yes. I'll revert it and make it greppable.

bradallred:
BTW, `PyArg_ParseTuple` can be used for conversions. Particularly useful for strings that need to be in the filesystem encoding.

MarcelHB:
> BTW, `PyArg_ParseTuple` can be used for conversions. Particularly useful for strings that need to be in the filesystem encoding.

I'm only aware of `SetPlayerSound` that is `s` and this goes into an `ieVariable`, something multibyte, and that's Ok since it goes through VFS with its own conversion handling on Windows.

MarcelHB:
> right, but the cost of doing the conversion with python is greater since you must allocate a buffer, convert into that buffer, then copy the result into the string. If we already have a means of converting from any of the possible encodings to UTF-16 it would be less code and quicker to just toss the raw data plus encoding at it since it would be just allocate the string, then convert in place.

Then have a look again, this approach requires some more whens-and-ifs.

lynxlynxlynx:
Windows build error is odd, considering it doesn't have problems with the same header in other files:
> PythonConversions.cpp(19,10): fatal error C1083: Cannot open include file: 'iconv.h': No such file or directory

MarcelHB:
> Windows build error is odd, considering it doesn't have problems with the same header in other files:

Maybe it needs explicit reference in GUIScript's CMakeLists.txt, can't replay it locally but I'll try it out.

lynxlynxlynx:
Ok, it builds everywhere. @bradallred any other comments? If not, let's merge. :)

bradallred:
go ahead!

PR #1875, opened 2023-08-04
RFC: added preliminary .clang-format file

lynxlynxlynx:
Started working on #161. I've updated a previous draft to version 17 (just about to be released), but 16 is fine as well.

The rule file is in the commit, while the effect of `clang-format -i gemrb/**/*.[ch] gemrb/**/*.cpp` can be seen here: https://gist.github.com/lynxlynxlynx/87ca0bb6dec4633ef07d4b9a523078a6
As expected some things are not ideal (eg. when to inline functions), but in general it looks ok to me.

TODO:
- [x] .clang-format: set a max column width at all?
  - [x] move some of the comments, so they don't cause silly wrapping
- [x] exclude fmt
- [x] ~pass  --Wno-error=unknown so it doesn't choke if ran by older version~
- [x] test the difference between 16 (currently available runners) and 17 (current rules)

Infra stuff mentioned in the bug
  - [x] PRs: write runner
    - use https://github.com/marketplace/actions/c-c-linter disabling tidy and only checking changed lines
      - fallback: https://github.com/marketplace/actions/git-clang-format-check-per-commit
    - git clang-format, clang-format-diff are shipped with clang
  - [x] write a pre-commit hook
    - [x] have cmake install it
  - [x] add blame ignore file
    - [x] document use: git config --add blame.ignoreRevsFile some_ignore_file
    - [x] add to cmake target
    - [ ] add reformatting sha together with the reformatting commit
 
Docs:
  - [x] update CONTRIBUTING
  - [x] remove gemrb/docs/en/CodingStyle.txt
  - [ ] update website

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,176 @@
+---
+# last update based on clang 17 docs
+AccessModifierOffset: -8
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: false
+AlignConsecutiveMacros: true
+#AlignConsecutiveShortCaseStatements: false, 17+
+AlignEscapedNewlines: DontAlign
+AlignOperands: false
+AlignTrailingComments: Never
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: Empty
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortLambdasOnASingleLine: All
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: Yes
+AttributeMacros: ['__declspec', '__unused', '__attribute__']
+BinPackArguments: true
+BinPackParameters: true
+BitFieldColonSpacing: Both
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterCaseLabel: true
+  AfterClass: false
+  AfterControlStatement: Never
+  AfterEnum: false
+  AfterExternBlock: true
+  AfterFunction: true
+  AfterNamespace: false
+  AfterStruct: false
+  AfterUnion: false
+  BeforeCatch: false
+  BeforeElse: false
+  BeforeLambdaBody: false
+  BeforeWhile: false
+  IndentBraces: false
+  SplitEmptyFunction: false
+  SplitEmptyNamespace: false
+  SplitEmptyRecord: false
+BracedInitializerIndentWidth: 8
+BreakAfterAttributes: Leave
+BreakBeforeBinaryOperators: None
+BreakBeforeConceptDeclarations: Allowed
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: BeforeColon
+BreakConstructorInitializersBeforeComma: false
+BreakInheritanceList: AfterComma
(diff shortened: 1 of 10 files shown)


MarcelHB:
I scrolled through the diff and I like it :)

bradallred:
The only thing that has caught my eye is the lack of space when using braced initialization eg:
```diff
std::vector<Ambient *> ambients;
-	std::atomic_bool active {false};
+	std::vector<Ambient*> ambients;
+	std::atomic_bool active{false};
```

I would have an easier time reading if the identifier was separated from the initialization like it was before.

lynxlynxlynx:
Pretty sure that can be changed, yeah. I went with the currently prevailing style.

lynxlynxlynx:
Yeah, it's `SpaceBeforeCpp11BracedList`. It also applies to cases like these, but that's fine by me:
```diff
-          Log(DEBUG, "FindPath", "Pathing failed for {}", fmt::WideToChar{caller->GetShortName()});
+          Log(DEBUG, "FindPath", "Pathing failed for {}", fmt::WideToChar {caller->GetShortName()});
-  auto key = std::string{tablename.c_str()};
+  auto key = std::string {tablename.c_str()};
```

bradallred:
yeah, that looks good

lynxlynxlynx:
Ok, all seems to be working fine. The pre-commit hook gets installed and makes you comply before committing anything and if anyone circumvents  that the PR style checker will catch it.

Made it use annotations instead of spamming comments like sonar:
https://github.com/gemrb/gemrb/pull/1875/commits/d14014b338968193577bfc3eabe63e18f27f9327

lynxlynxlynx:
I've cleaned up the history, but I'll wait with the big reformat stuff while all the other PRs are open. You can cherry-pick [9fc535d](https://github.com/gemrb/gemrb/pull/1875/commits/9fc535dd17b74e24f0c638a071173ec47372fce8) and [a9eeeb7](https://github.com/gemrb/gemrb/pull/1875/commits/a9eeeb7b1ee19648117df05329c240b3d2ffdaf6) (and move to .git/hooks) if you want it sooner.

MarcelHB:
Given I have so many different projects, or at least memories of them, and their presumable conventions, I'd be happy to get this at hands.

lynxlynxlynx:
I've been using it for the past year. Sometimes it seems to suggest wrong things when just looking at the changed diff or pulls in too many other changes, but this problem will go away once we reformat (--no-verify can be used when needed until then).

I suggest we merge it after #1876.

lynxlynxlynx:
In the interim, it looked like the generated diff somehow got 5x smaller. But I just didn't have globstar enabled in the shell. In reality it's only 5k less. :)

lynxlynxlynx:
There will be some minor regressions (below), but a relatively quick review of the changes, after some tweaks, shows nothing really problematic. It also looks very stable between versions (tested 16-18). So we're going live. :)
```diff
diff --git a/gemrb/core/Game.cpp b/gemrb/core/Game.cpp
index dd186dbdc..f765f431b 100644
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -1992,7 +1992,7 @@ void Game::CastOnRest() const
                                std::sort(wholeparty.begin(), wholeparty.end());
                        }
                        ps = ps2;
-               // Gather rest of the spells
+                       // Gather rest of the spells
                } else if (specialSpell.flags & SpecialSpell::Rest) {
                        while (ps--) {
                                Actor* tar = GetPC(ps, true);
diff --git a/gemrb/GemRB.cpp b/gemrb/GemRB.cpp
index 41452e689..18ef19981 100644
--- a/gemrb/GemRB.cpp
+++ b/gemrb/GemRB.cpp
@@ -42,9 +42,9 @@ int main(int argc, char* argv[])
        // if the threshold is too low, free() starts to permanently ask the kernel
        // about shrinking the heap.
        #if defined(HAVE_UNISTD_H)
-               int pagesize = sysconf(_SC_PAGESIZE);
+       int pagesize = sysconf(_SC_PAGESIZE);
        #else
-               int pagesize = 4*1024;
+       int pagesize = 4 * 1024;
        #endif
        setenv("MALLOC_TRIM_THRESHOLD_", fmt::format("{}", 5 * pagesize), 1);
 #endif
```

Issue #1878, opened 2023-08-08
BG2: Segmentation fault entering Waukeen's Promenade


rsantellan:
#### Bug description
Traveling to Waukeen's Promenade gives me a Segmentation Fault. The stranger thing is that happens sometimes, I can' reproduce it 100% of the time.

```
=thread-group-added,id="i1"
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
Warning: Debuggee TargetArchitecture not detected, assuming x86_64.
=cmd-param-changed,param="pagination",value="off"
Stopped due to shared library event (no libraries added or removed)
Loaded '/lib64/ld-linux-x86-64.so.2'. Symbols loaded.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Breakpoint 1, main (argc=0, argv=0x5088) at ~/games/BaldurGate2/gemrb-myfork-git/gemrb/GemRB.cpp:32
32	{
Loaded '~/games/BaldurGate2/gemrb-myfork-git/build/gemrb/core/libgemrb_core.so.0.9.2-git'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libpthread.so.0'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libstdc++.so.6'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libm.so.6'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libc.so.6'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libgcc_s.so.1'. Symbols loaded.
Loaded '/lib/x86_64-linux-gnu/libdl.so.2'. Symbols loaded.
[...]

PR #1880, opened 2023-08-08
Check if killer is initialized.

rsantellan:
## Description
Fix for issue  #1878 create another check if killer is initialized.

## Checklist

### DIFF
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -5168,11 +5168,7 @@ void Actor::Die(Scriptable *killer, bool grantXP)
 		// TODO: is this right?
 		killer = area->GetActorByGlobalID(LastHitter);
 	}
-	
-	if (killer) {
-		killer->LastKilled = GetGlobalID();
-	}
-	
+	if (killer) killer->LastKilled = GetGlobalID();
 	Actor* act = Scriptable::As<Actor>(killer);
 	bool killerPC = false;
 	if (act) {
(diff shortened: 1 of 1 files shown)


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -5168,9 +5168,12 @@ void Actor::Die(Scriptable *killer, bool grantXP)
 		// TODO: is this right?
 		killer = area->GetActorByGlobalID(LastHitter);
 	}
-	killer->LastKilled = GetGlobalID();
-	Actor* act = Scriptable::As<Actor>(killer);
```
lynxlynxlynx, line 5172: This works if killer is null, so no need to split it up and make it conditional. Just the previous line is problematic.

rsantellan: So **area->GetActorByGlobalID(LastHitter);** should return always a correct killer if not it will always fail. Is that correct ?

lynxlynxlynx: I'm not sure what you're saying. What I'm saying is that `Actor* act = Scriptable::As<Actor>(killer);` is fine regardless of the value of killer.

lynxlynxlynx: Basically the fix should only change one line.

rsantellan: But Actor don't accept **nullptr**
```
matching function for call to ‘GemRB::Actor::As<GemRB::Actor>(std::nullptr_t)’
[build]  5171 |  Actor* act = Scriptable::As<Actor>(nullptr);
```
Will I not be moving the error ?

lynxlynxlynx:
Thanks!

Issue #1611, opened 2022-05-18
[bg1][bg2][p:t] Some ingame strings have a bad encoding when using Polish dialog.tlk


Emwues:
#### Bug description
While a lot of strings display fine, both with "Encoding=polish" cfg option set as well as without, some do not(again, regardless of encoding setting), mostly in GUI.

#### Steps to reproduce
1. Download [Polish tlks](https://github.com/gemrb/gemrb/files/8721217/dialogs.zip)
2. unzip
3. replace original dialog.tlk and dialogf.tlk(where applicable) with the Polish ones in GamePath
4. run GemRB

#### Expected behavior
GemRB should display strings correctly

#### Screenshots
P:T is an interesting one, game menu has both correctly("Zapisz grę" = save game) and incorrectly(should be "Wczytaj grę" = load game) encoded strings:
![busted-pt](https://user-images.githubusercontent.com/105691089/169165128-90cb909f-3024-4f78-9095-049b0817ea94.png)
classic for comparison:
![correct-pt](https://user-images.githubusercontent.com/105691089/169165419-2b3908f2-0fb4-4e84-a9bc-5b19bb77261d.png)
Record screen is almost unreadable:
![busted-pt2](https://user-images.githubusercontent.com/105691089/169165688-82933124-6bb8-4f02-b36b-98a7fe9b3c55.png)
classic:
![correct-pt2](https://user-images.githubusercontent.com/105691089/169165833-c2fe0693-f27e-48c2-aeb4-9b35d0f4527b.png)
Same in BG1:
![busted-bg](https://user-images.githubusercontent.com/105691089/169165981-15e74744-a640-4a10-8184-bbbaa3d7a9f1.png)
classic:
![correct-bg](https://user-images.githubusercontent.com/105691089/169166068-41519b29-6288-4947-98b7-a4464cd9f859.png)
and ToB:
![busted-tob2](https://user-images.githubusercontent.com/105691089/169166174-f95fb7b7-048a-4e65-9053-b4cecaaaefe4.png)

GemRB version (check as many as you know apply)
- [X] master as of this issue
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many
 as you know apply)
- [ ] SDL1.2
- [X] SDL2 built with `USE_OPENGL_BACKEND`
- [X] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
[...]

PR #1883, opened 2023-08-13
Savegames and encoding

MarcelHB:
## Description

Actually this was somewhat painful but it works. What did I do here:

* Make sound sets of IWD2 work with non-ASCII characters, so it's the first time I hear the PCs talking in savegames
* Throwing out some `mbstowcs` and some direct C I/O stuff by realizing it doesn't work this way on Windows for non-ASCII paths.
* Creating save games that work across the original and GemRB. This _may_ break compatibility for non-English savegames.
* By-fixing stuff that crashed or needed rework by now.

Interestingly I noticed how I was unable to type in non-ASCII chars in the originals so some things needed testing over hex-editing game data.

Some rationale over the worst-possible scenario by example: sound sets, as they are directories that may contain language-specific stuff:

* We need to get their name from the FS, over C API (multi byte) or Windows API (wide char)
* We probably want to emit logging here and there (multi byte)
* We need them displayed in Python (Python-encoding)
* We need them for sound playback among the resources (...)
* We need to store them in save games (TLK-encoding)

So my decisions were roughly like:
* Let path handling remain `path_t == std::string` and deal with Windows' `wchar_t*` bytes only at the very end.
* Handle all the game data strings (not `ResRef`, `ieVariable`) as `String == std::u16string` so let the importers deal with TLK encoding.
* Keep Python API out of automatically converting anything for us, `GUIScript` is supposed to feed explicit conversions into both directions.
* There is some interference between the concepts when it comes to some sound files that are game data, paths and resources all at once so I've put conversions there. This shouldn't be too hot of a code path.

## Checklist

### DIFF
--- a/gemrb/core/Audio.cpp
+++ b/gemrb/core/Audio.cpp
@@ -118,4 +118,11 @@ float Audio::GetReverb(unsigned int channel) const
 	return channels[channel].getReverb();
 }
 
+Holder<SoundHandle> Audio::Play(const String& resource, unsigned int channel, const Point& p, unsigned int flags, tick_t *length) {
+	auto mbString = MBStringFromString(resource);
+	auto mbResource = StringView{mbString.c_str()};
+
+	return Play(mbResource, channel, p, flags, length);
+}
+
 }
--- a/gemrb/core/Audio.cpp
+++ b/gemrb/core/Audio.cpp
@@ -120,7 +120,7 @@ float Audio::GetReverb(unsigned int channel) const
 
 Holder<SoundHandle> Audio::Play(const String& resource, unsigned int channel, const Point& p, unsigned int flags, tick_t *length) {
 	auto mbString = MBStringFromString(resource);
-	auto mbResource = StringView{mbString.c_str()};
+	auto mbResource = StringView{mbString};
 
 	return Play(mbResource, channel, p, flags, length);
 }
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -481,7 +481,7 @@ void DisplayStringCoreVC(Scriptable* Sender, size_t vc, int flags)
 		actor->GetVerbalConstantSound(soundRef, vc);
 		std::string sound;
 		if (actor->PCStats && actor->PCStats->SoundFolder[0]) {
-			sound = fmt::format("{}/{}", actor->PCStats->SoundFolder, soundRef);
+			sound = fmt::format("{}{}{}", fmt::WideToChar{actor->PCStats->SoundFolder}, PathDelimiter, soundRef);
 		} else {
 			sound = soundRef.c_str();
 		}
--- a/gemrb/core/KeyMap.cpp
+++ b/gemrb/core/KeyMap.cpp
@@ -122,7 +122,12 @@ bool KeyMap::ResolveKey(unsigned short key, int group) const
 {
 	// FIXME: key is 2 bytes, but we ignore one. Some non english keyboards won't like this.
 	char keystr[2] = {(char)key, 0};
-	Log(MESSAGE, "KeyMap", "Looking up key: {}({}) ", key, keystr);
+	if (key < 128) {
+		Log(MESSAGE, "KeyMap", "Looking up key: {} ({}) ", key, keystr);
+	} else {
+		// We'd have to convert the codepoint into UTF-8 for non-ASCII numbers
+		Log(MESSAGE, "KeyMap", "Looking up key: {}", key);
+	}
 
 	return ResolveName(keystr, group);
 }
--- a/gemrb/core/SaveGameIterator.cpp
+++ b/gemrb/core/SaveGameIterator.cpp
@@ -35,13 +35,13 @@
 #include "GUI/WindowManager.h"
 #include "Scriptable/Actor.h"
 #include "Streams/FileStream.h"
+#include "System/VFS.h"
 
 #include <cassert>
 #include <set>
 #include <ctime>
 
 #include <fmt/chrono.h>
-#include <sys/stat.h>
 
 #ifdef VITA
 #include <dirent.h>
@@ -121,8 +121,8 @@ static std::string ParseGameDate(DataStream *ds)
 	}
 }
 
-SaveGame::SaveGame(path_t path, path_t name, const ResRef& prefix, std::string slotname, int pCount, int saveID)
-: Path(std::move(path)), Name(std::move(name)), Prefix(prefix), SlotName(std::move(slotname))
+SaveGame::SaveGame(path_t path, const path_t& name, const ResRef& prefix, std::string slotname, int pCount, int saveID)
+: Path(std::move(path)), Prefix(prefix), SlotName(std::move(slotname))
 {
 	static const auto DATE_FMT = FMT_STRING("{:%a %Od %b %T %EY}");
 	PortraitCount = pCount;
@@ -136,7 +136,8 @@ SaveGame::SaveGame(path_t path, path_t name, const ResRef& prefix, std::string s
 	} else {
 		Date = fmt::format(DATE_FMT, fmt::localtime(my_stat.st_mtime));
 	}
-	manager.AddSource(Path, Name.c_str(), PLUGIN_RESOURCE_DIRECTORY);
+	manager.AddSource(Path, name, PLUGIN_RESOURCE_DIRECTORY);
+	Name = StringFromUtf8(name.c_str());
 }
 
 Holder<Sprite2D> SaveGame::GetPortrait(int index) const
@@ -245,7 +246,7 @@ static bool IsSaveGameSlot(const path_t& Path, const path_t& slotname)
 
 	path_t ftmp = PathJoinExt(dtmp, core->GameNameResRef, "bmp");
 
-	if (access(ftmp.c_str(), R_OK)) {
+	if (!FileExists(ftmp)) {
 		Log(WARNING, "SaveGameIterator", "Ignoring slot {} because of no appropriate preview!", dtmp);
 		return false;
 	}
@@ -254,13 +255,13 @@ static bool IsSaveGameSlot(const path_t& Path, const path_t& slotname)
 	if (core->HasFeature(GFFlags::HAS_EE_EFFECTS)) return true;
 
 	ftmp = PathJoinExt(dtmp, core->WorldMapName[0], "wmp");
-	if (access(ftmp.c_str(), R_OK)) {
+	if (!FileExists(ftmp)) {
 		return false;
 	}
 
 	if (core->WorldMapName[1]) {
 		ftmp = PathJoinExt(dtmp, core->WorldMapName[1], "wmp");
-		if (access(ftmp.c_str(), R_OK)) {
+		if (!FileExists(ftmp)) {
 			Log(WARNING, "SaveGameIterator", "Ignoring slot {} because of no appropriate second worldmap!", dtmp);
 			return false;
 		}
@@ -299,7 +300,10 @@ bool SaveGameIterator::RescanSaveGames()
 	} while (++dir);
 
 	for (const auto& slot : slots) {
-		save_slots.push_back(BuildSaveGame(slot));
+		auto saveGame = BuildSaveGame(slot);
+		if (saveGame) {
+			save_slots.push_back(saveGame);
+		}
 	}
 
 	return true;
@@ -312,12 +316,12 @@ const std::vector<Holder<SaveGame> >& SaveGameIterator::GetSaveGames()
 	return save_slots;
 }
 
-Holder<SaveGame> SaveGameIterator::GetSaveGame(StringView name)
+Holder<SaveGame> SaveGameIterator::GetSaveGame(const String& name)
 {
 	RescanSaveGames();
 
 	for (const auto& saveSlot : save_slots) {
-		if (saveSlot->GetName().compare(name.c_str()) == 0)
+		if (saveSlot->GetName() == name)
 			return saveSlot;
 	}
 	return NULL;
@@ -630,6 +634,11 @@ int SaveGameIterator::CreateSaveGame(int index, bool mqs) const
 		displaymsg->DisplayMsgCentered(HCStrings::SaveSuccess, FT_ANY, GUIColors::XPCHANGE);
 	}
 	return GEM_OK;
+}
+
+int SaveGameIterator::CreateSaveGame(Holder<SaveGame> save, const String& slotname, bool force) const {
+	auto mbSlotName = MBStringFromString(slotname);
+	return CreateSaveGame(save, StringView{mbSlotName}, force);
 }
 
 int SaveGameIterator::CreateSaveGame(Holder<SaveGame> save, StringView slotname, bool force) const
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -8606,7 +8606,7 @@ String Actor::GetSoundFolder(int full, const ResRef& overrideSet) const
 	String soundset;
 	if (core->HasFeature(GFFlags::SOUNDFOLDERS)) {
 		if (full) {
-			soundset = fmt::format(u"{}/{}", PCStats->SoundFolder, wSet);
+			soundset = fmt::format(u"{}{}{}", PCStats->SoundFolder, PathDelimiterW, wSet);
 		} else {
 			soundset = fmt::format(u"{}", PCStats->SoundFolder);
 		}
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -3463,8 +3463,8 @@ bool Actor::VerbalConstant(int start, int count, int flags) const
 		do {
 			count--;
 			GetVerbalConstantSound(soundRef, start + count);
-			std::string chrsound = GetSoundFolder(1, soundRef);
-			if (gamedata->Exists(chrsound, IE_WAV_CLASS_ID, true) || gamedata->Exists(chrsound, IE_OGG_CLASS_ID, true)) {
+			auto soundFolder = GetSoundFolder(1, soundRef);
+			if (gamedata->Exists(soundFolder, IE_WAV_CLASS_ID, true) || gamedata->Exists(soundFolder, IE_OGG_CLASS_ID, true)) {
 				DisplayStringCoreVC((Scriptable *) this, start + RAND(0, count), flags|DS_CONST);
 				found = true;
 				break;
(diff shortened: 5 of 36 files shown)


gemrb/core/Streams/FileStream.h
```diff
@@ -98,6 +99,93 @@ struct File {
...
+	}
+
+	bool _SetFilePointer(stroff_t offset, DWORD method) {
+		LARGE_INTEGER liOffset;
+		liOffset.QuadPart = offset;
+
+		return SetFilePointerEx(file, liOffset, nullptr, method) != 0;
+	}
+};
+#endif
```
MarcelHB, line 188: Was there any good reason to have this `File` thing completely in the header? For now I ifdef'ed that into the Windows sction but that also increases the amount of compiler noise and time.

gemrb/core/Streams/DataStream.h
```diff
@@ -153,7 +153,7 @@ class GEM_EXPORT DataStream {
 	
 	template <typename STR>
 	strret_t WriteString(const STR& src, size_t len) {
-		return Write(src.begin(), len);
+		return Write(src.c_str(), len);
```
MarcelHB, line 156: Here, at some point the compiler started complaining of how `begin()` won't go into `void*`.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8600,16 +8600,18 @@ std::string Actor::GetSoundFolder(int full, const ResRef& overrideSet) const
 		set = overrideSet;
 	}
 
-	std::string soundset;
+	String wSet = StringFromResRef(set);
+	String soundset;
 	if (core->HasFeature(GFFlags::SOUNDFOLDERS)) {
 		if (full) {
-			soundset = fmt::format("{}/{}", PCStats->SoundFolder, set);
+			soundset = fmt::format(u"{}{}{}", PCStats->SoundFolder, PathDelimiterW, wSet);
```
MarcelHB, line 8607: Here I wasn't too sure if we want to work with resource or paths.

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -481,7 +481,7 @@ void DisplayStringCoreVC(Scriptable* Sender, size_t vc, int flags)
 		actor->GetVerbalConstantSound(soundRef, vc);
 		std::string sound;
 		if (actor->PCStats && actor->PCStats->SoundFolder[0]) {
-			sound = fmt::format("{}/{}", actor->PCStats->SoundFolder, soundRef);
+			sound = fmt::format("{}/{}", fmt::WideToChar{actor->PCStats->SoundFolder}, soundRef);
```
MarcelHB, line 484: Note to self: double-check (see similar thing below).

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -3779,7 +3779,7 @@ static PyObject* GemRB_VerbalConstant(PyObject * /*self*/, PyObject* args)
 	}
 
 	//get soundset based string constant
-	std::string sound = fmt::format("{}/{}{:02d}", actor->PCStats->SoundFolder, actor->PCStats->SoundSet, str);
+	std::string sound = fmt::format("{}/{}{:02d}", fmt::WideToChar{actor->PCStats->SoundFolder}, actor->PCStats->SoundSet, str);
```
MarcelHB, line 3782: Also need to check this again.

gemrb/core/Strings/StringConversion.cpp
```diff
@@ -123,16 +129,37 @@ std::string MBStringFromString(const String& string)
...
 	}
 
-	return buffer;
+	return newString;
+}
+
+std::string TLKStringFromString(const String& string)
+{
+	// TODO: if any TLK Encoding is actually wide, this may need further considerations
+	// w. r. t. the type or cut-off
```
lynxlynxlynx, line 149: The CJK ones are multibyte. https://gemrb.org/Text-encodings.html#korean

gemrb/core/SaveGameIterator.cpp
```diff
@@ -254,13 +255,13 @@ static bool IsSaveGameSlot(const path_t& Path, const path_t& slotname)
...
 	ftmp = PathJoinExt(dtmp, core->WorldMapName[0], "wmp");
-	if (access(ftmp.c_str(), R_OK)) {
+	if (!FileExists(ftmp)) {
 		return false;
 	}
 
 	if (core->WorldMapName[1]) {
 		ftmp = PathJoinExt(dtmp, core->WorldMapName[1], "wmp");
-		if (access(ftmp.c_str(), R_OK)) {
+		if (FileExists(ftmp)) {
```
lynxlynxlynx, line 264: Should be negated.

gemrb/core/Streams/FileStream.h
lynxlynxlynx: I doubt it and yeah, let's change it. I suggest we split out a File.h/File.cpp and add the Windows versions under platforms/. If there are many places that need it, maybe also a shell header, so the ifdefing doesn't have to be repeated too much.

MarcelHB:
So there is one thing left that is related to both umlaute and VFS case sensitivity, so essentially the best of portability:

The original IWD2 saves its soundset uppercased, apparently. Unfortunately https://github.com/gemrb/gemrb/blob/441408df5d4dfc891ad73f3881ada2cb6c7a5688/gemrb/core/System/VFS.cpp#L222 does not care about non-ASCII upcase and downcase differences, they are not considered and thus GemRB won't find some of the sound set folders.

We'd need something that (in case of German) is aware of `ä` <> `Ä` and `ö` <> `Ö` (there are more chars, but they do not appear in IWD2, like `ü` or `ß`).

Also, other languages _may_ suffer from similar stuff (GOG says there is Russian available) although I can't tell exactly to what extent. The full theory of case mappings over Unicode codepoints is [here](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/unidata/CaseFolding.txt).

What do we want to do here? I'd say that if the string contains a char >= 128, we may need to switch to something more advanced than `strcasecmp`. We can maintain a minor mapping file with actual cases, starting with German and waiting for whoever complains for other languages? I guess that something like ICU lib tends to be a bit too much for that purpose.

lynxlynxlynx:
We had these mappings in the language ini files before. They were for the bam fonts, so I guess with different values. Removed here:
https://github.com/gemrb/gemrb/commit/daf149eab80ba88723f393a0cbd347881776ff3f

bradallred:
You need to set a locale to get locale aware behavior; default is just the "C" locale. After, you use stuff like `std::strcoll`. `std::tolower` is also locale aware, hence why we don't need those mappings.

lynxlynxlynx:
We don't use the default, but immediately query the current locale, which I guess is de_DE.UTF-8 for him. So we just need to switch to std::strcoll or std::strxfrm there.

lynxlynxlynx:
Hmm, so your VFS encoding is UTF-8, while the TLK one is ISO-8859-1. Setting LC_COLLATE=C.UTF-8 might fix it for all cases, since we don't actually care for the collation to be correct, just that all the characters can be tested.

MarcelHB:
Yes. I'll give it a try.

bradallred:
> I'd like to skip any locale-dependent stuff since my Linux setup is English (only with appropriate keyboard settings) and my games may be German or English.

POSIX locals work on a per program basis. We just need to set one from a config. it wont change your system locale.

MarcelHB:
Sure, but is that free of side effects (dates, formatting, ...) for us, or do we better set and revert this setting around the respective calls, just to be safe?

lynxlynxlynx:
Just setting LC_COLLATE won't affect LC_TIME and others, only sorting (which we do on the py side for custom sounds and scripts).

bradallred:
you need to use the [correct](https://en.cppreference.com/w/cpp/locale/tolower) `tolower`

MarcelHB:
I've made it working. You cannot get around `wchar_t` from what I've tried but hey, also no messing around with env vars and configs.

gemrb/core/System/VFS.cpp
```diff
@@ -256,11 +258,95 @@ void PathAppend(path_t& target, const path_t& name)
...
+		// ASCII range
+		if (iA < 128) {
+			if (std::tolower(iA) != std::tolower(iB)) {
+				return false;
+			}
+
+			itA++, itB++;
+		} else {
+			// multibyte range (assuming no broken encoding)
+			auto getWidth = [](unsigned char c) -> uint8_t {
```
bradallred, line 290: Is this not exactly what `std::mbrlen` does?

gemrb/core/System/VFS.cpp
```diff
@@ -256,11 +258,95 @@ void PathAppend(path_t& target, const path_t& name)
...
+	}
+
+	// equal
+	return true;
+}
+
 static bool FindMatchInDir(const char* dir, char* item)
 {
+	bool expensiveCheck =
+		std::any_of(item, item + strlen(item) + 1, [](char c) { return c < 0; });
```
bradallred, line 342: if you change the parameter type then we can avoid `strlen`

lynxlynxlynx:
I suggest you move strlicmp to its own header file in case it becomes useful elsewhere later and since it doesn't really have to do anything with the VFS.

gemrb/core/System/VFS.cpp
MarcelHB: Seems to depend on `LC_ALL`, so for UTF-8 case I consider this easy enough.

gemrb/core/System/VFS.cpp
```diff
@@ -256,11 +258,95 @@ void PathAppend(path_t& target, const path_t& name)
...
+
+		// ASCII range
+		if (iA < 128) {
+			if (std::tolower(iA) != std::tolower(iB)) {
+				return false;
+			}
+
+			itA++, itB++;
+		} else {
+			// multibyte range (assuming no broken encoding)
```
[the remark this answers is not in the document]
MarcelHB: Not sure about non-UTF8 encodings but for that, the 128-rule is safe. And as I can't speak for platforms other than Linux (virtually UTF-8 nowadays) and Windows (UTF-16) that are both addressed by this: wait and see?

gemrb/core/Audio.cpp
```diff
@@ -118,4 +118,11 @@ float Audio::GetReverb(unsigned int channel) const
 	return channels[channel].getReverb();
 }
 
+Holder<SoundHandle> Audio::Play(const String& resource, unsigned int channel, const Point& p, unsigned int flags, tick_t *length) {
+	auto mbString = MBStringFromString(resource);
+	auto mbResource = StringView{mbString.c_str()};
```
bradallred, line 123: why does `StringView(mbString)` not work?

MarcelHB: Hm, don't remember any more, will check again.

gemrb/core/KeyMap.cpp
```diff
@@ -120,9 +120,13 @@ bool KeyMap::InitializeKeyMap(const path_t& inifile, const ResRef& tablefile)
 //main gamecontrol
 bool KeyMap::ResolveKey(unsigned short key, int group) const
 {
-	// FIXME: key is 2 bytes, but we ignore one. Some non english keyboards won't like this.
 	char keystr[2] = {(char)key, 0};
-	Log(MESSAGE, "KeyMap", "Looking up key: {}({}) ", key, keystr);
+	if (key < 128) {
```
bradallred, line 124: this doesnt actually address the `FIXME` you removed. We ought to work with keycodes, not characters.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -8591,7 +8591,7 @@ void Actor::SetSoundFolder(const ieVariable& soundset) const
 	}
 }
 
-std::string Actor::GetSoundFolder(int full, const ResRef& overrideSet) const
+String Actor::GetSoundFolder(int full, const ResRef& overrideSet) const
```
bradallred, line 8594: shouldn't this return `path_t`? or is the name wrong?

gemrb/core/KeyMap.cpp
MarcelHB: No, it primarily fixes the crash when libfmt encounters some invalid UTF-8 byte.

I was really thinking if there is any case with keyboard codes > 255, but yes, I can restore the FIXME note.

gemrb/core/Scriptable/Actor.cpp
MarcelHB: That's one of these case when I wasn't sure what to use. Either choice has implications for the code so (converting for Python vs. converting for path uses ...) so I choose against the bigger sigh for something.

gemrb/core/System/VFS.cpp
```diff
@@ -62,10 +68,18 @@
...
 
 struct dirent {
-	dirent() : buffer(_MAX_PATH, L'\0'), d_name(const_cast<char*>(buffer.data())) {}
+	dirent() : buffer(_MAX_PATH, '\0'), d_name(const_cast<char*>(buffer.data())) {}
 
 	std::string buffer;
-	const char *d_name;
+	char *d_name;
+
+	void assign(std::string&& entryName) {
```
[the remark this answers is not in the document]
MarcelHB: As you like.

gemrb/core/System/VFS.cpp
```diff
@@ -110,11 +124,13 @@ static dirent* readdir(DIR *dir) {
 		}
 	}
 
-	auto& de = dir->entry;
-	auto n = wcstombs(const_cast<char*>(de.d_name), c_file.name, de.buffer.length() - 1);
-	de.buffer[n] = 0;
+	char16_t *c = reinterpret_cast<char16_t*>(c_file.name);
+	size_t n = 0;
+	for (; n < _MAX_PATH && *c != u'\0'; ++c, ++n) {}
```
bradallred, line 129: deems like there should be a std function for this, no?

MarcelHB:
> I suggest you move strlicmp to its own header file in case it becomes useful elsewhere later and since it doesn't really have to do anything with the VFS.

Is there really such as case? My hope is that there is nothing else but VFS where there is a need to emulate a little Windows behaviour.

gemrb/core/System/VFS.cpp
MarcelHB: If there was a `std::find_n` ... maybe, I'm not aware of one r/n.

gemrb/core/System/VFS.cpp
bradallred: we have `MutableStringView`

gemrb/core/Audio.cpp
bradallred: I think there is at least one other case here. if this really doesnt work we should fix it, or at least pass the length to avoid `strlen`

gemrb/core/Scriptable/Actor.cpp
bradallred: not really sure. does it return an actual path, or just the name of the directory?

lynxlynxlynx:
I hope not, but that's one of the reasons why I didn't suggest putting it into CString.h. We had a similar strlcpy file for a while.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Depends on the game.

gemrb/core/SaveGameIterator.cpp
```diff
@@ -136,7 +136,8 @@ SaveGame::SaveGame(path_t path, path_t name, const ResRef& prefix, std::string s
 	} else {
 		Date = fmt::format(DATE_FMT, fmt::localtime(my_stat.st_mtime));
 	}
-	manager.AddSource(Path, Name.c_str(), PLUGIN_RESOURCE_DIRECTORY);
+	manager.AddSource(Path, name.c_str(), PLUGIN_RESOURCE_DIRECTORY);
```
bradallred, line 139: not your fault, but it looks like we should drop the `c_str`

gemrb/core/System/VFS.cpp
MarcelHB: Ok will try, did not have this on screen yet.

gemrb/core/System/VFS.cpp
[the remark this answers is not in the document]
MarcelHB: I've renamed into something making it more clear we are at UTF-8 for now.

gemrb/core/System/VFS.cpp
MarcelHB: See above.

gemrb/core/Audio.cpp
MarcelHB: Did work, not sure any more what I was facing.

gemrb/core/KeyMap.cpp
MarcelHB: Restored

gemrb/core/System/VFS.cpp
MarcelHB: I still haven't come up with anything std'ish that takes both conditions into account. Let me know if you find something but let's leave that for now.

gemrb/core/System/VFS.cpp
MarcelHB: No more `strlen`  anything.

MarcelHB:
So except from design decisions maybe, I hope that all suggestions are in.

And I just noticed that we are not alone with how we are doing, I guess: https://www.phoronix.com/news/Bcachefs-Case-Folding-Patches

MarcelHB:
Hm, we don't see the smoke test output on AppVeyor? Not sure what the problem is then.

lynxlynxlynx:
We do, this problem is unique to this PR.

lynxlynxlynx:
It worked in this PR before too, but clearly some of the refactoring is not sitting well with it. I guess somehow logging broke for there to be no feedback whatsoever. 
You can stick `Set-PSDebug -Trace 1` into that last `.appveyor` block to make sure one of the builtins didn't fail before even getting to gemrb. Unlikely, but you never know.

MarcelHB:
Is there otherwise a chance to get the build artifact? I'd test it locally.

lynxlynxlynx:
Not easily. You could temporarily add an uploader to the on_failure step:
https://www.appveyor.com/docs/packaging-artifacts/#pushing-artifacts-from-scripts

MarcelHB:
Ok, I got it. `C.UTF-8` has low chances to exist in any of the Windows runtimes. Since we don't need it for now, I cut the method off on Windows.

lynxlynxlynx:
Might be cleaner to handle the exception, but it doesn't really matter. Does this then also address #340? I guess it does #1611, but we can have the natives check.

MarcelHB:
I've seen the same problems as the people have screenshotted in #340 and #1611 and they are fixed by now (unless you continue to play an old savegame of GemRB).

lynxlynxlynx:
Excellent! :)
@bradallred: any further comments? Merge otherwise.

bradallred:
thanks!

bradallred:
so [this](https://github.com/libsdl-org/SDL/issues/8129) just dropped.

MarcelHB:
What a timing 😄

Issue #1885, opened 2023-08-18
 Regression Testing Setup


MarcelHB:
I don't want to do TDD or refactor anything for code coverage but I think there is space to think about some automated testing and CI.

There are already smaller units and ones that are already properly isolated that can easily get a bunch of tests to make sure that the next wave of refactoring will not leave us blind towards a bunch of regressions. Worst case to me is corrupting savegames.

I'd like to discuss anything related to how and what, testing libs/frameworks, mildly convenient CMake integration.

bradallred:
cmake has [CTest](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html)

lynxlynxlynx:
Some of the low-level stuff we can definitely cover and it can help with making our code reusable (people struggled before). 

The first things that come to mind are VFS, FixedSizedString and other string stuff (Explode was a nightmare to get right), maybe some importers, and the minimal test we use on CI could be made into a CTest test, so it's easier to run.

MarcelHB:
Yeah, CTest as driver makes sense for us. Preferences regarding invidual tests? I'm mildly familiar with GoogleTest but maybe there is also something more convincing.

> The first things that come to mind are VFS, FixedSizedString and other string stuff (Explode was a nightmare to get right), maybe some importers

That's what I'm also thinking about for the first cases.

lynxlynxlynx:
I haven't used any in practice. A quick search also suggests a fast https://github.com/doctest/doctest (inlining support may make it easier to deploy) and https://github.com/catchorg/Catch2, both of which also come in a single header-only version.

MarcelHB:
Hm, I'm playing around with all three of them since they are somewhat different (and GT comes with mocking facilities) and let's then have a look what looks scalable for us.

bradallred:
Thanks for doing the research @MarcelHB. Tests will be very helpful, even if limited in scope.

MarcelHB:
I've done a little setup in all three libs with CMake Test integration with cases of fresh memory. I even found the first bug this way (see VFS.cpp and Win32 test case) :slightly_smiling_face: 

What is nice: all three libs support putting a test section anywhere and one runner will include them all. So we only need one test in terms of CMake, not one for each unit. That also helps compilation speed.

[...]

PR #1886, opened 2023-08-19
Test demo: doctest

MarcelHB:
Some general and common remarks will follow as comment in #1885.

This is trying out [doctest](https://github.com/doctest/doctest).

* Really easy to embed: just one header
* Nice and easy hierarchy of test cases to share data without DRY.
* Not that big but still [many features](https://github.com/doctest/doctest/blob/master/doc/markdown/features.md#other-features) like fixture classes, template type iteration
* Needs one designated runner file but the rest will auto-register (here: `TestRunner.cpp`)

Impressions:
doctest


CMakeLists.txt
```diff
@@ -599,6 +599,10 @@ CONFIGURE_FILE(
 	IMMEDIATE @ONLY
 )
 
+# Test setup
+ENABLE_TESTING()
+LIST(APPEND CMAKE_CTEST_ARGUMENTS "--verbose")
```
lynxlynxlynx, line 604: In non-verbose mode I imagine there are no details printed for failed tests?

Can we make this a parameter to `make test`? Most of the time we won't care for the wall of text, just that the tests pass. Then we could do something like this in CI: `make test || make test --verbose --rerun-failed`

MarcelHB: I justed want to see whether my tests are registered properly (and not simply went missing) because otherwise no (new) tests also means green here.

There is also something like `--output-on-failure` and then CMake refrains from suppressing output of failure cases by running once again.

PR #1888, opened 2023-08-19
Test demo: GoogleTest

MarcelHB:
Some general and common remarks will follow as comment in #1885.

This is trying out [GoogleTest](https://github.com/google/googletest).

* Fumbles around with CMake and is set up via download (by default)
* Test organization is a little more stiff but allows control over more advanced customizations
* Wide range of matchers and assertions
* Most helpful default output
* Only one with mocking facilities, although I think we are still very far away from a suitable code base
* Apparently the most widely used library of all candidates

Impression:
gest

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -599,6 +599,19 @@ CONFIGURE_FILE(
 	IMMEDIATE @ONLY
 )
 
+# Test setup
+ENABLE_TESTING()
+LIST(APPEND CMAKE_CTEST_ARGUMENTS "--verbose")
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/refs/tags/v1.14.0.zip
+)
+
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+
 ADD_SUBDIRECTORY( gemrb )
 IF (NOT APPLE)
 	INSTALL( FILES "${CMAKE_CURRENT_BINARY_DIR}/gemrb.6" DESTINATION ${MAN_DIR} )
@@ -701,3 +714,4 @@ if ((NOT DISABLE_WERROR) AND GIT_VERSION)
 	message(STATUS "If you hit one, let us know and rerun cmake with -DDISABLE_WERROR=1 to continue.")
 	message(STATUS "")
 endif()
+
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -599,6 +599,19 @@ CONFIGURE_FILE(
...
 
+# Test setup
+ENABLE_TESTING()
+LIST(APPEND CMAKE_CTEST_ARGUMENTS "--verbose")
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/refs/tags/v1.14.0.zip
+)
```
lynxlynxlynx, line 610: _And a thousand packagers cried out in pain._ We'd have to ship a copy for everyone that needs offline builds and that's a blocker for me — gtest is huge in comparison to the other two.

MarcelHB: It's taken from the `Get started quickly` and works.

But yes, we had to find a way to vendor this (huge) thing for offline builds.

MarcelHB: On the other hand, it appears to be available in all major package systems.

bradallred:
Maybe its just due to being a quick trial, but I'd much prefer to not mix tests into the source.

MarcelHB:
🤷 

Helpful for static stuff and doesn't require additional bookkeeping in the build system. Then, yes, large things may quickly turn ugly.

CMakeLists.txt
lynxlynxlynx: It's of no use for most users, so we can just remove this part and rely on developers getting it themselves.

gemrb/core/TestRunner.cpp
```diff
@@ -0,0 +1,2 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include "doctest.h"
```
lynxlynxlynx, line 2: I guess this is just a remnant from the other test.

MarcelHB: Yeah, ignore this outside of the doctest branch.

lynxlynxlynx:
CMake overhead would be minimal. We're relatively small, so I don't think we need to worry about tests going stale if they're not in the same file either. CI would complain about problems.
Statics are a good point, but if it ever comes to that, we can just make them full class members as needed.

Issue #1879, opened 2023-08-08
BG2: Quests are reinitialized. 


rsantellan:
#### Bug description
I have a save game pretty advanced on the second chapter, going back to the wizard sphere on the slums after updating to master I'm receive by Gaelan Bayle like the first time, but the quest is already on my journal for a long time.

#### Steps to reproduce
1. Go to Slums after getting the introduction of Gaelan Bayle
2. Be receive by Gaelan Bayle
3. See error

#### Expected behavior
Only be greeted once by him.

#### Screenshots
Quest on journal
![image](https://github.com/gemrb/gemrb/assets/1820598/ed93f8ca-6016-45e6-9c1c-e35fa8d1b663)
Again him
![image](https://github.com/gemrb/gemrb/assets/1820598/0efa9675-8da3-4930-a49d-1fe44c90f793)

This is my save game.

[000000157-Traveling error.tar.gz](https://github.com/gemrb/gemrb/files/12286335/000000157-Traveling.error.tar.gz)

GemRB version (check as many as you know apply)
- [X] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `USE_OPENGL_BACKEND`
- [X] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
This is critical and likely caused by the dictionary refactor, since it's probably a problam with some variable or variable timer. If you can find a simpler reproducer in some other circumstance, that would be even better.

MarcelHB:
The problem appears to be something else: `SaveGameAREExtractor` isn't extracting the saved `AR0400` here or any more so you end up with a fresh instance.

MarcelHB:
Probably; from the corner of my eye, `areLocations` wasn't filled as it is supposed to be.

MarcelHB:
The problem is a little different than expected: When I look at the save games generated from GemRB, all area files but the currently loaded one lack the file extension in the sav file.

lynxlynxlynx:
All good now?

MarcelHB:
No, something is even more broken by now. If I save something now, I see one only file with a broken name saved in the SAV file:

![iwd2_sav_file](https://github.com/gemrb/gemrb/assets/238558/a7459e1f-e6a1-4a26-a8f2-ab3855a976fe)

[...]

PR #1889, opened 2023-08-20
Google Test setup (WIP)

MarcelHB:
## Description
See #1885.

Currently under experimenting with runners.

## Checklist


CMakeLists.txt
```diff
@@ -599,6 +599,15 @@ CONFIGURE_FILE(
 	IMMEDIATE @ONLY
 )
 
+# Tests (need init'd before subdirectories)
+FIND_PACKAGE(GTest)
+
+IF (GTest_FOUND)
```
lynxlynxlynx, line 605: It'd be worth noting somewhere that people can pass CMAKE_DISABLE_FIND_PACKAGE_GTest=1 to avoid the tests even if they have gtest installed. Even better if you add an option like USE_SDLMIXER.

MarcelHB:
I lack the Mac knowledge to see this can be fixed:
```
Test project /Users/runner/work/gemrb/gemrb/build
    Start 1: Test_gemrb_core
1/1 Test #1: Test_gemrb_core ..................Subprocess aborted***Exception:   0.10 sec
Errors while running CTest
dyld[9001]: Library not loaded: '@rpath/libgemrb_core.0.9.2-git.dylib'
make: *** [test] Error 8
  Referenced from: '/Users/runner/work/gemrb/gemrb/build/gemrb/Test_gemrb_core'
  Reason: tried: '/Users/runner/work/gemrb/gemrb/build/gemrb/../Frameworks/libgemrb_core.0.9.2-git.dylib' (no such file), '/Users/runner/work/gemrb/gemrb/build/gemrb/../Frameworks/libgemrb_core.0.9.2-git.dylib' (no such file), '/usr/local/lib/libgemrb_core.0.9.2-git.dylib' (no such file), '/usr/lib/libgemrb_core.0.9.2-git.dylib' (no such file)
```

MarcelHB:
AppVeyor looks as if Google Test dll isn't near by. Will have to investigate.

bradallred:
> I lack the Mac knowledge to see how this can be fixed:
> 
> ```
> Test project /Users/runner/work/gemrb/gemrb/build
>     Start 1: Test_gemrb_core
> 1/1 Test #1: Test_gemrb_core ..................Subprocess aborted***Exception:   0.10 sec
> Errors while running CTest
> dyld[9001]: Library not loaded: '@rpath/libgemrb_core.0.9.2-git.dylib'
> make: *** [test] Error 8
>   Referenced from: '/Users/runner/work/gemrb/gemrb/build/gemrb/Test_gemrb_core'
>   Reason: tried: '/Users/runner/work/gemrb/gemrb/build/gemrb/../Frameworks/libgemrb_core.0.9.2-git.dylib' (no such file), '/Users/runner/work/gemrb/gemrb/build/gemrb/../Frameworks/libgemrb_core.0.9.2-git.dylib' (no such file), '/usr/local/lib/libgemrb_core.0.9.2-git.dylib' (no such file), '/usr/lib/libgemrb_core.0.9.2-git.dylib' (no such file)
> ```

the problem is `libgemrb_core` is built and installed as part of an app bundle. I thought there was a cmake option you could pass to get a unix build instead, but I cant check now.

gemrb/core/System/VFS.cpp
```diff
@@ -247,18 +247,17 @@ void PathAppend(path_t& target, const path_t& name)
 		return;
 	}
 
-	if (!target.empty() && target.back() != PathDelimiter && name.front() != PathDelimiter) {
-		target.push_back(PathDelimiter);
+	auto scope = StringView{name.c_str(), name.length()};
+	// TotL has '\data\zcMHar.bif' in the key file
+	if (name[0] == '\\') {
```
bradallred, line 252: I don't think this should be part of the VFS. is there more than one place that has this issue?

MarcelHB: I agree about the misplacement. But I don't know any else r/n. This is mostly by-work since I wanted to address the first finding from the prework.

CMakeLists.txt
MarcelHB: I've added `USE_TESTS` with then-required GTest package.

MarcelHB:
> the problem is `libgemrb_core` is built and installed as part of an app bundle. I thought there was a cmake option you could pass to get a unix build instead, but I cant check now.

I'll be glad if you have a look at this. I did not find anything conclusive here.

Aside from some ongoing HTTP-trouble around AppImage right now, Windows and Linux are running with GTest now.

For me, this is the tiny little ground work to start with core-testing. Not exactly sure how to proceed with plugin code but let's see on demand.

By starting I mean: We have a layout to incrementally add tests cases; tests are isolated, optional units, and there is a place to test disk-based interaction or to place fixtures and bigger expectations as files, maybe (`tests/resources`).

So if we are bored, by now, i. e. after this setup PR, we can start picking elemental units for tests and start fixing and cleaning up the code.

bradallred:
> I'll be glad if you have a look at this. I did not find anything conclusive here.

If you look at the root CmakeLists.txt there is a `LAYOUT` variable. if you manually specify whatever linux uses instead of "bundle" then it should work.

> Not exactly sure how to proceed with plugin code

Probably should change `PluginLoader` to expose functionality for loading a single plugin and creating a test harness around that.

bradallred:
> > I'll be glad if you have a look at this. I did not find anything conclusive here.
> 
> If you look at the root CmakeLists.txt there is a `LAYOUT` variable. if you manually specify whatever linux uses instead of "bundle" then it should work.

Tho, it shouldn't matter unless you install. Tests shouldn't require an install.

MarcelHB:
I copied the core library into that folder that it expects on Mac and it works.

Which raises the next interesting question: Do you have any idea for alternatives regarding this locale trouble? i. e. what to do on Mac here? I can also disable the test, but I think this is some general concern.

bradallred:
`C.UTF-8` is a glibc thing, so that is not portable to systems that don't use glibc.

> Both FreeBSD and OS X define a "UTF-8" locale that has only one locale category defined in it: LC_CTYPE

so you can only set `LC_CTYPE` to UTF-8

MarcelHB:
Thanks, I'll get myself a FreeBSD VM later.

MarcelHB:
GhostBSD accepts the code without complaints. Back to solving this via GH Mac runner ...

bradallred:
> GhostBSD accepts the code without complaints and behaves like the glibc. Back to solving this via GH Mac runner ...

are you sure that you aren't using glibc on this? Be sure to build with clang, not gcc. There are other flavors of BSD you can use too.

MarcelHB:
I was using clang++ here.

I'll try another one if get stuck over what Mac OS manpage [says about](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/tolower.3.html) tolower.

bradallred:
`std::locale` has a constructor: `locale( const locale& other, const char* std_name, category cats );`, so you should be able to create a static function to return a locale setup with `C` for everything but `LC_CTYPE` set to `UTF-8`, which should be what we are looking for with `C.UTF-8`

bradallred:
> I'll try another one if I get stuck over what Mac OS manpage [says about](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/tolower.3.html) tolower.

this may or may not be relevant, but you ought to be looking at C++ documentation.

MarcelHB:
> `std::locale` has a constructor: `locale( const locale& other, const char* std_name, category cats );`, so you should be able to create a static function to return a locale setup with `C` for everything but `LC_CTYPE` set to `UTF-8`, which should be what we are looking for with `C.UTF-8`

I've tried this:
```cpp
#ifdef __APPLE__
			static const std::locale l1;
			static const std::locale l2{"UTF-8"};
			static const auto C_LOCALE = std::locale{l1, l2, std::locale::ctype};
#else
```
Since this is what appears to configuring the desired locale according [to the example](https://en.cppreference.com/w/cpp/locale/locale/locale). But no luck, UTF-8 is rejected.

bradallred:
if you are just going to `#ifdef` it then just use `std::setlocale` with `LC_CTYPE`

MarcelHB:
> if you are just going to `#ifdef` it then just use `std::setlocale` with `LC_CTYPE`. The problem with `std::locale` with the single argument for the locale name is that there is no `UTF-8` locale. you would need to use something like `en_US.UTF-8`

That (`en_US`) works, at least for my German stuff ... then let's leave it like this for now.

MarcelHB:
So two test groups for designing the setup and up to three issues already. I'm convinced this will help us save time and bug reports in the long :slightly_smiling_face:

MarcelHB:
So if you are fine with this, I'd target VFS (also because of Windows and localized data) and also try to catch the reason behind #1879

gemrb/tests/core/Strings/Test_UTF8Comaprison.cpp
```diff
@@ -0,0 +1,49 @@
...
+	EXPECT_TRUE(UTF8_stricmp("abc", "abc"));
+	EXPECT_TRUE(!UTF8_stricmp("abc", "ab"));
+	EXPECT_TRUE(!UTF8_stricmp("ab", "abc"));
+	EXPECT_TRUE(!UTF8_stricmp("abc", "def"));
+	EXPECT_TRUE(UTF8_stricmp("abc", "ABC"));
+	EXPECT_TRUE(UTF8_stricmp("ABC", "abc"));
+
+	EXPECT_TRUE(UTF8_stricmp("äbc", "äbc"));
+	EXPECT_TRUE(UTF8_stricmp("äbc", "ÄBC"));
+	EXPECT_TRUE(UTF8_stricmp("", ""));
```
lynxlynxlynx, line 44: We could also test for null parameters, since it could happen and the function actually appears to handle it well.

gemrb/tests/core/System/Test_VFS.cpp
```diff
@@ -0,0 +1,94 @@
...
+	path_t path{"dir\\"};
+	PathAppend(path, "subdir");
+	EXPECT_EQ(path, "dir\\subdir");
+#else
+	path_t path{"dir/"};
+	PathAppend(path, "subdir");
+	EXPECT_EQ(path, "dir/subdir");
+#endif
+}
+
```
lynxlynxlynx, line 93: It would make sense to add PathJoin tests, since the first test already relies on it.

MarcelHB: I even had them for a moment, until I noticed that `PathJoin` is broken on Windows. :)

Will have to do this a a new approach with whatever it takes, for not blowing up this first PR.

gemrb/tests/core/Strings/Test_UTF8Comaprison.cpp
MarcelHB: Sure but `nullptr == nullptr` or `nullptr != nullptr` (as in NaN != NaN)? :wink:

lynxlynxlynx: Currently it treats them as equal.

Mingun: You also have a misprint in the file name

MarcelHB: Very good catch

gemrb/tests/core/System/Test_VFS.cpp
bradallred: I assume you mean `PathAppend` or possibly `ResolveCase`

MarcelHB: I was facing issue with `PathJoin`. See the case:
```cpp
auto baseDir = PathJoin("tests", "resources", "VFS", "encoding");
EXPECT_TRUE(FileExists(PathJoin(baseDir, "file.txt")));
```
When I try this on Win, I see some chopped path string that does not work.

MarcelHB: I saw that it hit `access` on its code path which I have eliminated elsewhere already for causing problems by now. But I haven't verified this to be the exact cause yet.

bradallred: well `ResolveCase` calls `access`, but only if `core->config.CaseSensitive` is false

bradallred: ah, ill wager that `core` is `nullptr` and that is your issue. you can workaround it by using `PathJoin<false>`.

bradallred: we should probably change that conditional to `if (!core || !core->config.CaseSensitive)`

gemrb/core/Strings/UTF8Comparison.cpp
```diff
@@ -84,9 +87,14 @@ bool UTF8_stricmp(const char *a, const char *b) {
 				}
 			};
 
-			uint8_t widthA = getWidth(iA), widthB = getWidth(iB);
-
+#ifdef __APPLE__
```
bradallred, line 90: we should probably add a cmake detection for `C.UTF-8`... then again, why not just always do it this way? this should work for everything, no?

MarcelHB: You mean we check for `C.UTF-8` in some cmake step, and what is currently `#ifdef __APPPLE__` turns into `#ifndef HAVE_C_UTF8`?

bradallred: yes, but im also asking why not just drop the `#ifdef` and just do it the one way that works for everything.

MarcelHB: Simply because I've started with `C.UTF-8` as it sounds to have a wider compatibility, language-wise. If it works for the cases we have, we can give it a try, sure.

bradallred: well if they aren't equivalent (for our purposes), then we have a problem anyway. But all this is doing is creating a `C` locale, then setting `LC_CTYPE` to `UTF-8`, which is the same things as `C.UTF-8` as far as I can tell.

MarcelHB: Let's do it like this for now.

MarcelHB:
I still vote to defer it in favor of some new VFS cleanup PR. I don't want to branch for too long while there is work on master.

Issue #1892, opened 2023-08-25
Add filesystem case sensitivity as a build time option


bradallred:
See commit 0869613

Lots of platforms use case insensitive filesystems, and many have both as an option.

lynxlynxlynx:
I disagree, we should just detect it. It's simple to test.

bradallred:
sure, I'll edit this to reflect that path.

lynxlynxlynx:
I think three path groups will be enough. One for our files, one for game files and one for saves (more likely to be stored on a different partition, especially on ees). Then we can do three simple checks on startup and have almost no overhead in later calls. It would require more VFS API adjustments though. Any better ideas?

PR #1898, opened 2023-09-01
Test cases: VFS, DataStreams' reading

MarcelHB:
## Description
* Covering VFS related stuff (`PathJoin` was no longer a problem on Win)
* Covering reading of `DataStream`s (File + mmap)
* Minor cleanups and fixes around here

Next on the list is probaly something like writing of data streams & strings.

## Checklist


lynxlynxlynx:
Nice. 👍🏾 
All the path ifdefing is annoying. What do you think about using fmt to unify, something like `fmt::format("tests{0}resoUrces{0}vfs{0}file_ÄÖÜ.txt", SPathDelimeter)` for each path?

MarcelHB:
> Nice. 👍🏾 All the path ifdefing is annoying. What do you think about using fmt to unify, something like `fmt::format("tests{0}resoUrces{0}vfs{0}file_ÄÖÜ.txt", SPathDelimeter)` for each path?

I can reduce these cases by this, but not all of them since Win does not expect case resolution in some cases in contrast to all the rest (simplified assumption for now).

gemrb/tests/core/System/Test_VFS.cpp
```diff
@@ -23,11 +23,33 @@
 
 #include "System/VFS.h"
 
+#define IS_CASE_INSENSITIVE (WIN32 || __APPLE__)
```
lynxlynxlynx, line 26: I guess we will change this later in #1892, though it's just the test suite, so maybe we don't care. Just today I was talking to a macos user with build problems due to using a case sensitive FS.

MarcelHB: Agree. Just don't have any better concept to deal with that r/n, also since this appears to be even a matter of per-path settings such as on ext4.

MarcelHB:
Any concerns left?

gemrb/core/Streams/DataStream.cpp
```diff
@@ -135,20 +135,32 @@ strret_t DataStream::ReadLine(std::string& buf, strpos_t maxlen)
 	buf.clear();
 	buf.reserve(maxlen);
 	strpos_t i = 0;
-	//TODO: fix this to handle any combination of \r and \n
-	//Windows: \r\n
-	//Old Mac: \r
-	//otherOS: \n
+
+	bool isCR = false;
 	while (Pos < size && i < (maxlen - 1)) {
```
bradallred, line 140: this whole loop is really hard to read now.

MarcelHB: My mind works in state machines. Unless we want to introduce a `Peak` or something, that's what I found for compressing that a little:
```cpp
	while (Pos < size && i < (maxlen - 1)) {
		char ch;
		i += Read(&ch, 1);

		if (ch == '\r') {
			i += Read(&ch, 1);

			if (ch == '\n') {
				break;
			} else {
				i -= 1;
				Seek(-1, GEM_CURRENT_POS);
				break;
			}
		} else if (ch == '\n') {
			break;
		}

		if (ch == '\t') {
			ch = ' ';
		}

		buf.push_back(ch);
	}
```

bradallred: `Peek` would be useful. It's just read followed by seek.

I'm also fine with that change. Thanks.

lynxlynxlynx:
Thanks!

Issue #1882, opened 2023-08-11
 PST: Assertion violation on loading savegame in AR1202


MarcelHB:
#### Bug description
One of my all-time test savegames in PST isn't starting properly anymore.

#### Steps to reproduce
1. Load PST savegame from the attachment.

#### Screenshots
```gdb
(gdb) bt
#0  0x00007ffe17aff1e7 in msvcrt!abort () from C:\Windows\System32\msvcrt.dll
#1  0x00007ffe17afda39 in msvcrt!_wassert () from C:\Windows\System32\msvcrt.dll
#2  0x00007ffdef4c8eaf in GemRB::AREImporter::GetMap (this=0x13999e70, resRef=..., day_or_night=true) at D:/Sources/gemrb/gemrb/plugins/AREImporter/AREImporter.cpp:1592
#3  0x00007ffda779027b in GemRB::Game::LoadMap (this=0x12b616d0, resRef=..., loadscreen=true) at D:/Sources/gemrb/gemrb/core/Game.cpp:852
#4  0x00007ffda778f7b6 in GemRB::Game::GetMap (this=0x12b616d0, areaName=..., change=true) at D:/Sources/gemrb/gemrb/core/Game.cpp:690
#5  0x00007ffda7735c38 in GemRB::GameControl::ChangeMap (this=0x1281b340, pc=0x126d3340, forced=true) at D:/Sources/gemrb/gemrb/core/GUI/GameControl.cpp:2499
#6  0x00007ffda78321e8 in GemRB::Interface::HandleFlags (this=0x14f2e0) at D:/Sources/gemrb/gemrb/core/Interface.cpp:809
#7  0x00007ffda7832f49 in GemRB::Interface::Main (this=0x14f2e0) at D:/Sources/gemrb/gemrb/core/Interface.cpp:973
#8  0x0000000140001667 in main (argc=3, argv=0xa84890) at D:/Sources/gemrb/platforms/windows/GemRB.cpp:62
(gdb) f 2
#2  0x00007ffdef4c8eaf in GemRB::AREImporter::GetMap (this=0x13999e70, resRef=..., day_or_night=true) at D:/Sources/gemrb/gemrb/plugins/AREImporter/AREImporter.cpp:1592
1592            assert(ExploredBitmapSize <= mapSize);
(gdb) p ExploredBitmapSize
$1 = 495
(gdb) p mapSize
$2 = 435
```
[...]

PR #1905, opened 2023-09-06
AREImporter: dropping an assertion

MarcelHB:
## Description
See #1882.

## Checklist

### DIFF
--- a/gemrb/plugins/AREImporter/AREImporter.cpp
+++ b/gemrb/plugins/AREImporter/AREImporter.cpp
@@ -1596,7 +1596,6 @@ Map* AREImporter::GetMap(const ResRef& resRef, bool day_or_night)
 
 	Log(DEBUG, "AREImporter", "Loading explored bitmap");
 	ieDword mapSize = ieDword(map->ExploredBitmap.Bytes());
-	assert(ExploredBitmapSize <= mapSize);
 	mapSize = std::min(mapSize, ExploredBitmapSize);
 	str->Seek(ExploredBitmapOffset, GEM_STREAM_START);
 	str->Read(map->ExploredBitmap.begin(), mapSize);


bradallred:
this doesn't address my comment:
https://github.com/gemrb/gemrb/issues/1882#issuecomment-1677932178

so what is the extra data? is it just padding or what?

lynxlynxlynx:
Same 0xff values like the rest of the block

bradallred:
so why are we sure that we don't have a bug in our understanding of `Map::FogMapSize`? PST uses `GFFlags::SMALL_FOG`, but maybe it's a per area thing.

we should at least look at the value of `TMap->XCellCount` and `TMap->YCellCount` and see what that would produce without `SMALL_FOG`. 60 bytes is a lot more than any padding should introduce.

lynxlynxlynx:
The buggy state is unreproducable both in the original and gemrb, so I don't know why you're so adamant.

bradallred:
Because, if it's not reproducible then what is the point of removing the assert? As I said it's there for a reason.

MarcelHB:
Savegames don't run but we have no idea where this comes from and how to reproduce -- and having this removed, there is no apparent side-effect so far.

bradallred:
I understand removing it wont cause a problem other than using the wrong size for a fog tile. It sounds like a corrupt save game if it's not reproducible. What is the save date on the game? Maybe it corresponds to some relevant refactoring.

if its not corruption, then it feels like we are just hiding something we don't understand. Maybe some graphics setting affects it.

lynxlynxlynx:
PST doesn't store real time in the save, so we can't be sure. Best approximation would be mtime/ctime of the file on Marcel's disk. It's at least 2019 or older though, since the bug in #163 is visible in the thumbnail.

I now suspect it's a gemrb save, since NI doesn't display the journal.

bradallred:
ok, then one last issue:
```
ieDword mapSize = ieDword(map->ExploredBitmap.Bytes());
assert(ExploredBitmapSize <= mapSize);
mapSize = std::min(mapSize, ExploredBitmapSize);
str->Seek(ExploredBitmapOffset, GEM_STREAM_START);
str->Read(map->ExploredBitmap.begin(), mapSize);
```

seems like the data is supposed to contain `ExploredBitmapSize`. We aren't skipping the possible remainder. I'm supposing this is ok? it seems like this is the last bit of data read so it doesn't matter.

if thats true, then just go ahead and merge it I guess.

lynxlynxlynx:
We seek when reading the next structure (WEDImporter::GetWallGroups), since there's no guarantee that they would be in order or packed tightly.

Issue #1902, opened 2023-09-04
 Memory leak around font code


MarcelHB:
#### Bug description
Another one found while debugging with Valgrind.

#### Steps to reproduce
1. Can be reproduced playing some PST game.

#### Screenshots
```
==11222== 77,808 (88 direct, 77,720 indirect) bytes in 1 blocks are definitely lost in loss record 2,707 of 2,711
==11222==    at 0x4843FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==11222==    by 0x4C31431: std::__new_allocator<std::_Rb_tree_node<std::pair<unsigned short const, GemRB::Region> > >::allocate(unsigned long, void const*) (new_allocator.h:137)
==11222==    by 0x4C31085: std::allocator_traits<std::allocator<std::_Rb_tree_node<std::pair<unsigned short const, GemRB::Region> > > >::allocate(std::allocator<std::_Rb_tree_node<std::pair<unsigned short const, GemRB::Region> > >&, unsigned long) (alloc_traits.h:464)
==11222==    by 0x4C30B28: std::_Rb_tree<unsigned short, std::pair<unsigned short const, GemRB::Region>, std::_Select1st<std::pair<unsigned short const, GemRB::Region> >, std::less<unsigned short>, std::allocator<std::pair<unsigned short const, GemRB::Region> > >::_M_get_node() (stl_tree.h:563)
==11222==    by 0x4C300F2: std::_Rb_tree_node<std::pair<unsigned short const, GemRB::Region> >* std::_Rb_tree<unsigned short, std::pair<unsigned short const, GemRB::Region>, std::_Select1st<std::pair<unsigned short const, GemRB::Region> >, std::less<unsigned short>, std::allocator<std::pair<unsigned short const, GemRB::Region> > >::_M_create_node<unsigned short&, GemRB::Region&>(unsigned short&, GemRB::Region&) (stl_tree.h:613)
[...]

Issue #1901, opened 2023-09-04
 Memory leak in Map.cpp


MarcelHB:
#### Bug description
While looking at #1900 with Valgrind, I get the reports of some memory leaks in Map.cpp:

#### Steps to reproduce
1. Start a game and load a map while looking at Valgrind

#### Screenshots

```
==11222== 2,144 (824 direct, 1,320 indirect) bytes in 1 blocks are definitely lost in loss record 2,657 of 2,711
==11222==    at 0x4843FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==11222==    by 0x4D87A18: GemRB::Map::AddContainer(GemRB::FixedSizeString<32ul, &strncasecmp> const&, unsigned short, std::shared_ptr<GemRB::Gem_Polygon> const&) (Map.cpp:3495)
==11222==    by 0x96315F6: GemRB::AREImporter::GetContainer(GemRB::DataStream*, int, GemRB::Map*) (AREImporter.cpp:789)
==11222==    by 0x9635BB7: GemRB::AREImporter::GetMap(GemRB::FixedSizeString<8ul, &strncasecmp> const&, bool) (AREImporter.cpp:1512)
==11222==    by 0x4C6AEB6: GemRB::Game::LoadMap(GemRB::FixedSizeString<8ul, &strncasecmp> const&, bool) (Game.cpp:842)
==11222==    by 0x4C6A2EB: GemRB::Game::GetMap(GemRB::FixedSizeString<8ul, &strncasecmp> const&, bool) (Game.cpp:680)
==11222==    by 0x4C0A10C: GemRB::GameControl::ChangeMap(GemRB::Actor const*, bool) (GameControl.cpp:2501)
==11222==    by 0x4D172F8: GemRB::Interface::HandleFlags() (Interface.cpp:808)
==11222==    by 0x4D18209: GemRB::Interface::Main() (Interface.cpp:972)
==11222==    by 0x110ACA: main (GemRB.cpp:66)
==11222==
==11222== 4,712 (824 direct, 3,888 indirect) bytes in 1 blocks are definitely lost in loss record 2,690 of 2,711
==11222==    at 0x4843FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

Issue #1877, opened 2023-08-04
Segmentation fault when quitting the game (SDL) 


MarcelHB:
#### Bug description
When I quit the game, I'm left with a segfault in SDL, probably because there is still `Sprite2D`s around in `Interface` (`Cursors`, `GroundCircles`) even after we have explicitly taken down SDL, so that the d'tor is running too late.

#### Steps to reproduce
1. Start the game
2. Quit the game.

#### Screenshots
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff9611cbacc in ?? () from C:\MSYS2\mingw64\bin\SDL2.dll
(gdb) bt
#0  0x00007ff9611cbacc in ?? () from C:\MSYS2\mingw64\bin\SDL2.dll
#1  0x00007ff9613dc2d7 in GemRB::SDLTextureSprite2D::~SDLTextureSprite2D (this=0x3e903b0, __in_chrg=<optimized out>) at D:/Sources/gemrb/gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp:295
#2  0x00007ff9613da843 in std::_Destroy<GemRB::SDLTextureSprite2D> (__pointer=0x3e903b0) at C:/MSYS2/mingw64/include/c++/13.2.0/bits/stl_construct.h:151
#3  0x00007ff9613da1c2 in std::allocator_traits<std::allocator<void> >::destroy<GemRB::SDLTextureSprite2D> (__p=0x3e903b0) at C:/MSYS2/mingw64/include/c++/13.2.0/bits/alloc_traits.h:674
#4  std::_Sp_counted_ptr_inplace<GemRB::SDLTextureSprite2D, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x3e903a0) at C:/MSYS2/mingw64/include/c++/13.2.0/bits/shared_ptr_base.h:613
#5  0x00007ff919753b8b in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x3e903a0) at C:/MSYS2/mingw64/include/c++/13.2.0/bits/shared_ptr_base.h:346
#6  0x00007ff919754b17 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x3e942a8, __in_chrg=<optimized out>) at C:/MSYS2/mingw64/include/c++/13.2.0/bits/shared_ptr_base.h:1071
[...]

Issue #1871, opened 2023-07-24
 Can't compile master anymore, no option -Wstringop-overread


rsantellan:
#### Bug description
I just updated the latest master branch and I can't compile it any more. I believe that is because my gcc version but can't find the documentation to be really sure.

**cmake .. -DCMAKE_BUILD_TYPE=Debug -DSDL_BACKEND=SDL2 -DPYTHON_VERSION=3.8**
```
-- The C compiler identification is GNU 9.4.0
-- The CXX compiler identification is GNU 9.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detected version: 0.9.2-git
-- Git version bool: 1
-- Performing Test NO_STRINGOP_TRUNCATION
-- Performing Test NO_STRINGOP_TRUNCATION - Success
-- Performing Test NO_STRINGOP_OVERREAD
-- Performing Test NO_STRINGOP_OVERREAD - Failed
-- Performing Test FILE_PREFIX_MAP
-- Performing Test FILE_PREFIX_MAP - Failed
-- Performing Test VISIBILITY_HIDDEN
-- Performing Test VISIBILITY_HIDDEN - Failed
-- Performing Test FALLTHROUGH
-- Performing Test FALLTHROUGH - Failed
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython3.8.so (found suitable version "3.8.10", minimum required is "3.8") 
-- Looking for Python libraries and headers: found
-- Found SDL2_mixer: /usr/lib/x86_64-linux-gnu/libSDL2_mixer.so (found version "2.0.4") 
-- Looking for SDL_mixer: found
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11") 
-- Looking for Zlib: found
-- Performing Test Iconv_IS_BUILT_IN

-- Performing Test Iconv_IS_BUILT_IN - Success
-- Found Iconv: /usr/lib/x86_64-linux-gnu/libc.so  
[...]

PR #1907, opened 2023-09-09
Some fixes on memory leaks and segfault

MarcelHB:
## Description
Not a refactoring but cleaning up the current Memcheck reports a little.

Addressing #1877, #1901 and #1902.

## Checklist

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -1193,7 +1193,14 @@ void Interface::LoadFonts()
 		if (!fnt) {
 			error("Core", "Unable to load font resource: {} for ResRef {} (check fonts.2da)", font_name, resref);
 		} else {
-			fonts[resref] = fnt;
+			auto result = fonts.find(resref);
+			if (result != fonts.cend()) {
+				delete result->second;
+				result->second = fnt;
+			} else {
+				fonts[resref] = fnt;
+			}
+
 			Log(MESSAGE, "Core", "Loaded Font: {} for ResRef {}", font_name, resref);
 		}
 	}
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -608,9 +608,6 @@ Interface::~Interface() noexcept
 	delete worldmap;
 	delete keymap;
 
-	ResRefMap<Font*>::iterator fit = fonts.begin();
-	for (; fit != fonts.end(); ++fit)
-		delete (*fit).second;
 	// fonts need to be destroyed before TTF plugin
 	PluginMgr::Get()->RunCleanup();
 
@@ -940,7 +937,7 @@ void Interface::Main()
 		SetMouseScrollSpeed(scrollSpeed(config.CapFPS));
 	}
 
-	const Font* fps = GetTextFont();
+	auto fps = GetTextFont();
 	// TODO: if we ever want to support dynamic resolution changes this will break
 	Region fpsRgn(0, config.Height - 30, 80, 30);
 	String fpsstring = u"???.??? fps";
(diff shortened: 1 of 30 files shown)


gemrb/core/Interface.cpp
```diff
@@ -1193,7 +1203,14 @@ void Interface::LoadFonts()
 		if (!fnt) {
 			error("Core", "Unable to load font resource: {} for ResRef {} (check fonts.2da)", font_name, resref);
 		} else {
-			fonts[resref] = fnt;
+			auto result = fonts.find(resref);
```
bradallred, line 1206: I'd rather us refactor `fonts` to not hold raw pointers

gemrb/core/Interface.cpp
```diff
@@ -631,7 +631,17 @@ Interface::~Interface() noexcept
 
 	// Removing all stuff from Cache, except bifs
 	if (!config.KeepCache) DelTree(config.CachePath, true);
-	
+
+	Cursors.clear();
```
bradallred, line 635: this is unnecessary.

gemrb/core/Interface.cpp
```diff
@@ -631,7 +631,17 @@ Interface::~Interface() noexcept
...
+	for (int size = 0; size < MAX_CIRCLE_SIZE; size++) {
+		for (int i = 0; i < 6; i++) {
+			GroundCircles[size][i] = nullptr;
+		}
+	}
+
+	// also maintains sprites somewhere
+	guiscript.reset();
```
bradallred, line 643: so is this

gemrb/core/Interface.cpp
```diff
@@ -631,7 +631,17 @@ Interface::~Interface() noexcept
 
 	// Removing all stuff from Cache, except bifs
 	if (!config.KeepCache) DelTree(config.CachePath, true);
-	
+
+	Cursors.clear();
+	for (int size = 0; size < MAX_CIRCLE_SIZE; size++) {
```
bradallred, line 636: if you just make `GroundCircles` a `std::array` then you dont need this either.

gemrb/core/Interface.cpp
MarcelHB: No because their dtors then run after https://github.com/gemrb/gemrb/blob/c77003fe0735869b00429e1118a09a307098d579/gemrb/core/Interface.cpp#L635

If you look at the Valgrind/gdb report in #1871 then it's apparently the reason of calling a freed SDL internals: https://github.com/gemrb/gemrb/blob/c77003fe0735869b00429e1118a09a307098d579/gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp#L295

MarcelHB: We can maybe also try to reorder the members of Interface but I'm not sure if there is a reason to explicitly shut down the video driver. DLL-stuff or something?

bradallred: `VideoDriver` isn't a member, so the answer would be to not do that there.

bradallred: Actually, I just remembered I added that because it was preventing GemRB from terminating on macOS. But that is probably something that can just get moved into the Mac specific code.

MarcelHB:
Have a look again. Font refactoring takes some more lines now.

gemrb/core/Interface.cpp
```diff
@@ -631,8 +628,6 @@ Interface::~Interface() noexcept
 
 	// Removing all stuff from Cache, except bifs
 	if (!config.KeepCache) DelTree(config.CachePath, true);
-	
-	VideoDriver.reset();
```
bradallred, line 635: can you put this in CocoaWrapper.mm around line 210 to avoid the shutdown hang?

bradallred (commented):
everything else looks great; thanks.

gemrb/core/Interface.cpp
MarcelHB: There you go

lynxlynxlynx:
Eh, branches don't update the cache.

PR #1916, opened 2023-09-26
Tests: writing of data stream, some string functions

MarcelHB:
## Description
Another set of test cases with one bug fix, I assume. One thought below.

## Checklist


gemrb/tests/core/Strings/Test_String.cpp
```diff
@@ -23,6 +23,114 @@
...
+	result = FindNotOf<String>(unit.cbegin(), unit.cend(), StringViewT<String>{});
+	EXPECT_EQ(result, unit.cbegin());
+}
+
+TEST(String_Test, FindFirstNotOf) {
+	auto result = FindFirstNotOf(String{}, StringViewT<String>{});
+	EXPECT_EQ(result, String::npos);
+
+	result = FindFirstNotOf(String{}, StringViewT<String>{}, String::npos);
+	EXPECT_EQ(result, String::npos);
```
MarcelHB, line 70: Writing one of these test cases, I was wondering whether the second check here makes any sense: https://github.com/gemrb/gemrb/blob/82e2b53b5fb6b0e60d2fe4dad964c609a0b3922c/gemrb/core/Strings/String.h#L75-L82

bradallred: it looks like it would yield a different return if it wasn't there: `return iter == s.end() ? STR::npos`. don't know if that matters.

bradallred: id check what the std lib does for its `std::string::find_first_not_of` and follow that.

gemrb/core/Strings/StringView.h
```diff
@@ -50,7 +50,7 @@ class StringViewImp {
 
 	// explicit because strlen is inefficient
 	explicit StringViewImp(CharT* cstr) noexcept
-	: StringViewImp(cstr, std::strlen(cstr)) {}
+	: StringViewImp(cstr, std::char_traits<CharT>::length(cstr)) {}
```
bradallred, line 53: could you just add something like `Traits = std::char_traits<CharT>` and use it where appropriate?

gemrb/core/Strings/StringView.h
```diff
@@ -110,14 +110,22 @@ class StringViewImp {
 		return *(data + i);
 	}
 
+	bool operator==(const StringViewImp<CharT>& other) const noexcept {
+		if (other.length() != length()) {
+			return false;
+		}
+
+		return std::char_traits<CharT>::compare(data, other.data, 0) == 0;
```
bradallred, line 118: this `0` is a mistake, no?

bradallred: also I believe `compare`  returns 0 when they are equal

gemrb/tests/core/Strings/Test_String.cpp
MarcelHB: It says: _Position of the found character or [std::string::npos](https://en.cppreference.com/w/cpp/string/basic_string/npos) if no such character is found._

But I was looking at it logically, to check whether this was supposed to mean `sv` maybe: Since `pos` is non-negative, so at a length of 0 of `s`, the first check has already taken care of it.

gemrb/core/Strings/StringView.h
MarcelHB: Yeah, that's something I was running over as well. [Docs ](https://en.cppreference.com/w/cpp/string/char_traits/compare) say:
> Compares the first `count` characters of the character strings `s1` and `s2`. The comparison is done lexicographically.
If `count` is zero, strings are considered equal. 

I am assuming _equal_ as _equal length_. I can check this later again but for what other reason would it reasonably be designed this way.

gemrb/tests/core/Strings/Test_String.cpp
bradallred: then we should probably also return `npos`

> the first check has already taken care of it.

For our known cases, probably. its template and `STR::size_type` doesn't have to be unsigned.

gemrb/core/Strings/StringView.h
MarcelHB: Ok, yes

gemrb/core/Strings/StringView.h
bradallred: its just saying if you pass a 0, no checking is done. you should pass one of the lengths.

bradallred: This is probably a good opportunity to add a test that compares two strings of the same length and different chars :)

MarcelHB: Indeed ...

bradallred:
not something to address here, but now that we are adding comparison to `StringView` we might eventually find ourselves in a position where some case insensitive string (`ResRef`, `ieVariable`) gets converted to a view and then a comparison somewhere is suddenly case sensitive. We might need to add traits to `FixedSizeString` for case insensitive comparison and pass that as a template parameter to `StringView`. `std::string` and friends already do this.

MarcelHB:
Ok, tests are good now:

* I don't know how to make `ASSERT_DEATH` work on MSVC-Windows, so I'm skipping this for now since it's tedious to find out.
* AppImage works locally, so no covert breaking by this, I hope.

> not something to address here, but now that we are adding comparison to `StringView` we might eventually find ourselves in a position where some case insensitive string (`ResRef`, `ieVariable`) gets converted to a view and then a comparison somewhere is suddenly case sensitive. We might need to add traits to `FixedSizeString` for case insensitive comparison and pass that as a template parameter to `StringView`. `std::string` and friends already do this.

I was also thinking of other reasonable comparison criteria while writing this. But yes, let's wait until there is a need.

lynxlynxlynx:
That's odd, there's no crash on master and we already have newer builds.

MarcelHB:
Rebase to the rescue

lynxlynxlynx:
Thanks!

PR #1918, opened 2023-09-28
Add StringMap

### LINKED ISSUE #1829: Check map types after switching to C++14
now that we are using `std::unordered_map` over custom hash maps, there is one downside to that: searching the map requires constructing a new key type. Not usually a big deal, but not ideal for a game engine.

for `std::map` c++14 has [overloads](https://en.cppreference.com/w/cpp/container/map/find) that avoid this, but for whatever reason `std::unordered_map` did not gain them until c++20.

I wonder if many of our `std::unordered_map`s would be better as `std::map`, since we probably are mostly just building them all at once and then just doing find operations. This is probably extra true if the key is a string, since you can compare many strings in the time it takes to hash just one.

we already have an issue (#1768) for moving to C++14

bradallred:
just a std::unordered_map wrapper with heterogeneous key lookup. Testing builds before merging.

resolves #1829


lynxlynxlynx:
Can you add test cases? Such classes are a perfect fit.

bradallred:
I'd like to, and plan on it, but I cant get tests to work for me.

I have gtest installed, cmake finds it, but when I run: `cmake -DDISABLE_WERROR=1 -DUSE_TESTS=ON ..` I dont get the output for running the tests and they indeed don't run. trying manually just yields:

```
Test project /Users/brad/Projects/gemrb/build
    Start 1: Test_gemrb_core
1/1 Test #1: Test_gemrb_core ..................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.01 sec
```

bradallred:
@lynxlynxlynx do you know why is sonar reporting issues that have nothing to do with these changes?

gemrb/core/CMakeLists.txt
```diff
@@ -174,7 +175,7 @@ IF (APPLE)
 		LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/gemrb/${PROJECT_NAME}.app/Contents/Frameworks
 	)
 
-	IF (BUILD_TESTING)
+	IF (USE_TESTS)
```
lynxlynxlynx, line 178: One more at gemrb/CMakeLists.txt:173

lynxlynxlynx:
Re sonar: not really. Is this branch based on latest master?

gemrb/core/CMakeLists.txt
bradallred: bah, yeah, trying to get tests to work without luck

bradallred:
yeah, its branched from the latest commit there

PR #1922, opened 2023-10-01
Plugin test cases: INI, 2DA, IDS

MarcelHB:
## Description

This time I wanted to try out what it takes to make simple plugin code testable. So here we are:

* INI, 2DA, IDS since test cases are easy to build and classes have an easy-to-use interface
* Simplistic approach that shouldn't be concerned too much with linking options and loading (although we need to check all runners now since the lib paths may differ)
* Some minor admin stuff around that
* I'll post some additional comments below at the code level

## Checklist


gemrb/tests/2DAImporter/Test_2DAImporter.cpp
```diff
@@ -0,0 +1,119 @@
...
+protected:
+	FileStream* stream;
+	p2DAImporter unit;
+public:
+	void SetUp() override {
+		this->stream = new FileStream{};
+
+		stream->Open(GetParam());
+		assert(unit.Open(stream));
+		// p2DAImporter deletes stream internally after use
```
MarcelHB, line 40: I dislike this design. I vote for RAII or a `std::unqiue_ptr` signature to make this clear at code level.

gemrb/tests/IDSImporter/Test_IDSImporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+		this->stream = new FileStream{};
+
+		stream->Open(GetParam());
+		assert(unit.Open(stream));
+		// IDSImporter deletes stream internally after use
+		this->stream = nullptr;
+	}
+};
+
+TEST_P(IDSImporter_Test, GetValueByString) {
```
MarcelHB, line 47: `IDSImporter` uses some homemade `Pair` struct internally. I'd refactor this into `std::pair`.

gemrb/tests/INIImporter/Test_INIImporter.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+TEST_F(INIImporter_Test, GetTagNameByIndex) {
+	EXPECT_EQ(StringView{"sectionA"}, unit.GetTagNameByIndex(0));
+	EXPECT_EQ(StringView{"sectionB"}, unit.GetTagNameByIndex(1));
+	EXPECT_EQ(StringView{"sectionC"}, unit.GetTagNameByIndex(2));
+	EXPECT_EQ(StringView{"sectionC"}, unit.GetTagNameByIndex(3));
+
+	// undefined: std::vector::operator[] over index
+	// EXPECT_EQ(StringView{}, unit.GetTagNameByIndex(4));
```
MarcelHB, line 55: The other loaders have checks enabled that return some default values in such a case. Maybe we should adopt this here, too.

bradallred: probably not a need. the collection should probably just be a `set` or something

gemrb/tests/IDSImporter/Test_IDSImporter.cpp
bradallred: probably not a need. the collection should probably just be a `set` or something

CMakeLists.txt
```diff
@@ -602,6 +602,34 @@ MACRO(ADD_GEMRB_PLUGIN plugin_name)
...
 ENDMACRO(ADD_GEMRB_PLUGIN)
 
+# TODO: think of a way to add additional libs for linking
+MACRO(ADD_GEMRB_PLUGIN_TEST plugin_name)
+	IF (BUILD_TESTING)
+		ADD_EXECUTABLE(Test_${plugin_name} ${ARGN})
+
+		TARGET_LINK_LIBRARIES(Test_${plugin_name} GTest::gtest GTest::gtest_main gemrb_core)
+
+		# copy test executables to where the core lib is
```
bradallred, line 612: I don't understand why we do this. seems like we are making/have mad more work for ourselves by not being able to just use the build dir. I've never had to manually juggle things in other cmake+ctest projects. Is it because we aren't using cpack?

MarcelHB: I've never used CPack, so I don't know.

It's just that plugins (and their tests now) end up in subdirectories where there is no `gemrb_core.dll` in Windows' lookup paths.

gemrb/tests/IDSImporter/Test_IDSImporter.cpp
MarcelHB: There are interfaces for lookup by integer, by string value and by index so I guess a set alone won't help all these cases; I did not check yet whether all of them are still in use.

gemrb/tests/INIImporter/Test_INIImporter.cpp
lynxlynxlynx: The other two have explicit default values while the INI format doesn't. So I don't think we should do anything.

MarcelHB: The concern is more about UB since whatever bad access may end up here, it may go unnoticed or not.

gemrb/tests/resources/INIImporter/sample.ini
```diff
@@ -0,0 +1,23 @@
+[sectionA]
+stringValue=abc
+;oldValue=def
+intValue=123
+floatValue=2.5
+noValue=
+garbage
+trueValue=true
+falseValue=FALSE
```
lynxlynxlynx, line 9: I think it would be useful to check for a duplicate key here as well.

gemrb/tests/IDSImporter/Test_IDSImporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+	EXPECT_EQ(unit.GetStringIndex(10), std::string{"arghxxl"});
+
+	EXPECT_EQ(unit.GetStringIndex(100), std::string{});
+}
+
+TEST_P(IDSImporter_Test, GetValueIndex) {
+	EXPECT_EQ(unit.GetValueIndex(0), 8);
+	EXPECT_EQ(unit.GetValueIndex(10), 45);
+
+	EXPECT_EQ(unit.GetValueIndex(100), 0);
```
lynxlynxlynx, line 76: This is a bit dodgy in the core design, but I'll open a separate issue.

gemrb/tests/INIImporter/Test_INIImporter.cpp
lynxlynxlynx: Currently both users ensure the lookup is within the realm of possibility. But the section counts are small, so it wouldn't really hurt if we asserted in the function itself too.

MarcelHB: We can also refactor this into using iterators instead.

lynxlynxlynx: Yeah, either works.

MarcelHB: I drop the iterator idea, that involves refactoring upto that abstract `DataFileMgr`.

gemrb/tests/INIImporter/Test_INIImporter.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+TEST_F(INIImporter_Test, GetKeyAsString) {
+	auto sectionA = StringView{"SectionA"};
+
+	EXPECT_EQ(StringView{"abc"}, (unit.GetKeyAsString(sectionA, StringView{"stringValue"})));
+	EXPECT_EQ(StringView{"abc"}, (unit.GetKeyAsString(sectionA, StringView{"Stringvalue"})));
+	EXPECT_EQ(StringView{"n/a"}, (unit.GetKeyAsString(sectionA, StringView{"strongValue"}, StringView{"n/a"})));
+	// first match
+	EXPECT_EQ(StringView{"123"}, (unit.GetKeyAsString(sectionA, StringView{"intValue"})));
```
lynxlynxlynx, line 91: I don't find this terribly intuitive — I'd expect the second key to override the value.

MarcelHB: Agree, needs some rework though

gemrb/tests/INIImporter/Test_INIImporter.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+	EXPECT_EQ(StringView{"xyz"}, (unit.GetKeyAsString(StringView{"sectionB"}, StringView{"anotherStringValue"})));
+}
+
+TEST_F(INIImporter_Test, GetKeyAsInt) {
+	auto sectionA = StringView{"SectionA"};
+
+	EXPECT_EQ(123, (unit.GetKeyAsInt(sectionA, StringView{"intValue"}, 0)));
+	EXPECT_EQ(123, (unit.GetKeyAsInt(sectionA, StringView{"Intvalue"}, 0)));
+	EXPECT_EQ(456, (unit.GetKeyAsInt(sectionA, StringView{"intValueX"}, 456)));
```
lynxlynxlynx, line 103: Why not consider this a bug? The key is not present, so one should be able to get a consistent result, not something else in case a similar key exists.

gemrb/tests/INIImporter/Test_INIImporter.cpp
MarcelHB: Ok, done. But only for keys inside of the same section. Later sections of the same name don't merge anything so they are available via iteration but not by direct query.

gemrb/core/DataFileMgr.h
```diff
@@ -38,13 +43,135 @@ namespace GemRB {
  * Abstract loader for .INI files
  */
 
+class KeyValueGroup {
+public:
+	using KeyValue = std::pair<std::string, std::string>;
```
lynxlynxlynx, line 48: unused

gemrb/core/Interface.cpp
```diff
@@ -2180,15 +2180,14 @@ bool Interface::InitializeVarsWithINI(const path_t& iniFileName)
...
-			//skip any existing entries. GemRB.cfg has priority
+	for (auto& group : *defaults) {
+		for (auto& pair : group) {
+			// skip any existing entries. GemRB.cfg has priority
+			const auto& key = pair.first;
 			auto lookup = vars.Get(key);
 			if (lookup == nullptr) {
-				ieDword defaultVal = defaults->GetKeyAsInt(tag, key, 0);
-				vars.Set(key, overrides->GetKeyAsInt(tag, key, defaultVal));
+				auto value = defaults->GetKeyAsInt(group.GetName(), key, 0);
```
lynxlynxlynx, line 2189: Maybe simpler as `auto value = atoi(pair.second);`?

gemrb/tests/resources/INIImporter/sample.ini
```diff
@@ -2,7 +2,7 @@
 stringValue=abc
 ;oldValue=def
 intValue=123
-intValue=0
+intValue=1
```
lynxlynxlynx, line 5: Rereading now how we parse, should we add another test and fix section detection? Anything without an equals sign is currently treated as one, not requiring being in brackets.

gemrb/tests/INIImporter/Test_INIImporter.cpp
lynxlynxlynx: That sounds treacherous. I think it should behave the same as keys, overriding any previous section (not merging).

gemrb/tests/resources/INIImporter/sample.ini
MarcelHB: Can you give me an example what you mean? I don't get it.

lynxlynxlynx:
```
[notASection
notASection]
notASection
```

MarcelHB: The `garbage` below doesn't for fall this. But I'll check the other stuff.

gemrb/core/Interface.cpp
MarcelHB: Yeah, just goes around the interface, will think again of this.

gemrb/core/Strings/String.h
```diff
@@ -109,6 +109,17 @@ void RTrim(STR& string, StringViewT<STR> chars = WHITESPACE_STRING_VIEW(STR))
 	}
 }
 
+template <typename STR>
+STR RTrimOf(const STR& string, StringViewT<STR> chars = WHITESPACE_STRING_VIEW(STR))
+{
+	auto pos = FindLastNotOf(string, chars);
```
bradallred, line 115: why not just implement in terms of `RTrim`?

```
template <typename STR>
 STR RTrimOf(STR string, StringViewT<STR> chars = WHITESPACE_STRING_VIEW(STR))
 {
    return RTrim(string, chars);
}
```

bradallred: I do realize that you aren't always needing to copy the entire string, but given how memory allocation works it's not really worth fussing over. The use case is going to have only a few bytes max that its loping off and ultimately the allocation will be the same.

gemrb/core/DataFileMgr.h
```diff
@@ -38,13 +43,134 @@ namespace GemRB {
  * Abstract loader for .INI files
  */
 
+class KeyValueGroup {
+public:
+	using KeyValueMap = std::unordered_map<std::string, std::string>;
```
bradallred, line 48: Why use `std::unordered_map` over `StringMap`?

MarcelHB: That's so new I haven't yet realized what it does.

gemrb/core/Strings/String.h
bradallred: I think naming to `RTrimCopy` might be a better name too.

lynxlynxlynx: Then an extra function like that is not needed at all, it would just be an alias.

gemrb/tests/resources/INIImporter/sample.ini
lynxlynxlynx: Ah, I see, it's handled outside AddLine.

gemrb/core/Strings/String.h
bradallred: thats not true, it creates a copy unlike `RTrim`

gemrb/plugins/IDSImporter/IDSImporter.h
```diff
@@ -24,20 +24,14 @@
...
 private:
-	struct Pair {
-		int val;
-		std::string str;
-		
-		Pair(int val, std::string str)
-		: val(val), str(std::move(str))
-		{}
-	};
+	using Pair = std::pair<int, std::string>;
```
bradallred, line 34: could you use `HeterogeneousStringKey`? It looks like we have a bunch of `stricmp` that can go away.

bradallred: bah, my mistake. that isn't part of the key. Maybe we need a sase insensitive `char_traits` after all.

gemrb/core/Strings/String.h
MarcelHB: Can do. `RTrim` just doesn't return anything.

gemrb/core/DataFileMgr.h
MarcelHB: I've changed that but the code now fails over this assertion that looks odd since lookup was negative but insertion wasn't done either:
https://github.com/gemrb/gemrb/blob/ec6f1282354625d5146e73fb77210d676342d6c8/gemrb/core/Strings/StringMap.h#L60-L62
Can continue tomorrow, hopefully.

gemrb/core/DataFileMgr.h
```diff
@@ -38,13 +42,127 @@ namespace GemRB {
...
+			return true;
+		}
+
+		auto valueEnd = FindLastNotOf(iniLine, " ");
+		if (valueEnd == std::string::npos) {
+			return true;
+		}
+
+		std::string value = SubStr(iniLine, valueStart, valueEnd - valueStart + 1).MakeString();
+		kvMap.Set(key, std::move(value));
```
bradallred, line 99: you can instead do `StringView value` and `kvMap.Set(key, value.MakeString())`

gemrb/core/DataFileMgr.h
bradallred: Interesting, what is `ins.first` pointing to?. It uses `CstrHashCI` and `CstrEqCI` to match nodes so let's examine where those might go wrong.

I see that `CstrHash` is erroneously terminating when encountering a '\0', but that would at worst generate hash collisions.

`CstrEqCI` is using `stricmp`, but should probably be using something else, otherwise equality will break for anything containing a '\0'. Also, `StringView` isn't guaranteed to be null terminated so that can explode. Does ASAN complain here for you?

bradallred: you can try rebasing on master

CMakeLists.txt
MarcelHB: For future reference, this is the layout of MSVC. 3rd party DLLs are not where the tests need them.
msvc

gemrb/core/DataFileMgr.h
MarcelHB: That's better

gemrb/core/DataFileMgr.h
bradallred: I just thought it was better readability

```
std::string value = SubStr(iniLine, valueStart, valueEnd - valueStart + 1).MakeString();
kvMap.Set(key, std::move(value));
```
vs
```
StringView value = SubStr(iniLine, valueStart, valueEnd - valueStart + 1);
kvMap.Set(key, value.MakeString());
```

bradallred:
> I just remembered that similar tests can later be done also on our CFG parser.

We also have `GemMarkupParser`

gemrb/core/Strings/StringView.h
```diff
@@ -101,6 +101,14 @@ class StringViewImp {
...
 	}
 
+	void clear() {
+		erase(0);
+	}
+	
+	void erase(size_type index) {
+		this->len = index;
+	}
+
```
bradallred, line 111: Why is this needed? `StringView` is almost never mutable.

gemrb/plugins/INIImporter/INIImporter.h
```diff
@@ -26,67 +26,40 @@
...
-	StringView GetKeyAsString(StringView Key, StringView Default) const;
-	int GetKeyAsInt(StringView Key, const int Default) const;
-	float GetKeyAsFloat(StringView Key, const float Default) const;
-	bool GetKeyAsBool(StringView Key, const bool Default) const;
-};
-
 class INIImporter : public DataFileMgr {
 private:
-	std::vector<INITag> tags;
+	std::vector<KeyValueGroup> tags;
```
bradallred, line 36: why are we using `vector` instead of `set`?

gemrb/plugins/INIImporter/INIImporter.h
```diff
@@ -26,67 +26,40 @@
...
+private:
+	template<typename T>
+	T GetAs(StringView Tag, StringView Key, const T Default) const {
+		auto result = find(Tag);
+		if (result != end()) {
+			return result->GetAs<T>(Key, Default);
+		}
+
+		return Default;
+	}
```
bradallred, line 62: this looks exactly like `StringMap::Get`. cant it be implemented in terms of that?

gemrb/core/DataFileMgr.h
MarcelHB: Ok. Thought I was missing something here.

gemrb/core/Strings/StringView.h
MarcelHB: You suggested using `RTrim` in `RTrimCopy`, so we need to ducktype this.

gemrb/plugins/INIImporter/INIImporter.h
MarcelHB: I guess that helps us represeting the order in which sections appear in the INI file, there is some comment inside of the code with edge cases.

bradallred: didn't realize we cared about order (seems odd tho)

gemrb/plugins/INIImporter/INIImporter.h
MarcelHB: The lookup is over the name of the tags in the list, so no StringMap. The rest is the proxy to the other class that does that.

gemrb/core/Strings/StringView.h
bradallred: Huh, I pulled the code down and there doesnt seem to be any purpose for `RTrimCopy`.

```diff
-		auto key = RTrimCopy(SubStr(iniLine, keyStart, equalsPos - keyStart));
+		auto key = SubStr(iniLine, keyStart, equalsPos - keyStart);
+		RTrim(key);
```

CMakeLists.txt
MarcelHB: Great, now the plugin tests are 100% wrong here but at least no more library loading failures. I'll leave it this way for now until we have some better approach to cover all uses and build systems.

gemrb/core/Strings/StringView.h
MarcelHB: By now, yes. When I started it, I took other paths until here. 😉 We can leave it, I personally like chaining, or rewrite it again.

CMakeLists.txt
lynxlynxlynx: At least disable these tests on windows then, otherwise we won't get any successful builds.

MarcelHB: I'm checking this since I want to rule out that there is actually something wrong now.

lynxlynxlynx: Ok. I remember before we ironed out vcpkg people used to have problems building/running with VS for a similar reason, it couldn't find the dlls. If you can make it work with the dlls near the targets, I see no problem with changing that aspect. They're all dumped together when packaging anyway.

MarcelHB:
Eh, no clue what unholy combo of MSVC, GTest and our code made `assert` derail the programs. I could not even properly see anything in MSVC debugging. With `NDEBUG` unset as well, everything is good.

lynxlynxlynx:
Ok, everything passes now. :) Any further plans (still marked as a draft) or shall we merge?

MarcelHB:
I try to address some open things. Opinions regarding my very first comment on top?

lynxlynxlynx:
No preference, but yeah, manual memory management should be avoided.

gemrb/core/Interface.cpp
MarcelHB: Found something better.

lynxlynxlynx:
Now it's just begging for `GameData::GetResourceStream`, `FileStream::OpenFile` and others to get converted to return a unique_ptr. But that's something for a separate PR.

MarcelHB:
Yes, step by step. I still have to investigate what's breaking the no-INI test run.

Also I was thinking whether `run-or-dump.sh` should return a bad exit code on failure.

lynxlynxlynx:
It already tries, but it appears gdb doesn't have any documented return codes. I don't know if this core dumping ever worked either.

MarcelHB:
> It already tries, but it appears gdb doesn't have any documented return codes. I don't know if this core dumping ever worked either.

Could be rewritten like:
```sh
if ! gemrb ...; then
  gdb ...
  exit 1
fi
```

MarcelHB:
For me it's Ok now.

bradallred:
> It already tries, but it appears gdb doesn't have any documented return codes. I don't know if this core dumping ever worked either.

Maybe its not what you are talking about, but we do get this on master:
```
...
[ResourceManager]: Searching for 'NUMTEST'...egmentation fault (core dumped)

warning: Can't open file /SYSV00000000 (deleted) during file-backed mapping note processing
[New LWP 8070]
[New LWP 8071]
[New LWP 8075]
[New LWP 8076]
[New LWP 8077]
[New LWP 8078]
[New LWP 8079]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `gemrb/gemrb -q --color 2 -c ../gemrb/GemRB.cfg.noinstall.sample'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005638fdaaa9fa in GemRB::ScrollView::SetVScroll(GemRB::ScrollBar*) ()
[Current thread is 1 (Thread 0x7f72135fb1c0 (LWP 8070))]
#0  0x00005638fdaaa9fa in GemRB::ScrollView::SetVScroll(GemRB::ScrollBar*) ()
#1  0x00005638fdaaafa2 in GemRB::ScrollView::ScrollView(GemRB::Region const&) ()
#2  0x00005638fdace565 in GemRB::Window::Window(GemRB::Region const&, GemRB::WindowManager&) ()
#3  0x00005638fdad2351 in GemRB::WindowManager::WindowManager(std::shared_ptr<GemRB::Video>, std::shared_ptr<GemRB::GUIFactory>) ()
#4  0x00005638fdb525cd in GemRB::Interface::Interface(GemRB::CoreSettings&&) ()
#5  0x00005638fda25de3 in main ()
```

gemrb/core/Interface.cpp
```diff
@@ -2157,20 +2163,17 @@ bool Interface::InitializeVarsWithINI(const path_t& iniFileName)
 	PluginHolder<DataFileMgr> ini = MakePluginHolder<DataFileMgr>(IE_INI_CLASS_ID);
 	FileStream* iniStream = FileStream::OpenFile(iniFileName);
 	// if filename is not set we assume we are creating defaults without an INI
-	bool opened = ini->Open(iniStream);
+	bool opened = iniStream != nullptr && ini->Open(std::unique_ptr<DataStream>{iniStream});
```
bradallred, line 2166: these two checks seem redundant. If they aren't then that implies `Open` cannot handle a nullptr, which it probably should.

gemrb/core/Interface.cpp
```diff
@@ -2157,20 +2163,17 @@ bool Interface::InitializeVarsWithINI(const path_t& iniFileName)
...
 	}
-	if (!opened || iniFileName[0] == 0) {
-		delete iniStream; // Open deletes it itself on success
-	}
 
 	PluginHolder<DataFileMgr> gemINI = MakePluginHolder<DataFileMgr>(IE_INI_CLASS_ID);
 	DataStream* gemINIStream = gamedata->GetResourceStream("defaults", IE_INI_CLASS_ID);
 
-	if (!gemINIStream || !gemINI->Open(gemINIStream)) {
+	if (!gemINIStream || !gemINI->Open(std::unique_ptr<DataStream>{gemINIStream})) {
```
bradallred, line 2176: ditto this one.

gemrb/plugins/2DAImporter/2DAImporter.cpp
```diff
@@ -32,11 +32,8 @@ static bool StringCompKey(const std::string& str, TableMgr::key_t key)
 
 TableMgr::index_t p2DAImporter::npos = TableMgr::npos;
 
-bool p2DAImporter::Open(DataStream* str)
+bool p2DAImporter::Open(std::unique_ptr<DataStream> str)
 {
-	if (str == NULL) {
-		return false;
-	}
```
bradallred, line 37: I guess this is why.

If we dont want to accept `nullptr`, why use a pointer at all? Why not `Open(DataStream& str)`?

MarcelHB: That's the alternative I was thinking about, who's supposed to own that?

bradallred: yeah, might be something for later. We'd have to look at what is going on. Seems like the file shouldn't need to persist past the `Open` call, but I wouldn't be surprise if that isn't universally true. But if some plugin does need to take ownership, it can always `std::move` it from within `Open` itself.

MarcelHB: Technically of course. I just try to have signatures that tell me, as a caller, what to expect w. r. t. even potential ownership transfer.

These three things here have their individual interface class each and previously did `delete`, and I guess in the long term we will drop all raw pointers of `DataStream`. That's the rationale.

bradallred: it can also just take a `const&` and then copy if need be. Thats assuming there is even a need to own the stream at all.

MarcelHB: `const` ref, if I'm getting this right, won't work since the regular operations mutate the stream.

Regardless of the ownership concept, somebody just needs to delete it unambiguously.

bradallred: you're right. Of course because of all that it seems like `Open` taking ownership is indistinguishable from any other way an `Open` call taking nonconst reference might manipulate the stream, including just closing it. So I don't see why that would be more objectionable than what we have now with unchecked pointers. We could use `&&` if that felt better. Does anything expect to read (or do anything) with a stream after passing it to `Open`? Seems like `Open` is a logical owner of the stream to me, but I haven't looked closely.

MarcelHB: In _these_ cases: no, nothing after `Open`.

bradallred: no matter, what you have is good enough for now. if we care later we can figure it out.

lynxlynxlynx: Yeah, I'll just merge now. Btw, do tests still don't work for you? All I had to do was rebuild with `-DUSE_TESTS=on`.

lynxlynxlynx:
Thanks!

gemrb/plugins/2DAImporter/2DAImporter.cpp
bradallred: they do now:

```bash
➜  build git:(master) ✗ ctest                                     
Test project /Users/brad/Projects/gemrb/build
    Start 1: Test_gemrb_core
1/4 Test #1: Test_gemrb_core ..................   Passed    0.62 sec
    Start 2: Test_2DAImporter
2/4 Test #2: Test_2DAImporter .................   Passed    0.30 sec
    Start 3: Test_IDSImporter
3/4 Test #3: Test_IDSImporter .................   Passed    0.12 sec
    Start 4: Test_INIImporter
4/4 Test #4: Test_INIImporter .................   Passed    0.13 sec

100% tests passed, 0 tests failed out of 4

Total Test time (real) =   1.18 sec
```

Issue #93, opened 2019-02-03
[feature] fully externalising char animations


lynxlynxlynx:
**Description**
This was a big ugly mess in the originals, except for pst, which used ini files. We should move everything over to the same system and abandon avatars.2da, unless it somehow remains relevant. EEs opted for the same thing.

The individual files can then contain all the exceptions and tweaks that are now sprinkled in CharAnimations. Several of the animation types will be obsolete, as we have a few just due to missing frames and/or orientations.

lynxlynxlynx:
If I read correctly, EEs ship with inis for all games now, even iwd2. If that's true, only the walkscale values would need verification for different games — luckily the scaling factors between them are nice.

Issue #92, opened 2019-02-03
[feature] rndbase bitmap based combat timing


lynxlynxlynx:
**Description**
the original used several bitmaps to decide when attacks happen, when to animate what and so on. Should also fix arrows sometimes appearing behind the shooter.

  * combat round timings are off: the times for weapons to launch and for animations to trigger and so forth http://gibberlings3.net/forums/index.php?showtopic=17946&page=8#entry177690 (RE notes) (initiative is the y coord, round timer x (combatcounter?))
    * http://www.shsforums.net/topic/54572-interesting-research-on-attacks-in-the-round/ (tobex thread) (iwd2: looks like one per attack count and one for launchers)

lynxlynxlynx:
This will also enable a proper implementation of the Swing() and SwingOnce() actions. Recoil probably as well, but it's not used anywhere.

Quoting Bubb:

> Given the current RNDBASE* of the character, Swing() and SwingOnce() starts a processing of the BMP, (x coord is current combat frame, y coord is character's m_speedFactor, which appears to be a randomized variable at the start of Swing() and SwingOnce()):

```
Color processing:
// Note: If cosmetic attacks are disabled via options, Teal is treated as Black.

Black =>
    if (isSwingOnce && alreadySwung) // Note: alreadySwung is stored on the action as the first int param
        return ACTION_DONE

    if (curSequence != SEQ_READY)
        // do SEQ_READY

    return ACTION_INTERRUPTABLE

Teal (cosmetic attack) => 
    if (isSwingOnce)
        return ACTION_NORMAL

    if (curSequence != SEQ_ATTACK)
        // do SEQ_ATTACK

    return ACTION_NORMAL

Light Grey =>
    if (curSequence != SEQ_ATTACK)
        // do SEQ_ATTACK

    if (isSwingOnce)
        alreadySwung = true

    return ACTION_NORMAL

Bright Red =>
    if (curSequence != SEQ_ATTACK)
        // do SEQ_ATTACK
    return ACTION_NORMAL

White (only appears in the final column) =>
    return ACTION_DONE

default =>
    return ACTION_NORMAL
```

> So, in brief, Swing() goes through a normal attack round with a randomized speed factor. It plays cosmetic attacks (if they aren't disabled) and does the normal amount of attack swings the character would do in a round.
> SwingOnce() is the same, but ignores cosmetic attacks and ends once the first real attack swing has been performed.

lynxlynxlynx:
[...]

PR #1931, opened 2023-11-21
Bug fixes

Dreamkins:
## Description
Hello! I'm currently playing through BG1 and encountering bugs from time to time. I will correct it if possible.

## Checklist


lynxlynxlynx:
Thanks, I'll take a look. You haven't documented the reasoning behind many of the changes, but let's address that per-commit. Some of the things have been fixed already (eg. Game::RestParty), since it seems you worked on older code.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -529,7 +529,7 @@ Actor::stat_t Actor::GetSafeStat(unsigned int StatIndex) const
 	if (StatIndex >= MAX_STATS) {
 		return 0xdadadada;
 	}
-	if (PrevStats) return PrevStats[StatIndex];
+	//if (PrevStats) return PrevStats[StatIndex];
```
lynxlynxlynx, line 532: definitely bad

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3338,7 +3338,7 @@ int Actor::NewStat(unsigned int StatIndex, ieDword ModifierValue, ieDword Modifi
 	return Modified[StatIndex] - oldmod;
 }
 
-int Actor::NewBase(unsigned int StatIndex, ieDword ModifierValue, ieDword ModifierType)
+int Actor::NewBase(unsigned int StatIndex, ieDwordSigned ModifierValue, ieDword ModifierType)
```
lynxlynxlynx, line 3341: The stats are words, so while this can solve the warnings, it's not really correct. It'd be better to add a new modifier type, eg. MOD_ADDITIVE_NEGATIVE and pass positive values. Now if there are legitimate high values of the stats, they would overflow.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4927,59 +4927,46 @@ void Actor::Turn(Scriptable *cleric, ieDword turnlevel)
...
-	if (turnlevel >= level + turnDeathLevelMod) {
+	if (casterTurnLevel >= targetTurnLevel + turnDeathLevelMod) {
 		if (evilcleric) {
 			Effect* fx = EffectQueue::CreateEffect(control_creature_ref, GEN_UNDEAD, 3, FX_DURATION_INSTANT_LIMITED);
 			if (!fx) {
 				fx = EffectQueue::CreateEffect(control_undead_ref, GEN_UNDEAD, 3, FX_DURATION_INSTANT_LIMITED);
 			}
 			if (fx) {
-				fx->Duration = core->Time.round_sec;
+				fx->Duration = core->Time.round_sec * caster->GetXPLevel(true);
```
lynxlynxlynx, line 4959: Nice catch, how did you get this value? The bg wiki is not the best resource, claiming a fixed 10 rounds, but the testing could have easily been bad (in the sense of testing the wrong variables).

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7457,7 +7444,17 @@ void Actor::ApplyModal(const ResRef& modalSpell)
...
+			if (!area) return;
+			std::vector<Actor*> targets = area->GetAllActorsInRadius(Pos, GA_NO_LOS|GA_NO_DEAD|GA_NO_NEUTRAL|GA_NO_ALLY|GA_NO_SELF|GA_NO_UNSCHEDULED, GetSafeStat(IE_VISUALRANGE));
+			for (const auto& target : targets) {
+				if (GameScript::ID_Alignment(this, AL_EVIL) || target->GetStat(IE_GENERAL) == GEN_UNDEAD) {
+					core->ApplySpell(modalSpell, target, this, 0);
+				}
+			}
+		} else {
+			core->ApplySpellPoint(modalSpell, GetCurrentArea(), Pos, this, 0);
+		}
```
lynxlynxlynx, line 7457: Why is this needed at all? This can all be done in the spell itself (and should be already). However it's not clear if turning paladins still works properly under this change.

gemrb/core/CharAnimations.cpp
```diff
@@ -2079,13 +2079,13 @@ void CharAnimations::AddMHRSuffix(ResRef& dest, unsigned char StanceID,
...
-			Cycle = 8 + Orient;
+			Cycle = Orient;
 			break;
 
 		case IE_ANI_CONJURE://ending
 			dest.Append("ca");
 			EquipData.Suffix = "ca";
-			Cycle = Orient;
+			Cycle = 8 + Orient;
 			break;
```
lynxlynxlynx, line 2089: Which animation / creature did you test with?

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -864,9 +864,11 @@ static void UpdateHappiness(Actor *actor) {
 	if (!core->HasFeature(GFFlags::HAPPINESS)) return;
 
 	ieWordSigned newHappiness = GetHappiness(actor, core->GetGame()->Reputation);
-	if (newHappiness == actor->PCStats->Happiness) return;
+	if (actor->PCStats) {
```
lynxlynxlynx, line 867: We can just return  here if it's not available, since the rest of the function doesn't make sense without the PCStats block.

gemrb/core/Projectile.cpp
```diff
@@ -250,6 +250,8 @@ void Projectile::Setup()
 		ZPos = ProHeights::Flying;
 	}
 
+	ZPos /= 2;
+
```
lynxlynxlynx, line 254: What projectile does this address? And the ZPos halving is mighty suspicious.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
```diff
@@ -4690,16 +4690,7 @@ int fx_cast_spell (Scriptable* Owner, Actor* target, Effect* fx)
...
-		// no deplete, no interrupt, caster or provided level
-		// ForceSpell doesn't have a RES variant, so there's more work
-		std::string tmp = fmt::format("ForceSpell([-1],{})", ResolveSpellNumber(fx->Resource));
-		Action *forceSpellAction = GenerateActionDirect(std::move(tmp), target);
-		if (fx->Parameter1 != 0) {
-			// override casting level
-			forceSpellAction->int1Parameter = fx->Parameter1;
-		}
-		Owner->AddActionInFront(forceSpellAction);
-		Owner->ImmediateEvent();
```
lynxlynxlynx, line 4702: What was the problem? This effect deals with a lot more use cases ...

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7331,7 +7331,7 @@ void Actor::UpdateModalState(ieDword gameTime)
 	//actually, iwd2 has autosearch, also, this is useful for dayblindness
 	//apply the modal effect about every second (pst and iwds have round sizes that are not multiples of 15)
 	// FIXME: split dayblindness out of detect.spl and only run that each tick + simplify this check
-	if (InParty && core->HasFeature(GFFlags::AUTOSEARCH_HIDDEN) && (third || (roundFraction % core->Time.defaultTicksPerSec == 0))) {
+	if (InParty && core->HasFeature(GFFlags::AUTOSEARCH_HIDDEN) && third && (roundFraction % core->Time.defaultTicksPerSec == 0)) {
```
lynxlynxlynx, line 7323: This is now bad for non-3ed games; nothing will check for hidden doors.

Dreamkins:
> Thanks, I'll take a look. You haven't documented the reasoning behind many of the changes, but let's address that per-commit. Some of the things have been fixed already (eg. Game::RestParty), since it seems you worked on older code.

I'm working on 0.9.2, since the latest version is very broken (interface buttons don't work, at least)

gemrb/includes/ie_stats.h
```diff
@@ -151,7 +151,7 @@ namespace GemRB {
 
 // can't walk or attack - helpless
 // FIXME: why is STATE_BERSERK on the list?
-#define STATE_CANTMOVE ((STATE_MINDLESS|STATE_STUNNED|_DEAD_STATES|STATE_DEACTIVATED)^(STATE_CHARMED|STATE_CONFUSED)) // 0x180fef
+#define STATE_CANTMOVE ((STATE_MINDLESS|STATE_STUNNED|_DEAD_STATES|STATE_DEACTIVATED)^(STATE_CHARMED|STATE_CONFUSED|STATE_PANIC|STATE_BERSERK)) // 0x180fef
```
lynxlynxlynx, line 154: justification? Clearly one can move while panicking or berserking.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7876,7 +7876,10 @@ bool Actor::AdvanceAnimations()
 
 bool Actor::IsDead() const
 {
-	return InternalFlags & IF_STOPATTACK;
+	if (InternalFlags & IF_REALLYDIED) return true;
+	if (Modified[IE_STATE_ID] & STATE_DEAD) return true;
+
+	return false;
```
lynxlynxlynx, line 7879: This is also bad without adjusting the users. The internal flags check will catch fewer states and the state check will catch some extra. For some that may be fine, but that's to be proven first. IF_STOPATTACK is clearly there with a specific intention, even if the function name is a bit deceptive.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: > Nice catch, how did you get this value? The bg wiki is not the best resource, claiming a fixed 10 rounds, but the testing could have easily been bad (in the sense of testing the wrong variables).

I'm not sure about this point, but the rest should be correct. Before the edits, turn undead threw one projectile at the caster himself.

lynxlynxlynx:
> I'm working on 0.9.2, since the latest version is very broken (interface buttons don't work, at least)

Can you report a bug with the details? bg1 still works fine here.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: > Why is this needed at all? This can all be done in the spell itself (and should be already). However it's not clear if turning paladins still works properly under this change.

As I already said, it cast turn undead at the caster himself. And this is not an effect with a sub-effect. The paladin made save throw from turn undead.

gemrb/core/CharAnimations.cpp
Dreamkins: > Which animation / creature did you test with?

Animation of a humanoid (human). In the original, he spins his arms in circles and at the end pushes his arms forward. Before the edits, he first did pushes with his hands, and at the end he did a circle.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: That's not necessarily bad if it's an aoe (which this is, it's using inareans.pro that ignores the center). But the code you removed was not for when paladins are turning, but when evil clerics are turning — they can turn paladins as well, besides the undead.

gemrb/core/Projectile.cpp
Dreamkins: > What projectile does this address? And the ZPos halving is mighty suspicious.

All arrows/bullets. They flew from the character's head. Division by two is visually correct.

gemrb/core/Scriptable/Door.cpp
```diff
@@ -324,13 +324,13 @@ void Highlightable::TryDisarm(Actor* actor)
...
 
-	int skill = actor->GetStat(IE_TRAPS);
+	int skill = actor->GetStat(IE_LOCKPICKING);
 	int roll = 0;
 	int bonus = 0;
 	int trapDC = TrapRemovalDiff;
 
 	if (core->HasFeature(GFFlags::RULES_3ED)) {
-		skill = actor->GetSkill(IE_TRAPS);
+		skill = actor->GetSkill(IE_LOCKPICKING);
```
lynxlynxlynx, line 333: These two are outright wrong even by the in-game description.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
Dreamkins: > What was the problem? This effect deals with a lot more use cases ...

Look at the cleric spell "find traps". This is an effect with delayed sub-effects (20 pieces). For each sub-effect, the character stopped and did a full cast animation.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -7280,7 +7280,7 @@ void Actor::UpdateActorState()
 		//IN BG1 and BG2, this is at the ninth frame... (depends on the combat bitmap, which we don't handle yet)
 		// however some critters don't have that long animations (eg. squirrel 0xC400)
 		if ((frameCount > 8 && currentFrame == 8) || (frameCount <= 8 && currentFrame == frameCount/2)) {
-			GetCurrentArea()->AddProjectile(attackProjectile, Pos, LastTarget, false);
+			GetCurrentArea()->AddProjectile(attackProjectile);
```
lynxlynxlynx, line 7272: Was the intention to launch it even if the creature got obliterated in the meanwhile or what? Wondering if LastTargetPersistent would work instead for whatever you want to achieve with this commit.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: > This is now bad for non-3ed games; nothing will check for hidden doors.

Spell "find traps" and modal find traps of rogues did this.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -9273,10 +9273,14 @@ bool Actor::UseItem(ieDword slot, ieDword header, const Scriptable* target, ieDw
...
+	}
+
+	if (pro) {
 		attackProjectile = pro;
-	} else { // launch it now as we are not attacking
-		GetCurrentArea()->AddProjectile(pro, Pos, tar->GetGlobalID(), false);
+		attackProjectile->MoveTo(tar->GetCurrentArea(), ranged ? Pos : tar->Pos);
+		attackProjectile->SetTarget(tar->GetGlobalID(), false);
 	}
+
```
lynxlynxlynx, line 9285: This chunk is very suspicious. For fake launches the attack projectile is pointing to freed memory.

gemrb/includes/ie_stats.h
Dreamkins: > justification? Clearly one can move while panicking or berserking.

Yes, before the edits, the frightened skeletons did not move.

gemrb/core/Projectile.cpp
lynxlynxlynx: It should be fixed in the attack projectile then, when it's created, not here where it affects also spells.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Somewhere the original check did not work for me. I looked at the reference in your code
```
if (ga_flags&GA_NO_DEAD) {
if (InternalFlags&IF_REALLYDIED) return false;
if (Modified[IE_STATE_ID] & STATE_DEAD) return false;
}
```

gemrb/core/Scriptable/Actor.cpp
Dreamkins: I have attack hits without damage before edits sometime.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: The new code should also work on evil clerics.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Ok.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Ok.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
lynxlynxlynx: Hmm, ok, needs a closer look then, thanks.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Not really, you explicitly exclude almost all good creatures (GA_NO_ALLY) when iterating over potential targets.

gemrb/core/Projectile.cpp
lynxlynxlynx: I highly doubt it. The casting offsets for the casting glow positions were hardcoded and the community managed to recover that fully, which we now implement. The spells are launched from the same points — if some aren't that's a bug that needs to be addressed. Hence the perpetual need for examples, which also help when code gets revisited.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: You mean visual hits or something else? Because the games animate fake attacks to make it seem more is going on. I think we don't implement that part though, it's basically #92.

What weapons and circumstances were involved?

gemrb/includes/ie_stats.h
lynxlynxlynx: Sorry, you're right — I misread the parentheses placement.

gemrb/core/Inventory.cpp
```diff
@@ -1929,35 +1929,33 @@ HCStrings Inventory::WhyCantEquip(int slot, int twohanded, bool ranged) const
...
-			if (slot>=SLOT_MELEE&&slot<=LAST_MELEE && (slot-SLOT_MELEE)&1) {
-				return HCStrings::NotInOffhand;
+			if (slot + 1 <= LAST_MELEE && !IsSlotEmpty(slot + 1)) {
+				return HCStrings::OffhandUsed;
 			}
-		} else if (slot == SLOT_LEFT) {
-			return HCStrings::NotInOffhand;
-		}
-		if (IsSlotBlocked(slot)) {
-		//cannot equip two handed while shield slot is in use?
```
lynxlynxlynx, line 1960: So what was the broken test case here? 
Now we don't complain with HCStrings::NoRangedOffhand and HCStrings::NotInOffhand any more.

gemrb/core/Inventory.cpp
```diff
@@ -1090,7 +1090,7 @@ int Inventory::FindTypedRangedWeapon(unsigned int type) const
 		//always look for a ranged header when looking for a projectile/projector
 		const ITMExtHeader *ext_header = itm->GetWeaponHeader(true);
 		int weapontype = 0;
-		if (ext_header && (ext_header->AttackType == ITEM_AT_BOW || ext_header->AttackType == ITEM_AT_PROJECTILE)) {
+		if (ext_header && (ext_header->AttackType == ITEM_AT_BOW || (ext_header->AttackType == ITEM_AT_PROJECTILE && !ext_header->Charges))) {
```
lynxlynxlynx, line 1093: Nice catch!

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Done rework it

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Done revert duration

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Done fix it

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Done rework it

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: It's probably still wrong, since I doubt people would report 10 rounds if it was only 1. So we had a bug before. Just needs some testing to see if it scaled with level (if so, it should probably only take the cleric/paladin level into account, so eg. fighter/clerics don't get unjust power).

gemrb/core/Projectile.cpp
Dreamkins: Done refix it

gemrb/core/Scriptable/Actor.cpp
Dreamkins: This can happen if ClearActions was called (which resets LastTarget) between the creation of the Attack projectile and the call to GetCurrentArea()->AddProjectile.

gemrb/core/Inventory.cpp
Dreamkins: It was really broken. Just try the tested combinations.

gemrb/core/Inventory.cpp
```diff
@@ -1929,35 +1929,33 @@ HCStrings Inventory::WhyCantEquip(int slot, int twohanded, bool ranged) const
...
+			}
 		}
 	}
 
+	//can't equip twohand if offhand equipped
 	if (twohanded) {
 		if (IWD2) {
-			if (slot>=SLOT_MELEE&&slot<=LAST_MELEE && (slot-SLOT_MELEE)&1) {
-				return HCStrings::NotInOffhand;
+			if (slot + 1 <= LAST_MELEE && !IsSlotEmpty(slot + 1)) {
```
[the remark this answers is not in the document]
Dreamkins: Didn't work on the last IWD2 weapon set.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Yes, I'll try to test it

gemrb/core/Inventory.cpp
lynxlynxlynx: Seems to work fine here, what's a broken example you found? I tried bg1 and bg2 (easier to tell, since it has all the strings).

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
lynxlynxlynx: Ok, but that was an implementation problem. If you believe slot can't ever be SLOT_LEFT here, just add an assert.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: ok, LastTargetPersistent is immune to that.

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
Dreamkins: This is an IWD2 check. IWD2 has 4 weapon sets with 4 offhand. SLOT_LEFT is not intended for IWD2.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Like I said at the start, I don't believe this whole chunk is needed at all (and what's up with the random alignment check?). What's actually broken now?

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
lynxlynxlynx: I know, but the function can still take it, it's not excluded like most of the rest of the slots.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Take a paladin and try to use turn undead. By the way, why is he from level 5 if he should be from level 3? The paladin will roll a saving throw against his own turn undead and the surrounding undead will not be affected. This effect has no sub-effect.

gemrb/core/Projectile.cpp
lynxlynxlynx: Testing a bit, it might have actually used the same position as spells. Try using Projectile::GetStartOffset to adjust Pos (no need to touch ZPos manually then) and see if it looks ok.

lynxlynxlynx:
I've untangled and pushed some of your changes, while we work out the rest. In the future please use rebase judiciously and amend commits on the fly.

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
Dreamkins: In IWD2, you can put a shield in the last set while wearing a two-handed weapon. My check works more reliably. I don't understand how SLOT_LEFT will help here.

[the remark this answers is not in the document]
Dreamkins: Oh, that's another check. But the stated problem is present.

gemrb/core/Inventory.cpp
Dreamkins: I can't repeat this now. Perhaps I messed something up myself. But the described problem with IWD2 is present.

gemrb/core/Projectile.cpp
Dreamkins: Ok

Dreamkins:
> I've untangled and pushed some of your changes, while we work out the rest. In the future please use rebase judiciously and amend commits on the fly.

Ok, i will try)

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
lynxlynxlynx: I'm not questioning your intent to fix the 4th set, I'm just saying you should add the assert so we don't have to think about that possibility ever again.

gemrb/core/Projectile.cpp
Dreamkins: start.y += ZPos/2 ? GetStartOffset static function, i cant read ZPos. Also, it's not very pretty code

lynxlynxlynx: Something like `attackProjectile->Pos += Projectile::GetStartOffset(this);`.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Done refix it

gemrb/core/Scriptable/Door.cpp
Dreamkins: My mistake. I always thought that lockpicking was used there.

gemrb/core/Projectile.cpp
Dreamkins: GetStartOffset returns negative values; if you add it, it raises the projectile even higher. You can subtract minus by minus, it seems true, but you still need to change the final position on the target, otherwise it flies much higher. Working with ZPos made editing easier.

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
Dreamkins: In IWD2 SLOT_LEFT is always 44 for any incoming slot value.

gemrb/core/Scriptable/Actor.cpp
Dreamkins: Okay, so be it, as long as it's not a cheat. ;)

gemrb/core/Projectile.cpp
lynxlynxlynx: ok, let's try the halving route then.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Hmm, I can reproduce this just walking on some random traps. The core issue is likely somewhere else though, this used to work fine. I can confirm it did in 0.9.1.

gemrb/core/Projectile.cpp
Dreamkins: Projectile::GetStartOffset has almost no effect on the height of the attacking projectile (+1/-1).
But have effect pos.y-=GetZPos(); in DrawTravel, it does not separate caster projectiles and attacking ones.
I think I need to work with ZPos in this situation.

Dreamkins: 1) GetStartOffset has almost no effect on projectile height (+1/-1)
2) The height of the projectile is affected by pos.y-=GetZPos(); in the DrawTravel function
3) With ZPos 0 projectile height on center of the character circle.
4) With default ZPos projectile height on above the character's head.

I think the problem is in ZPos.

lynxlynxlynx: Since there's no Z axis, ZPos is just an offset to Pos.y (your second point), that's why it was worth a shot. I was expecting Projectile::MoveTo to apply it already, because the caster should be set by then. So I don't understand what's different. Are you getting the normal or flying height assigned to ZPos? The latter is almost double the first, which could explain things ... does adding `attackProjectile->SFlags|= PSF_FLYING;` work too?

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: All the games but iwd2 had it. That's why the fixme is there, since we're doing two things at once; now the game flag is effectively useless.

gemrb/core/Inventory.cpp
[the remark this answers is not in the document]
lynxlynxlynx: That doesn't matter, the reader of the code shouldn't have to worry about it.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: I'm resolving this note, but this change is not going in. If you encounter where you had a faulty check again, document it and we can look at the root problem.

gemrb/core/Projectile.cpp
Dreamkins: Attack projectiles already have the PSF_FLYING flag set, it reads it from bullet.pro. Resetting this flag reduces ZPos from ProHeights::Flying to ProHeights::Normal, which looks good for normal sized characters, but is still bad for hobbits and dwarves.

lynxlynxlynx: Aren't smaller/bigger creatures worse off anyway, since we use a constant? Yuan-ti archers are the biggest ranged unit that comes to mind (it'd be amazing if they used special ammo just to compensate).

Dreamkins: I looked at the original and there is the height of the arrow for hobbits and dwarves
bad too)

lynxlynxlynx: We can fix this once we externalize all the animation data and can just add another key. Now only pst uses it. #93 
Ok, we could add another table in the interim if anyone cared that much.
For now then it's simpler if we just remove the PSF_FLYING flag, since it's a smaller change and you say it works as well. Thanks for the investigation!

Dreamkins: I do what I can)

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: Ok, found the issue — in the effect itself. It works fine unless you get poisoned the same way twice, which is not unlikely when getting jumped by several spiders.

gemrb/core/Scriptable/Actor.cpp
lynxlynxlynx: I tried both a cleric and a paladin and both turned skeletons fine. It was ToB, so the level difference was enough that when they were affected (only one managed a save once in about 10 rolls), they got destroyed. Also means high saves on the paladin already. Maybe you were unlucky. Maybe something has been fixed since. Either way I don't see anything needing fixing here.

bradallred:
I’m going to go out on a limb and say this is because you are using 0.9.1 GUIScripts with 0.9.2. If not then there should be a python error you can open a bug about. On Nov 21, 2023, at 08:28, Dreamkins ***@***.***> wrote:﻿

Thanks, I'll take a look. You haven't documented the reasoning behind many of the changes, but let's address that per-commit. Some of the things have been fixed already (eg. Game::RestParty), since it seems you worked on older code.

I'm working on 0.9.2, since the latest version is very broken (interface buttons don't work, at least)

Dreamkins:
> I’m going to go out on a limb and say this is because you are using 0.9.1 GUIScripts with 0.9.2. If not then there should be a python error you can open a bug about. On Nov 21, 2023, at 08:28, Dreamkins ***@***.***> wrote:﻿ Thanks, I'll take a look. You haven't documented the reasoning behind many of the changes, but let's address that per-commit. Some of the things have been fixed already (eg. Game::RestParty), since it seems you worked on older code. I'm working on 0.9.2, since the latest version is very broken (interface buttons don't work, at least) —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

How can this be if it comes as part of the source code and is compiled along with the project?

lynxlynxlynx:
It happens when people `make install` over existing installations without uninstalling first. Now and then we remove a file, reorder things and then you get some files from several versions at once.

For anyone building themselves, this is all completely unnecessary, since you can run from the build dir without installing.

gemrb/plugins/FXOpcodes/FXOpcodes.cpp
lynxlynxlynx: The casting problem has been gone, but the spell payload itself indeed does not work. Probably a projectile thing, still investigating.
This code change could not have fixed it though; the spell uses Parameter2 == 1 and self-targeting, so this branch doesn't get used.

Dreamkins:
I am compiling with MSVC according to your instructions.
In commit 2e78ef6, the wrong save file began to load.
In commit 9f14c25, the interface buttons stopped working.

lynxlynxlynx:
Do you get the same results if you try the buildbot binary?

Dreamkins:
> Do you get the same results if you try the buildbot binary?

No, this does not happen in the buildbot binary(

lynxlynxlynx:
Well, that's also using msvc to build, so something is broken in your setup. Maybe restart with a clean build dir.

Dreamkins:
I'm guessing it's a python issue. After a clean installation the game does not start with the following message:
```
[ResourceManager/WARN]: Invalid path given: C:\gamesPython path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = 'C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\gemrb\\gemrb.exe'
  sys.base_prefix = ''
  sys.base_exec_prefix = ''
  sys.platlibdir = 'lib'
  sys.executable = 'C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\gemrb\\gemrb.exe'
  sys.prefix = ''
  sys.exec_prefix = ''
  sys.path = [
    'C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\gemrb\\python310_d.zip',
    '.\\DLLs',
    '.\\lib',
    'C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\gemrb',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00004dec (most recent call first):
  <no Python frame>
```

I have to add environment variables to CMakeSettings_schema.json
```
	"environments": [
		{
			"PYTHONHOME": "C:\\workspace\\bgtb\\vcpkg\\packages\\python3_x64-windows\\tools\\python3",
			"PYTHONPATH": "C:\\workspace\\bgtb\\vcpkg\\packages\\python3_x64-windows\\tools\\python3"
		}
	]
```

lynxlynxlynx:
You shouldn't need to set either variable; that itself can cause problems. Do you have several pythons installed?

Dreamkins:
> You shouldn't need to set either variable; that itself can cause problems. Do you have several pythons installed?

No, it is included in vcpkg and cmake sees it
```
1> [CMake] The package python3 is compatible with built-in CMake targets:
1> [CMake] 
1> [CMake]     find_package(Python3 COMPONENTS Development REQUIRED)
1> [CMake]     target_link_libraries(main PRIVATE Python3::Python)
1> [CMake] 
1> [CMake] The package python3 provides a python interpreter that supports virtual environments:
1> [CMake] 
1> [CMake]     >tools\python3\python3.10 -m venv c:\path\to\venv
1> [CMake]     >set VIRTUAL_ENV=c:\path\to\venv
1> [CMake]     >set PATH=c:\path\to\venv\bin;%PATH%
1> [CMake]     >set PYTHONHOME=
1> [CMake] 
1> [CMake]     See https://docs.python.org/3/library/venv.html for more details.
```

lynxlynxlynx:
In the package we copy over the modules dir from the python distribution, since that's its standard lib that wouldn't be present otherwise ("encodings" is one of those modules) . Shouldn't be needed in a local install though, but I guess only if it's in a standard location.

What problems do you get when you set the paths?

Dreamkins:
> In the package we copy over the modules dir from the python distribution, since that's its standard lib that wouldn't be present otherwise ("encodings" is one of those modules) . Shouldn't be needed in a local install though, but I guess only if it's in a standard location.
> 
> What problems do you get when you set the paths?

I didn't set any path. What kind of path do you mean?

lynxlynxlynx:
PYTHONHOME & PYTHONPATH

Dreamkins:
> PYTHONHOME & PYTHONPATH

Here is my CMakeSettings.json now
```
{
	"environments": [
		{
			"PYTHONHOME": "C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\vcpkg_installed\\x64-windows\\tools\\python3",
			"PYTHONPATH": "C:\\workspace\\bgtb\\gemrb\\out\\build\\x64-Debug\\vcpkg_installed\\x64-windows\\tools\\python3"
		}
	],
	"configurations": [
		{
			"name": "x64-Debug",
			"generator": "Ninja",
			"configurationType": "Debug",
			"inheritEnvironments": [ "msvc_x64_x64" ],
			"buildRoot": "${projectDir}\\out\\build\\${name}",
			"installRoot": "${projectDir}\\out\\install\\${name}",
			"cmakeCommandArgs": "",
			"buildCommandArgs": "",
			"ctestCommandArgs": "",
			"cmakeToolchain": "C:/workspace/bgtb/vcpkg/scripts/buildsystems/vcpkg.cmake"
		},
		{
			"name": "x64-Release",
			"generator": "Ninja",
			"configurationType": "RelWithDebInfo",
			"buildRoot": "${projectDir}\\out\\build\\${name}",
			"installRoot": "${projectDir}\\out\\install\\${name}",
			"buildCommandArgs": "",
			"ctestCommandArgs": "",
			"inheritEnvironments": [ "msvc_x64_x64" ],
			"cmakeToolchain": "C:/workspace/bgtb/vcpkg/scripts/buildsystems/vcpkg.cmake"
		}
	]
}

```

The game starts, but the wrong saves are loaded (I click on one, another one loads) and the interface buttons do not work.

lynxlynxlynx:
I have nothing to add, sorry. This is the invocation for our build bot, but nothing really stands out and your problems are with the environment anyway:
https://github.com/gemrb/gemrb/blob/master/.appveyor.yml#L86

Dreamkins:
It looks like my build continues to reference external python, but yours can work without python installed.

lynxlynxlynx:
In the package we ship a stripped down version of python, so users don't have to install it themselves. Python is an unavoidable dependency.

Dreamkins:
Lol everything works in release, but not in debug.

lynxlynxlynx:
We copy the files in the cmake file under platforms/windows/ . I guess msvc doesn't run cmake twice, so since I think the bin dir is different for debug and other build modes, only the first one configured will be fine. Patches welcome, this is too annoying to test through a build bot.

Dreamkins:
The problem may not be in python at all. Try the MSVC debug build. You can build it via buildbot if you don't have MSVC.

lynxlynxlynx:
Well yeah, that's what I said, I suspect it's a build system issue. Though I don't see how msvc could really get a proper release and debug build without running cmake twice.

Dreamkins:
> Well yeah, that's what I said, I suspect it's a build system issue. Though I don't see how msvc could really get a proper release and debug build without running cmake twice.

This does not correspond to the fact that everything works in 0.9.2. 
In my experience, this is more likely to do with an uninitialized variable.

lynxlynxlynx:
Both debug and release? I find it unlikely with your current symptoms, but have you even checked if the python stdlib was copied to your debug build path? 
Anyway, please open a new issue or PR, since it's unrelated to the work done here. I've now pushed a fix for the last of the identified problems.

PR #1932, opened 2023-11-21
Fix build for certain SDL Video versions

DarkMalex:
## Description
Project fails to build with SDL Video for versions between  >2.0.18 and <2.0.22.

When trying to build with SDL 2.0.18 I get the following error:
```
lcc: "gemrb/gemrb/plugins/SDLVideo/SDL20Video.cpp", line 597: error #167:
          argument of type "const SDL_Color *" is incompatible with parameter
          of type "const int *"
    reinterpret_cast<const SDL_Color*>(colors.data()),
    ^
``` 

This happens because prior to libsdl-org/SDL@b7885abc449250fc733a756239d670394d01fe62 `SDL_RenderGeometryRaw()` passes colors as `int*` instead of `SDL_Color*`.

We could simply change the check in `SDL20VideoDriver::DrawRawGeometry()` from 
`#if SDL_VERSION_ATLEAST(2, 0, 18)`
to
`#if SDL_VERSION_ATLEAST(2, 0, 22)`
or check the specific way we pass color based on SDL Video version.

Big thanks to @a1batross for figuring this out!

## Checklist

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -589,7 +589,7 @@ void SDL20VideoDriver::DrawRawGeometry(
 		SDL_SetRenderDrawBlendMode(renderer, SDL_BLENDMODE_BLEND);
 	}
 
-	static_assert(sizeof(int) == sizeof(SDL_Color));
+	static_assert(sizeof(int) == sizeof(SDL_Color), "Incompatible types to cast");
 
 	SDL_RenderGeometryRaw(
 		renderer,
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -589,7 +589,9 @@ void SDL20VideoDriver::DrawRawGeometry(
 		SDL_SetRenderDrawBlendMode(renderer, SDL_BLENDMODE_BLEND);
 	}
 
+	#if !SDL_VERSION_ATLEAST(2, 0, 22)
 	static_assert(sizeof(int) == sizeof(SDL_Color), "Incompatible types to cast");
+	#endif
 
 	SDL_RenderGeometryRaw(
 		renderer,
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, LGTM! @bradallred any preference on which way this simple matter is resolved?

gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -594,7 +594,11 @@ void SDL20VideoDriver::DrawRawGeometry(
 		nullptr,
 		vertices.data(),
 		2 * sizeof(float),
+		#if SDL_VERSION_ATLEAST(2, 0, 22)
 		reinterpret_cast<const SDL_Color*>(colors.data()),
+		#else
+		reinterpret_cast<const int*>(colors.data()),
```
bradallred, line 600: could you add this here: `static_assert(sizeof(int) == sizeof(SDL_Color));`

bradallred (changes requested):
This is fine, but let's explicitly fail to compile when things wont work.

gemrb/plugins/SDLVideo/SDL20Video.cpp
DarkMalex: I'm not sure I understand correctly. How do I put a block declaration in an argument list? You meant after or before SDL_RenderGeometryRaw() call?

DarkMalex: Done!
fad94a78f8c28bf01d64219aaf391f5561f3dfcd

DarkMalex:
I was able to build project with the latest commit, but it broke AppVeyor build. Mb revert it to the way it was?

lynxlynxlynx:
> error C2429: language feature 'terse static assert' requires compiler flag '/std:c++17'

You need to add a second parameter with the error message, we're still on c++14. Eg. from some other code:
`static_assert(F::is_integer, "From must be integral");`

lynxlynxlynx:
Thanks!

lynxlynxlynx:
Turns out this was actually changed in SDL 2.0.20 already (that's why our coverity build failed):
https://github.com/gemrb/gemrb/actions/runs/6959260102/job/18936094507#step:6:2321

I'll fix it manually.

PR #1936, opened 2023-11-26
Install does not copy the Python lib directory

Dreamkins:
## Description
Install does not copy the Python lib directory.


lynxlynxlynx:
Thanks!

Issue #1937, opened 2023-11-26
 Problems in MSVC Debug build


Dreamkins:
#### Bug description
BG1. In MSVC, in the Debug build, the interface buttons do not work, the Journal partially does not work, and the wrong save files are loaded. Everything works fine in Release. This creates problems when debugging.
The journal did not work also in 0.9.2.
In commit https://github.com/gemrb/gemrb/commit/2e78ef6c5576c58c3e3295ae7a2c655f07534054, the wrong save file began to load.
In commit https://github.com/gemrb/gemrb/commit/9f14c25007a01e743ad97f30056eb4fde5ca56c8, the interface buttons stopped working.

lynxlynxlynx:
Typically it's the other way around and release builds don't work due to reliance on some side effects in asserts (that get optimized out). When you say the buttons don't work for you, what do you mean and which buttons?

Dreamkins:
![Screenshot 2023-11-26 131432](https://github.com/gemrb/gemrb/assets/119408536/905c64d7-fe21-4257-81c1-12dde255b410)
![Screenshot 2023-11-26 131342](https://github.com/gemrb/gemrb/assets/119408536/7286bb11-fbac-4b24-a471-2ff28f00888c)
![Screenshot 2023-11-26 131309](https://github.com/gemrb/gemrb/assets/119408536/cd70b828-8509-4f53-864b-e9eb40407247)

Dreamkins:
Wrong save loaded
![Uploading Screenshot 2023-11-26 131759.png…]()
![Uploading Screenshot 2023-11-26 131821.png…]()

Dreamkins:
In my experience this may be due to an uninitialized variable. In the Release it is cleared to 0, and garbage remains in the Debug.

Dreamkins:
It's very strange, but that's how it works
![Screenshot 2023-11-26 163217](https://github.com/gemrb/gemrb/assets/119408536/a6a5b74e-4e1c-4484-9678-fc2f44e4d2a6)
![Screenshot 2023-11-26 163313](https://github.com/gemrb/gemrb/assets/119408536/53cc836d-8477-4165-a6ef-a2a831946886)

Dreamkins:
```
	operator StringView() const noexcept {
		return StringView(str, len);
	}
```
It looks like the object is deleted by the destructor before it can be returned.

Dreamkins:
[...]

PR #1938, opened 2023-11-26
Fix lost data on string convertion

Dreamkins:
## Description
Issue https://github.com/gemrb/gemrb/issues/1937

This fixes all the problems. I couldn't think of anything better, but it works.

### DIFF
--- a/gemrb/plugins/GUIScript/GUIScript.cpp
+++ b/gemrb/plugins/GUIScript/GUIScript.cpp
@@ -660,7 +660,7 @@ static PyObject* GemRB_Table_GetValue(PyObject* self, PyObject* args)
 	auto GetIndex = [&tm](PyObject* obj, bool isRow) -> TableMgr::index_t {
 		if (PyUnicode_Check(obj)) {
 			auto str = PyString_AsStringView(obj);
-			return isRow ? tm->GetRowIndex(str.ToStringView()) : tm->GetColumnIndex(str.ToStringView());
+			return isRow ? tm->GetRowIndex(str) : tm->GetColumnIndex(str);
 		} else if (PyLong_Check(obj)) {
 			return static_cast<TableMgr::index_t>(PyLong_AsLong(obj));
 		}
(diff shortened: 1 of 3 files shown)


bradallred:
I can’t examine the code because I’m on mobile, but this doesn’t seem like the correct fix. PyStringAsStringView should just return a StringView.On Nov 26, 2023, at 08:23, sonarcloud[bot] ***@***.***> wrote:﻿
Kudos, SonarCloud Quality Gate passed!    
  0 Bugs
  0 Vulnerabilities
  0 Security Hotspots
  0 Code Smells
 No Coverage information
 0.0% Duplication

Dreamkins:
> I can’t examine the code because I’m on mobile, but this doesn’t seem like the correct fix. PyStringAsStringView should just return a StringView.>

Look at Issui please

lynxlynxlynx:
I agree, it's an unwieldy fix.  And since there are no known problems with other encodings any more, what @bradallred  suggested is the way to go. The TODO in PyStringAsStringView and PyStringWrapper can also be removed then. It will also enable making that StringView operator explicit.

bradallred:
I did. C++ garantes that temporaries are valid until the end of an expression. Anywhere we are violating that can be addressed by saving the temporary to a variable like you are doing, but you still shouldn’t need to change the conversion operator.We shouldn’t have to make two function calls to get what we need from the PyObject.On Nov 26, 2023, at 08:52, Dreamkins ***@***.***> wrote:﻿

I can’t examine the code because I’m on mobile, but this doesn’t seem like the correct fix. PyStringAsStringView should just return a StringView.>

Look at Issui please

Dreamkins:
In my opinion, the conversion operator is dangerous, since it is easy to pass the result by reference and not notice it. I returned the conversion statement but the additional variables remained to hold the result.

bradallred:
what is the value of your `core->config.SystemEncoding`?

Dreamkins:
> what is the value of your `core->config.SystemEncoding`?
windows-1251

Dreamkins:
> I think we can close this. @Dreamkins is there any remaining issue you are having that this resolves for you?

The journal still doesn't work for me on your fixes.

bradallred:
which of the fixes in this PR are relevant to that?

Dreamkins:
> which of the fixes in this PR are relevant to that? I just fixed another, maybe that was it?

My PR fixes it together with buttons.

Dreamkins:
> that didn't answer the question. Is it working with that last change then?

The buttons work with the latest changes, but the Journal is completely empty.

Dreamkins:
Yes, with the latest changes to the Joutnal it worked!

bradallred:
I've opened a new issue for you. I'm closing this PR. thanks.

PR #1941, opened 2023-11-27
Size of colors token may be 3

Dreamkins:
## Description
Size of colors token may be 3. Caught an overflow in BG2.

Screenshot 2023-11-27 212447
Screenshot 2023-11-27 213741

### DIFF
--- a/gemrb/core/GUI/TextSystem/GemMarkup.cpp
+++ b/gemrb/core/GUI/TextSystem/GemMarkup.cpp
@@ -25,11 +25,11 @@ namespace GemRB {
 static Color ParseColor(const String& colorString)
 {
 	Color color = ColorWhite;
-	uint8_t values[4] = {0, 0, 0, 0};
+	uint8_t values[4] = { 0, 0, 0, 0xff };
 
 	// That's a short way to interpret 4 groups of %02hhx,
 	// since `swscanf` does not work on our string type unless converting first
-	for (uint8_t i = 0; i < 4; ++i) {
+	for (uint8_t i = 0; i < colorString.size() / 2; ++i) {
 		uint8_t value = 0;
 
 		for (uint8_t j = 0; j < 2; ++j) {


Dreamkins:
You can leave support with or without an alpha channel.

lynxlynxlynx:
We actually never pass the alpha. Come to think of it, would translucent text have any use? Effectively it's just a different way to arrive at another opaque color. Can't think of anything useful, but just padding the input color would work as well here.

gemrb/core/GUI/TextSystem/GemMarkup.cpp
```diff
@@ -25,11 +25,11 @@ namespace GemRB {
 static Color ParseColor(const String& colorString)
 {
 	Color color = ColorWhite;
-	uint8_t values[4] = {0, 0, 0, 0};
+	uint8_t values[4] = { 0, 0, 0, 0xff };
 
 	// That's a short way to interpret 4 groups of %02hhx,
 	// since `swscanf` does not work on our string type unless converting first
-	for (uint8_t i = 0; i < 4; ++i) {
+	for (uint8_t i = 0; i < colorString.size() / 2; ++i) {
```
bradallred, line 32: I don't understand why `/ 2` is here. shouldn't `colorString.size()` be the number of nibbles represented in the string? I guess not, but I dont know what `colorString` looks like, is it padded with `0x00` for some reason? Maybe we should fix that.

Dreamkins: Screenshot 2023-11-29 224227

bradallred: right, so it is just the number of nibbles. we should iterate them all we just have some funky offset voodoo that we can probably get rid of instead.

bradallred:
I've refactored `ParseColor` to handle arbitrary string length.

PR #1956, opened 2023-12-07
Feature minimize windows misprint. PC Panel start minimized.

Dreamkins:
Feature minimize windows misprint. PC Panel start minimized.


Dreamkins:
wrong fix

PR #1960, opened 2023-12-08
IWD monster summoning amount fix

Dreamkins:
IWD monster summoning amount fix


lynxlynxlynx:
You have the parameter order reversed, but this should be fixed differently anyway. Just add EFFECT_DICED to the effect flags and it will be taken care of automatically.

lynxlynxlynx:
Thanks!

Issue #1958, opened 2023-12-07
Spells from scrolls not worked 


Dreamkins:
#### Bug description
Spells from scrolls not worked (IWD least)

Acton is null. Point spell similar.
![Screenshot 2023-12-07 114811](https://github.com/gemrb/gemrb/assets/119408536/c94a99fa-dbe1-457e-8596-b7d42eb4a693)

If action not null then action will be removed. May be It's better to swap these lines.
![Screenshot 2023-12-07 114843](https://github.com/gemrb/gemrb/assets/119408536/5e30ab77-3dfe-4381-8f86-a158d752bee3)

GemRB version (check as many as you know apply)
- [x] master as of this issue

lynxlynxlynx:
We need to add ForceSpellRes to our gemaction.ids for pst and iwd1, I didn't notice they don't have it. Or use it stealthily, by using ForceSpell directly.

And good catch with UseItem as well, thanks!

Dreamkins:
> We need to add ForceSpellRes to our gemaction.ids for pst and iwd1, I didn't notice they don't have it.

I added them, but the parser misses the first letter here
```
			case 's': //String
			{
				SKIP_ARGUMENT();
				src++;
```
You can comment out src++; and then it works, but won't it affect something else?

PR #1961, opened 2023-12-08
 Spells from scrolls

Dreamkins:
[Spells from scrolls issue](https://github.com/gemrb/gemrb/issues/1958)

### DIFF
--- a/gemrb/core/GameScript/Actions.cpp
+++ b/gemrb/core/GameScript/Actions.cpp
@@ -6864,8 +6864,8 @@ void GameScript::UseItemPoint(Scriptable* Sender, Action* parameters)
 		return;
 	}
 
-	act->UseItemPoint(Slot, header, parameters->pointParameter, flags);
 	Sender->ReleaseCurrentAction();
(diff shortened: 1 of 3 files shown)


gemrb/core/GameScript/Actions.cpp
```diff
@@ -6814,8 +6814,8 @@ void GameScript::UseItem(Scriptable* Sender, Action* parameters)
 		return;
 	}
 
-	act->UseItem(Slot, header, tar, flags);
 	Sender->ReleaseCurrentAction();
+	act->UseItem(Slot, header, tar, flags);
```
lynxlynxlynx, line 6818: This change should be applied to GameScript::UseItemPoint as well and ideally this PR would be in 3 commits, for the 3 different things it does.

lynxlynxlynx:
Thanks! Can you rebase, so there's no extra commits (the first two)? It's easiest to do `git rebase -i HEAD~5` (on this branch) and then you can just delete the first two lines. Git will do the rest and clean up the history, all you have to do then is force-push the branch again.

PR #1964, opened 2023-12-10
Fix PST Black Rat animations

Dreamkins:
Fix PST blackr.cre (are0301.are) animations


lynxlynxlynx:
Nice, thanks! Hard to believe this went past all our testers before, but here we are.

PR #1966, opened 2023-12-11
Fix creature spawn script name

Dreamkins:
Empty critter script name can remove worked cre script name. 
As a result, PST script 1500CS1.bsc step MoveToObject("Hargrim") does not work because it doesn’t find a character with this script name.

### DIFF
--- a/gemrb/core/IniSpawn.cpp
+++ b/gemrb/core/IniSpawn.cpp
@@ -729,7 +729,9 @@ void IniSpawn::SpawnCreature(const CritterEntry& critter) const
 	cre->SetPosition(critter.SpawnPoint, true);
 	cre->SetOrientation(ClampToOrientation(critter.Orientation), false);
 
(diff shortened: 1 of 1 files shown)


gemrb/core/IniSpawn.cpp
```diff
@@ -729,7 +729,9 @@ void IniSpawn::SpawnCreature(const CritterEntry& critter) const
 	cre->SetPosition(critter.SpawnPoint, true);
 	cre->SetOrientation(ClampToOrientation(critter.Orientation), false);
 
-	cre->SetScriptName(critter.ScriptName);
+	if (!critter.ScriptName.empty()) {
+		cre->SetScriptName(critter.ScriptName);
+	}
```
lynxlynxlynx, line 734: Please add a comment explaining why this is needed and the test case, since it's not really intuitive (and as such at risk of being removed later down the line).

lynxlynxlynx:
Thanks!

PR #1967, opened 2023-12-13
Support for Tracy

MarcelHB:
## Description

RPi 4 performance is again down the drain a little. And also, profiling is still bad in general this age:

* AMD/Intel tools only work on their platforms
* AMD µProf is turning into a more malfunctioning mess (for GCC) with every release
* `gprof` dislikes dynamic linking and leaves little option to see anything related to external libraries (like SDL).
* These tools aren't well designed around FPS
* Profiling on RPi is a pain to perform

So, I'd like to try out [tracy](https://github.com/wolfpld/tracy). Open source, works well on what I've tested so far and allows some easy annotation and helps developing and profiling over network, even in real time. GPU support varies though.

I've annotated some sections that I know take their time for this. Also I need to check CI here.

See how RPi roughly spends its 100ms per frame (menu, loading screen, then game):

## Checklist


bradallred:
cool. profiling is always useful. 

Not sure I'm a fan of annotating code like this tho. Have you tried `perf` to see if it gets what you want without annotating?

we're already aware of some graphics performance issues which will be addressed by #1771. Please don't spend a lot of time trying to fix something that is already going to fundamentally change. If you see something interesting, open an issue and let's talk about it there.

MarcelHB:
I can also drop my annotations (except the frame mark, that's probably essential) and just leave the CMake/header parts to make it available for devs. I just found it handy to quickly dig through things during development, way more than `perf` of which the data crunching/presentation is also fairly slow on a slow device, not to mention building and displaying a HTML flame graph.

Like how bad is `FindPath` on this device and did that really drop the frame time, or some I/O, or is it all just rendering.

CMakeLists.txt
```diff
@@ -655,6 +656,26 @@ IF (USE_TESTS)
...
+		tracy
+		GIT_REPOSITORY https://github.com/wolfpld/tracy.git
+		GIT_TAG v0.10
+		GIT_SHALLOW TRUE
+		GIT_PROGRESS TRUE
+	)
+	FETCHCONTENT_MAKEAVAILABLE(tracy)
+
+	ADD_DEFINITIONS("-DUSE_TRACY")
+	INCLUDE_DIRECTORIES(${tracy_SOURCE_DIR}/public)
```
lynxlynxlynx, line 676: FYI, this should end up in `${CMAKE_BINARY_DIR}/_deps` meaning we don't need to add an exclusion for `make dist` purposes.

lynxlynxlynx:
I wouldn't mind the annotations if they can be parameterized, so it's clear that they're for tracy. Eg. another macro indirection `TRACY(ZoneScoped);` that would be a noop without tracy and just the macro in the param otherwise.

CMakeLists.txt
MarcelHB: `${tracy_SOURCE_DIR}` is `build/_deps/tracy-src`, so there is not interference (I just run `make dist`) or I don't get what you mean.

lynxlynxlynx: Just documented that nothing more needs to be done.

PR #1962, opened 2023-12-10
Fix spells from scrolls

Dreamkins:
[Spells from scrolls issue](https://github.com/gemrb/gemrb/issues/1958)

Issue #1913, opened 2023-09-15
Crash due to invalid read in scripting


lynxlynxlynx:
I can reproducibly crash by clicking on a free boulder in the iwd2 ar5004 puzzle. 
I'm surprised nobody hit this sooner, since it seems any cutscene that doesn't have just blocking actions should trigger it. It looks like it started being a problem from the cache rewrite, maybe 28564536. ~Definitely after the last release.~

<details><summary>Valgrind output</summary>
<p>

```c++
Invalid read of size 4
   GemRB::Action::GetRef() const (GameScript.h:281)
   GemRB::Response::~Response() (GameScript.h:316)
   GemRB::Response::~Response() (GameScript.h:323)
   GemRB::Response::Release() (GameScript.h:326)
   GemRB::ResponseSet::~ResponseSet() (GameScript.h:339)
   GemRB::ResponseSet::~ResponseSet() (GameScript.h:342)
   GemRB::ResponseSet::Release() (GameScript.h:345)
   GemRB::ResponseBlock::~ResponseBlock() (GameScript.h:362)
   GemRB::ResponseBlock::~ResponseBlock() (GameScript.h:365)
   GemRB::ResponseBlock::Release() (GameScript.h:368)
   GemRB::Script::~Script() (GameScript.h:381)
   GemRB::RCCache<GemRB::FixedSizeString<8ul, &strncasecmp>, GemRB::Script, GemRB::CstrHash< GemRB::FixedSizeString<8ul, &strncasecmp>, &tolower> >::Value::~Value() (Cache.h:41)

Address 0x1cb3ecec is 204 bytes inside a block of size 208 free'd
   operator delete(void*, unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   GemRB::Action::~Action() (GameScript.h:255)
   GemRB::Action::Release() (GameScript.h:295)
   GemRB::Scriptable::ReleaseCurrentAction() (Scriptable.cpp:481)
   GemRB::GameScript::ExecuteAction(GemRB::Scriptable*, GemRB::Action*) (GameScript.cpp:2593)
   GemRB::Scriptable::AddAction(GemRB::Action*) (Scriptable.cpp:395)
   GemRB::Response::Execute(GemRB::Scriptable*) (GameScript.cpp:2441)
   GemRB::GameScript::EvaluateAllBlocks(bool) (GameScript.cpp:2192)
[...]

PR #1968, opened 2023-12-16
fix UseItemPoint, parameters cleared after release action

Dreamkins:
for https://github.com/gemrb/gemrb/pull/1962

### DIFF
--- a/gemrb/core/GameScript/Actions.cpp
+++ b/gemrb/core/GameScript/Actions.cpp
@@ -6864,8 +6864,10 @@ void GameScript::UseItemPoint(Scriptable* Sender, Action* parameters)
 		return;
 	}
 
+	Point targetPos = parameters->pointParameter;
+
 	Sender->ReleaseCurrentAction();
-	act->UseItemPoint(Slot, header, parameters->pointParameter, flags);
+	act->UseItemPoint(Slot, header, targetPos, flags);
 }
 
 //addfeat will be able to remove feats too


bradallred:
hmmm, aren't `Action`s refcounted? Ideally this would be fixed by increasing the refcount before calling `ReleaseCurrentAction` and decreasing it at the end, but maybe that should wait until we switch that to something less manual, like `std::shared_ptr` so its automatic.

lynxlynxlynx:
Thanks!

gemrb/core/GameScript/Actions.cpp
```diff
@@ -6864,8 +6864,10 @@ void GameScript::UseItemPoint(Scriptable* Sender, Action* parameters)
 		return;
 	}
 
+	Point targetPos = parameters->pointParameter;
+
 	Sender->ReleaseCurrentAction();
-	act->UseItemPoint(Slot, header, parameters->pointParameter, flags);
+	act->UseItemPoint(Slot, header, targetPos, flags);
```
Mingun, line 6870: It is better to document such changes explicitly in the code so them wouldn't be refactored back.

lynxlynxlynx: It is pretty obvious if you're working on actions code.

Mingun: And yet, the bug appeared, for all the obvious

bradallred: I agree with you @Mingun, but I already have a WIP patch to use `shared_ptr` for `Actions` (see #1913) which would solve the possibility of this sort of bug appearing again which is probably better than a comment.

PR #1969, opened 2023-12-18
Clang MSVC enabling + Top-level CMake cleanup

czarny247:
## Checklist


.gitignore
```diff
@@ -37,3 +37,7 @@ build_debug/
 
 #clangd cache dir
 .cache
+
+#patches, diffs, etc.
+*.patch
+*.diff
```
lynxlynxlynx, line 43: this is a bit arbitrary — why the addition?

gemrb/core/GameScript/Actions.cpp
```diff
@@ -7118,6 +7118,30 @@ void GameScript::ProtectObject(Scriptable* Sender, Action* parameters)
...
+		// auto visActors = area->GetAllActorsInRadius(
+		// 	Pos, flag, seenby ? VOODOO_VISUAL_RANGE / 2 : GetSafeStat(IE_VISUALRANGE) / 2, this);
+		// actor->AttackedBy(const Actor *actor)
+		// scr->GetCurrentArea()->get
+		//get all creatures in seeing radius
+		//GetAllActorsInRadius
+		//check if protectee is attacked by any of them
+		//if so, then attack nearest one
+	}
+
```
lynxlynxlynx, line 7144: This has no place in this PR, besides being unfinished.

CMakeLists.txt
```diff
@@ -1,88 +1,44 @@
-cmake_minimum_required(VERSION 3.11)
-if(COMMAND cmake_policy)
-	cmake_policy(SET CMP0003 NEW)
-	cmake_policy(SET CMP0005 NEW)
-endif(COMMAND cmake_policy)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.11)
+IF(COMMAND cmake_policy)
+	CMAKE_POLICY(SET CMP0003 NEW)
+	CMAKE_POLICY(SET CMP0005 NEW)
+ENDIF(COMMAND cmake_policy)
```
lynxlynxlynx, line 5: Are these policy statements still needed? We have a much more recent minimum version now, so perhaps they're the default.

CMakeLists.txt
```diff
@@ -1,88 +1,44 @@
...
 
-# If the user specifies -DCMAKE_BUILD_TYPE on the command line, take their definition
-# and dump it in the cache along with proper documentation, otherwise set CMAKE_BUILD_TYPE
-# to Release prior to calling PROJECT()
-IF(DEFINED CMAKE_BUILD_TYPE)
-	SET(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
-ELSE(DEFINED CMAKE_BUILD_TYPE)
-	SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
-ENDIF(DEFINED CMAKE_BUILD_TYPE)
+# SET_INTERNAL(ABC "DEF")
```
lynxlynxlynx, line 17: useless comment

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+
+FUNCTION(SET_WERROR_POLICY)
+    SET(RC "")
+    STRING(REGEX MATCH "-git$" RC ${GEMRB_VERSION})
+    IF(RC STREQUAL "")
+        set (ENABLE_WERROR 0)
+    ELSEIF(NOT DISABLE_WERROR)
+        set (ENABLE_WERROR 1)
+    ENDIF()
+    MESSAGE(STATUS "ENABLE_WERROR: ${ENABLE_WERROR}")
```
lynxlynxlynx, line 507: This will be deceptive, as users are passing DISABLE_WERROR that you now ignore.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+            STRING(JOIN " "
+                CMAKE_CXX_FLAGS
+                ${CMAKE_CXX_FLAGS}
+                "-Werror"
+                "-Wno-inline"
+                "-Wno-error=cast-align"
+                "-Wmissing-declarations"
+            )
+        ELSE ()
+            ADD_FLAG_IF_SUPPORTED(
```
lynxlynxlynx, line 42: This shouldn't be contingent on -Werror, especially not without any justification.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+            CMAKE_CXX_FLAGS
+            ${CMAKE_CXX_FLAGS}
+            "-Wall"
+            "-W"
+            "-Wpointer-arith"
+            "-Wno-format-y2k"
+            "-Wno-long-long"
+        )
+
+        ADD_FLAG_IF_SUPPORTED(FLAG "-pedantic")
```
lynxlynxlynx, line 59: Does some compiler not support it? Haven't got any complaints to that effect.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+        SET(TEST_FLAG "${_FLAG}")
+    ENDIF ()
+    
+    CHECK_CXX_COMPILER_FLAG("${TEST_FLAG}" SUPPORTED)
+
+    IF(SUPPORTED)
+        IF(${_VALUE})
+            SET(_FLAG " ${_FLAG}=${_VALUE}")
+        ENDIF ()
+        STRING(APPEND CMAKE_CXX_FLAGS "${_FLAG}")
```
lynxlynxlynx, line 25: Pretty sure this is missing a space.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+            STRING(APPEND CMAKE_MODULE_LINKER_FLAGS " -Wl,--enable-auto-import")
+        ENDIF (WIN32)
+        # Ensure all plugin symbols exist.
+        IF(NOT APPLE AND NOT UNSAFE_PLUGIN)
+            string(APPEND CMAKE_MODULE_LINKER_FLAGS " -Wl,--no-undefined")
+        ENDIF (NOT APPLE AND NOT UNSAFE_PLUGIN)
+    ENDIF ()
+
+    IF(CMAKE_COMPILER_ID MATCHES "GNU")
+        IF(ENABLE_WERROR)
```
lynxlynxlynx, line 80: This check is redundant.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+            string(APPEND CMAKE_CXX_FLAGS " -Wno-error=enum-float-conversion")
+
+            #string(APPEND CMAKE_CXX_FLAGS " -Wno-error=")
+
+            #FIXME Warnings occuring in multiple files
+            string(APPEND CMAKE_CXX_FLAGS " -Wno-error=sign-conversion")
+            string(APPEND CMAKE_CXX_FLAGS " -Wno-error=zero-as-null-pointer-constant")
+            string(APPEND CMAKE_CXX_FLAGS " -Wno-error=exit-time-destructors")
+            string(APPEND CMAKE_CXX_FLAGS " -Wno-error=missing-variable-declarations")
+
```
lynxlynxlynx, line 212: What's all this new stuff about?

CMakeLists.txt
```diff
@@ -207,522 +79,35 @@ if(APPIMAGE)
...
-# as of 2.8 cmake does not support anything like EVAL
-MACRO(PRINT_OPTION option)
-if (${option})
-	message(STATUS "  ${option}: ${${option}}")
-else()
-	message(STATUS "  ${option}: disabled")
-endif()
-ENDMACRO(PRINT_OPTION)
+CONFIGURE_COMPILER()
+INCLUDE(config)
```
lynxlynxlynx, line 97: Isn't this now too early to be properly populated?

cmake/FindDependencies.cmake
```diff
@@ -0,0 +1,76 @@
...
+
+IF(USE_OPENAL)
+INCLUDE(FindOpenAL)
+IF(OPENAL_FOUND)
+	MESSAGE(STATUS "Looking for OpenAL: found")
+ELSE()
+	MESSAGE(WARNING "Looking for OpenAL: not found!")
+	MESSAGE(WARNING "If you want to build the OpenAL plugin, get OpenAL from www.openal.org.")
+	MESSAGE(WARNING "If it just wasn't found, try setting the OPENALDIR environment variable.")
+ENDIF()
```
lynxlynxlynx, line 19: misindented

cmake/FindDependencies.cmake
```diff
@@ -0,0 +1,76 @@
...
+ENDIF()
+
+IF(USE_LIBVLC AND NOT MSVC)
```
lynxlynxlynx, line 22: This is worth retesting, since the reason for the MSVC inclusion here was lifted.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+            IF(NOT (FOUND EQUAL "-1"))
+                LIST(APPEND CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+            ENDIF()
+        ENDIF()
+    ENDIF()
+ENDFUNCTION()
+
+FUNCTION(CONFIGURE_FMT_INCLUDES)
+    # supress static analysis of fmt
+    INCLUDE_DIRECTORIES(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/includes/fmt)
```
lynxlynxlynx, line 709: Nice idea, I guess this is for clang-tidy?

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+        FIND_LIBRARY(COCOA_LIBRARY_PATH Cocoa)
+        FIND_LIBRARY(COREFOUNDATION_LIBRARY CoreFoundation)
+        # for objective-c++
+        STRING(APPEND CMAKE_CXX_FLAGS " -DTARGET_OS_MAC")
+        # for pure objective-c
+        STRING(APPEND CMAKE_C_FLAGS " -x objective-c -DTARGET_OS_MAC")
+    ELSE(APPLE)
+        SET(BAKE_ICON 1)
+    ENDIF()
+    SET(BAKE_ICON ${BAKE_ICON} CACHE INTERNAL "")
```
lynxlynxlynx, line 627: No need to cache it, as it's part of config.h.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+    ENDIF(VITA)
+ENDFUNCTION()
+
+FUNCTION(CONFIGURE_UNIX_SPECIFICS)
+    IF(UNIX)
+        SET(CMAKE_THREAD_PREFER_PTHREAD true)
+        SET(CMAKE_THREAD_PREFER_PTHREAD ${CMAKE_THREAD_PREFER_PTHREAD} CACHE INTERNAL "")
+        FIND_PACKAGE(Threads REQUIRED)
+
+        # TODO ONLY FOR UBUNTU
```
lynxlynxlynx, line 642: This new comment should be removed. If it bothers you, you can  remove the Ubuntu mention in the line below.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,790 @@
...
+        INSTALL( FILES "${CMAKE_CURRENT_BINARY_DIR}/gemrb.6" DESTINATION ${MAN_DIR} )
+        IF (NOT HAIKU)
+            INSTALL( FILES artwork/gemrb-logo.png DESTINATION ${ICON_DIR} RENAME gemrb.png )
+            INSTALL( FILES artwork/logo04-rb_only.svg DESTINATION ${SVG_DIR} RENAME gemrb.svg )
+            INSTALL( FILES platforms/linux/gemrb.desktop DESTINATION ${MENU_DIR} )
+            INSTALL( FILES platforms/linux/org.gemrb.gemrb.metainfo.xml DESTINATION ${METAINFO_DIR} )
+        ENDIF()
+        INSTALL( FILES README.md INSTALL COPYING NEWS AUTHORS DESTINATION ${DOC_DIR} )
+        INSTALL( FILES admin/extend2da.py DESTINATION ${BIN_DIR}
+            PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
```
lynxlynxlynx, line 388: This does not belong in this function.

lynxlynxlynx:
Thanks, that was quite a load to review, since you didn't split the work. In addition to the above, for the platform specific functions you can use the return statement to reduce the amount of indentation.

gemrb/includes/exports.h
```diff
@@ -90,14 +90,14 @@
...
-/// Make sure we don't like to static libraries
+/// Make sure we don't link to static libraries
 /// This causes hard to debug errors due to multiple heaps.
 #if defined(_MSC_VER) && !defined(_DLL)
 #	error GemRB must be dynamically linked with runtime libraries on win32.
 #endif
 
 /// Silence some persistent unused warnings (supported since gcc 2.4)
-#ifdef __GNUC__
+#if defined __GNUC__ or defined __clang__
```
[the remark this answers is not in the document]
czarny247: Are you sure about it in case of clang for msvc, a.k.a clang-cl?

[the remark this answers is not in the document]
lynxlynxlynx: That's my impression from the docs, but it's trivial for you to check with an actual install available.

cmake/Helpers.cmake
lynxlynxlynx: I meant the message, but misread. However the combination of git version and DISABLE_WERROR leaves the variable undefined, I guess an empty string. It would be better if it was consistent.

.gitignore
czarny247: I suppose that these kind of files are meant to be used locally for work in progress or for sending them to another dev.
I just noticed that they're recognized by a git in this repo so I automatically added them to the ignore list.
I think if there will ever be any patch or (less likely) diff to be tracked by version control, then it can be directly un-ignored here. What do you think?

gemrb/includes/exports.h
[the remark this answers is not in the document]
czarny247: Ok so here's the proof (compiled with clang-cl for msvc):

image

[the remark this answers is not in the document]
czarny247: image

cmake/Helpers.cmake
czarny247:
```
if ((NOT DISABLE_WERROR) AND GIT_VERSION)
```
This was an original check.
I just simplified it and got rid of incorrect name - `GIT_VERSION`.
Maybe I misunderstood something or made an error.
If so then please keep in mind my intentions and tell me how do you think it should work then.

cmake/Helpers.cmake
czarny247:
```
    IF(SUPPORTED)
        IF(${_VALUE})
            SET(_FLAG " ${_FLAG}=${_VALUE}")
        ENDIF ()
        STRING(APPEND CMAKE_CXX_FLAGS "${_FLAG}")
    ENDIF ()
 ```
 Which line exactly?

File: gemrb/includes/exports.h (hunk 18 above)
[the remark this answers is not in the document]
Comment (bradallred) [reply gemrb/includes/exports.h:100]: IIRC, `__GNUC__` is a macro to indicate the compeer supports GNU extensions. I dont think we should be using `__GNUC__` at all.

File: cmake/Helpers.cmake (hunk 9 above)
Comment (czarny247) [reply cmake/Helpers.cmake:80]: I just only rearanged original code:

https://github.com/gemrb/gemrb/blob/82b1b9cbbb9c5866e2d0cab15496221036cddfb1/CMakeLists.txt#L220

I wanted to make it like this:

```
if(GNU | Clang)
 //common stuff
endif()

if(GNU)
 //GNU specific
endif()
```

File: .gitignore (hunk 1 above)
Comment (lynxlynxlynx) [reply .gitignore:43]: This would hide them from git status, making it easier for them to remain around. I actually do have a couple laying around, so it'd be impacted, plus if there ever was a need to ship a patch, one step less would be needed.

File: gemrb/includes/exports.h (hunk 18 above)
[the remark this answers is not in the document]
Comment (lynxlynxlynx) [reply gemrb/includes/exports.h:100]: Ah, cool, I guess other uses need to be reviewed then as well.

File: cmake/Helpers.cmake (hunk 6 above)
Comment (lynxlynxlynx) [reply cmake/Helpers.cmake:42]: Yes, that's basically the problem — you merged two different statements (is this release code? Should compiler warnings be errors?) into one boolean. In this case we want to disable debug info only for release code, regardless of -Werror state, so it remains a "reproducible build".

File: cmake/Helpers.cmake (hunk 8 above)
Comment (lynxlynxlynx) [reply cmake/Helpers.cmake:25]: The one I commented under. If ${_VALUE} does not evaluate to true, what will ensure _FLAG has a space in front?

File: cmake/Helpers.cmake (hunk 9 above)
Comment (lynxlynxlynx) [reply cmake/Helpers.cmake:80]: -Wno-error subflags will be noops if -Werror is not enabled, so there's no harm in always adding them (that I know of).

File: .gitignore (hunk 1 above)
Comment (czarny247) [reply .gitignore:43]: Ok, I will remove it then ;)

File: cmake/Helpers.cmake (hunk 8 above)
Comment (czarny247) [reply cmake/Helpers.cmake:25]: Ok I thought it was about indentation. I get it now :D

File: gemrb/includes/exports.h (hunk 18 above)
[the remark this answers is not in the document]
Comment (bradallred) [reply gemrb/includes/exports.h:100]: Also, It's not clear if `__GNUC__` is defined if you disable GNU extensions even if your compiler supports it. maybe we should open an issue to figure out what we really ought to be doing for these checks.

File: CMakeLists.txt (hunk 11 above)
Comment (czarny247) [reply CMakeLists.txt:97]: Only this is needed

```
LIST(APPEND CMAKE_MODULE_PATH
	"${CMAKE_SOURCE_DIR}/cmake"
	"${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/"
)
```

Before ```INCLUDE(config)```.

Also
```CMAKE_CXX_COMPILER``` and ```CMAKE_C_COMPILER```
needs to be set before. I can made a sanity check for it.
I don't think it's needed since "everything" else will fail if you don't specify compilers at all.

File: cmake/Helpers.cmake (hunk 14 above)
Comment (czarny247) [reply cmake/Helpers.cmake:709]: For `clangd` but yeah, it embeds `clang-tidy`.

https://clangd.llvm.org/features#clang-tidy-checks

File: CMakeLists.txt (hunk 11 above)
Comment (lynxlynxlynx) [reply CMakeLists.txt:97]: I'm asking because some defines are set based on cmake variables set later in the file. If the actual processing is deferred, then good, but do check your config.h looks as expected.

File: cmake/Helpers.cmake (hunk 5 above)
Comment (czarny247) [reply cmake/Helpers.cmake:414]: The logic seems to be correct now. What name for this variable do you suggest then?

File: cmake/Helpers.cmake (hunk 6 above)
Comment (czarny247) [reply cmake/Helpers.cmake:53]: Fixed

File: gemrb/includes/exports.h (hunk 18 above)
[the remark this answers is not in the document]
Comment (czarny247) [reply gemrb/includes/exports.h:100]: @bradallred I opened an issue for it https://github.com/gemrb/gemrb/issues/1971

File: cmake/Helpers.cmake [hunk 19]
```diff
@@ -269,7 +269,13 @@ FUNCTION(CONFIGURE_SDL SDL_BACKEND)
         INCLUDE(FindSDL)
         IF(USE_SDLMIXER)
             FIND_PACKAGE(SDL_mixer MODULE)
+            SET_INTERNAL(SDL_MIXER_INCLUDE_DIR ${SDL2_MIXER_INCLUDE_DIRS})
+            SET_INTERNAL(SDL_MIXER_LIBRARIES ${SDL2_MIXER_LIBRARIES})
         ENDIF()
+
+        SET_INTERNAL(SDL_INCLUDE_DIR ${SDL_INCLUDE_DIRS})
+        SET_INTERNAL(SDL_LIBRARY "${SDL2_LIBRARIES}")
```
lynxlynxlynx: Won't all these be empty? SDL1 does not set SDL2_MIXER_INCLUDE_DIRS etc.

cmake/Helpers.cmake
```diff
@@ -56,7 +56,7 @@ FUNCTION(CONFIGURE_COMPILER)
             "-Wno-long-long"
         )
 
-        ADD_FLAG_IF_SUPPORTED(FLAG "-pedantic")
+        STRING(APPEND CMAKE_CXX_FLAGS "-pedantic")
```
lynxlynxlynx, line 70: space

gemrb/includes/exports.h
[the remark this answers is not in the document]
bcmhercun: Can it be resolved? What do you think @lynxlynxlynx ?

cmake/Helpers.cmake
```diff
@@ -593,86 +492,73 @@ endif()
...
 FUNCTION(CONFIGURE_TARGET_PLATFORM_SPECIFICS)
-    #EACH OF THEM DOES NOTHING IF TARGET PLATFORM IS NOT MATCHING
-    CONFIGURE_APPLE_SPECIFICS()
-    CONFIGURE_VITA_SPECIFICS()
-    CONFIGURE_UNIX_SPECIFICS()
-    CONFIGURE_RPI_SPECIFICS()
+    SET_INTERNAL(BAKE_ICON 0)
+    IF(APPLE)
+        CONFIGURE_APPLE_SPECIFICS()
+        RETURN()
```
lynxlynxlynx, line 559: These returns are bad. At a glance they're only redundant, saving a few checks, but actually the blocks are not independant — UNIX also matches on APPLE.

cmake/Helpers.cmake
```diff
@@ -1,7 +1,7 @@
 #Made for convenience of setting variables
```
lynxlynxlynx, line 1: What's the point, benefit of this commit? It seems to just complicate matters.

CMakeLists.txt
```diff
@@ -131,8 +131,8 @@ message(STATUS "")
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 message(STATUS "Target bitness: ${CMAKE_SIZEOF_VOID_P}*8")
 message(STATUS "")
-if (ENABLE_WERROR)
+IF((NOT DISABLE_WERROR) AND IS_RELEASE_VERSION)
```
lynxlynxlynx, line 134: you got this the other way around

CMakeLists.txt
bcmhercun: I checked it also with comparing to the printed values and they're the same.

cmake/Helpers.cmake
```diff
@@ -40,7 +40,7 @@ ENDFUNCTION()
 FUNCTION(CONFIGURE_COMPILER)
     INCLUDE(CheckCXXCompilerFlag)
     IF(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
-        IF(ENABLE_WERROR)
+        IF((NOT DISABLE_WERROR) AND IS_RELEASE_VERSION)
```
lynxlynxlynx, line 43: same here

cmake/Helpers.cmake
```diff
@@ -49,7 +49,7 @@ FUNCTION(CONFIGURE_COMPILER)
                 "-Wno-error=cast-align"
                 "-Wmissing-declarations"
             )
-        ELSE ()
+        ELSEIF(NOT IS_RELEASE_VERSION)
```
lynxlynxlynx, line 52: same here

cmake/Helpers.cmake
bcmhercun: You're right. I noticed that. It didn't happened locally since I forgot to wipe out CMake cache ;)

cmake/Helpers.cmake
bcmhercun: Ok so what did you mean with returns previously?
I rarely use returns in CMake so I don't see profits of them at all.

gemrb/includes/exports.h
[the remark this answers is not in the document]
lynxlynxlynx: Do you have two accounts or what? The __clang__ can remain for now.

CMakeLists.txt
lynxlynxlynx: We don't print things like SUPPORTS_MEMSTREAM.

cmake/Helpers.cmake
lynxlynxlynx: They would make sense if they were in the subfunctions themselves, not here. This way is cleaner though, so I suggest you just remove them.

gemrb/includes/exports.h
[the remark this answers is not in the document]
czarny247: @lynxlynxlynx Yeah, sorry for that. The other one is a work account. I just forgot to switch.

CMakeLists.txt
czarny247: Yeah I know I meant only ones which are printed.
The rest of them were checked by comparing the `config.h` and the code itself.
Do you need some extra prints for it? (We can run this via CI and then I can remove them in subsequent commit).

cmake/FindDependencies.cmake
czarny247: Not sure what do you mean by that.

cmake/Helpers.cmake
czarny247: Now I get it. I might have done it as an early exits... <facepalm>
But yeah, I will remove them.

CMakeLists.txt
lynxlynxlynx: No, just making sure you tested fully.

cmake/FindDependencies.cmake
lynxlynxlynx: the "AND NOT MSVC" should not be needed any more. It was added since our vlc use relied on pthreads previously.

cmake/Helpers.cmake
lynxlynxlynx: This has not been fixed.

lynxlynxlynx:
Ok, so it now builds everywhere again. Can you clean up the commit history into something sensible?

czarny247:
> Ok, so it now builds everywhere again. Can you clean up the commit history into something sensible?

Sure, will do it soon ;)

czarny247:
> Ok, so it now builds everywhere again. Can you clean up the commit history into something sensible?

Done, please take a look if it is fine for you.

CMakeLists.txt
```diff
@@ -1,87 +1,38 @@
...
-	IF(RPI AND NOT DISABLE_VIDEOCORE)
-		SET(HINT_GLES_LIBNAME brcmGLESv2)
-		SET(HINT_GLES_INCDIR "/opt/vc/include")
-		SET(HINT_GLES_LIBDIR "/opt/vc/lib")
-		SET(OPENGL_BACKEND GLES)
-		SET(SDL_BACKEND SDL2)
-	ENDIF()
-ENDIF()
+SET(CMAKE_BUILD_RPATH_USE_ORIGIN ON)
+SET(CMAKE_BUILD_RPATH_USE_ORIGIN ON)
```
lynxlynxlynx, line 35: doubled

cmake/Helpers.cmake
```diff
@@ -0,0 +1,685 @@
...
+
+FUNCTION(CACHE_BUILD_TYPE)
+    # If the user specifies -DCMAKE_BUILD_TYPE on the command line, take their definition
+    # and dump it in the cache along with proper documentation, otherwise set CMAKE_BUILD_TYPE
+    # to Release prior to calling PROJECT()
+    IF(DEFINED CMAKE_BUILD_TYPE)
+        SET(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
+    ELSE(DEFINED CMAKE_BUILD_TYPE)
+        SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
+    ENDIF(DEFINED CMAKE_BUILD_TYPE)
```
lynxlynxlynx, line 386: As a further improvement, this can also be deduped using the `$<IF>` construct.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,685 @@
...
+                SET_INTERNAL(SDL_FOUND ${SDL_FOUND})
+            ENDIF()
+        ENDIF()
+        IF(NOT (SDL_FOUND OR SDL2_FOUND))
+            MESSAGE(WARNING "No SDL or SDL2 libraries found.")
+            MESSAGE(FATAL_ERROR "Please get SDL from www.libsdl.org")
+        ENDIF()
+    ENDIF()
+
+    IF(SDL_BACKEND STREQUAL "SDL" OR SDL_BACKEND_DETECTED STREQUAL "SDL")
```
lynxlynxlynx, line 180: I don't know why you haven't answered me on this, but I see no good reason for these doubled variables.

cmake/Helpers.cmake
```diff
@@ -0,0 +1,685 @@
...
+        SET_INTERNAL(HINT_GLES_INCDIR "/opt/vc/include")
+        SET_INTERNAL(HINT_GLES_LIBDIR "/opt/vc/lib")
+        SET_INTERNAL(OPENGL_BACKEND GLES)
+        SET_INTERNAL(SDL_BACKEND_DETECTED SDL2)
+        SET(SDL_BACKEND ${SDL_BACKEND_DETECTED} CACHE STRING FORCE "")
+    ENDIF()
+ENDFUNCTION()
+
+FUNCTION(CONFIGURE_TARGET_PLATFORM_SPECIFICS)
+    SET_INTERNAL(BAKE_ICON 0)
```
lynxlynxlynx, line 549: this looks wrong, should be 1

CMakeLists.txt
```diff
@@ -208,451 +75,37 @@ if(APPIMAGE)
...
+CONFIGURE_FMT_INCLUDES()
+
 # Tests (need init'd before subdirectories)
 IF (USE_TESTS)
-  FIND_PACKAGE(GTest REQUIRED)
-
+	FIND_PACKAGE(GTest REQUIRED)
 	INCLUDE(CTest)
 	ENABLE_TESTING()
+	SET_INTERNAL(BUILD_TESTING 1)
```
lynxlynxlynx, line 108: why are you forcing a value here?

cmake/Helpers.cmake
```diff
@@ -0,0 +1,685 @@
...
+                ${CMAKE_CXX_FLAGS}
+                "-Werror"
+                "-Wno-inline"
+                "-Wno-error=cast-align"
+                "-Wmissing-declarations"
+            )
+        ELSEIF(IS_RELEASE_VERSION)
+            ADD_FLAG_IF_SUPPORTED(
+                FLAG "-ffile-prefix-map"
+                TEST_VALUE "a=b"
```
lynxlynxlynx, line 54: This is redundant, you can just use the final VALUE and then also slightly simplify ADD_FLAG_IF_SUPPORTED, since this is the only TEST_VALUE user.

cmake/Helpers.cmake
czarny247: It is to keep `SDL_BACKEND` value equal to user's choice and `SDL_BACKEND_DETECTED` equal to "real" detected value. It is made specifically for case with `Auto` as a chosen value.

lynxlynxlynx: Auto is just the default, there's no reason for anyone to pass it. If one wanted to change the backend used in a subsequent build, a reconfigure is needed either way, so there's nothing to gain.

And same with OpenGL, except there's not even an auto there.

cmake/Helpers.cmake
czarny247: If I understand it correctly, cmake generator expressions are evaluated during the cmake build step, whereas variable are set earlier, at the configuration step, so this is not possible. Or maybe I don't get your idea. :D

lynxlynxlynx: Ah, ok. We could still reduce the duplication by using a temporary var though.

lynxlynxlynx:
Thanks!

lynxlynxlynx:
Actually there is a sneaky problem now. The flag tests (eg. -fsigned-char) are broken; if you look at the cmake log you can see it also passes a define -D-FSIGNED-CHAR, which then causes the test build to fail (macro names must be identifiers) and the flag gets rejected.

I propose you strip any leading dashes and not to uppercase the display name, since it just looks weird.

czarny247:
> Actually there is a sneaky problem now. The flag tests (eg. -fsigned-char) are broken; if you look at the cmake log you can see it also passes a define -D-FSIGNED-CHAR, which then causes the test build to fail (macro names must be identifiers) and the flag gets rejected.
> 
> I propose you strip any leading dashes and not to uppercase the display name, since it just looks weird.

Oh, sorry for that. Will be fixed today ;)

czarny247:
> -D-FSIGNED-CHAR

I cannot find this, checked logs from CI.

lynxlynxlynx:
It's in the log file, not printed.

czarny247:
> It's in the log file, not printed. But you can see the symptoms in CI as well: https://github.com/gemrb/gemrb/actions/runs/7438831835/job/20238250738#step:4:57

Fix: https://github.com/gemrb/gemrb/pull/1974

PR #1974, opened 2024-01-08
Compiler flags testing fixed

czarny247:
## Checklist


lynxlynxlynx:
It doesn't appear to be good enough. You can still see -fsigned-char and -ffast-math failing everywhere and it's not like they're new flags that only fresh compilers would support. Before we were adding all three unconditionally.

I think the dashes remain the same problem, just that it's now the second one.

bcmhercun:
> It doesn't appear to be good enough. You can still see -fsigned-char and -ffast-math failing everywhere and it's not like they're new flags that only fresh compilers would support. Before we were adding all three unconditionally.
> 
> I think the dashes remain the same problem, just that it's now the second one.

Yes, it looks like no dashes/hyphens are allowed in variable names: https://stackoverflow.com/questions/50423280/cmake-indicates-compiler-does-not-support-flag-but-it-does

Removed all of them.

bradallred:
I don't know if it's a problem or not, but I always write it `IF(HAS_FLAG)` instead of `IF("${HAS_FLAG}")`. I'm not sure how an empty string evaluates in cmake.

lynxlynxlynx:
Please just sanitize it completely, so it doesn't break later with some other problematic char. I suggest turning anything not in [:alnum:] into an underscore.

lynxlynxlynx:
Thanks!

PR #1975, opened 2024-01-12
Fix linking on FreeBSD.

dstolfa:
Address a few blockers when building gemrb on FreeBSD & aarch64.

I've replaced all occurrences of `CMAKE_THREAD_LIBS_INIT` with `Threads::Threads` however I don't have all the systems to test on. The ones I *did* test on work, however.

## Checklist

- [ x ] Commit messages are descriptive and explain the rationale for changes
- [ x ] I used the same coding style as the surrounding code 
- [ x ] I have tested the proposed changes


lynxlynxlynx:
Thanks, the first commit looks fine. However the second shouldn't be needed and it opens a can of worms (what all other places sneakily break?). For this very reason we pass `-fsigned-char`:
https://github.com/gemrb/gemrb/blob/042096f0506a4623a645bf22e4d820cae4055725/cmake/Helpers.cmake#L67C1-L67C1

Did you use some uncommon compiler for that to not get added? We do it for gcc and clang.

dstolfa:
> Thanks, the first commit looks fine. However the second shouldn't be needed and it opens a can of worms (what all other places sneakily break?). For this very reason we pass `-fsigned-char`: https://github.com/gemrb/gemrb/blob/042096f0506a4623a645bf22e4d820cae4055725/cmake/Helpers.cmake#L67C1-L67C1

Ah, I missed this in the build. Thanks for pointing it out.

> Did you use some uncommon compiler for that to not get added? We do it for gcc and clang.

It was clang, albeit a modified one for Morello (https://www.morello-project.org/). It might have something to do with the flag not being set. I'll take a look at the CMake file and see what's going on. In the meantime, would you prefer me to force push to remove that commit and open a separate PR if necessary once I have this bit working, so that the pthread one can be tested?

lynxlynxlynx:
Well, the first commit doesn't build on windows, so both will need more work. But yeah, force pushing is fine.

Looks easy to fix:
https://ci.appveyor.com/project/lynxlynxlynx/gemrb/builds/48941490#L389

dstolfa:
This seems to be caused by FindThreads not actually finding any threading library for those plugins on Windows. I'm not quite why that might be happening, but maybe it's expected? I could work around the issue by just checking for `WIN32` however I'd rather not do that without knowing why the threading library is missing in the first place. When using the CMake variable (`CMAKE_THREAD_LIBS_INIT`) over the target, the string is simply empty so CMake silently does not link any threading library into the plugins while using the target makes it fail.

lynxlynxlynx:
No, it's simply not called, like the error states.

lynxlynxlynx:
Why not just move the find_package call instead of making every use conditional?

dstolfa:
> Why not just move the find_package call instead of making every use conditional?

Do you mean moving [this](https://github.com/gemrb/gemrb/blob/master/cmake/Helpers.cmake#L538) call out of the UNIX conditional/adding it into an `IF(WIN32)` conditional? I wasn't sure if linking threads on Windows would break anything as they weren't being linked before, so I just went for what I felt was the safer approach and in-line with the current build. I'm not familiar enough with the details of GemRB to confidently judge what effect that would have, but I'm happy to do it if it's the preferred approach.

I could also move it out of the conditional all together, but I see that there is vita support and I have no way to check what happens to that if I move it out.

lynxlynxlynx:
The engine doesn't work without threading at all, so yeah, if it's possible to do the call unconditionally, we should just do it and let the module deal with platform differences. Vita support is currently at a standstill unfortunately, since it lacks a python3.

lynxlynxlynx:
Yay, it works, thanks!

PR #1977, opened 2024-01-13
Fix the cmake warning (CMP0148)

K1ngst0m:
## Description

According to the [cmake policy CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html):

`FindPythonLibs` have been deprecated since CMake 3.12. So we should update to use the new `FindPython`.

## Checklist

- [X] The proposed change builds also on our build bots (check after submission)


lynxlynxlynx:
Thanks. Please also bump the minimum version to 3.12 then, since it appears FindPython is only available from that point onward.

Something goes wrong on windows:

> -- Could NOT find Python (missing: Development Development.Module Development.Embed) (found suitable version "3.12", minimum required is "3")

lynxlynxlynx:
VCPKG is versioned and we use a snapshot, so the py version shouldn't matter. The problem is somewhere between cmake and the host. Since 3.12 is on the host, I bet it's being picked up, but with just the runtime available, thus failing the requirement. So we need a different way to enforce the found version — before we used that version define to short-circuit multiple version problems.

lynxlynxlynx:
You can try with a ranged version: https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature

Something like `-DPYTHON_VERSION="3.10<3.11"`.

lynxlynxlynx:
Any plans on completing this work?

czarny247:
> Thanks. Please also bump the minimum version to 3.12 then, since it appears FindPython is only available from that point onward.
> 
> Something goes wrong on windows:
> 
> > -- Could NOT find Python (missing: Development Development.Module Development.Embed) (found suitable version "3.12", minimum required is "3")

I checked [https://vcpkg.io/en/packages](https://vcpkg.io/en/packages)
And it looks like:
image
(also here: https://github.com/microsoft/vcpkg/issues/36735)

So @K1ngst0m you can update python at most to `3.11.5`.

lynxlynxlynx:
This is not about bumping the python version though. And the minimum cmake one was already sufficiently bumped in a different PR, so that part is not relevant any more.

lynxlynxlynx:
Closing in favour of #2070

PR #1978, opened 2024-01-13
Report and set SDL_FOUND when -DSDL_BACKEND specified

K1ngst0m:
## Description

1. Report the SDL version and handle the `SDL_FOUND` only happen in the default path.
2. When using `-DSDL_BACKEND=SDL2`, cmake goes to the findSDL2 path and it does not set `SDL_FOUND` but `SDL2_FOUND`, so we need to handle that.

## Checklist

- [X] The proposed change builds also on our build bots (check after submission)


lynxlynxlynx:
Thanks, a recent regression!

Issue #1982, opened 2024-02-06
Stoneskin changes dialog name color


burner1024:
#### Bug description
Iron skin and probably others too.

(Also dialog lines colors are different, grey vs blue. Not sure if it's intentional?)

#### Steps to reproduce
1. Cast stoneskin
2. Select/command the character.

#### Expected behavior
Character name using main character color.

#### Screenshots
BG2 (switched Imoen's color to red for more contrast)
![Captura desde 2024-02-07 04-11-11](https://github.com/gemrb/gemrb/assets/1279852/059eee80-7b44-43bd-972e-2864d0fa47c4)

Gemrb, after casting stoneskin color changed.
![Captura desde 2024-02-07 04-11-37](https://github.com/gemrb/gemrb/assets/1279852/9a93fadd-1740-498f-9f0d-db3bbe2a0299)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Yes, we compute the name color instead of hardcoding it (IIRC the original had a fixed-sized list to pick from). It takes the color stats into account, so when effects change them, it shows.

How did you switch the color in the original?

burner1024:
The normal UI 
![1](https://github.com/gemrb/gemrb/assets/1279852/4c0a9975-8d46-4dd9-a257-72baa9a19bd1)

I probably wouldn't mind dynamic name color, but the issue is that later in game many of the characters are stoneskinned, and the dialog window starts to look really bland and hard to distinguish who says what.
![Captura desde 2024-02-07 03-58-01](https://github.com/gemrb/gemrb/assets/1279852/685feb58-2e8d-4328-b323-50bdd39f0f2e)

Also the light grey text color doesn't help when names are grey too. Is that intentional, after all?

lynxlynxlynx:
Ok, so it's the major color. As for the grey, I don't know, but that's configurable already (probably DIALOG in colors.2da).

burner1024:
I tried changing DIALOG and DIALOG_PARTY, that works when in dialog mode, but has no effect in "world" mode, basically selection sounds are always white/light grey. Should I file a separate issue about that?

[...]

PR #1987, opened 2024-02-08
Console colors fix

burner1024:
## Description
unhardcode console colors, fix bg2 party console and dialog color, ref #1982

## Checklist

### DIFF
--- a/gemrb/core/DisplayMessage.cpp
+++ b/gemrb/core/DisplayMessage.cpp
@@ -169,12 +169,15 @@ Color DisplayMessage::GetSpeakerColor(String& name, const Scriptable *&speaker) 
 	Color speaker_color {0x80, 0, 0, 0xff};
 	// NOTE: name color was hardcoded to a limited list in the originals;
 	// the 1PP mod tackled this restriction by altering the exe to use a bigger list.
-	// We just generate a colour by looking at the existing palette instead.
+	// We used to generate a colour by looking at the existing palette instead.
+	// That proved to be inconvenient in-game when many characters are stoneskinned and all get the same color;
+	// also dynamic name colors do not have any real use, they just add confusion.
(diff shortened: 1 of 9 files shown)


burner1024:
Captura desde 2024-02-08 23-58-33

Fixed dialog console color. Hopefully without collateral damage...

lynxlynxlynx:
Thanks. I suggest a different name for the colors, so they don't get confused for the actual console colors. Maybe MSG and MSG_PARTY.

Please also add the two new entries to other colors.2da, so the structure remains the same and they don't default to full white. 

The name color fix is a oneliner in `DisplayMessage::GetSpeakerColor` btw.

burner1024:
> Thanks. I suggest a different name for the colors, so they don't get confused for the actual console colors. Maybe MSG and MSG_PARTY.

Maybe straight DISPLAYSTRING + _PARTY? If this 2da is supposed to be self-documenting for modders or something.
Because from the first glance I can't tell what some of those mean, like MAPICNBG or ALTDOOR.

lynxlynxlynx:
DISPLAYSTRING isn't very telling either, since almost all text goes through those functions.

burner1024:
Where'd you get [PST party color](https://github.com/gemrb/gemrb/commit/c75bfb46fcca9dc22aa356d28a81d716e9ec9047#diff-eefe647a01562f0dc95ef8bec8e4f6a1729e3e4e7c2ebe4ad2497c32a54c8222R18)? Looks very similar, maybe I should use that one

lynxlynxlynx:
Same inaccurate color picker approach.

burner1024:
OK, I copied whatever colors there were from DIALOG_* to MSG_* for all games except BG2, where I used the new value for MSG_PARTY.

Captura desde 2024-02-09 07-08-00

lynxlynxlynx:
Thanks!

PR #1990, opened 2024-02-11
Brightness and contrast

burner1024:
## Description
Added brightness and contrast to the existing shader #1098.

Actually, there's some confusion about the settings, they called it contrast in the interface and gamma in the configs. But from my testing, it seems that it is also brightness. Just the first one is additive, the other one is multiplicative. (Maybe a better name for 2nd one would be saturation, my color theory is very weak.)

I tried actual contrast formula, as well as actual gamma correction formula, and the results are different, but I can't say that they are better.
There are other options for brightness too, sigmoid and stuff. Maybe worth exploring.

Also fixes brightness not applying on start.

## Checklist

### DIFF
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -473,8 +473,8 @@ int SDL20VideoDriver::RenderCopyShaded(SDL_Texture* texture, const SDL_Rect* src
 
 	blitRGBAShader->SetUniformValue("u_greyMode", 1, greyMode);
 
-	blitRGBAShader->SetUniformValue("u_brightness", 1, this->brightness);
-	blitRGBAShader->SetUniformValue("u_contrast", 1, this->contrast);
+	blitRGBAShader->SetUniformValue("u_brightness", 1, brightness);
+	blitRGBAShader->SetUniformValue("u_contrast", 1, contrast);
 
 	GLint channel = 3;
 	if (flags & BlitFlags::STENCIL_RED) {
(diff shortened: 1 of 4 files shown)


gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -1012,10 +1012,10 @@ bool SDL20VideoDriver::CanDrawRawGeometry() const {
 #endif
 }
 
-void SDL20VideoDriver::SetGamma(int brightness, int contrast)
+void SDL20VideoDriver::SetGamma(int new_brightness, int new_contrast)
 {
-	this->brightness = 1.0 + (float)brightness*0.0015;
-	this->contrast = 1.0 + (float)contrast*0.05;
+	this->brightness = 1.0 + (float)new_brightness*0.0015;
+	this->contrast = 1.0 + (float)new_contrast*0.05;
```
lynxlynxlynx, line 1018: You can lose the `this` and please use camel case like everything else does.

burner1024: Done.

burner1024:
There probably can be some improvements over original with more complicated functions, but this at least gets the job done.
Also looks like there's SDL3 on the horizon, then shaders would need to be re-written anyway?
In the long run, it would probably be good to put shaders into override and compile them at runtime, letting users to experiment. Different games have different gammas and might benefit from different corresponding functions.

lynxlynxlynx:
Can you rebase the fixups into the original commit? Or we can just squash merge.
@bradallred any comments?

burner1024:
I think squash merge is fine, there's not a lot of changes anyway.

bradallred:
LGTM, ship it!

lynxlynxlynx:
Thanks!

Issue #1989, opened 2024-02-09
Can't launch in full screen 


burner1024:
#### Bug description
This is probably something stupid on my part, but... can't get the game to start in full screen.
```bash
$ grep -i "full screen" *baldur.ini
baldur.ini:Full Screen=1
gem-baldur.ini:Full Screen = 1

$ grep -i "fullscreen" ~/games/GemRB-bg2-test.cfg 
#Fullscreen [Boolean]
Fullscreen=1

$ /opt/gemrb/bin/gemrb -c ~/games/GemRB-bg2-test.cfg
# launches windowed
```

Then Ctrl-F switches to full screen just fine.

#### Steps to reproduce
1. Set full screen = 1 in all inis and cfgs.
2. Launch the game.

#### Expected behavior
Launched in full screen.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Wrong key in `gemrb/core/Interface.cpp:370`. I think we should just retire the cfg key instead, as everything else relies on the ini version (name) and that's also what toggling it ingame changes.

If not, we need another lookup, so both options would work.

burner1024:
It is kinda confusing to have two.

Maybe also save the last value in gem-baldur.ini?
Clicking 
![Captura desde 2024-02-10 00-10-13](https://github.com/gemrb/gemrb/assets/1279852/8a6c238a-9d93-4def-934d-6cd94c7af596)
immediately switches between full screen and windowed.

But switching with CTRL-F on the same page does not toggle the option.

burner1024:
#1990 actually will fix this as well. At least the proper startup mode.

I don't think the value from GemRB.cfg is used anywhere. It can probably be retired, and I think it would be consistent if CTRL-F also set and saved the option in gem-baldur.ini.

burner1024:
OK, actually `Fullscreen` appears to be used in some other places:
```bash
$ grep -R Fullscreen *
...
platforms/vita/README.Vita.md:Use "Fullscreen=1" to scale game area to native Vita resolution or "Fullscreen=0" to keep game area at the center of the screen.
[...]

PR #1997, opened 2024-02-15
Remove old Fullscreen option from GemRB.cfg, update docs. Ref #1989.

burner1024:
## Description
Remove old Fullscreen option from GemRB.cfg, update docs. Ref #1989.

Other video options are still read from cfg, though. Can be confusing too. Not sure why they were duplicated in the first place.

## Checklist

### DIFF
--- a/gemrb/core/Video/Video.cpp
+++ b/gemrb/core/Video/Video.cpp
@@ -20,6 +20,8 @@
 
 #include "Video.h"
 
+#include "Interface.h"
+
 #include "Palette.h"
 #include "Sprite2D.h"
 
@@ -169,6 +171,8 @@ void Video::SetScreenClip(const Region* clip)
 
 bool Video::ToggleFullscreenMode()
 {
+	auto& vars = core->GetDictionary();
+	vars.Set("Full Screen", !fullscreen);
 	return SetFullscreenMode(!fullscreen);
 }
 
--- a/gemrb/core/Video/Video.cpp
+++ b/gemrb/core/Video/Video.cpp
@@ -20,8 +20,6 @@
 
 #include "Video.h"
 
-#include "Interface.h"
-
 #include "Palette.h"
 #include "Sprite2D.h"
 
@@ -171,8 +169,6 @@ void Video::SetScreenClip(const Region* clip)
 
 bool Video::ToggleFullscreenMode()
 {
-	auto& vars = core->GetDictionary();
-	vars.Set("Full Screen", !fullscreen);
 	return SetFullscreenMode(!fullscreen);
 }
 
--- a/platforms/vita/README.Vita.md
+++ b/platforms/vita/README.Vita.md
@@ -68,7 +68,7 @@ Keyboard input is done with D-Pad (on character creation and game saves. 'a-z', 
 
 Pointer movement speed can be changed with 'GamepadPointerSpeed' parameter in GemRB.cfg.
 
-Use "Fullscreen=1" to scale game area to native Vita resolution or "Fullscreen=0" to keep game area at the center of the screen.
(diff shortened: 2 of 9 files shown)


platforms/vita/README.Vita.md
```diff
@@ -68,7 +68,7 @@ Keyboard input is done with D-Pad (on character creation and game saves. 'a-z',
 
 Pointer movement speed can be changed with 'GamepadPointerSpeed' parameter in GemRB.cfg.
 
-Use "Fullscreen=1" to scale game area to native Vita resolution or "Fullscreen=0" to keep game area at the center of the screen.
+Use `Full Screen = 1` in game ini config (e.g. `gem-baldur.ini`) to scale game area to native Vita resolution or `Full Screen = 0` to keep game area at the center of the screen.
```
lynxlynxlynx, line 71: This is for people that just installed it, so the ini wouldn't exist yet. So it's better this way:
```suggestion
Use `Full Screen = 1` in game ini config (e.g. `baldur.ini`) to scale game area to native Vita resolution or `Full Screen = 0` to keep game area at the center of the screen.
```

lynxlynxlynx:
I bet that if you looked at git history, you'd find that these settings predate any reading of the game inis. There is value in keeping eg. the resolution in our own cfg.

This deprecates the key, but doesn't solve the issue. Do you plan to work on that in this PR, elsewhere or leave it for later?

lynxlynxlynx:
They get all loaded into the same dictionary? Ok, then, didn't know.
That leaves just ctrl-f (`WindowManager::HotKey`) not using the same mechanism.

lynxlynxlynx:
It should also toggle the variable. I don't think we need to make it permanent (write the config) though.

burner1024:
Looks like toggling the variable **is** what writes it to config.
(Or rather, set it up to be written, actual writing happens on game exit as far as I can see).

lynxlynxlynx:
Do you plan to finish this? Only that var toggling is missing.

burner1024:
You said you don't want write to config, toggling the var writes it.

lynxlynxlynx:
It doesn't, we only save if you go to options. Now if you hotkeyed to fullscreen and went there, the button state would be wrong.

burner1024:
If you say so.

gemrb/core/Video/Video.cpp
```diff
@@ -169,6 +171,8 @@ void Video::SetScreenClip(const Region* clip)
 
 bool Video::ToggleFullscreenMode()
 {
+	auto& vars = core->GetDictionary();
+	vars.Set("Full Screen", !fullscreen);
```
lynxlynxlynx, line 175: This should be done in the culprit, WindowManager::HotKey, which will also avoid tangling Video with Interface.

burner1024: Sorry, I can't get it to work.
Tried this and variants.
```diff
 bool WindowManager::HotKey(const Event& event) const
 {
        if (event.type == Event::KeyDown && event.keyboard.repeats == 1) {
+               auto& vars = core->GetDictionary();
+               bool fullscreen;
                switch (event.keyboard.keycode) {
                        case 'f':
                                video->ToggleFullscreenMode();
+                               fullscreen = (bool) vars.Get("Full Screen");
+                               vars.Set("Full Screen", !fullscreen);
                                return true;
                        case GEM_GRAB:
                                video->ToggleGrabInput();
```

lynxlynxlynx: You're using the wrong `Get` — the one without a fallback returns a pointer, so that's why `fullscreen` remains always true.

burner1024:
Oh ok.

lynxlynxlynx:
Thanks!

Issue #1991, opened 2024-02-11
Perfectionist's hell 


burner1024:
#### Bug description
Misaligned buttons in main menu.

#### Steps to reproduce
1. Start BG2.
2. Click on Quit/Options. Watch the button borders.

#### Expected behavior
Aligned! Buttons! Aligned!

#### Screenshots
Open the attached files in full screen mode, and flip between them fast.
[menu.zip](https://github.com/gemrb/gemrb/files/14233237/menu.zip)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

burner1024:
I looked in GUIscripts, but didn't find any coordinates.

bradallred:
we have some functions to do this sort of alignment in the CharGen. You can do something similar here.

https://github.com/gemrb/gemrb/blob/672e8315c9f8a3aaea71db0839f5e4bea849634d/gemrb/GUIScripts/bg2/CharGenCommon.py#L33-L38

burner1024:
That seems to be for the position of the window itself, not buttons within the window?

bradallred:
The problem is that the window is not aligned with the one below it. Buttons are almost always baked into the CHU background.

burner1024:
I think something else is going on here.
Here's comparison of the main screen, original and GemRB. Different screenshot resolution, but same game resolution, fullscreen view should align them perfectly.
As far as I understand, there are no overlaid windows here, just the one main window.
3 buttons are off.

[cmp.zip](https://github.com/gemrb/gemrb/files/14312518/cmp.zip)

bradallred:
right, the problem I thought you were describing we fixed in #433 using the technique I described.

But you can see for yourself in NI that the "buttons" are baked into the CHU background. The background is `GUIVERB.MOS`

<img width="559" alt="image" src="https://github.com/gemrb/gemrb/assets/1126744/9b51eada-46c1-4aa9-b898-5351e06c37b4">

> As far as I understand, there are no overlaid windows here, just the one main window.

No, that is not correct. You can see for yourself in NI or look in the Start2.py script.

burner1024:
Yes, I saw the CHUs and #433, and Start2.py.

[...]

PR #1998, opened 2024-02-17
Custom menu button fixes for BG2

burner1024:
## Description
Menu alignment fixes for BG2:  #1991.

Alignment is correct now, but there's still some barely noticeable twiching on these screens, in these areas:

a
b

Not sure where it comes from, center art overlay order? Scaling/rendering?
But you need to really look for it to notice.

Anyway, this is better than before.
 
## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg2/Start2.py
+++ b/gemrb/GUIScripts/bg2/Start2.py
@@ -19,6 +19,7 @@
 #this is essentially Start.py from the SoA game, except for a very small change
 import GemRB
 import GameCheck
+from StartFix import ShiftWindow
 
 def OnLoad():
 	StartWindow = GemRB.LoadWindow (0, "START")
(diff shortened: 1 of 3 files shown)


bradallred:
I'm still not sure why we feel like this is an issue with button positions and not the MOS rendering or upscaling. I haven't seen a per pixel comparison with the original at the same resolution (without hdpi active) but you should compare both sides of the button when doing so.

It's hard to believe the originals would have shifted buttons around and it makes me feel like this is going to turn up more places. Your patch appears to just shift everything left by one so have we checked other UIs to see if maybe this is just a general issue with rounding?

```cpp
if (Image) {
	// FIXME: maybe it's useless...
	Point offset((frame.w / 2) - (Image->Frame.w / 2), (frame.h / 2) - (Image->Frame.h / 2));
	VideoDriver->BlitSprite(Image, rgn.origin + offset);
}
```

maybe just tweak that?

burner1024:
Huh? I did think it's something else, and was told to just move the buttons.
I haven't noticed this kind of shifting elsewhere.

lynxlynxlynx:
It can't be the MOS, since only 3 of the buttons look misaligned. But Brad may be on to something, as the top right and bottom left buttons do have the same x position, but the top right is one pixel wider than its image. 

I didn't know we do centering, but the alignment flags are just for text. It seems it was added on a whim in 2004: 4c378e053. :)
I don't have python2 handy anywhere anymore, but it would be easy to check all the CHUs with iesh and find any other buttons whose size does not match their image.

burner1024:
I tried removing offset from
```cpp
if (Image) {
	// FIXME: maybe it's useless...
	Point offset((frame.w / 2) - (Image->Frame.w / 2), (frame.h / 2) - (Image->Frame.h / 2));
	VideoDriver->BlitSprite(Image, rgn.origin + offset);
}
```
It achieves the same result as this pull request. On the first glance, I don't see any side effects.

bradallred:
We would also need to make the same changes to `Button::HitTest`.

burner1024:
Should I make the changes and send a new PR?

lynxlynxlynx:
Can you check if the rest of the guis look ok, also in other games?

burner1024:
I do not see any interface issue in BG1, BG2, PST, IWD1. That's all I have at the moment.
But my experience is mostly limited to BG2.

lynxlynxlynx:
ok, let's try it then.

Issue #2010, opened 2024-02-22
BG1: Tooltips delay option isn't properly loaded


bsdcode:
#### Bug description
When the tooltips delay is set to maximum, tooltips should never pop up on its own, only through pressing TAB. It works in your current play session, but when you load the game later on the delay is set to maximum in the options screen visually but tooltips show up after ~3s. I assume the option value is off a little bit when loading a game.

Original Baldur's Gate has `Tooltips=2147483647` in `baldur.ini` for the maximum delay value. GemRB stores it as `Tooltips = 100` in `gem-baldur.ini`.

#### Steps to reproduce
1. Set tooltips option to maximum delay
2. No tooltips show up, only by pressing TAB (as intented)
3. Load another game
4. Tooltips slider is at the maximum position in the option screen
5. But tooltips show up

#### Expected behavior
Tooltips option set to maximum prevents automatic tooltips.

#### Screenshots
![bg1_gemrb](https://github.com/gemrb/gemrb/assets/155195419/7473d3ff-c776-41ff-9405-b4524c99925e)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

burner1024:
Uhm #1797 was supposed to fix it.
I just tried master in BG1, and I don't see tooltips on max setting.

PR #2016, opened 2024-02-22
Fix tooltips max setting on game restart, ref #2010.

burner1024:
## Description

GUIopt.py increase only applies if you played around in Options windows.
With cold start, delay setting would pass directly to WindowManager and never get the multiplier.
This moves the final multiplier into WindowManager so it would apply on game start as well.

## Checklist

### DIFF
--- a/gemrb/core/GUI/WindowManager.cpp
+++ b/gemrb/core/GUI/WindowManager.cpp
@@ -22,7 +22,6 @@
 #include "GameData.h"
 #include "Interface.h"
 #include "ImageMgr.h"
-#include "Tooltip.h"
 #include "Window.h"
 #include "GUI/GameControl.h"
 #include "GUI/GUIFactory.h"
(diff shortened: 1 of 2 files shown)


gemrb/core/GUI/WindowManager.cpp
```diff
@@ -41,7 +42,8 @@ Holder<Sprite2D> WindowManager::CursorMouseDown;
 
 void WindowManager::SetTooltipDelay(int delay)
 {
-	ToolTipDelay = delay;
+	// Max setting disables tooltips in originals, we just set it to an extremely high value for simplicity.
+	ToolTipDelay = (delay == 10 * Tooltip::DELAY_FACTOR) ? 1e7 : delay;
```
lynxlynxlynx, line 46: Please just write the number out, which will also avoid a pointless cast.

burner1024: Like this?

gemrb/core/GUI/WindowManager.cpp
```diff
@@ -43,7 +42,8 @@ Holder<Sprite2D> WindowManager::CursorMouseDown;
 void WindowManager::SetTooltipDelay(int delay)
 {
 	// Max setting disables tooltips in originals, we just set it to an extremely high value for simplicity.
-	ToolTipDelay = (delay == 10 * Tooltip::DELAY_FACTOR) ? 1e7 : delay;
```
lynxlynxlynx: No no, I meant the `1e7` double, of course named constants are fine.

burner1024: okay

lynxlynxlynx:
Thanks!

Issue #2012, opened 2024-02-22
BG1: Opening doors doesn't play soundeffect


bsdcode:
#### Bug description
When opening doors you should hear a soundeffect, but GemRB stays silent. "Opening" shelves, like the shelves in the Candlekeep Inn, have the proper soundeffect, so it seems to only affect doors.

#### Steps to reproduce
1. Open Candlekeep Inn door and expect "door opening"-soundeffect
2. You won't hear it

#### Expected behavior
Door opening play soundeffect.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
This door has no opening or closing sound defined. I guess we need a fallback in `Door::ToggleTiles`. It's the now unused default sounds:
```
OPEN    AMB_D03A    
CLOSE   AMB_D03B
```
Their failure variants are used in gemrb/core/GameScript/Actions.cpp:2623 .

lynxlynxlynx:
Ah, even better, fixed in the importer. :)
That would be then one of the first thing to check — what happens when loading door2616. The other would be what goes on in Door::ToggleTiles (are we calling it without playsound?).

bsdcode:
I traced it. The `openResRef` is correctly set to the AMB_D03A resource. What hinders the sound to play is `GameScript::ToogleDoor`. There we have
```
door->SetDoorOpen(!door->IsOpen(), false, actor->GetGlobalID());
```
Actually, all occurences of SetDoorOpen, ToggleDoor, etc... have playsound=false. Weird, door sounds never play ;)

Now the question is if we can toggle from false to true in that snippet above? Seven lines above it we have the `core->PlaySound` line which plays a sound when opening a door is not successful, so this function already doesn't care about being silent. Maybe just switch?

PR #2022, opened 2024-02-23
Fix missing door soundeffects, ref #2012

bsdcode:
## Description

Opening or closing a door is supposed to play a corresponding soundeffect. The responsible function `GameScript::ToggleDoor` currently calls `SetDoorOpen` with `playsound=false`. Because the function already does play a sound when opening the door fails, it seems sensible to change to `playsound=true` when calling `SetDoorOpen`.

## Checklist


lynxlynxlynx:
Sorry, but it's more complicated than that. It was disabled in 9a50e3d269 / 0ee054de27, so for iwds it should definitely remain as false. Likely all scripted uses should be like that.

It just happens to be that we currently map the user clicking on a door with the same action (via an alias, in gemrb/core/GUI/GameControl.cpp:1950). So I think it's safer to just play it for interactive uses for now.

Issue #2017, opened 2024-02-23
BG2 multiplayer button exception 


burner1024:
#### Bug description
```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "/opt/gemrb/share/gemrb/GUIScripts/bg2/Start2.py", line 107, in MultiPlayerPress
[Python/ERROR]:     OptionsButton.SetText ("")
[Python/ERROR]: NameError: name 'OptionsButton' is not defined
```

#### Steps to reproduce
1. Launch BG2
2. Click on multiplayer button.

#### Expected behavior
Should it be just disabled/greyed out? Since it doesn't do anything.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Both things can be done, fixing the code, and disabling its execution. It's unlikely to ever change, so it could also just be removed.

PR #2023, opened 2024-02-24
Disable MultiPlayer button in BG2, cleanup related GUI code. Ref #2017.

burner1024:
## Description
Conservatively disabled MultiPlayer button, #2017.

Captura desde 2024-02-24 19-40-12
Captura desde 2024-02-24 19-40-20
Captura desde 2024-02-24 19-40-26

## Checklist


lynxlynxlynx:
Thanks!

burner1024:
I think maybe worth repurposing it for "load last save" or "quick load".

lynxlynxlynx:
Quick load is much easier, but I think we should add it to the first screen (twice), so it's even faster.

Issue #2009, opened 2024-02-22
BG1: New game is missing intro text


bsdcode:
#### Bug description
When starting a new game you start outside Candlekeep Inn and you should see an intro text in the text panel. With GemRB it isn't there.

#### Expected behavior
Intro text is there.

#### Screenshots
![bg1_wine](https://github.com/gemrb/gemrb/assets/155195419/45e3e3f6-7fb2-4e13-8727-121aeef3629d)

![bg1_gemrb](https://github.com/gemrb/gemrb/assets/155195419/8c6abe58-be72-4113-a89d-bb48df1de8b7)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
That's trivial to add in bg1's Game.py or somesuch. We can also fix the color of GOT_GOLD .

bsdcode:
Ok, with the help from NearInfinity I found the StrRef number for the text: `16214`. Now I need to find the right place to do a `DisplayString` with it. Looking through share/GUIScripts/bg1...

lynxlynxlynx:
Actually gemrb/GUIScripts/bg1/Start.py might be a better place than Game.py; if it works there after the LoadGame (in NewSingle), that's a one-liner. Otherwise we'd need extra logic to display it only once.

bsdcode:
Yeah, doesn't work in NewSingle form Start.py. At this point we haven't even loaded the character creation screen.

It works when putting `GemRB.DisplayString(16214, ColorWhite)` after `MessageWindow.OnLoad()` in Game.py. Of course it also displays when loading a saved game, so we must restrict it to a new game. According to Near Infinity the original script checks the chapter number and displays this message only when it's still chapter 0. Maybe we can do a similar check?

lynxlynxlynx:
Ah, it was scripted in the area script. Then we run that script already and shouldn't duplicate it. I guess the intro story screen resets Chapter to a higher number already, so when the area script runs again, the block doesn't execute:
```
IF
  Global("Chapter","GLOBAL",-1)
THEN
  RESPONSE #100
    IncrementChapter("Chptxt0")
    DisplayString(Myself,16214)
[...]

PR #2025, opened 2024-02-24
Fix missing intro text, ref #2009

bsdcode:
In accordance to similar actions, DisplayString should also emit the DS_CONSOLE flag when not called by an ACTOR.
DisplayStringCore then should not cancel the DS_CONSOLE flag again when not called by an ACTOR.

## Checklist

### DIFF
--- a/gemrb/core/GameScript/Actions.cpp
+++ b/gemrb/core/GameScript/Actions.cpp
@@ -1728,8 +1728,9 @@ void GameScript::DisplayString(Scriptable* Sender, Action* parameters)
 
 	int flags = DS_CONSOLE;
 
-	if (Sender->Type != ST_ACTOR)
+	if (Sender->Type != ST_ACTOR) {
 		flags |= DS_AREA;
+	}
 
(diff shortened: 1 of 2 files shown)


gemrb/core/GameScript/Actions.cpp
```diff
@@ -1725,11 +1725,13 @@ void GameScript::DisplayString(Scriptable* Sender, Action* parameters)
...
-	if (Sender->Type==ST_ACTOR) {
-		DisplayStringCore(target, ieStrRef(parameters->int0Parameter), DS_CONSOLE);
-	} else {
-		DisplayStringCore(target, ieStrRef(parameters->int0Parameter), DS_AREA);
-	}
+
+	int flags = DS_CONSOLE;
+
+	if (Sender->Type != ST_ACTOR)
+		flags |= DS_AREA;
```
lynxlynxlynx, line 1732: Please either use braces or put it into a single line.

bsdcode: Sorry I forgot. Added missing braces.

lynxlynxlynx:
Thanks!

Issue #2011, opened 2024-02-22
BG1: Doors and shelves are highlighted on complete surface not just the outline


bsdcode:
#### Bug description
Pointing on a door or a shelf highlights the whole object surface. It should only highlight the outline.

#### Expected behavior
Only highlight outline of doors etc.

#### Screenshots
![bg1_wine0](https://github.com/gemrb/gemrb/assets/155195419/d4b1d9e8-de6e-4656-8ff8-9da68101e18a)
![bg1_wine1](https://github.com/gemrb/gemrb/assets/155195419/79de8d64-e8df-4779-92da-b697538cfb70)

![bg1_gemrb0](https://github.com/gemrb/gemrb/assets/155195419/4b1ade9c-282e-4bdc-a131-01bb8b8f45b0)
![bg1_gemrb1](https://github.com/gemrb/gemrb/assets/155195419/a9f3c807-02c9-40a9-8150-b2fe4f0c6a27)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
It's easy to change (GameControl::Draw or thereabout), since we have to draw them separately anyway. But it would make it different from everyone else, so a new game flag would be needed.

bsdcode:
I only know Baldur's Gate 1. Does this mean that the other Infinity Engine games are highlighting the complete surface instead of just the outline?

lynxlynxlynx:
Yes, like we do now.

bsdcode:
Ok. I actually like the aesthetics of just the outline more. Personally I would rather see it in the other games too.

Can I add a new option in gem-baldur.ini for controlling the highlight behaviour, defaulting to the current way, i.e. highlight the whole surface by default. This way everyone can decide how they want it to be highlighted for all games.

lynxlynxlynx:
That's what I meant with game flag. You can grep for `GFFlags` to get started

bsdcode:
Alright.

PR #2027, opened 2024-02-24
Add flag to only highlight outline of doors and containers, ref #2011

bsdcode:
## Description

Add the 'HighlightOutlineOnly'-flag, which toggles whether the game highlights only the outline of doors and containers (like in BG1), or highlights the complete surface of the object (all except BG1). In accordance with the original games, this flag is enabled for BG1 and disabled for the other games by default.

HighlightOutlineOnly=0
outline0

HighlightOutlineOnly=1
outline1

## Checklist

### DIFF
--- a/gemrb/core/Scriptable/Scriptable.cpp
+++ b/gemrb/core/Scriptable/Scriptable.cpp
@@ -1727,11 +1727,17 @@ void Highlightable::DrawOutline(Point origin) const
 	}
 	origin = outline->BBox.origin - origin;
 
-	if (core->HasFeature(GFFlags::PST_STATE_FLAGS)) {
-		VideoDriver->DrawPolygon(outline.get(), origin, outlineColor, true, BlitFlags::MOD | BlitFlags::HALFTRANS);
-	} else {
-		VideoDriver->DrawPolygon( outline.get(), origin, outlineColor, true, BlitFlags::BLENDED|BlitFlags::HALFTRANS );
-		VideoDriver->DrawPolygon( outline.get(), origin, outlineColor, false );
+	bool highlightOutlineOnly = core->HasFeature(GFFlags::HIGHLIGHT_OUTLINE_ONLY);
+	bool pstStateFlags = core->HasFeature(GFFlags::PST_STATE_FLAGS);
+
+	if (!highlightOutlineOnly) {
+		BlitFlags flag = BlitFlags::HALFTRANS | (pstStateFlags ? BlitFlags::MOD : BlitFlags::BLENDED);
+
+		VideoDriver->DrawPolygon(outline.get(), origin, outlineColor, true, flag);
+	}
+
+	if (highlightOutlineOnly || !pstStateFlags) {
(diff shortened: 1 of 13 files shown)


bsdcode:
#2011

gemrb/core/Scriptable/Scriptable.cpp
```diff
@@ -1727,11 +1727,17 @@ void Highlightable::DrawOutline(Point origin) const
...
+	bool highlightOutlineOnly = core->HasFeature(GFFlags::HIGHLIGHT_OUTLINE_ONLY);
+	bool pstStateFlags = core->HasFeature(GFFlags::PST_STATE_FLAGS);
+
+	if (!highlightOutlineOnly) {
+		BlitFlags flag = BlitFlags::HALFTRANS | (pstStateFlags ? BlitFlags::MOD : BlitFlags::BLENDED);
+
+		VideoDriver->DrawPolygon(outline.get(), origin, outlineColor, true, flag);
+	}
+
+	if (highlightOutlineOnly || !pstStateFlags) {
```
lynxlynxlynx, line 1739: The pst check is now redundant here, so it can be removed.

bsdcode: Why is it redundant exactly? According to the previous code we should not draw the outline if the pst flag is set.

lynxlynxlynx: Now there's a configuration option available with the same effect. If anyone wanted to change the behaviour, it would unexpectedly have no effect, which is bad.

bsdcode: Ah ok, now I understand where you're coming from. Still, as long as we want to have the previous behaviour with the pst flag available alongside the new option, we need such a test:

- HOO=0, PST=0 : (fill with BlitFlags::BLENDED) + (outline)
- HOO=0, PST=1 : (fill with BlitFlags::MOD)
- HOO=1, PST=0 : (outline)
- HOO=1, PST=1 : (outline)

The first two cases represent the previous behaviour for which we need the test. It is the last case which generates a conflict in which option has precedence. But how to resolve? Always draw the outline? Would that be ok for PST?

bsdcode: Now, what does the PST_STATE_FLAG actually control in the context here? It controls the BlitFlag. That's actually all it does here. So I think we can resolve the conflict with always drawing the outline, but we also apply the BlitFlags to the outline which is depending on the pst flag.

With this we can seperate the concerns of both flags:
- HOO: only draw outline
- PST: different blending/color

lynxlynxlynx: Ah, I see, I was just looking at pst, not everyone else.  So it's the highlight check that's redundant in that line comparing to the old behaviour. But this way it does become configurable.

lynxlynxlynx:
Thanks!

Issue #2007, opened 2024-02-22
 BG1: Voice selection shows only the default voice in the character creation


bsdcode:
#### Bug description
Window to choose the character's voice only shows the default entry.
(And the inactive window in the back should actually be darker than the active foreground voice selection window).

#### Expected behavior
Window shows all available voices.

#### Screenshots
![bg1_wine](https://github.com/gemrb/gemrb/assets/155195419/baa18eca-2f60-4e0e-8c28-aa67eb23ab1c)

![bg1_gemrb](https://github.com/gemrb/gemrb/assets/155195419/8f467a4f-ba4b-4c9a-9898-ff7b2d360699)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Thanks, can reproduce — it's not a German-related problem.

burner1024:
And the "play" volume is soooo low. At first I thought it didn't play at all.

burner1024:
Try loading a save first, the exiting to main menu and starting new one.

bsdcode:
Yes. Can reproduce this way.

bsdcode:
I haven't looked deeper, but maybe something is wrong here regarding the volume problem:

https://github.com/gemrb/gemrb/blob/253e06cac9eb8247e101235198442278bfda422d/gemrb/GUIScripts/bg1/GUICG19.py#L80-L81

burner1024:
Might be same as this https://github.com/gemrb/gemrb/issues/1993#issuecomment-1962349636... sound system appears to have some sort of lingering.

bsdcode:
Missing voices in list:

https://github.com/gemrb/gemrb/blob/253e06cac9eb8247e101235198442278bfda422d/gemrb/plugins/GUIScript/GUIScript.cpp#L4686

borks `dirit` somehow. I'm investigating.

lynxlynxlynx:
I don't see anything in valgrind, so it's less ominous than you make it sound. You could just bisect it, without giving it much thought, since it did work at the time of the last release.

bsdcode:
[...]

PR #2028, opened 2024-02-25
Fix predicate EndsWithFilter, ref #2007

bsdcode:
## Description

Replace stricmp with strnicmp in order to compare only as much characters as the predicate string has. Otherwise the predicate may be never fulfilled.

## Checklist


lynxlynxlynx:
Great, thanks!

Issue #2005, opened 2024-02-22
 BG1: Big delay by loading single player main menu when using Game Text Update by Baldurdash


bsdcode:
#### Bug description
I use the original german 5-disc-set version of Baldur's Gate 1 with the `dialog.tlk` replaced by Baldurdash's Game Text Update `dialog.tlk`. When I start gemrb and click "single player" in the main menu then there's a delay of ~24s before the program continues to show the next main menu screen. The log shows
```
[Streams/ERROR]: Invalid seek position 1093201284 in file dialog.tlk (limit: 3100217)
```
The program probably tries to scan through this or more files before it continues.
It doesn't happen with the original `dialog.tlk`. It also doesn't happen with the 0.9.2 release version.

#### Steps to reproduce
1. Replace `dialog.tlk` with Baldurdash's Game Text Update
2. Start gemrb
3. Click "Single Player"
4. Experience long delay before it continues

#### Expected behavior
No delay.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Sounds like a mod data problem, but we should be more resilient.

lynxlynxlynx:
BTW, the G3 Fixpack has the latest GTU.
Does Baldurdash have a German GTU? If not, that could explain wrong strings, but I don't see how it could cause any issues that aren't just display related — unless the file is malformed.

bsdcode:
No it doesn't have a German GTU.

The cause of the big delay is a huge returned string length `l`, about ~1.5GB. And this big `mbstr` then wants to be converted to StringView in `StringFromTLK`'s constructor.

bsdcode:
Is the G3 Fixpack also applicable to BG1? If so I would be happy to try it.

bsdcode:
It's the StringRef `24110`. NI shows "No such index".

lynxlynxlynx:
In the official bg1 that's "mission pack" for its menu button (the shipped save at the start of TotSC). But that suggests indeed that the file is corrupted.

bsdcode:
[...]

PR #2029, opened 2024-02-25
Make TLKImporter more resilient, ref. #2005

bsdcode:
## Description

Do not process invalid seek results. These are garbage anyway and can slow down the subsequent processing considerably.

This affects mostly mods, e.g. Baldurdash's GameTextUpdate for BG1.

#2005

## Checklist


lynxlynxlynx:
How does the main menu look after this for you? Did you identify which strings it had problems with?

bsdcode:
The main menu looks perfectly fine. I don't think that it were real strings, because the seek position is way beyond the limit. `mbstr` was empty after `str->Read(&mbstr[0], l);`. Nothing to show.

Issue #2030, opened 2024-02-25
BG & OpenAL: Speech handling broken in CharGen in some circumstances


MarcelHB:
#### Bug description
In BG or BG 2, when going through CharGen and the PC's sound table, the playback is almost mute. What we know:

* It's only observable BG/2, IWD games look good (Python script is using another method there)
* When going directly into GC without loading a game first, it's Ok.
* It does not appear to be related to any volume or position corruption, debugging says it's 100% at (0, 0), as expected.
* It is independent of having EFX enabled.
* It is located around `GEM_SND_SPEECH` since if you set non-replacing playback, i. e. the speech flag, in the Python script, everything is Ok.
* When you return to an actual game, all PC voices are good again.
* There is no error returned by the AL implementation around the sound settings before queuing the buffers.

I've spent some time debugging this without any findings so far, yet is remarkable in some way as it is fully reproducible and the code in question appears to work for every other case but _returning_ to the GC from a game – which itself is somewhat an edge case, so maybe only for reference to report here.

#### Steps to reproduce
0. Use OpenAL as sound driver.
1. Start a BG2 game by loading a save game.
2. Exit the game.
3. Start a new game via CG.
4. When choosing any of the sound sets, click the playback button.

#### Expected behavior
Full volume for the voice.

#### Screenshots
You can only hear somthing on the lowest possible volume in the far left channel. Virtually nothing on bad speakers.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

lynxlynxlynx:
Is it just about the actor soundset? It looks obvious to me that the two PlaySound calls should use 5 instead of 4, so there's no spatial positioning done.

MarcelHB:
That makes sense, the position of the listener is probably just left somewhere by the game ... that's why we don't notice in the first run.

Issue #1993, opened 2024-02-12
Reverbation should not apply to all sounds 


burner1024:
#### Bug description
It seems that reverbation applies to all sounds.
It shouldn't apply at least to:
- inventory sounds (moving stuff around)
- interface sounds (clicking buttons and stuff)

#### Steps to reproduce
1. Start a new BG2 game.
2. Click on any interface button: options, record, etc. Move some stuff in inventory.

#### Expected behavior
No reverbation for inventory sounds, interface sounds.
Not sure what kinds/channels are else there. It definitely should apply to ambients, voices.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

burner1024:
Reverbation remains even after you exit from game to main menu.

lynxlynxlynx:
We can probably get away with limiting it to certain sound channels, ie. not using it for SFX_CHAN_GUI and SFX_CHAN_NARRATOR.

Issue #1715, opened 2022-11-14
BG/BG2/PST: menu music keeps playing on first load 


MarcelHB:
#### Bug description
For at least in PST, BG & BG2, you keep hearing the menu music playing even after entering the first savegame. A new load, or changing the map finally makes it stop.

IWD & IWD2 look good. Using OpenAL.

#### Steps to reproduce
1. Enter BG/BG2/PST and load a save game.

#### Expected behavior
No menu music after entering the game

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

bradallred:
looks to be a dupe of #1643

MarcelHB:
I'm sure this wasn't a problem until some time ago. Currently into something else, but maybe find what made it broke for what reason.

MarcelHB:
What about doing it like this?
```diff
@@ -2541,10 +2541,11 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
        if (IsStar(*poi) && !MasterArea && SongType == SONG_BATTLE && core->HasFeature(GF_BREAKABLE_WEAPONS)) {
                poi = &core->GetMusicPlaylist(SongType);
                pl = SongType;
+       } else if (IsStar(*poi)) {
+               core->GetMusicMgr()->HardEnd();
+               return;
        }

-       if (IsStar(*poi)) return;
-
        //check if restart needed (either forced or the current song is different)
        if (!restart && core->GetMusicMgr()->IsCurrentPlayList(*poi)) return;
        int ret = core->GetMusicMgr()->SwitchPlayList(*poi, hard);
```

If `***` really encodes *no music*, then there is no music on this map and we stop what's playing, like the menu music (this is why it only affects some savegames: those without MUS file on the map), with exceptions from that rule to be encoded right before that.

lynxlynxlynx:
I think it's better to just call Soft/HardEndPL from python or before loading a map object for the first time. Menu music is not map music, so let's keep them separated. TextScreen.py already does it, but that's relevant only for new games, not loading.

PR #2031, opened 2024-02-25
Audio improvements/fixes

MarcelHB:
## Description
Some things to address #2030, #1993 and apparently something regressed/unfixed after #1715.

## Checklist

### DIFF
--- a/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
+++ b/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
@@ -515,7 +515,7 @@ Holder<SoundHandle> OpenALAudioDriver::Play(StringView ResRef, unsigned int chan
 #ifdef HAVE_OPENAL_EFX_H
 	ieDword efxSetting = core->GetDictionary().Get("Environmental Audio", 0);
 
-	if (efxSetting && hasReverbProperties && (!p.IsZero() || (flags & GEM_SND_RELATIVE))) {
+	if (efxSetting && hasReverbProperties && (flags & GEM_SND_RELATIVE)) {
 		alSource3i(Source, AL_AUXILIARY_SEND_FILTER, efxEffectSlot, 0, 0);
 	} else {
 		alSource3i(Source, AL_AUXILIARY_SEND_FILTER, 0, 0, 0);
--- a/gemrb/GUIScripts/GUILOAD.py
+++ b/gemrb/GUIScripts/GUILOAD.py
@@ -162,7 +162,7 @@ def LoadGamePress (btn):
 		LoadScreen.StartLoadScreen ()
 		# it will close windows, including the loadscreen
 		GemRB.EnterGame ()
-		GemRB.SoftEndPL ()
+		GemRB.HardEndPL ()
 	return
 
 def GetQuickLoadSlot():
--- a/gemrb/GUIScripts/bg1/GUICG19.py
+++ b/gemrb/GUIScripts/bg1/GUICG19.py
@@ -78,7 +78,7 @@ def PlayPress():
 	while (not GemRB.HasResource (CharSound + SoundSeq[SoundIndex], RES_WAV)):
 		NextSound()
 	# play the sound like it was a speech, so any previous yells are quieted
-	GemRB.PlaySound (CharSound + SoundSeq[SoundIndex], CHAN_CHAR1, 0, 0, 4)
+	GemRB.PlaySound (CharSound + SoundSeq[SoundIndex], CHAN_CHAR1, 0, 0, 5)
 	NextSound()
 	return
 
--- a/gemrb/GUIScripts/bg2/GUICG19.py
+++ b/gemrb/GUIScripts/bg2/GUICG19.py
@@ -78,7 +78,7 @@ def PlayPress():
 	while (not GemRB.HasResource (CharSound + SoundSequence[SoundIndex], RES_WAV)):
 		NextSound()
 	# play the sound like it was a speech, so any previous yells are quieted
-	GemRB.PlaySound (CharSound + SoundSequence[SoundIndex], CHAN_CHAR1, 0, 0, 4)
+	GemRB.PlaySound (CharSound + SoundSequence[SoundIndex], CHAN_CHAR1, 0, 0, 5)
 	NextSound()
 	return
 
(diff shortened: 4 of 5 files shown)


gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -2944,7 +2944,7 @@ void AddXPCore(const Action *parameters, bool divide)
 		// give xp to everyone
 		core->GetGame()->ShareXP(atoi(xpvalue), 0);
 	}
-	core->PlaySound(DS_GOTXP, SFX_CHAN_ACTIONS);
+	core->PlaySound(DS_GOTXP, SFX_CHAN_GUI);
```
lynxlynxlynx, line 2947: Did you figure these out with testing via sndchann,2da? It's quite unintuitive that actions wouldn't use the actions channel.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
```diff
@@ -515,7 +515,7 @@ Holder<SoundHandle> OpenALAudioDriver::Play(StringView ResRef, unsigned int chan
 #ifdef HAVE_OPENAL_EFX_H
 	ieDword efxSetting = core->GetDictionary().Get("Environmental Audio", 0);
 
-	if (efxSetting && hasReverbProperties && (!p.IsZero() || (flags & GEM_SND_RELATIVE))) {
+	if (efxSetting && hasReverbProperties && channel != SFX_CHAN_GUI && channel != SFX_CHAN_NARRATOR) {
```
lynxlynxlynx, line 518: Come to think of it, shouldn't we also check for GEM_SND_RELATIVE, just in the opposite way (not present)? I can't think of any non-spatial sound that should have spatial effects applied. Something like:
```c++
if (efxSetting && hasReverbProperties && ((channel != SFX_CHAN_GUI && channel != SFX_CHAN_NARRATOR) || !(flags & GEM_SND_RELATIVE))) {
```

gemrb/core/GameScript/GSUtils.cpp
MarcelHB: Hm, no. I thought of things that happen on the map (like lock picking etc.) vs. things help you notice something on the GUI for information.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
MarcelHB: Then again we can also drop the whole channel check? Unless there are some flags set wrong, like the sound set thing in CG, none of these two (GUI, narrator) is spatial? I guess that simply many cases had (0, 0) for the default value.

gemrb/core/GameScript/GSUtils.cpp
lynxlynxlynx: I suggest you test it in the original then. Set the various channels to 0 and you should be able to confirm which one is used. Actions can be ran through the cheat console too.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
lynxlynxlynx: Yeah, we can simplify. I think GEM_SND_RELATIVE casts a bigger net, so that seems better.

gemrb/core/GameScript/GSUtils.cpp
MarcelHB: I think this can be reverted then with the other idea of looking at the relative flag. The idea was to keep some noise out of the reverb case spectrum.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
MarcelHB: Not so useful r/n, since a lot of things go through this and `PlayRelative` applies that flag regardless: https://github.com/gemrb/gemrb/blob/d0af7b1c9952e00489cfe0c39056760eb0e8c3dd/gemrb/core/Interface.cpp#L3396-L3402

And this defsound.2da is both GUI and actions.

lynxlynxlynx: Sure, but isn't the check currently reversed?

MarcelHB: That's master: 
```cpp
... && (!p.IsZero() || (flags & GEM_SND_RELATIVE))
```
And by then by that `PlayRelative` many sounds turn into GEM_SND_RELATIVE.

lynxlynxlynx: That's what I'm saying, shouldn't it be `!(flags & GEM_SND_RELATIVE)`? Relative ie. non-spatial.

MarcelHB: Eh, ok. My understanding was relative ie. relative to the listener position.

MarcelHB: Took me a second what was going on there. But I guess we have no good alternative to relying on channels: If you click the PCs, they are fixed towards the listener (GEM_SND_RELATIVE) but since they are on the map, we want them to get the filtered effect. Whereas other actors are subject to the camera position.

lynxlynxlynx: Yeah.

I've also checked how we use GEM_SND_RELATIVE and if I assume its name makes sense, don't we have these bugs? The way AL_SOURCE_RELATIVE works suggests I was wrong and the name is fine, but check these out:

reversed check:
gemrb/plugins/SDLAudio/SDLAudio.cpp:272 <-- plain wrong
gemrb/core/GameScript/GSUtils.cpp:522 <-- perhaps just a wrong comment, this would need more testing
gemrb/plugins/GUIScript/GUIScript.cpp:7459 <-- we're in a shop, why should this be relative? Also no pos

empty pos:
GemRB_VerbalConstant <-- we could it get from the actor trivially ... or not be relative at all
gemrb/plugins/TLKImporter/TLKImporter.cpp:375 <-- without a clear object to attach it to — should probably be non-spatial anyway

MarcelHB: I propose this: We remove GEM_SND_RELATIVE and introduce GEM_SND_SPATIAL for everything that is depending on the viewport's position _and_ EFX, and GEM_SND_EFX e. g. for PCs and dialog speaker sounds that are not spatial but that are subject to EFX, and fix these findings? Gets rid of this OAL name, too.

That GSUtils.cpp:522 is good to me in how it is thought of, since I'd expect always full volume on my party feedback, even if the PC is off-screen.

lynxlynxlynx: I agree, was thinking about renaming the constant as well. It'd be easier with NONSPATIAL, since there's fewer users to inspect, I guess, but the name is worse.

lynxlynxlynx: Any opposition to merging this now and dealing with the rest in a separate issue/PR?

MarcelHB: Do it. A little broken then but I'm going through it already.

lynxlynxlynx: It fixes two other things, so that's why I'm keen. :)

PR #2033, opened 2024-02-26
Brew fallback to libpng@1.6.40

czarny247:
## Checklist


czarny247:
I looked through the last succesfull builds for mac os and they worked with libpng 1.6.40, so for now maybe the fallback makes sense.

PR #2034, opened 2024-02-26
Libnpg include dirs fixed

czarny247:
## Checklist


czarny247:
@lynxlynxlynx Could you please kill this action https://github.com/gemrb/gemrb/actions/runs/8053767452 ?
It failed to finish somehow.

lynxlynxlynx:
Thanks, this was annoying. Didn't realize we hardcode the version.

.github/workflows/builder.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       run: |
         brew install sdl2 openal-soft sdl2_mixer libpng libvorbis sccache googletest
         brew reinstall libpng
+        brew --cellar libpng
+        brew list --versions libpng
+        brew list --versions libpng | egrep -o '[0-9]+\.[0-9]+\.[0-9]+'
```
lynxlynxlynx, line 25: These don't hurt, but now the pointless reinstall can be removed.

cmake/FindDependencies.cmake
```diff
@@ -42,8 +42,10 @@ ENDIF()
 
 IF(USE_PNG)
 	IF(APPLE AND DEFINED ENV{GITHUB_SHA})
-		SET(PNG_INCLUDE_DIRS "/usr/local/Cellar/libpng/1.6.40/include")
-		SET(PNG_LIBRARIES "/usr/local/Cellar/libpng/1.6.40/lib/libpng16.16.dylib")
+		#brew --cellar libpng -> /usr/local/Cellar/libpng
+		#brew list --versions libpng -> libpng 1.6.42
+		SET(PNG_INCLUDE_DIRS "/usr/local/Cellar/libpng/1.6.42/include")
+		SET(PNG_LIBRARIES "/usr/local/Cellar/libpng/1.6.42/lib/libpng16.16.dylib")
```
lynxlynxlynx, line 48: This way we're just kicking the ball down the line. Ideally we could just use globbing, but I'm sceptical that would work.

find_path can be used to locate the dylib, since that name doesn't change:
https://cmake.org/cmake/help/latest/command/find_path.html
(I suggest using NO_DEFAULT_PATH for greater robustness)

... and then you can append "../../include" and bam, we also have the value for the include dir.

.github/workflows/builder.yml
czarny247: Removed

cmake/DeduceLibpngPath.cmake
```diff
@@ -0,0 +1,17 @@
+FUNCTION(DEDUCE_LIBPNG_PATH)
```
lynxlynxlynx, line 1: this is platform-specfic, so should be in platforms/... instead.

cmake/DeduceLibpngPath.cmake
```diff
@@ -0,0 +1,17 @@
...
+	execute_process(
+		COMMAND brew list --versions libpng
+		OUTPUT_VARIABLE LIBPNG_WITH_VERSION
+		OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+	set(regex_pattern "([0-9]+\\.[0-9]+\\.[0-9]+)")
+	string(REGEX MATCH "${regex_pattern}" LIBPNG_VERSION "${LIBPNG_WITH_VERSION}")
```
lynxlynxlynx, line 8: what happens if brew returns more than one version?

cmake/FindDependencies.cmake
```diff
@@ -42,8 +42,15 @@ ENDIF()
 
 IF(USE_PNG)
 	IF(APPLE AND DEFINED ENV{GITHUB_SHA})
-		SET(PNG_INCLUDE_DIRS "/usr/local/Cellar/libpng/1.6.40/include")
-		SET(PNG_LIBRARIES "/usr/local/Cellar/libpng/1.6.40/lib/libpng16.16.dylib")
+		INCLUDE(DeduceLibpngPath)
+		DEDUCE_LIBPNG_PATH()
+
+		if("${LIBPNG_PATH}" STREQUAL "")
```
lynxlynxlynx, line 48: I'm pretty sure cmake treats empty strings as falsy, so this can be simplified.

cmake/FindDependencies.cmake
```diff
@@ -42,8 +42,15 @@ ENDIF()
...
+			message(ERROR "Cannot deduce LIBPNG_PATH!")
```
lynxlynxlynx, line 49: If this happens we shouldn't execute the following lines.

cmake/DeduceLibpngPath.cmake
lynxlynxlynx: `brew --prefix libpng` probably prints the path directly.

czarny247:
> Thanks, this was annoying. Didn't realize we hardcode the version.

Me too, I started with workaround until realized it :D

cmake/FindDependencies.cmake
czarny247: I failed making it with `find_path`. It is not so obvious to me and hard to test without macOS locally.

cmake/DeduceLibpngPath.cmake
lynxlynxlynx: It's a symlink to the actual location, so it should resolve just fine.

CMakeLists.txt
```diff
@@ -1,12 +1,23 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 3.11)
 
+SET(CMAKE_PROJECT_INCLUDE_BEFORE
+	"${CMAKE_SOURCE_DIR}/cmake/IncludeBeforeProject.cmake"
+)
+
+PROJECT(gemrb)
+
```
lynxlynxlynx, line 8: why did you move this around?

CMakeLists.txt
```diff
@@ -1,12 +1,23 @@
...
 SET(CMAKE_CXX_STANDARD 14)
 
 LIST(APPEND CMAKE_MODULE_PATH
-	"${CMAKE_SOURCE_DIR}/cmake"
+	"${CMAKE_CURRENT_SOURCE_DIR}/cmake"
 	"${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/"
 )
 
+IF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
lynxlynxlynx, line 16: it would be cleaner to test APPLE here, since that's what we use everywhere else.

CMakeLists.txt
czarny247: 1. Note mentioned here: https://cmake.org/cmake/help/latest/command/project.html#usage
2. `CMAKE_SYSTEM_NAME` is empty before `project` call so i couldn't set this properly:

```
IF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
	LIST(APPEND CMAKE_MODULE_PATH
		"${CMAKE_CURRENT_SOURCE_DIR}/platforms/apple/cmake")
ENDIF()
```

cmake/Helpers.cmake
```diff
@@ -515,9 +515,9 @@ FUNCTION(CONFIGURE_APPLE_SPECIFICS)
 	FIND_LIBRARY(COCOA_LIBRARY_PATH Cocoa)
 	FIND_LIBRARY(COREFOUNDATION_LIBRARY CoreFoundation)
 	# for objective-c++
-	STRING(APPEND CMAKE_CXX_FLAGS " -DTARGET_OS_MAC")
+	SET(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} " -DTARGET_OS_MAC")
 	# for pure objective-c
-	STRING(APPEND CMAKE_C_FLAGS " -x objective-c -DTARGET_OS_MAC")
+	SET(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} " -x objective-c -DTARGET_OS_MAC")
```
lynxlynxlynx, line 520: How are these two things different?
Anyway, since I expect this to fail, the cmake way to add definitions is https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions

CMakeLists.txt
lynxlynxlynx: Your own CACHE_BUILD_TYPE says it should happen before the project call, so it needs to be moved if so, now that it's ran later.

I see now that CMAKE_PROJECT_INCLUDE_BEFORE requires 3.15, so please update the minimum version as well.

lynxlynxlynx:
I see you're struggling with the CFLAGS changes. Considering a bunch of the changes in this PR were removed via simplification, I suggest you check if the previous simple string appending still works instead.

czarny247:
> I see you're struggling with the CFLAGS changes. Considering a bunch of the changes in this PR were removed via simplification, I suggest you check if the previous simple string appending still works instead.

I made a search trough the github and realized how it should be done. It works now ;)

lynxlynxlynx:
Hmm. But does the semicolon actually work? I know it compiles, but doesn't the compiler then get a mangled flag?

Anyway, there might be a better way still — perhaps it works without those two lines, since we do reference the files for compilation: https://cmake.org/cmake/help/v3.16/release/3.16.html#languages

czarny247:
> Hmm. But does the semicolon actually work? I know it compiles, but doesn't the compiler then get a mangled flag?
> 
> Anyway, there might be a better way still — perhaps it works without those two lines, since we do reference the files for compilation: https://cmake.org/cmake/help/v3.16/release/3.16.html#languages

I simplified it, take a look at the main CMakeLists.txt

lynxlynxlynx:
Yeah that's much nicer, thanks. Just the minimum version bump is missing.

czarny247:
> Yeah that's much nicer, thanks. Just the minimum version bump is missing.

Done ;)

lynxlynxlynx:
argh, it's 3.16 — remember the obj-c release notes.

lynxlynxlynx:
Thanks!

Issue #1988, opened 2024-02-09
Sound frequency settings get reset 


burner1024:
#### Bug description
"Selection Sounds Frequency" and "Command Sounds Frequency" get reset to "Never", if you just visit the corresponding settings page and hit "Cancel", then check the settings again.
Like #1762, but for these.
![Captura desde 2024-02-09 20-36-27](https://github.com/gemrb/gemrb/assets/1279852/a5c231ca-3f4d-43c5-851d-ff66ffe528a1)

I thought it would be a similar easy fix, but there's some radio button mumbo jumbo that I can't comprehend.

#### Steps to reproduce
1. Go to character sound settings page.
2. Set frequencies to "always".
3. Click OK, return to main menu.
4. Go to the same settings page, return to main menu hitting Escape.
5. Go to the same page for the 3rd time. Settings are set to "never".

#### Expected behavior
Settings preserved.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
Radio buttons are the most annoying GUI element since textarea editing has been implemented.
Pressing escape and clicking on cancel are not necessarily the same, but I can't think of how this would be a problem here.
What is important is whether the actual values in gem-balduri.ini remain sane. IOW is it just a display problem or does data get corrupted?

burner1024:
Values get reset to 1.

burner1024:
Looks like it was supposed to be fixed in #1823 but wasn't.

bsdcode:
This bug affects RADIOBUTTON controls in general. There is also the `BitsPerPixel` option in the graphics option window which suffers from the same problem. There the value can get reset to `32`. It seems the erreonous reset-value is the value from the highest `button_id` inside one `name_id` group:
https://github.com/gemrb/gemrb/blob/d0af7b1c9952e00489cfe0c39056760eb0e8c3dd/gemrb/GUIScripts/GUIOPT.py#L129-L133
https://github.com/gemrb/gemrb/blob/d0af7b1c9952e00489cfe0c39056760eb0e8c3dd/gemrb/GUIScripts/GUIOPT.py#L231-L236

[...]

PR #2035, opened 2024-02-27
Fix radiobutton preselection, ref. #1988

bsdcode:
## Description

Add setting the dictionary value back to its previous value, otherwise the internal state of the radiobutton is set wrongly to the last processed button. Also generalize it for all radiobuttons.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/GUIOPT.py
+++ b/gemrb/GUIScripts/GUIOPT.py
@@ -108,6 +108,20 @@ def TrySavingConfiguration():
 
 ###################################################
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
LGTM, just please get rid of the underscores in the names — we're trying to unify the styles, not diverge them further.

bsdcode:
Ok, I will fix it. The variable names were inspired by the function definition GUIOPTControls.OptRadio.

lynxlynxlynx:
Thanks!

PR #2036, opened 2024-02-27
Audio spatial flags & improvements

MarcelHB:
## Description
See discussion in #2031. Also some minor improvments I found while testing.

## Checklist


gemrb/GUIScripts/GUIRECCommon.py
```diff
@@ -372,7 +372,7 @@ def PlaySoundPressed():
 			break
 	else:
 		NextSound()
-	GemRB.PlaySound (VoiceSet + SoundSeq[SoundIndex], "CHARACT" + str(pc - 1), 0, 0, 5)
+	GemRB.PlaySound (VoiceSet + SoundSeq[SoundIndex], "CHARACT" + str(pc - 1), 0, 0, 4)
```
lynxlynxlynx, line 375: Ideally we should just add the constants to ie_sounds.py or somesuch and use the named ones.

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -514,19 +514,20 @@ void DisplayStringCore(Scriptable* const Sender, ieStrRef Strref, int flags, con
...
-		// disable position, but only for party
+
 		Actor* actor = Scriptable::As<Actor>(Sender);
-		if (!actor || actor->InParty ||
-			core->InCutSceneMode() || core->GetGameControl()->InDialog()) {
-			speech |= GEM_SND_RELATIVE;
-			pos.reset();
+		// Spatial unless PC, cutscene or dialog
+		if (actor && !actor->InParty && !core->InCutSceneMode() && !core->GetGameControl()->InDialog()) {
+			Point pos = Sender->Pos;
```
lynxlynxlynx, line 526: Should be without Point, so it affects the outer pos.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -3657,7 +3657,7 @@ void Actor::PlaySelectionSound(bool force)
 	if (!found) {
 		ResRef sound;
 		GetSoundFromFile(sound, Verbal::Select);
-		core->GetAudioDrv()->Play(sound, SFX_CHAN_MONSTER, Pos);
+		core->GetAudioDrv()->Play(sound, SFX_CHAN_MONSTER, Pos, GEM_SND_SPATIAL);
```
lynxlynxlynx, line 3660: I'm suspicious whether most of these Actor ones should be spatial. Armor clangs are fine as that, but if your pc is getting mauled off-screen, the player should know about it. For selection is likely doesn't matter, for regular combat sounds no idea, should be tested.

gemrb/core/Scriptable/Door.cpp
```diff
@@ -116,12 +116,12 @@ void Door::ToggleTiles(int State, int playsound)
...
 		state = !closedIndex;
 		if (playsound && !OpenSound.IsEmpty()) {
-			core->GetAudioDrv()->PlayRelative(OpenSound, SFX_CHAN_ACTIONS);
+			core->GetAudioDrv()->Play(OpenSound, SFX_CHAN_ACTIONS, toOpen[0], GEM_SND_SPATIAL);
 		}
 	} else {
 		state = closedIndex;
 		if (playsound && !CloseSound.IsEmpty()) {
-			core->GetAudioDrv()->PlayRelative(CloseSound, SFX_CHAN_ACTIONS);
+			core->GetAudioDrv()->Play(CloseSound, SFX_CHAN_ACTIONS, toOpen[0], GEM_SND_SPATIAL);
```
lynxlynxlynx, line 124: shouldn't it be 1 instead of 0 here and below?

MarcelHB: Not exactly sure, looked at the docs and it says that it's one of two door points (both sides?), and then I was thinking if there are examples of points with huge distances, or if it is neglectable and just pick any of the two.

gemrb/core/Scriptable/Actor.cpp
MarcelHB: Good point, trying out and maybe make it conditional.

gemrb/core/GameScript/GSUtils.cpp
MarcelHB: Thanks

gemrb/core/Scriptable/Door.cpp
lynxlynxlynx: It's for when it's open and when it's closed, but you're right, it is usually close enough that it probably makes no difference.

gemrb/GUIScripts/TextScreen.py
```diff
@@ -194,8 +194,8 @@ def EndTextScreen ():
 def ReplayTextScreen ():
 	global TextArea, TableName, Row
 
-	# stop any current speech, flag of 5 = GEM_SND_RELATIVE|GEM_SND_SPEECH
-	GemRB.PlaySound(None, CHAN_GUI, 0, 0, 5)
+	# stop any current speech, flag of 4 = GEM_SND_SPEECH
+	GemRB.PlaySound(None, CHAN_GUI, 0, 0, 4)
```
lynxlynxlynx, line 198: you can replace it here as well and remove that part of the comment.

MarcelHB: Urgh, thanks

lynxlynxlynx:
Thanks!

PR #2039, opened 2024-02-29
BG1: fix main menu buttons, #2005

bsdcode:
## Description

Add IE_GUI_BUTTON_NO_TEXT to the MoviesButton in the single player menu when playing without TotSc expansion, otherwise the menu would show a disabled and borderless button with text "Movies".

Apply it also to the ExitButton in the multi player menu in order to avoid using SetText("").

Also reset the Movies- and ExitButton flags to IE_GUI_BUTTON_NORMAL in BackToMain, otherwise these buttons loose their soundeffect when clicking on them. The ExitButton actually already lost its soundeffect before when going into the multi player menu.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg1/Start.py
+++ b/gemrb/GUIScripts/bg1/Start.py
@@ -62,7 +62,7 @@ def SinglePlayerPress():
 		MoviesButton.OnPress (MissionPack)
 		MoviesButton.SetText(24110)
 	else:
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Does that match the original? Or should an actual movies list be accessible this way?

bsdcode:
Without TotSc there should be no Movie-button at all:

menu

lynxlynxlynx:
Ok, thanks!

Issue #2008, opened 2024-02-22
 BG1: Music keeps on playing


bsdcode:
#### Bug description
Whenever a new background music is playing it keeps on playing even when the "situation" is over.

#### Steps to reproduce
1. Start gemrb, start a new game, hear the main theme music at the character creation (still ok)
2. New game starts before the Candlekeep Inn, main theme music still plays
3. Go into Candlekeep Inn, hear the tavern music
4. Go outside Candlekeep Inn, tavern music still plays
5. Go to house in the north where Skunk is attacking you, hear battle music
6. Battle is over, go outside, battle music still plays

#### Expected behavior
Music stops playing when appropriate.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Basically a duplicate of #1854.

PR #2042, opened 2024-03-01
Fix textscreen not playing chapter music

bsdcode:
## Description

Use HardEndPL in order to stop the currently playing music when entering the chapter introductions. This allows the actual chapter music to play.

This doesn't fix the underlying issues with music not playing or stopping properly (e.g. #2008), it merely covers up the symptoms in this special case. But I think that using HardEndPL in this case is actual the right thing to do regardless, especially when it's called already if `MusicName == "*"` (e.g. demo) and the experience with the original games is the same.

## Checklist


lynxlynxlynx:
Yeah, thanks!

Issue #2037, opened 2024-02-28
Projectiles/Spell effects burning down too fast on higher FPS 


MarcelHB:
#### Bug description
I don't know whether it's recent regression or an obvious oversight for some time already, but I think there's something wrong with at least some area spell projectiles and effects on FPS > 30: They manifest shorter by some factor while being accelerated to the inverse apperently, and by listening to the annoying sounds they make.

#### Steps to reproduce
1. `CapFPS=0`
2. Cast something like Web or Entangle
3. Watch the time until it's over

#### Expected behavior
Manifestation duration should adhere to the description regardless of the FPS.

#### Screenshots

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
That's odd, didn't you successfully decouple the drawing from the game speed? What's your "Maximum Frame Rate" in your gem-*.ini?

MarcelHB:
I looked at effects but probably not suitable ones for this. `Projectile.cpp` (for the whole Web thing) has some code parts that looked suspicious at first glance.

`Maximum Frame Rate` is 30, and capping the FPS to 30 prolonges the whole thing. Although that setting still looked to short for the description (30s vs. 60s), but I will test that again as well.

MarcelHB:
Yep, the draw code of projectiles also manages the state transitions.

MarcelHB:
Trying to rewrite this a little, the otherwise simple state machine make its transitions a little hard to figure out from the outside.

MarcelHB:
Yes, like it's supposed to be. It's just that after the first rewrite I still had little idea what I refactored and why it worked that way to the degree of _almost_.

MarcelHB:
I fixed the projectiles, but I'd also fix sparkles/particles. But that currently suffers from things that look otherwise broken to me (see skull trap in IWD2), so let's check them in a different issue.

PR #2045, opened 2024-03-02
Projectiles: decoupling state from drawing

MarcelHB:
## Description
See #2037. It's rewritten to the extent that _most_ of the state transition are now left in `::Update()`, with some expiry requests left.

Feel free to test, I checked arrows and a bunch of spells.

## Checklist

### DIFF
--- a/gemrb/core/Projectile.cpp
+++ b/gemrb/core/Projectile.cpp
@@ -204,7 +204,7 @@ void Projectile::CreateIteration()
 	}
 
 	// added by fuzzie, to make magic missiles instant, maybe wrong place
-	pro->Setup();
+	pro->Update();
 }
 
 void Projectile::GetSmokeAnim()
@@ -340,7 +340,6 @@ void Projectile::Setup()
 		light = CreateLight(Size(LightX, LightY), LightZ);
 	}
 
-	phase = P_TRAVEL;
 	unsigned int flags = GEM_SND_SPATIAL;
 	if (SFlags & PSF_LOOPING) {
 		flags |= GEM_SND_LOOPING;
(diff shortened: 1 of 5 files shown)


gemrb/core/Projectile.cpp
```diff
@@ -1196,48 +1200,65 @@ void Projectile::SecondaryTarget()
...
+		case ProjectileState::TRAVELLING:
+		case ProjectileState::HIT:
+			state = DoStep();
+			break;
+		case ProjectileState::AWAITING_HIT:
+			state = CheckTrigger(Extension->TriggerRadius);
+			break;
+		case ProjectileState::EXPLODING:
+		case ProjectileState::EXPLODING_AGAIN:
+			state = CheckTrigger(Extension->TriggerRadius);
```
lynxlynxlynx, line 1239: This assignment is redundant the way it is now. But before what is now `GetNextExplosionState` didn't get called if the trigger got revoked, so I suspect we're actually missing more conditional logic here.

Before:
```c++
            CheckTrigger(Extension->TriggerRadius);
            if (phase == P_EXPLODING1 || phase == P_EXPLODING2) {
                DrawExplosion(viewport, flags);
            }
```

gemrb/core/Projectile.h
```diff
@@ -198,6 +188,18 @@ struct ProjectileExtension
 
 class GEM_EXPORT Projectile
 {
+private:
+	enum class ProjectileState {
+		NEW,
+		TRAVELLING,       // moving towards target
+		AWAITING_HIT,     // arrived
+		HIT,              // hit
```
lynxlynxlynx, line 196: These name and comment combinations are confusing. I propose we use AWAITING_TRIGGER instead, since it's only relevant for projectiles with triggers (this is the same nomenclature other tools use). "hit" and "arrived" are true for both values, so I propose to change the trigger comment to "trap set".

lynxlynxlynx: Sorry, maybe I wasn't clear enough, but it's AWAITING_HIT that is for the skulltrap behaviour (was P_TRIGGER before), not the other way around.

MarcelHB: Please write an edit suggestion for these lines, so I'm not left confused.

lynxlynxlynx: AWAITING_HIT,     // arrived
		HIT,              // hit

to

		AWAITING_TRIGGER,     // hovering over target, waiting to trigger
		HIT,              // just hit target

gemrb/core/Projectile.cpp
MarcelHB: Right, I've added a check whether `GetNextExplosionState ` is still applicable after `CheckTrigger`. But the assignment being redundant?

lynxlynxlynx: Looks fine now and subsequently there's no more redundant assignment (before the first one was unused).

lynxlynxlynx:
Thanks, didn't see anything bad in testing here either.

Issue #2004, opened 2024-02-22
FreeBSD: Segfault on program exit


bsdcode:
#### Bug description
Exiting GemRB generates a segmentation fault on FreeBSD. Tested with the demo and Baldur's Gate I 5-disc-set version. It doesn't happen with the 0.9.2 release version.
First it would be interesting if this is a isolated FreeBSD problem or if it also happens on other systems.

#### Steps to reproduce
1. Open a terminal
2. Start gemrb with either the demo or Baldur's Gate 1
3. Exit from the main menu
4. Terminal should show that a segmentation fault occured

#### Expected behavior
No segfault on exit

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
On the demo a freetype related segfault can happen, but that can't be the case for bg1. Please gdb/lldb it, so we can see where it's crashing for you.

bsdcode:
I realized the segfault is in SDLAudio, but I use OpenAL. I then recompiled GemRB without SDL2_mixer support and now I get a different segfault:
```
Process 63569 stopped
* thread #1, name = 'gemrb', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
    frame #0: 0x0000000821bfdae5 libthr.so.3`_pthread_mutex_lock + 37
```
Now it becomes strange because I then recompiled with SDL2_mixer support again and don't get the previous segfault in SDLAudio.h, but keep segfaulting in `_pthread_mutex_lock`.

Hmmm, race condition anywhere? Let's see if it's different with GCC again...

lynxlynxlynx:
Sounds like it could be something specific to BSD. Maybe try running it through valgrind or with -DSANITIZE=address first.

bsdcode:
I installed the separate debug symbols for FreeBSD binaries in order to get more info for the calls into e.g. libthr.so.3. I tried recompiling with GCC and tried building with `make -j1`, but it doesn't change anything. I also got the segfault in SDLAudio.so again, although I do not use SDL for the audio. In addition a new segfault can pop up in place of the previous two:
```
Process 25716 stopped
[...]

PR #2048, opened 2024-03-04
Fix segfault on shutdown

bsdcode:
## Description

Ensure destruction of OpenGL and SDL resources before quitting SDL and use
SDL_QuitSubSystem instead of SDL_Quit in case unloading a shared library
still depends on SDL. The subsystems audio, joystick and gamecontroller
are quitted explicitely now.

Ref. #2004

## Checklist

### DIFF
--- a/gemrb/plugins/SDLVideo/SDLVideo.cpp
+++ b/gemrb/plugins/SDLVideo/SDLVideo.cpp
@@ -28,12 +28,12 @@ using namespace GemRB;
 
 SDLVideoDriver::~SDLVideoDriver(void)
 {
-	SDL_Quit();
+	SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER | SDL_INIT_JOYSTICK | SDL_INIT_VIDEO);
 }
 
 int SDLVideoDriver::Init(void)
 {
-	if (SDL_InitSubSystem( SDL_INIT_VIDEO ) == -1) {
+	if (SDL_InitSubSystem(SDL_INIT_VIDEO) == -1) {
 		Log(ERROR, "SDLVideo", "InitSubSystem failed: {}", SDL_GetError());
 		return GEM_ERROR;
 	}
--- a/platforms/windows/GemRB.cpp
+++ b/platforms/windows/GemRB.cpp
@@ -66,6 +66,7 @@ int main(int argc, char* argv[])
 		ret = GEM_ERROR;
 	}
 
+	VideoDriver.reset();
 	PluginMgr::Get()->RunCleanup();
 
 	ToggleLogging(false); // Windows build will hang if we leave the logging thread running
--- a/gemrb/GemRB.cpp
+++ b/gemrb/GemRB.cpp
@@ -70,6 +70,7 @@ int main(int argc, char* argv[])
 		return GEM_ERROR;
 	}
 
+	VideoDriver.reset();
 	PluginMgr::Get()->RunCleanup();
 
 	return GEM_OK;
--- a/gemrb/plugins/SDLAudio/SDLAudio.cpp
+++ b/gemrb/plugins/SDLAudio/SDLAudio.cpp
@@ -81,12 +81,14 @@ SDLAudio::~SDLAudio(void)
 	Mix_HookMusic(NULL, NULL);
 	FreeBuffers();
 	Mix_ChannelFinished(NULL);
+
+	SDL_QuitSubSystem(SDL_INIT_AUDIO);
 }
 
 bool SDLAudio::Init(void)
 {
-	// TODO: we assume SDLVideo already got loaded
 	if (SDL_InitSubSystem(SDL_INIT_AUDIO) < 0) {
+		Log(ERROR, "SDLAudio", "InitSubSystem failed: {}", SDL_GetError());
 		return false;
 	}
 #ifdef RPI
--- a/gemrb/plugins/SDLVideo/SDL12Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL12Video.cpp
@@ -55,13 +55,6 @@ int SDL12VideoDriver::Init(void)
 		// this may limit people actually using very old single button mice, but who cares :)
 		setenv("SDL_HAS3BUTTONMOUSE", "SDL_HAS3BUTTONMOUSE", 1);
 #endif
-		if (SDL_InitSubSystem(SDL_INIT_JOYSTICK) == -1) {
-			Log(ERROR, "SDLJoystick", "InitSubSystem failed: {}", SDL_GetError());
-		} else {
-			if (SDL_NumJoysticks() > 0) {
-				gameController = SDL_JoystickOpen(0);
-			}
-		}
 	}
 
 	if (SDL_InitSubSystem(SDL_INIT_JOYSTICK) == -1) {
--- a/gemrb/plugins/SDLVideo/SDL20Video.cpp
+++ b/gemrb/plugins/SDLVideo/SDL20Video.cpp
@@ -82,6 +82,10 @@ SDL20VideoDriver::SDL20VideoDriver() noexcept
 
 SDL20VideoDriver::~SDL20VideoDriver() noexcept
 {
+#if USE_OPENGL_BACKEND
+	delete blitRGBAShader;
+#endif
+
 	if (SDL_GameControllerGetAttached(gameController)) {
 		SDL_GameControllerClose(gameController);
 	}
@@ -93,10 +97,6 @@ SDL20VideoDriver::~SDL20VideoDriver() noexcept
 
 	SDL_DestroyRenderer(renderer);
 	SDL_DestroyWindow(window);
-
-#if USE_OPENGL_BACKEND
-	delete blitRGBAShader;
-#endif
 }
 
 int SDL20VideoDriver::Init()
--- a/platforms/android/GemRB.cpp
+++ b/platforms/android/GemRB.cpp
@@ -81,5 +81,7 @@ int main(int argc, char* argv[])
 		return GEM_ERROR;
 	}
 
+	VideoDriver.reset();
+
 	return GEM_OK;
 }
--- a/platforms/vita/GemRB.cpp
+++ b/platforms/vita/GemRB.cpp
@@ -138,6 +138,8 @@ int main(int argc, char* argv[])
 		SDL_SetVideoModeScaling(vitaDestRect.x, vitaDestRect.y, vitaDestRect.w, vitaDestRect.h);
 	}
 #endif
+	VideoDriver.reset();
+
 	ToggleLogging(false);
 
 	return sceKernelExitProcess(0);


gemrb/includes/SDLInitSubSystemWrapper.h
```diff
@@ -0,0 +1,69 @@
...
+#include <SDL.h>
+
+#include "Logging/Logging.h"
+
+namespace GemRB {
+
+static bool SDLInitSubSystemWrapper(Uint32 flags) {
+	const char* owner = nullptr;
+
+	// FIXME: doesn't handle multiple subsystems in flags
```
lynxlynxlynx, line 34: Why is this a fixme? We never intialize more than one thing at a time, so this can go.

bsdcode: Ok.

gemrb/includes/SDLInitSubSystemWrapper.h
```diff
@@ -0,0 +1,69 @@
...
+	case SDL_INIT_JOYSTICK:
+		owner = "SDLJoystick";
+		break;
+	case SDL_INIT_VIDEO:
+		owner = "SDLVideo";
+		break;
+	default:
+		owner = "UNIMPLEMENTED";
+		break;
+	}
```
lynxlynxlynx, line 50: This will fail so rarely, I don't see much value in pretty printing it (there are more readable errors down the line anyway). You can just use `"SDL Init: {}", flags` instead of the owner.

bsdcode: Ok.

lynxlynxlynx:
Does the minimal test run for you locally? You can run it from the build dir this way: `gemrb/gemrb -c ../gemrb/GemRB.cfg.noinstall.sample`

bsdcode:
Yes it runs. No windows shown as expected and last line of the log is
`[GUIScript]: Loading Script Start.`

lynxlynxlynx:
Can you rebase this branch on master and force push it?

lynxlynxlynx:
Still doesn't work on windows. 🤷🏾‍♂️

MarcelHB:
Can reproduce this:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff808220984 in ?? () from C:\Windows\System32\Windows.Gaming.Input.dll
(gdb) bt
#0  0x00007ff808220984 in ?? () from C:\Windows\System32\Windows.Gaming.Input.dll
#1  0x00007fffa2ae554a in WGI_JoystickQuit () at D:/Sources/SDL/src/joystick/windows/SDL_windows_gaming_input.c:985
#2  0x00007fffa28a95fa in SDL_JoystickQuit () at D:/Sources/SDL/src/joystick/SDL_joystick.c:1333
#3  0x00007fffa2861938 in SDL_QuitSubSystem_REAL (flags=62001) at D:/Sources/SDL/src/SDL.c:387
#4  0x00007fffa2861b80 in SDL_Quit_REAL () at D:/Sources/SDL/src/SDL.c:483
#5  0x00007fffa2884934 in SDL_Quit () at D:/Sources/SDL/src/dynapi/SDL_dynapi_procs.h:92
#6  0x00007fffa570295c in _execute_onexit_table (table=0x7fffa5795000 <atexit_table>) from D:\Sources\gemrb\build\gemrb\plugins\SDLVideo.dll
#7  0x00007fffa568114c in _CRT_INIT (hDllHandle=hDllHandle@entry=0x7fffa5680000, dwReason=dwReason@entry=0, lpreserved=lpreserved@entry=0x1) at C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtdll.c:130
#8  0x00007fffa5681229 in __DllMainCRTStartup (hDllHandle=0x7fffa5680000, dwReason=0, lpreserved=0x1) at C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtdll.c:196
#9  0x00007ff812da9a1d in ntdll!RtlActivateActivationContextUnsafeFast () from C:\Windows\SYSTEM32\ntdll.dll
#10 0x00007ff812dedcda in ntdll!LdrShutdownProcess () from C:\Windows\SYSTEM32\ntdll.dll
#11 0x00007ff812deda8d in ntdll!RtlExitUserProcess () from C:\Windows\SYSTEM32\ntdll.dll
#12 0x00007ff8116de3bb in KERNEL32!FatalExit () from C:\Windows\System32\kernel32.dll
#13 0x00007ff8127fa155 in msvcrt!_exit () from C:\Windows\System32\msvcrt.dll
#14 0x00007ff8127fa7c5 in msvcrt!_initterm_e () from C:\Windows\System32\msvcrt.dll
#15 0x00007ff75e6513cf in __tmainCRTStartup () at C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:269
#16 0x00007ff75e651406 in mainCRTStartup () at C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:188
```

Sigh

MarcelHB:
My best explanation here is: SDL may have loaded joystick drivers, and on Win one of them also has a dynamically loaded library -- that may have gone already, since there is no guarantee to which DLL is unloaded first. So I assume the following is happening.

1. `videoDriver.reset()` -> cleanup of some of our resources
1. post exe cleanups, handles, libraries, ...
1. `FreeLibrary(Windows.Gaming.Input.dll)`
1. `FreeLibrary` ...
1. `FreeLibrary(SDLVideo.dll)` -> `atexit` -> `SQL_Quit` ->`WGI_JoystickQuit` still expecting stuff from `Windows.Gaming.Input.dll` to be there?

bradallred:
BTW, changes were made to `GemRB.cpp`, but Windows doesnt use this file. It has its own version in the platforms directory. Probably just need to make the same change there.

MarcelHB:
Hate to say it, but maybe we really need to have a hard linking to SDL in the main, at least for quitting? If there are safe alternatives, then I'm not sure if they are looking sane w. r. t. platform portability and linking flavor.

bradallred:
> Hate to say it, but maybe we really need to have a hard linking to SDL in the main, at least for quitting? If there are safe alternatives, then I'm not sure if they are looking sane w. r. t. platform portability and linking flavor.

I don't think so. Why wouldn't actually unloading our plugins when tearing down `Interface` solve this?

platforms/windows/GemRB.cpp
```diff
@@ -60,6 +60,7 @@ int main(int argc, char* argv[])
 
 		Interface gemrb(std::move(cfg));
 		gemrb.Main();
+		VideoDriver.reset();
```
lynxlynxlynx, line 63: please add it to the other two as well

MarcelHB:
> I don't think so. Why wouldn't actually unloading our plugins when tearing down `Interface` solve this?

Because it doesn't. It _does_ do the dtor, but we do not unload any dynamic libraries from the OS by that. And if we did that, we have to sort if this works fine with SDLVideo and SDLAudio order of operations on the shutdown.

platforms/windows/GemRB.cpp
bsdcode: ok, done.

bradallred:
> > I don't think so. Why wouldn't actually unloading our plugins when tearing down `Interface` solve this?
> 
> Because it doesn't. It _does_ do the dtor, but we do not unload any dynamic libraries from the OS by that. And if we did that, we have to sort if this works fine with SDLVideo and SDLAudio order of operations on the shutdown, or maybe also hardcode that order on exit.

Huh? Thats what I'm saying tho, we need to actually run `FREE_PLUGIN` (which calls `dlclose`) when the `PluginHolder` refcount drops to 0 and we need to make that happen by storing the plugins in a non static data structure that gets destroyed as part of `~Interface`.

You don't need to worry about ordering between SDLVideo and SDLAudio because each only shuts down their respective subsystems. SDL_Quit needs to be registered by them as an `std::atexit` handler which will then get called twice, but that should be harmless.

There are more robust options too (also discussed above), but we shouldn't need that complexity.

If we really did have dependency order issues between plugins, then the only sane solution is to create a method of enforcing load order and then ensuring destruction is the reverse of that. We do have `PluginFlagsType::DELAY` which may have been an attempt to resolve such issues in the past. I can't think of any plugins that obviously depend on another tho.

MarcelHB:
Yes, I didn't read all this out of your single line response above. But don't know what's best to do with this PR then.

bradallred:
@bsdcode  can you rebase this on master again? I've made some changes to fix crash on shutdown for our TTF plugin and the cause is very much related to this.

hopefully this is all good enough for now. At some point it would be nice to refactor `PluginMgr` a bit to make this automatic and safe.

bradallred:
@lynxlynxlynx the windows build didnt run as part of that last push. Not sure how to force it to run.

lynxlynxlynx:
It's because the last commit only touches things in unrelated platforms/. A rebase into a single commit would do.

gemrb/plugins/SDLVideo/SDLVideo.cpp
```diff
@@ -22,21 +22,21 @@
 
 #include "Interface.h"
 #include "Video/RLE.h"
+#include "SDLInitSubSystemWrapper.h"
 #include "SDLPixelIterator.h"
 
 using namespace GemRB;
 
 SDLVideoDriver::~SDLVideoDriver(void)
```
bradallred, line 30: you should remove the destructor altogether.

bsdcode: thanks, done.

bradallred:
looks like our `atexit` isnt running before libraries are unloaded so stuffing `SDL_Quit` there wont work. Maybe revert that part of this and the `VideoDriver` reset and cleaning up plugins at the right time will be enough to fix the immediate issue. 

The rest needs some kind of refactoring for controlling plugin unloading. But that still leaves a question of how to correctly handle `SDL_Quit` since `atexit` wont work. Linking magic to merge symbols (eg a counter) from two different plugins seems like it would be a pain and may not even be possible universally. I don't know that there is a way to get the refcount for a shared library to check the count on sdl2.so before calling `SDL_Quit` either. and again something that may not be universally achievable. Maybe the answer is that we don't need to call `SDL_Quit`?

MarcelHB:
Dropping `SDL_Quit` leaves a dirty Valgrind report. I tried `SDL_QuitSubSystem(SDL_INIT_VIDEO)` instead and it cleaned that up again. Yet SDL2 says:

> You still need to call [SDL_Quit](https://wiki.libsdl.org/SDL2/SDL_Quit)() even if you close all open subsystems with [SDL_QuitSubSystem](https://wiki.libsdl.org/SDL2/SDL_QuitSubSystem)().

lynxlynxlynx:
How about shutting down the joystick stuff before video?

bradallred:
> Dropping `SDL_Quit` leaves a dirty Valgrind report. I tried `SDL_QuitSubSystem(SDL_INIT_VIDEO)` instead and it cleaned that up again. Yet SDL2 says:
> 
> > You still need to call [SDL_Quit](https://wiki.libsdl.org/SDL2/SDL_Quit)() even if you close all open subsystems with [SDL_QuitSubSystem](https://wiki.libsdl.org/SDL2/SDL_QuitSubSystem)().

Right, I'm not suggesting a permanent fix ATM. Something for now to fix the segfaults on exit. We can open a new issue for whatever plugin lifecycle refactoring we need to do.

bsdcode:
Ok, so what's the way forward for this PR? At the moment we have these changes:

1. `VideoDriver.reset()` after `gemrb.Main()`
2. Wrap `SDL_InitSubSystem` and `atexit(SDL_Quit)`, remove `SDL_Quit` from dtor
3. `delete blitRGBAShader` before destroying renderer

From @bradallred:
> Maybe revert that part of this and the VideoDriver reset and cleaning up plugins at the right time will be enough to fix the immediate issue.

From @MarcelHB:
> Dropping SDL_Quit leaves a dirty Valgrind report. I tried SDL_QuitSubSystem(SDL_INIT_VIDEO) instead and it cleaned that up again.

So undo change `2` and replace `SDL_Quit` with `SDL_QuitSubSystem` in `~SDLVideoDriver`? Also put an `SDL_QuitSubSystem` for SDLAudio and for joystick and controller somewhere?

bradallred:
> So undo change 2 and replace SDL_Quit with SDL_QuitSubSystem in ~SDLVideoDriver? Also put an SDL_QuitSubSystem for SDLAudio and for joystick and controller somewhere?

Thats what I would try. Joystick/Controller are part of SDLVideo. Sorry, I have neither Windows nor BSD to experiment with.

> VideoDriver.reset() after gemrb.Main()

I think technically this should go right above `PluginMgr::Get()->RunCleanup()` to also execute in the case we fail to startup.

bsdcode:
Well, at least the Windows tests succeeded now.

lynxlynxlynx:
There's still the matter of calling SDL_Quit, though looking at what it does, it doesn't look like we'd be missing out:
https://github.com/libsdl-org/SDL/blob/SDL2/src/SDL.c#L497

lynxlynxlynx:
So what's the consensus on this, should we merge and do without SDL_Quit for now?

bradallred:
Sounds like it to me. I'll open an issue for the underlying plugin cleanup issue.

bsdcode:
As far as I understand it yes. SDL_Quit is problematic on Windows and going with SDL_QuitSubSystem instead seems good enough until the underlying issue is fixed.

lynxlynxlynx:
Alright, thanks!

Issue #848, opened 2020-06-27
Proper implementation of modal windows


bradallred:
GemRB has long treated modal windows as something special that floats above all other windows. This doesn't appear to jive with how the games are actually implemented and its even more noticeable now with the Enhanced Editions.

What we need is actually a window flag so that _multiple_ windows can be modal and they should only block events to _only_ windows below them as opposed to the current implementation of a single modal. This would greatly simplify the modal window code and free us from some of the hoops we have to jump though. For example if we have a modal that opens a new modal then we need to setup a callback on window close to restore the previous modal state (or unnecessarily close the first window when opening the second)

I believe Lynx mentioned there is actually a flag in the CHU that we currently ignore that identifies which windows are "modal".

here is an example of a side effect involving cursors/tooltips
![store](https://user-images.githubusercontent.com/238558/85931644-2d398080-b8c6-11ea-96a4-3bf31103281d.gif)

discussion of this can be found in #840

lynxlynxlynx:
Yeah, should be this from Taimon:
> Some info for the CHU file format: Window+0x1a (flags): Bit0 means don't shade the underlying window. (Easy to test with the confirmation dialog in GUILOAD, when you try to delete a game.) Bit1 skips the allocation/loading of the window ingame.

It's a word we currently don't read, right before the control table. Don't think it's very useful though, since eg. none of the windows in GUIINV have it set.

bradallred:
GUINV is a "top window" (we should probably rename that when we can have stacked modals) and we can set it ourselves. the main thing we need is to have the ability to flag multiple windows, not change the window ordering, and block events only to windows beneath a modal.

Also, I wonder if that word is just for the shadow type for the modal and maybe one of the surrounding words we don't read (if there is one) is actually the modal flag. Not that we need it.

lynxlynxlynx:
Even guirec has it only set for one window and it's not the error popups or any of the small customization ones. guiw8/guiload/guisave none. Pretty useless field, though of course we can store something else there. No other _control_ fields are known.

We could probably treat everything that has a frame after cg as modal to save some of the manual calls. Once we support multiple modals, of course.

bradallred:
yeah, that would be a good heuristic.

bradallred:
we allow stacked modal windows since e3c446. I dont know yet if they behave sufficiently for our needs.

PR #2051, opened 2024-03-05
BG1: add missing shadows for modal windows in character creation

bsdcode:
## Description

Need to call ShowModal with MODAL_SHADOW_GRAY instead of MODAL_SHADOW_NONE.

cc

It is correct for BG1, but this changes `gemrb/GUIScripts/GUICG13.py` and `gemrb/GUIScripts/LUSpellSelection.py` too. If this is troublesome for the other games then I would need to add game checks.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/GUICG13.py
+++ b/gemrb/GUIScripts/GUICG13.py
@@ -40,9 +40,10 @@ SkinColor = 0
 MinorColor = 0
 MajorColor = 0
 PDollButton = 0
+ModalShadow = MODAL_SHADOW_NONE
 
 def OnLoad():
-	global ColorWindow, DoneButton, PDollButton, ColorTable
+	global ColorWindow, DoneButton, PDollButton, ColorTable, ModalShadow
 	global HairButton, SkinButton, MajorButton, MinorButton
 	global HairColor, SkinColor, MinorColor, MajorColor
 	
@@ -98,7 +99,12 @@ def OnLoad():
 	DoneButton.OnPress (NextPress)
 	BackButton.OnPress (BackPress)
 	BGCommon.RefreshPDoll (PDollButton, MinorColor, MajorColor, SkinColor, HairColor)
-	ColorWindow.ShowModal (MODAL_SHADOW_GRAY)
+
+	if GameCheck.IsBG1 ():
+		ModalShadow = MODAL_SHADOW_GRAY
+
+	ColorWindow.ShowModal (ModalShadow)
+
 	return
 
 def DonePress():
@@ -107,7 +113,7 @@ def DonePress():
 	if ColorPicker:
 		ColorPicker.Close ()
 
-	ColorWindow.ShowModal (MODAL_SHADOW_GRAY)
+	ColorWindow.ShowModal (ModalShadow)
 
 	PickedColor = ColorTable.GetValue(ColorIndex, GemRB.GetVar("Selected"))
 	if ColorIndex == 0:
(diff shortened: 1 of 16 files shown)


bradallred:
there is actually a bit in the CHU we should be reading to determine this

bsdcode:
Ah ok, it's bit 0 in the flags word:
https://github.com/gemrb/gemrb/blob/192b88d9942d251fc47aff9ebfa1278729cbf0c7/gemrb/plugins/CHUImporter/CHUImporter.cpp#L419-L425
So when it's not set then we dim the background.

lynxlynxlynx:
No, it's garbage.

lynxlynxlynx:
See discussion here: #848

gemrb/GUIScripts/GUICG13.py
```diff
@@ -159,13 +159,13 @@ def GetColor():
...
 	ColorPicker.GetControl(0).MakeEscape ()
-	ColorPicker.ShowModal (MODAL_SHADOW_NONE)
+	ColorPicker.ShowModal (MODAL_SHADOW_GRAY)
 	return
 
 def HairPress():
 	global ColorIndex, PickedColor
 
-	ColorWindow.SetVisible(False)
+	ColorWindow.SetVisible(True)
```
lynxlynxlynx, line 168: What's the rationale here? These SetVisible changes look bogus.

gemrb/GUIScripts/LUSpellSelection.py
```diff
@@ -246,7 +246,7 @@ def OpenSpellsWindow (actor, table, level, diff, kit=0, gen=0, recommend=True, b
 		if recommend:
 			SpellsWindow.Focus()
 		else:
-			SpellsWindow.ShowModal (MODAL_SHADOW_NONE)
+			SpellsWindow.ShowModal (MODAL_SHADOW_GRAY)
```
lynxlynxlynx, line 249: this is fine, the others will always pick the other branch and iwd1 is still using its own code.

gemrb/GUIScripts/GUICG13.py
bsdcode: Ah sorry, I forgot it in the description. The colorwindow is not supposed to disappear when you select a color:

Original:
wine

GemRB:
gemrb

Thanks for making me checking it again. Now I discovered two more bugs with the color window, but that's for another PR (Major and Minor button press actions are swapped, and ESC-ing out of the color picking window changes the color to some default color).

gemrb/GUIScripts/LUSpellSelection.py
bsdcode: Thanks!

gemrb/GUIScripts/GUICG13.py
lynxlynxlynx: It might be from bg2 then, but we might get away without this there as well, since the two windows look like they overlap perfectly (once the placement bug is fixed). But that would mean we can remove all these calls, since there's nothing setting them to invisible.

bsdcode: Of course! Thanks, I removed them.

bsdcode:
Just to clarify: do we need a BG1 game check for the changes in GUICG13?

lynxlynxlynx:
Yes.

lynxlynxlynx:
Thanks. I've also checked that bg2 is fine without the hiding.

PR #1876, opened 2023-08-04
GUIScript portraits and more

bradallred:
This is the foundational pieces of #1788 along with some changes to address #1870

## Checklist

- [x] Commit messages are descriptive and explain the rationale for changes
- [x] I used the same coding style as the surrounding code 
- [x] I have tested the proposed changes
- [x] I extended the documentation, if necessary
- [x] The proposed change builds also on our build bots (check after submission)

PR #2052, opened 2024-03-06
Cleanup and fix colorwindow and -picker

bsdcode:
## Description

GUICG13:
- demote unneeded global PickedColor to local variable
- sequence MajorColor and MinorColor consistently across the sourcecode in order to prevent confusion for the reader
- fix swapped lookup in PortraitTable for initial MajorColor and MinorColor
- fix swapped assignment of ControlID for MajorButton and MinorButton
- fix swapped assignment of Gradient for initial MajorButton and MinorButton
- fix swapped assignment of ColorIndex for MajorColor and MinorColor in DonePress
- fix swapped assignment of Gradient for MajorColor and MinorColor in DonePress
- remove unneeded update of Selected-variable on matching color in GetColor; SetVarAssoc overrides it anyway afterwards
- remove MakeEscape for the first button in GetColor, otherwise pressing ESC always sets the color to the first button's color

GUIREC:
- demote unneeded global PickedColor to local variable
- remove unneeded update of Selected-variable on matching color in OpenColorPicker; SetVarAssoc overrides it anyway afterwards
- add a few whitespaces in DonePress (python-lint)

InventoryCommon:
- demote unneeded global PickedColor to local variable
- add a few whitespaces in ColorDonePress (python-lint)
- put demo buttontext behind the proper else branch in GetColor
- remove useless if/else branch in GetColor

The changes to `ColorWindow.GetControl` in GUICG13 are correct for BG1. If the other games have different ControlIDs (GUICG.CHU), then I need to add GameChecks.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/InventoryCommon.py
+++ b/gemrb/GUIScripts/InventoryCommon.py
@@ -766,53 +766,49 @@ def ColorDonePress():
 		ColorPicker.Close ()
 
 	ColorTable = GemRB.LoadTable ("clowncol")
-	PickedColor=ColorTable.GetValue (ColorIndex, GemRB.GetVar ("Selected"))
-	if ColorIndex==0:
-		GUICommon.SetColorStat (pc, IE_HAIR_COLOR, PickedColor)
-	elif ColorIndex==1:
-		GUICommon.SetColorStat (pc, IE_SKIN_COLOR, PickedColor)
-	elif ColorIndex==2:
-		GUICommon.SetColorStat (pc, IE_MAJOR_COLOR, PickedColor)
+	pickedColor = ColorTable.GetValue (ColorIndex, GemRB.GetVar ("Selected"))
+	if ColorIndex == 0:
+		GUICommon.SetColorStat (pc, IE_HAIR_COLOR, pickedColor)
+	elif ColorIndex == 1:
+		GUICommon.SetColorStat (pc, IE_SKIN_COLOR, pickedColor)
+	elif ColorIndex == 2:
+		GUICommon.SetColorStat (pc, IE_MAJOR_COLOR, pickedColor)
 	else:
-		GUICommon.SetColorStat (pc, IE_MINOR_COLOR, PickedColor)
+		GUICommon.SetColorStat (pc, IE_MINOR_COLOR, pickedColor)
 	UpdateInventoryWindow ()
 	return
 
 def HairPress():
-	global ColorIndex, PickedColor
+	global ColorIndex
 
 	pc = GemRB.GameGetSelectedPCSingle ()
 	ColorIndex = 0
-	PickedColor = GemRB.GetPlayerStat (pc, IE_HAIR_COLOR, 1) & 0xFF
 	GetColor()
 	return
 
 def SkinPress():
-	global ColorIndex, PickedColor
+	global ColorIndex
 
 	pc = GemRB.GameGetSelectedPCSingle ()
 	ColorIndex = 1
-	PickedColor = GemRB.GetPlayerStat (pc, IE_SKIN_COLOR, 1) & 0xFF
 	GetColor()
 	return
 
 def MajorPress():
 	"""Selects the major color."""
-	global ColorIndex, PickedColor
+	global ColorIndex
 
 	pc = GemRB.GameGetSelectedPCSingle ()
 	ColorIndex = 2
-	PickedColor = GemRB.GetPlayerStat (pc, IE_MAJOR_COLOR, 1) & 0xFF
 	GetColor()
 	return
 
 def MinorPress():
 	"""Selects the minor color."""
-	global ColorIndex, PickedColor
+	global ColorIndex
 
 	pc = GemRB.GameGetSelectedPCSingle ()
 	ColorIndex = 3
-	PickedColor = GemRB.GetPlayerStat (pc, IE_MINOR_COLOR, 1) & 0xFF
 	GetColor()
 	return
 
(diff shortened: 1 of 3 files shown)


bsdcode:
Wait, I just realized that we have another colorpicker in the inventory. And this one has the funny bug to just not display the currently selected color, see the `PickedColor == MyColor` section:
https://github.com/gemrb/gemrb/blob/8c6fd7afa02667a3e7514c263361e876ee3e784d/gemrb/GUIScripts/InventoryCommon.py#L836-L851
I highly doubt that this is desired ;) I will do an additional commit for InventoryCommon.py.

bradallred:
these are shared scripts, did you check to see that you aren't just inverting the issue for another game?

bsdcode:
Unfortunatly I didn't (can't) check for the other games at the moment.

The swapped Minor and Major buttons/colors thing is just pure insanity. If some of the other games actually depends on this behaviour then we need to put GameChecks there. The current state, e.g. writing MinorColor into MajorButton, is insane...

The update of the Selected-variable is always useless, regardless which game. SetVarAssoc always overrides it anyway.

That MakeEscape is useless. I'm confident. What's it going to achieve in the other games?

The if/else branch in InventoryCommon: again, what's it going to achieve in the other games? I have almost no fantasy what that should be :) Even if there's a usecase, it should probably behind a GameCheck, to save the sanity of all future readers of the code.

And for the ControlIDs in GUICG13 I wrote in the description that these could be different for different games and there should be GameChecks. Yes I need someone to verify the other games for this.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -829,24 +825,20 @@ def GetColor():
...
-			GemRB.SetVar ("Selected",i)
-			Button.SetState (IE_GUI_BUTTON_LOCKED)
-			Button.MakeEscape()
-		else:
-			Button.SetBAM ("COLGRAD", 2, 0, MyColor)
-			Button.SetFlags (IE_GUI_BUTTON_PICTURE|IE_GUI_BUTTON_RADIOBUTTON, OP_OR)
-			Button.SetState (IE_GUI_BUTTON_ENABLED)
+		Button.SetBAM ("COLGRAD", 2, 0, MyColor)
+		Button.SetFlags (IE_GUI_BUTTON_PICTURE|IE_GUI_BUTTON_RADIOBUTTON, OP_OR)
+		Button.SetState (IE_GUI_BUTTON_ENABLED)
```
lynxlynxlynx, line 841: Are these changes enough for the selected color button to actually appear selected (green outline)? If not it needs to be in a different state.

lynxlynxlynx:
GUICG13 is only also used by bg2 and I doubt it's different in this regard. GUIREC changes are almost cosmetic. InventoryCommon does need wider testing.

If the code is similar enough, you could try importing InventoryCommon in GUICG13 and reuse it.

gemrb/GUIScripts/InventoryCommon.py
bsdcode: Oh, thanks. We have some inconsistency here. BG1 doesn't use the green outline and doesn't indicate the currently selected color:
bg1_wine

We have a BG2 GameCheck in GUICG13 to set the buttons to radiobuttons. In GUIREC it doesn't set the buttons to radiobuttons. In InventoryCommon we set the buttons to radiobuttons unconditionally.

This brings me to the following conclusion:
- We need BG2 GameChecks in all three places and turn the buttons into radiobuttons for BG2. For BG1 these should not be radiobuttons.
- We need to apply the radiobutton fix we already did for the sound frequency: set the visible state of the currently selected button to IE_GUI_BUTTON_SELECTED and adjust the dictionary with an explicit `GemRB.SetVar("Selected", pickedColorIdx)`. Therefore we need PickedColor as a global variable again.

Does this sound correct to you?

bradallred:
I should probably inform you about this:
https://github.com/gemrb/gemrb/commit/c6ca9a8fdc9d604f9df73880cbe7f7608b42e82e

In case that influences your decisions.

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Yes and no, let's just do the same in bg1 if possible. If the button bams don't have a selected frame, too bad, we'll need to work around that, but if they do, having a visual indicator of what is the current choice is good.

lynxlynxlynx:
Oh right, forgot about that. That's in #1876 for reference @bsdcode.

bradallred:
Not quite, but it is in the GUIScript-refactor branch, and he can always make a PR to that branch that will get cherry-picked in very soon after #1876.

bsdcode:
Alright, I'll have a look at the GUIScript-refactor branch. Thanks for informing about it.

gemrb/GUIScripts/InventoryCommon.py
bsdcode: As far as I can see there is no BAM for a selected frame. I tried to unify the button handling and used radiobuttons for BG1 too, but that doesn't look good:

bg1_gemrb_000

You see that we can have some kind of broken image as an indicator, but I wouldn't do it. There's also some different handling of the button state between BG1 and 2. 2 disables unused buttons and 1 locks unused buttons. Without testing BG2 I don't know how that behaves. All in all I think we don't get around doing things differently for BG1.

Now I also realize that the position of that color picker window is wrong and there are three colors missing in the last row.

lynxlynxlynx: Yeah, offsetting the position will help with knowing what is already selected. Still, bg2 would need to be tested. But the whole thing is tangled in other ongoing work, so meh, not sure what's the best way to continue.

bsdcode: I wouldn't think too much about adding a color selection indicator for BG1, it wasn't there in the originals and people hadn't problems with it ;)

Yeah, let me look into the GUIScript-refactor branch. Maybe putting the bugfixes into this branch is the right way to go forward and lay this PR to rest. Let's see.

bsdcode:
The GUIScript-refactor branch already fixes many things and makes a lot of my changes obsolete. I'll rebase my changes on this branch and would make a PR against it.

So I think this PR can be closed.

lynxlynxlynx:
Very well, thanks.

PR #2058, opened 2024-03-10
Fix minor PaperDoll.py bugs

bsdcode:
## Description

- fix uninitialized color stats when OpenPaperDollWindow is called in GUICG13 (SelectColorForPC loaded stats unconditionally from ColorStatsFromPC which are uninitialized at this stage)
- restore PickedColor after Button.SetVarAssoc overrides it
- use lookup by name for colors in ColorStatsFromPortrait

~~remove trailing whitespaces left in the file~~
~~document minimum requirement python 3.10~~

~~The python 3.10 requirement comes from using `Py_IsNone` in gemrb/plugins/GUIScript/GUIScript.cpp.~~

## Checklist

### DIFF
--- a/gemrb/GUIScripts/PaperDoll.py
+++ b/gemrb/GUIScripts/PaperDoll.py
@@ -36,13 +36,11 @@ def ColorStatsFromPortrait(PortraitName):
 	else:
 		PortraitTable = GemRB.LoadTable("pictures")
 
-	PortraitIndex = PortraitTable.GetRowIndex(PortraitName) or 0
-
 	return {
-		IE_MINOR_COLOR : PortraitTable.GetValue (PortraitIndex, 3),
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
3.10 is quite new and if that's the only reason for the bump, we shouldn't do it. A simple comparison can be used instead.
Please don't throw unrelated changes into the same commit.

bsdcode:
I thought you are aware of it, it was already there when I tried to compile the branch. I only wanted to document it.

>Please don't throw unrelated changes into the same commit.

Yeah sorry, you are right.

bradallred:
re `Py_IsNone`, that isn't present in the reworked brach for #1876 so when the remaining commits are cherry picked it wont be an issue.

bsdcode:
Alight, I cleaned the commit up.

PR #2061, opened 2024-03-12
BG1: fix radiobutton preselections in GUICG with associated cleanups

bsdcode:
## Description

The radiobuttons in the GUICG windows suffer from the known radiobutton preselection issue, i.e. the last created button of a buttongroup is preselected, which is not the behaviour in the original and also leaves the controls in a wrong state, e.g. the Done-button isn't enabled despite a preselected radiobutton.

Steps to fix this issue:
- call SetState(enabled/disabled) after SetVarAssoc in order to reset the selected state
- call SetVar(DictVar) after radiobuttons are created in order to restore the value of DictVar which was overriden by SetVarAssoc

While applying the above steps, the surrounding code contexts need cleanups and simplifications:
- wrap male/female button into a function in GUICG1
- wrap Kit and KitName into functions in GUICG22 in order to improve readability of the for-loop
- group button settings (SetVarAssoc, SetFlags, SetState, ...) together and use the same sequence of settings for more clarity and consistency
- merge multiple for-loops over the same buttons into one for-loop
- simplify index and associated ControlID calculations
- add more comments to improve reasoning

Example screenshot of issue in gender window:
gender

Example screenshot of issue in race window:
race

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg1/GUICG1.py
+++ b/gemrb/GUIScripts/bg1/GUICG1.py
@@ -42,17 +42,21 @@ def OnLoad():
 	TextAreaControl = GenderWindow.GetControl(5)
 	TextAreaControl.SetText(17236)
 
-	MaleButton = GenderWindow.GetControl(2)
-	MaleButton.SetFlags(IE_GUI_BUTTON_RADIOBUTTON,OP_OR)
+	GemRB.SetVar("Gender", 0)
 
-	FemaleButton = GenderWindow.GetControl(3)
-	FemaleButton.SetFlags(IE_GUI_BUTTON_RADIOBUTTON,OP_OR)
-	
-	GemRB.SetVar("Gender",0)
-	MaleButton.SetVarAssoc("Gender",1)
-	FemaleButton.SetVarAssoc("Gender",2)
-	MaleButton.OnPress (ClickedMale)
-	FemaleButton.OnPress (ClickedFemale)
+	def SetGenderButton(controlID, var, onPress):
+		Button = GenderWindow.GetControl(controlID)
+		Button.SetVarAssoc("Gender", var)
+		Button.SetFlags(IE_GUI_BUTTON_RADIOBUTTON, OP_OR)
+		Button.SetState(IE_GUI_BUTTON_ENABLED) # reset from SELECTED after SetVarAssoc
+		Button.OnPress(onPress)
+
+	SetGenderButton(2, 1, ClickedMale)
+	SetGenderButton(3, 2, ClickedFemale)
+
+	# restore after SetVarAssoc
+	GemRB.SetVar("Gender", 0)
+
 	DoneButton.OnPress (NextPress)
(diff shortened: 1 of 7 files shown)


bradallred:
I think we probably prefer to solve the radio selection issue more generically (I'm pretty sure I have this fixed somewhere), but this is maybe fine for now. I think we would prefer work broken up into smaller commits tho.

I think we need to get #1876 and the remaining bits of #1788 merged for you. There is a lot of nonsense eliminated or made easier. I really don't want to discourage you from finding and fixing issues, but if you're hunting issues in GUIScripts it's certainly going to collide. I think it will be less frustrating for you and yield better results if you had those tools at your disposal. I'll try to finish those up this month and at least get a single branch up that you can make PRs against while we test it if you want.

bsdcode:
> I think we probably prefer to solve the radio selection issue more generically (I'm pretty sure I have this fixed somewhere), but this is maybe fine for now.

Of course solving it at the root is preferable. I don't want to do the same fix over and over again :) I already tried to get to the root cause for #1988, AFAIK it's entangled deep down in `BindDictVar` and the associated callstack, but that's a little bit too wild for me right now. If you have it fixed somewhere then that would be really great.

Be aware that one half of the fix in this PR, calling `SetState` after `SetVarAssoc`, you would do anyway to enable or disable the button. It's just the ordering which is important to be after `SetVarAssoc`.
Of course the second part, restoring with `SetVar` afterwards, is unnecessary if the final fix to the root problem is incorporated sometime in the future. At least it's not harmful.

> I think we would prefer work broken up into smaller commits tho.

Yes that's totally fine. I actually thought about breaking this commit up, but I think there isn't actually much one could break up here. The cleanups/simplifications/refactors per file are very atomic in my view, you either do them all or none of them.

> I think we need to get #1876 and the remaining bits of #1788 merged for you. There is a lot of nonsense eliminated or made easier. I really don't want to discourage you from finding and fixing issues, but if you're hunting issues in GUIScripts it's certainly going to collide. I think it will be less frustrating for you and yield better results if you had those tools at your disposal. I'll try to finish those up this month and at least get a single branch up that you can make PRs against while we test it if you want.

This would be really great! I already have more issues and fixes in the backlog, would be great if I had a good branch I can work against.

Don't worry about my motivation. I want to start a BG1 run with GemRB soon, but I have to squash as many bugs as I find before I can start with a good feeling. I am motivated to keep on working on your project ;)

lynxlynxlynx:
For commits you could easily split the refactoring from the fixes. Think about it as preparatory work. A refactor ideally shouldn't change any functionality, so if it turns out we need to revert the fix, the improvements from the refactor don't get wiped with it. Splitting stuff up also makes review easier.

If your commit message lists 7 different changes, that should already raise an internal alarm.

bsdcode:
I think this PR can be closed. I won't work on this version anymore because of @bradallred's work on the GUIScripts which will render my changes obsolete and I would need to recheck and rebase anyway.

Issue #1854, opened 2023-07-05
 [iwd+how] No battle music


gregariousmonk:
#### Bug description
Battle music is broken or disabled, no music plays during battle scenes.  

There are a number of other audio issues with iwd, but I'll do a separate report for each of them individually after I've trolled though more of the code. 

#### Steps to reproduce
1. Start a game of iwd+how
2. Get in a battle
3. Only weapon sounds and ambient sounds are heard, battle music is absent.

#### Expected behavior
Battle music from the music cue files should play during combat.

#### Screenshots
n/a I'm using gnome desktop video capture which has no sound recording :/

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.1
- [ ] 0.9.0
- [ ] 0.8.8

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `USE_OPENGL_BACKEND`
- [ ] SDL2 without `USE_OPENGL_BACKEND`

lynxlynxlynx:
I noticed the same now when testing bg1. It happened before, ChangeSong or wherever we end up has some poorly understood logic.

gregariousmonk:
I see periodic warnings about not finding 'cdplayer.bcs', I assume that those are for playing tracks from the CD(?)

lynxlynxlynx:
Might be worth looking at [CGameArea::PlaySong](https://github.com/search?q=repo%3Aalexbatalov%2Fiwd2-re%20PlaySong&type=code) and where it's used. Especially once Alex has them fleshed out more.

lynxlynxlynx:
For iwd2 looking at the code shows no fallbacks (SetSong is only called by a scripting action, which I verified with a disassembler). Looking at the data, no battle music is set in areas (typically 0 in master,  -1 in subareas). Even Targos is without and just continues playing the day/night theme.

Exceptions:
- ar5005: proper playlist
- ar5303, ar6201: battle6 / b6
- ar6305: battle5 / b5
Those 4 areas are the only ones with any battle music in iwd2.

Issue #1476, opened 2021-09-03
 Icewind Dale: Wrong music


dagdha2002:
#### Bug description

The game are not displaying combat music in most of the areas. Instead, the game is displaying Easthaven in peace theme. Checking the tracker at the start of a combat I get this:

[FindPath/DEBUG]: s = (1680, 540), d = (1608, 426), caller = Red Myconid, dist = 65, size = 2
[ResourceManager]: Searching for 'FS_DIRT1'...
[ResourceManager]: Found 'FS_DIRT1.wav' in 'chitin.key'.
[InitRound]: Name: Red Myconid | Attacks: 1 | Start: 8966772
[Actor/DEBUG]: Performattack for Red Myconid, target is: Dynaheir
[ResourceManager]: Searching for 'FS_STON1'...
[ResourceManager]: Found 'FS_STON1.wav' in 'chitin.key'.
[PathFinderWIP/DEBUG]: Abandoning because I'm close to the goal
[MUSImporter]: Loading G:\data\GemRB\IWD\music\MX1000.mus...
[ResourceManager]: Searching for 'MX1000\MX1000A'...
[ResourceManager]: Found 'MX1000\MX1000A.acm' in 'Music'.
[MUSImporter]: Playing MX1000\MX1000A...
[OpenAL]: Music in INITIAL State. AutoStarting
[KeyMap]: Looking up key:  ( )
[KeyMap]: Looking up key:  ( )
[InitRound]: Name: Khalid | Attacks: 3 | Start: 8966776
[Actor/DEBUG]: Performattack for Khalid, target is: Red Myconid
[ResourceManager]: Found 'itemsnd.2da' in 'GemRB Unhardcoded data'.
[ResourceManager]: Searching for 'sw_xbow4'...
[ResourceManager]: Found 'sw_xbow4.wav' in 'chitin.key'.
[Attack/COMBAT]: (Main) Left: 2 | Next: 8966811 [Critical Miss]

So, it´s calling to display Easthaven in Peace instead of a combat music. This is a general bug, only in a very few areas the combat music displays correctly.
Also, most of the time, instead of the music area, I have these theme looping:

[KeyMap]: Looking up key:  ( )
[KeyMap]: Looking up key:  ( )
[OpenAL]: Playing Next Music
[ResourceManager]: Searching for 'mx9000\MX9000A'...
[ResourceManager]: Found 'mx9000\MX9000A.acm' in 'Music'.
[...]

PR #2065, opened 2024-03-16
BG1: fix wrong battle music in subareas without songlist

bsdcode:
## Description
Fallback to the songlist from masterarea in order to play the correct battle music.

## Checklist

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -2392,14 +2392,23 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
 {
 	size_t pl = SongList[SongType];
 	const ieVariable* poi = &core->GetMusicPlaylist(pl);
-	// for subareas fall back to the main list
-	// needed eg. in bg1 ar2607 (intro candlekeep ambush south)
-	// it's not the correct music, perhaps it needs the one from the master area
-	// it would match for ar2607 and ar2600, but very annoying (see GetMasterArea)
-	// ... but this is also definitely wrong for iwd
+	Game* game = core->GetGame();
+
+	// Some subareas don't have their own songlist. It is currently unclear
+	// how the different games handle this situation and which music GemRB
+	// should play, if any. Further research is needed.
+	// At least for the battle music in BG1, e.g. AR2607 (intro candlekeep
+	// ambush south), there is a strong assumption to play the music from
+	// the masterarea's songlist.
+	// This assumption is definitely wrong for IWD, see #1476! Therefore we
+	// use a preliminary flag test to restrict it to BG1 for now.
 	if (IsStar(*poi) && !MasterArea && SongType == SONG_BATTLE && core->HasFeature(GFFlags::BREAKABLE_WEAPONS)) {
-		poi = &core->GetMusicPlaylist(SongType);
-		pl = SongType;
+		static constexpr int bc1Idx = 19; // fallback to first BG1 battle music
+
+		const Map* lastMasterArea = game->GetMap(game->LastMasterArea, false);
+
+		pl = lastMasterArea ? lastMasterArea->SongList[SongType] : bc1Idx;
+		poi = &core->GetMusicPlaylist(pl);
 	}
 
 	if (IsStar(*poi)) return;
@@ -2413,7 +2422,7 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
 		return;
 	}
 	if (SongType == SONG_BATTLE) {
-		core->GetGame()->CombatCounter = 150;
+		game->CombatCounter = 150;
 	}
 }
 
--- a/gemrb/core/Game.h
+++ b/gemrb/core/Game.h
@@ -360,8 +360,6 @@ public:
 	bool MasterArea(const ResRef &area) const;
 	/** Dynamically adding an area to master areas*/
 	void SetMasterArea(const ResRef &area);
-	/** Guess the master area of the given area*/
-	//Map* GetMasterArea(const char *area);
 	/** place persistent actors in the fresly loaded area*/
 	void PlacePersistents(Map *map, const ResRef &resRef);
 	/** Returns slot of the map, if it was already loaded,


gemrb/core/Map.cpp
```diff
@@ -2392,12 +2392,14 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
...
+
+	// Some subareas don't have their own songlist and GemRB would fail to play
+	// the expected music then. Currently this is only observed for the battle
+	// music in BG1, e.g. AR2607 (intro candlekeep ambush south).
+	// In this case GemRB falls back to the music referenced directly by SongType.
 	// it's not the correct music, perhaps it needs the one from the master area
 	// it would match for ar2607 and ar2600, but very annoying (see GetMasterArea)
-	// ... but this is also definitely wrong for iwd
-	if (IsStar(*poi) && !MasterArea && SongType == SONG_BATTLE && core->HasFeature(GFFlags::BREAKABLE_WEAPONS)) {
+	if (IsStar(*poi) && !MasterArea && SongType == SONG_BATTLE && core->config.GameType == "bg1") {
```
lynxlynxlynx, line 2403: 1. This is not better, we never check by game type. The proper fix is to add another game flag, if one is needed, which is still unclear.
2. The iwd comment is valuable testing info, why did you remove it?

gemrb/core/Map.cpp
```diff
@@ -2392,16 +2392,19 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
...
 
 	// Some subareas don't have their own songlist and GemRB would fail to play
 	// the expected music then. Currently this is only observed for the battle
 	// music in BG1, e.g. AR2607 (intro candlekeep ambush south).
-	// In this case GemRB falls back to the music referenced directly by SongType.
-	// it's not the correct music, perhaps it needs the one from the master area
-	// it would match for ar2607 and ar2600, but very annoying (see GetMasterArea)
+	// In this case GemRB now plays the music from the masterarea's songlist,
+	// or falls back to the music referenced directly by SongType if the masterarea
+	// can't be determined (almost always incorrect, but better than nothing).
```
lynxlynxlynx, line 2402: which areas does it not work for or what does the incorrectness comment refer to?

gemrb/core/Map.cpp
bsdcode: Both of your points result from me reading to much into the existing comment and test for the `BREAKABLE_WEAPON' flag. I thought that it is indeed kind of established that this is only relevant for BG1 and I could condense the comment away and use a more direct gamecheck.

Now I am wiser. I have to revert this commit because it is indeed not completely researched and understood how all the games behave in situations with subareas without songlists.

gemrb/core/Map.cpp
bsdcode: The incorrectness comment refers to the fallback if the masterarea can't be determined (`GetMap` returning `nullptr` for whatever reason). Here we have `SongType == SONG_BATTLE == 3`, so we alway access the same music `GetMusicPlaylist(3)` in this fallback case. For BG1 it is `CNITE.mus`. I bet this is almost always not the correct battle music ;)

But yeah, I should probably restructure the comment a little bit to clarify it.

gemrb/core/Map.cpp
lynxlynxlynx: Using BREAKABLE_WEAPON was just a lazy fix, since perhaps another flag isn't needed. And to limit the change to bg1, due to the difference in iwd. #1854 has some good iwd2 RE linked. I suspect this actually used the same logic in all games.

Interestingly I don't see any master area checks in the code. The song header data also isn't inherited if empty:
https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CGameArea.cpp#L1884

So it might be that it just continued playing whatever was outside (all these -1 checks), however that can't explain why the bg1 area does have battle music.

bsdcode: Thanks for the info and link. That's an interesting project.

Maybe the original Infinity Engine does different things for different games in this context. It could also be that the engine has the battle music hardcoded into the executable, i.e. in an area without songlist it just plays this one battle music if a battle starts.

It is possible that the fallback to the masterarea's songlist is actually not what's happening in the originals.

Hmm, in the end we don't really know. That's yet again a topic for wider and deeper research.

gemrb/core/Map.cpp
lynxlynxlynx: Yes, that GetMusicPlaylist call will always be bad. Our music.2da is most probably fine, so we should choose one of 19-29 instead in this fallback. That's also obviously not correct for other games.

gemrb/core/Map.cpp
lynxlynxlynx: strings doesn't suggest anything as bad as hardcoding battle music refs (which one would they pick anyway), while iwd2's version is straightforward https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CGameArea.cpp#L1333

Looking at why we added the game flag, the iwd problematic area from #1476 would pick up a song all the time. Even now that we conveniently store a close enough master area and could take it from there, that would still be bad, since that one does have a battlesong, but battles in the subarea don't have any. So we haven't learned anything new since then and the bg1 check should remain until further research shines more light on things.

bsdcode: Ok, the commit is reverted. Comment and flag test are back in place.

gemrb/core/Map.cpp
bsdcode: Indeed that's a good idea to just use a proper battlemusic in this case then. Defaults now to 19 for last resort fallback.

bsdcode:
According to the conversations I streamlined the commits into one. The comments and flag checks are back in place and a last-resort fallback to the first BG1 battlemusic with index 19 replaces the usage of SongType for direct reference into the musiclist.

lynxlynxlynx:
Thanks!

PR #2068, opened 2024-03-20
Refine some music handling

bsdcode:
## Description
The initial goal was to solve #2008. On the way to the solution I discovered some circumstances that could be improved:

- unnecessary heap allocation in `ChangeSong`
- `ChangeSong` is spammed if the area has no music set
- `MoveBetweenAreasCore` executes an ineffective `ChangeSong`
- fallback music in `PlayAreaSong` almost hid another not understood behaviour of the engine from me (topic for another issue/PR), so remove the fallback
- further opportunities for improved subarea music handling in BG1
- and finally #2008 itself

## Checklist

### DIFF
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -1566,6 +1566,12 @@ void Game::UpdateScripts()
 		CombatCounter--;
 		doChangeSong = !CombatCounter;
 		doChangeSongHard = false;
+	} else if (!core->GetMusicMgr()->IsPlaying()) {
+		// perhaps a StartMusic action stopped the area music?
+		// (we should probably find a less silly way to handle this,
+		// because nothing can ever stop area music now..)
+		doChangeSong = true;
+		doChangeSongHard = false;
 	} else {
 		// nothing to change
 		doChangeSong = false;
@@ -1587,13 +1593,6 @@ void Game::UpdateScripts()
 		for(unsigned int i=0;i<idx;i++) {
 			DelMap(i, false);
 		}
-	}
-
-	// perhaps a StartMusic action stopped the area music?
-	// (we should probably find a less silly way to handle this,
-	// because nothing can ever stop area music now..)
-	if (!core->GetMusicMgr()->IsPlaying()) {
-		ChangeSong(false,false);
 	}
 
 	//this is used only for the death delay so far
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -1570,8 +1570,18 @@ void Game::UpdateScripts()
 		// perhaps a StartMusic action stopped the area music?
 		// (we should probably find a less silly way to handle this,
 		// because nothing can ever stop area music now..)
-		doChangeSong = true;
+
+		// Call ChangeSong only once per round in order to prevent spam
+		// calls every tick in areas without music.
+		static int ticks = 0;
+		ticks++;
+
+		doChangeSong = ticks >= core->Time.round_size;
 		doChangeSongHard = false;
+
+		if (doChangeSong) {
+			ticks = 0;
+		}
 	} else {
 		// nothing to change
 		doChangeSong = false;
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -2158,10 +2158,9 @@ void Game::ChangeSong(bool always, bool force) const
 		BattleSong++;
 	} else {
 		//will select SONG_DAY or SONG_NIGHT
-		Trigger* parameters = new Trigger;
-		parameters->int0Parameter = 0; // TIMEOFDAY_DAY, while dusk, dawn and night we treat as night
-		Song = int(GameScript::TimeOfDay(nullptr, parameters) != 1);
-		delete parameters;
+		Trigger parameters;
+		parameters.int0Parameter = 0; // TIMEOFDAY_DAY, while dusk, dawn and night we treat as night
+		Song = int(GameScript::TimeOfDay(nullptr, &parameters) != 1);
 		BattleSong = 0;
 	}
 	//area may override the song played (stick in battlemusic)
--- a/gemrb/core/Game.cpp
+++ b/gemrb/core/Game.cpp
@@ -1552,18 +1552,27 @@ void Game::UpdateScripts()
 		Maps[idx]->UpdateScripts();
 	}
 
+	// determine if we should start some music
+	bool doChangeSong;
+	bool doChangeSongHard;
+
 	if (PartyAttack) {
 		//ChangeSong will set the battlesong only if CombatCounter is nonzero
-		CombatCounter=150;
-		ChangeSong(false, true);
+		CombatCounter = 150;
+		doChangeSong = true;
+		doChangeSongHard = true;
+	} else if (CombatCounter) {
+		//Change song if combatcounter goes down to 0
+		CombatCounter--;
+		doChangeSong = !CombatCounter;
+		doChangeSongHard = false;
 	} else {
-		if (CombatCounter) {
-			CombatCounter--;
-			//Change song if combatcounter went down to 0
-			if (!CombatCounter) {
-				ChangeSong(false, false);
-			}
-		}
+		// nothing to change
+		doChangeSong = false;
+		doChangeSongHard = false;
+	}
+	if (doChangeSong) {
+		ChangeSong(false, doChangeSongHard);
 	}
 
 	if (StateOverrideTime)
--- a/gemrb/core/GameScript/GSUtils.cpp
+++ b/gemrb/core/GameScript/GSUtils.cpp
@@ -1349,7 +1349,7 @@ void MoveBetweenAreasCore(Actor* actor, const ResRef &area, const Point &positio
 	Map* map1 = actor->GetCurrentArea();
 	Map* map2;
 	Game* game = core->GetGame();
-	bool newSong = false;
+
 	if (!area.IsEmpty() && (!map1 || area != map1->GetScriptRef())) { //do we need to switch area?
 		//we have to change the pathfinder
 		//to the target area if adjust==true
@@ -1358,7 +1358,6 @@ void MoveBetweenAreasCore(Actor* actor, const ResRef &area, const Point &positio
 			map1->RemoveActor( actor );
 		}
 		map2->AddActor( actor, true );
-		newSong = true;
 
 		// update the worldmap if needed
 		if (actor->InParty) {
@@ -1379,9 +1378,6 @@ void MoveBetweenAreasCore(Actor* actor, const ResRef &area, const Point &positio
 	if (actor->InParty) {
 		GameControl *gc=core->GetGameControl();
 		gc->SetScreenFlags(ScreenFlags::CenterOnActor, BitOp::OR);
-		if (newSong) {
-			game->ChangeSong(false, true);
-		}
 	}
 }
 
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -2392,7 +2392,10 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
 {
 	size_t pl = SongList[SongType];
 	const ieVariable* poi = &core->GetMusicPlaylist(pl);
+
+	bool isBG1 = core->HasFeature(GFFlags::BREAKABLE_WEAPONS); // preliminary BG1 test
 	Game* game = core->GetGame();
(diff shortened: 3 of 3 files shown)


bsdcode:
The linux build failed because a dependency couldn't be fetched. The network could not be reached: https://github.com/gemrb/gemrb/actions/runs/8363088961/job/22895236065?pr=2068#step:3:33
Probably a temporary problem.

gemrb/core/Game.cpp
```diff
@@ -1589,13 +1595,6 @@ void Game::UpdateScripts()
 		}
 	}
 
-	// perhaps a StartMusic action stopped the area music?
-	// (we should probably find a less silly way to handle this,
-	// because nothing can ever stop area music now..)
-	if (!core->GetMusicMgr()->IsPlaying()) {
-		ChangeSong(false,false);
-	}
-
```
lynxlynxlynx, line 1589: I'm not sure this is really equivalent. What if the previous ChangeSong call caused a failure for whatever reason?

gemrb/core/Game.cpp
```diff
@@ -1568,10 +1568,10 @@ void Game::UpdateScripts()
 		doChangeSongHard = false;
 	} else if (!core->GetMusicMgr()->IsPlaying()) {
 		// perhaps a StartMusic action stopped the area music?
-		// (we should probably find a less silly way to handle this,
-		// because nothing can ever stop area music now..)
-		doChangeSong = true;
+		doChangeSong = InterruptedMusic;
 		doChangeSongHard = false;
+
+		InterruptedMusic = false;
```
lynxlynxlynx, line 1574: This whole commit looks suspicious as well. Do you have any test case or did you just go from the ancient comment?

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -1379,9 +1378,6 @@ void MoveBetweenAreasCore(Actor* actor, const ResRef &area, const Point &positio
 	if (actor->InParty) {
 		GameControl *gc=core->GetGameControl();
 		gc->SetScreenFlags(ScreenFlags::CenterOnActor, BitOp::OR);
-		if (newSong) {
-			game->ChangeSong(false, true);
-		}
```
lynxlynxlynx, line 1384: What's the alternative solution then? ChangeSong could easily be made to take an optional Map* parameter.

gemrb/core/Map.cpp
```diff
@@ -2403,11 +2403,9 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
 	// This assumption is definitely wrong for IWD, see #1476! Therefore we
 	// use a preliminary flag test to restrict it to BG1 for now.
 	if (IsStar(*poi) && !MasterArea && SongType == SONG_BATTLE && core->HasFeature(GFFlags::BREAKABLE_WEAPONS)) {
-		static constexpr int bc1Idx = 19; // fallback to first BG1 battle music
-
 		const Map* lastMasterArea = game->GetMap(game->LastMasterArea, false);
 
-		pl = lastMasterArea ? lastMasterArea->SongList[SongType] : bc1Idx;
```
lynxlynxlynx, line 2410: I don't understand the motivation here. There's a minor chance this gets hit and still we intentionally make it worse?

gemrb/core/Game.cpp
bsdcode: Indeed this isn't fully equivalent. But does this really serve a purpose? When the ChangeSong failed in the previous call why should it succeed just a moment later? The internal parameters won't have changed then.

gemrb/core/Game.cpp
bsdcode: I actually don't have a test case. I started from that comment and looked when an area song change can happen, i.e. when is PlayAreaSong called: change maps, enter/leave combat, and StartMusic action. Setting a flag in StartMusic to signal this and using this flag here in order to prevent ChangeSong spams for no-music-areas seems sensible to me.

gemrb/core/GameScript/GSUtils.cpp
bsdcode: No alternative solution needed. A GetMap which actually changes the area executes a ChangeSong afterwards.

gemrb/core/Map.cpp
bsdcode: I missed the forest for the trees here... My original motivation was a situation in the temple on Friendly Arm map (AR2304), where I started a battle with Gellana and no battle music is starting. The reason is a yet not fully understood behaviour of the original games, because AR2304 is itself a masterarea with no own battle music but should play the battle music of the "parent" masterarea AR2300. This situation is currently not covered here. Somehow for this commit I thought it was something different and AR2304 isn't a masterarea and the fallback occluded this from me. But as it turned out this is not the case here...

Still, with this change it becomes more obvious if something went wrong and further investigation is needed from the dev team. A fallback might occlude this and keep a deeper problem undiscovered.

gemrb/core/Game.cpp
lynxlynxlynx: This out of band communication is ugly and a waste of a member variable. In what case would StartMusic get into a conflict here? PlayAreaSong will start the requested music, so IsPlaying will be true anyway.

Btw, StartMusic has only 3 uses in the originals, for boss music.

gemrb/core/Map.cpp
lynxlynxlynx: Either way you'd have had to step through the function, so I'd leave it in.

Checking PreviousArea instead should fix that case for most maps in bg1. The counterpoint to test would be Nashkel mines and similar, where you stay in the same worldmap location through several areas that are all likely in mastarea.2da.

bsdcode: > Either way you'd have had to step through the function, so I'd leave it in.

Ok. I'll drop this commit then.

> Checking PreviousArea instead should fix that case for most maps in bg1. The counterpoint to test would be Nashkel mines and similar, where you stay in the same worldmap location through several areas that are all likely in mastarea.2da.

Thanks for the info. I'll test this in the next round.

gemrb/core/Game.cpp
bsdcode: Ah ok, you are right, when StartMusic starts a new music it can only be stopped by either map change or combat change. But that also means that the comment there is obsolete, and it begs the question if the IsPlaying test and the associated ChangeSong call here do actual have a reason to exist? It can function as a failsafe for unexpected stopped music at most.

So I will drop this commit. But we need to stop the ChangeSong spam with some other measurement.

lynxlynxlynx: What do you mean with ChangeSong spam?

bsdcode: The !IsPlaying test and the associated ChangeSong call are executed on every UpdateScripts call in an area without music.

lynxlynxlynx: An easy mitigation would be to only check it once per round.

bsdcode: Thanks, done.

bsdcode:
I updated the commits accordingly:
- music fallback for subareas in PlayAreaSong is back in place
- call ChangeSong only once per round in UpdateScripts in areas without music

gemrb/core/Game.cpp
```diff
@@ -1570,8 +1570,18 @@ void Game::UpdateScripts()
 		// perhaps a StartMusic action stopped the area music?
 		// (we should probably find a less silly way to handle this,
 		// because nothing can ever stop area music now..)
-		doChangeSong = true;
+
+		// Call ChangeSong only once per round in order to prevent spam
+		// calls every tick in areas without music.
+		static int ticks = 0;
```
lynxlynxlynx, line 1576: The music plays when paused, so we can't just use GameTime here. However, if the game is paused, no interaction can occur, so nothing would change the song (at most it could run out if not looping). So you actually can use GameTime.

gemrb/core/Map.cpp
```diff
@@ -2397,12 +2397,12 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
 	// Some subareas don't have their own songlist. It is currently unclear
 	// how the different games handle this situation and which music GemRB
 	// should play, if any. Further research is needed.
-	// At least for the battle music in BG1, e.g. AR2607 (intro candlekeep
-	// ambush south), there is a strong assumption to play the music from
-	// the masterarea's songlist.
+	// At least for BG1 there is a strong assumption to play the music from
+	// the masterarea's songlist, e.g. AR2607 (intro candlekeep ambush south),
+	// or AR2302 (friendly arm inn 2nd floor).
```
lynxlynxlynx, line 2405: Is this really still true? The current working hypothesis is that the last worldmap area is used.

gemrb/core/Map.cpp
```diff
@@ -2413,11 +2416,24 @@ void Map::PlayAreaSong(int SongType, bool restart, bool hard) const
...
+		// It is currently unclear how the different games handle the change
+		// to "no music". Should GemRB stop the currently playing music?
+		// Further research is needed.
+		// At least for BG1 there is a strong assumption to just stop the
+		// music, e.g. transition from AR2616 to AR2617 (go to 2nd floor
+		// in Candlekeep Inn) or from AR2616 to AR2600 (leaving Candlekeep
+		// Inn).
+		if (isBG1) {
+			musicMgr->HardEnd();
+		}
```
lynxlynxlynx, line 2429: This block still makes me uneasy. Playlists have their own short interruption sounds, so they stop gracefully. This instead cuts them off hard. Does `End` really not suffice?

gemrb/core/Game.cpp
lynxlynxlynx: Looks like the originals tolerated 10 ai updates without music, before rechecking:
https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CGameArea.cpp#L1309

lynxlynxlynx:
Did some more investigating and testing: https://github.com/gemrb/gemrb/issues/1854#issuecomment-2027570631
TLDR; we do need a new gameflag, iwds didn't have continuation, at least not in the inheritance sense.

gemrb/core/Game.cpp
bsdcode: I don't fully understand. Do we need to bring GameTime into the mix or is it ok to replace the tick >= round comparision with a tick >= 10 comparison?

gemrb/core/Map.cpp
bsdcode: Haven't had the time to test this further unfortunately.

gemrb/core/Map.cpp
bsdcode: I agree `HardEnd` is brute force. But `End` doesn't suffice. The very first example is Candlekeep Inn. Leaving it should stop the tavern music immediately, but with `End` it would play a whole lot longer after leaving the Inn, which is just not right.

gemrb/core/Game.cpp
lynxlynxlynx: I'm saying you don't need an extra timer. Looking at the code again, yeah, 10 is the default instead of round size (75/90/105), but for day/night song it only checks every 600. And an extra complication for combat. So I think we can just simplify it and leave it at round size intervals.

gemrb/core/Map.cpp
lynxlynxlynx: It's easy to test:
- pick and area with subsequent subareas (eg. the store in BG, some tower, deep dungeon)
- make the first subarea a master area by adding it to mastarea.2da
- ensure the first two have different songs set
- test what happens in the third area

gemrb/core/Map.cpp
lynxlynxlynx: I see the problem. Tav2.mus refers to itself, while most do have an ending section (day1-2 have an invalid reference). So we need to fix `End` to revert to `HardEnd` in these cases. I'll look into it and you just use `End`.

lynxlynxlynx: It's done.

lynxlynxlynx:
Ok, so only the two small changes remain and adding a proper gameflag to exclude iwd2. Do you think you'll have time to finish this soon?

bsdcode:
Sorry for the long delay. Thank you for your investigations and further help. Unfortunately I won't have time to finish this anytime soon properly. I won't work on it in the foreseeable future.

Please feel free to use anything from this PR you find useful.

lynxlynxlynx:
No problem, I'll take over for the last few steps.

PR #2071, opened 2024-03-26
optimize some of our math

bradallred:
## Description
we should be using `float_t` for math operations since we don't care about scientific precision. This allows the compiler to select the native floating point type for the architecture.

additionally, we should always use `std` for invoking math functions to use an appropriate overload that can be reduced to a compiler builtin suitable for the native arch. non `std` math functions are the `C` functions with only one signature and probably cant always get inlined.

## Checklist

### DIFF
--- a/gemrb/core/GameScript/Triggers.cpp
+++ b/gemrb/core/GameScript/Triggers.cpp
@@ -1564,16 +1564,16 @@ int GameScript::InLine(Scriptable *Sender, const Trigger *parameters)
 		return 0;
 	}
 
-	double fdm1 = SquaredDistance(Sender, scr1);
-	double fdm2 = SquaredDistance(Sender, scr2);
-	double fd12 = SquaredDistance(scr1, scr2);
-	double dm1 = sqrt(fdm1);
-	double dm2 = sqrt(fdm2);
+	float_t fdm1 = SquaredDistance(Sender, scr1);
+	float_t fdm2 = SquaredDistance(Sender, scr2);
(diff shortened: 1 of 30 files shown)


gemrb/core/GameScript/Triggers.cpp
```diff
@@ -1564,11 +1564,11 @@ int GameScript::InLine(Scriptable *Sender, const Trigger *parameters)
...
-	double fdm1 = SquaredDistance(Sender, scr1);
-	double fdm2 = SquaredDistance(Sender, scr2);
-	double fd12 = SquaredDistance(scr1, scr2);
-	double dm1 = sqrt(fdm1);
-	double dm2 = sqrt(fdm2);
+	float_t fdm1 = SquaredDistance(Sender, scr1);
+	float_t fdm2 = SquaredDistance(Sender, scr2);
+	float_t fd12 = SquaredDistance(scr1, scr2);
+	float_t dm1 = std::sqrt(fdm1);
+	float_t dm2 = std::sqrt(fdm2);
```
lynxlynxlynx, line 1571: There's a bare acos a few lines below. And a suspicious comparison multiplying radians with degrees — surely that should be a division with M_PI.

lynxlynxlynx:
How come you didn't touch the plugins?

gemrb/core/GameScript/Triggers.cpp
bradallred: that is indeed suspicious. I'll add a commit to address that too.

lynxlynxlynx:
The build failure is not from this, will check.

Issue #2080, opened 2024-05-12
 Missing BMPv5 support


lynxlynxlynx:
Looking at arrows +2 (arow11.itm) in the inventory shows no image:
> [ResourceManager]: Searching for 'CAROW11'...
> [BMPImporter/ERROR]: Compressed 32-bits Image, not supported.
> [ResourceManager/WARN]: Couldn't find 'CAROW11'... Tried CAROW11.mos CAROW11.bmp CAROW11.pvrz CAROW11.png

Good format descriptions: https://www.gibberlings3.net/forums/topic/29498-32-bit-windows-5x-bmps-in-the-ees/

The file is attached and gimp does read this format.
[CAROW11.zip](https://github.com/gemrb/gemrb/files/15285464/CAROW11.zip)

PR #2081, opened 2024-05-13
BMPs: v3 and v5 format support

MarcelHB:
## Description
See #2080.

Anyway we still have some funny stuff by now, like the BMPv5 CARROW11.bmp vs. the better looking CARROW11.BAM next to each other, in contrast to the weird PAMUL27.BAM (some discolored background) and the slighly better PAMUL27.BMP.

bmp_bam

## Checklist


lynxlynxlynx:
In my old copy of bg2ee (2.1) AMUL27 actually uses the smudgy bam; I guess that's the old bg from the originals. On bgee it looks fine.
Here's a comment from Argent77:
>  BG(2)EE doesn't use BMP for description images anymore. Instead, many description BAMs use alpha blending to achieve the desired effect (mostly used in SoD, but by now also commonly used in BGEE and BG2EE). This is the current description icon for arrows +2
image

So we should prefer BAM over BMP in this case.

I noticed the off-centered nature as well. Perhaps the ees did some scaling (since they do support gui scaling and there's a relevant options toggle) or cropping.

lynxlynxlynx:
Did you change something else? I see InventoryCommon using Button_SetItemIcon, which then goes to GameData::GetAnySprite, which first checks for a BAM, not even trying a BMP.

bradallred:
> I noticed the off-centered nature as well. Perhaps the ees did some scaling (since they do support gui scaling and there's a relevant options toggle) or cropping.

The issue is the same as any other BAM vs BMP alignment issue. The BAMs have coordinates, other image formats do not. I guess the original would just center them.

lynxlynxlynx:
Maybe centering both, yeah, since some bams are misaligned as well.

MarcelHB:
No open code changes other than what you see. And just by looking at the log output, I don't see any obvious explanation. Other items look good, like CRING06 which has both a BAM and BMP but they look almost similar.

lynxlynxlynx:
What version of bg2ee are you testing on? We don't need to cater to old ones like mine (I'll go update). In my copy the arrow bam simply isn't present.

MarcelHB:
It's v2.6.6 of GOG.com.

lynxlynxlynx:
There, all solved. Thanks for making it possible!

Issue #2095, opened 2024-05-24
Another UI performance drop


MarcelHB:
#### Bug description
Feels somewhat similar to #2087. I'm here and there is nothing moving:

![cg_fps_drop](https://github.com/gemrb/gemrb/assets/238558/ab253eb0-6169-4346-8c87-8e059ce22c38)

Yet, the more text to the right, the bigger the FPS are going down. Also: scrollbar elements are missing despite being useful here. Not sure if related to this problem for now. Anyway, plenty of font rendering going on:
![tracy_font](https://github.com/gemrb/gemrb/assets/238558/e2e36e7d-6172-4435-9e9c-ade0dac21f30)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.2
- [ ] 0.9.1
- [ ] 0.9.0

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
That's interesting, the scrollbar actually works, we just don't display it. Maybe something goes wrong when we recreate it, since it's connected to the textarea.

I can imagine it being slow on load, since it has to layout to such short columns, but the actual rendering shouldn't be impacted by the invisible text — that sounds fishy.

bradallred:
> Maybe something goes wrong when we recreate it, since it's connected to the textarea.

we dont relocate scrollbars. We ignore the ones in the CHU if they are assigned to a `TextArea` and the `ScrollView` creates its own. They are supposed to get hidden if there is no scrollable content.

The trace is impossible to decipher, but I don't think we optimize away printing text. `Font::Print` will [optimize away the sprite blitt](https://github.com/gemrb/gemrb/blob/a8b88eed79da48a14f5aace7a3216cbe14516439/gemrb/core/GUI/TextSystem/Font.cpp#L320), but I don't think anything stops the `View` from running the `Draw` code unless the entire `View` is clipped:

https://github.com/gemrb/gemrb/blob/a8b88eed79da48a14f5aace7a3216cbe14516439/gemrb/core/GUI/View.cpp#L252

bradallred:
yeah, see this `TODO`: https://github.com/gemrb/gemrb/blob/a8b88eed79da48a14f5aace7a3216cbe14516439/gemrb/core/GUI/TextSystem/TextContainer.cpp#L370

[...]

PR #2096, opened 2024-05-25
Scrollbar performance pt II

MarcelHB:
## Description
See #2095.

## Checklist

### DIFF
--- a/gemrb/core/GUI/View.cpp
+++ b/gemrb/core/GUI/View.cpp
@@ -141,6 +141,14 @@ void View::InvalidateSubviews(const Region& rgn) const
 	}
 }
 
+void View::SetVisible(bool vis) {
+	SetFlags(Invisible, vis ? BitOp::NAND : BitOp::OR);
+
+	if (superView != nullptr) {
+		superView->MarkDirty();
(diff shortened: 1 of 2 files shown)


gemrb/core/GUI/View.cpp
```diff
@@ -141,6 +141,14 @@ void View::InvalidateSubviews(const Region& rgn) const
 	}
 }
 
+void View::SetVisible(bool vis) {
+	SetFlags(Invisible, vis ? BitOp::NAND : BitOp::OR);
+
+	if (!vis && superView != nullptr) {
```
bradallred, line 147: I think we need to do this regardless of `vis`. I cant think of a real world scenario, but if there were some case where a superview was tainted in the same region (eg overlapping view, or swapping a view with another which we do) as the view becoming visible, and the now visible view wasn't fully opaque, there could be artifacts.

I dont think there is much payoff optimizing it compared to difficulty debugging such a scenario.

MarcelHB: Alright

lynxlynxlynx:
Thanks!

PR #2110, opened 2024-07-28
Fixes around save game directories

MarcelHB:
## Description
Some issues fixed I encountered earlier today:

1. Crash when saving with non-ASCII characters since code path used open setting `SystemEncoding` when everything else is expected UTF-8 regarding paths for now, so invalid bytes.
1. Underlying VFS fixes on Windows after that was fixed.
1. Cannot delete directories on Windows with non-ASCII chars from FS.

## Checklist


MarcelHB:
Will look at the runners tomorrow.

lynxlynxlynx:
Thanks!

Issue #2104, opened 2024-07-18
 Icewind Dale Complete crashing upon entering overworld map when using GemRB


HJ000001:
#### Bug description
Hello devs,

I'm using the GoG version of Icewind Dale Complete with the latest version of GemRB.

The tavern loads fine after starting a new game, but as soon as I go outside it crashes. If I try to start the expansion directly it crashes right after character creation. Since I can go to the tavern cellar I am pretty certain that it is the game trying to load the overworld map which makes it crash.

Neither happens when I don't use GemRB, so I am certain that GemRB is the issues. I've tried changing several of the settings in the .cfg file, but it hasn't helped. I also couldn't see anything helpful in the log, although I don't know what exactly to look for either. 

#### Steps to reproduce
1. Select the Icewind Dale base game.
2. Create a character and start the game.
3. Leave the tavern.

OR 

1. Select the Icewind Dale HoW expansion.
2. Create a character and start the game.

#### Expected behavior
The overworld map should load.

GemRB version (check as many as you know apply)
- [ ] master as of this issue
- [X] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)

I don't know. I didn't change anything, so whatever is the default in GemRB 0.9.3 and a freshly installed Icewind Dale Complete Edition.

- [ ] SDL1.2
- [ ] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

**Here is the GemRB.log:**
http://paste.debian.net/1323609/

lynxlynxlynx:
This is on windows? Nothing sticks out from the log.

HJ000001:
Yes, this is on Windows 11 Pro.

It appears to me that the mere fact GemRB is used causes the crash. I've disabled every single line in the .cfg using # signs and it still happened.

HJ000001:
Yes, this is on Windows 11 Pro.

It happens even if every single line in the .cfg file is disabled using #, so it seems that the mere fact GemRB is used causes the crash.

lynxlynxlynx:
If you're using any sort of scaling, you can change the ScaleQuality config setting back from best to nearest, which will do less interpolation.

MarcelHB:
Hm, I can reproduce this using the official binaries. Compiling v0.9.3 on my own looks good.

lynxlynxlynx:
[...]

PR #2111, opened 2024-08-12
WEDImporter: getting rid of float math

MarcelHB:
## Description
Similar to #1790, #2104, Clang on Raspberry Pi will warn that `-frounding-math` is not supported on this platform, and the same crash will happen again.

So instead of this annoying flag handling, let's get rid of this `ceilf` here.

## Checklist


lynxlynxlynx:
Thanks!

PR #2114, opened 2024-08-17
CMakeLists: removes `SDL_RESOLUTION_INDEPENDANCE`

MarcelHB:
## Description
It appears that the original intent of this macro was never validated, and I've noticed that choosing _Full screen_ at the in-game options makes no difference to this code path. Can be tested by starting GemRB from TTY on a Wayland computer with something like a full HD screen.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,6 @@ OPTION(DISABLE_WERROR "Do not treat warnings as errors" OFF)
 OPTION(USE_TRACY "Build with Tracy support" OFF)
 
 OPTION(USE_SDL_CONTROLLER_API "Enable SDL controller APIs. (disable if you plan on handling controller input in an external program like gptokeyb)" ON)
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
The difference is that you now can't have fullscreen by default, since we deprecated the config key and only look at the game one (where it's unset if you don't run config.exe). We could add a flag like `-f` to take care of that.

MarcelHB:
All the games have something in their own settings menu, and when it's ticked then it's switching and persisted. Some of the old games, unfortunately, don't have it accessible from the main menu, only after entering the game. So an `-f` is more like a one-time requirement.

lynxlynxlynx:
Maybe, that's assuming the initial windowed run works fine enough. I don't know if that can be a problem.

MarcelHB:
There now is an `-f`, works for me like this.

lynxlynxlynx:
Thanks!

PR #2115, opened 2024-08-18
GAMImporter: fixes PST variable special case when saving

MarcelHB:
## Description
As reported by https://github.com/gemrb/gemrb/issues/1404#issuecomment-1966478997 PST original crashes again when loading one of our save games. Not sure when it broke but `MakeVariable` removes that critical whitespace from this variable, and the old fix did not do much help anymore.

## Checklist


lynxlynxlynx:
Thanks!

PR #2121, opened 2024-08-25
RPi performance optimization ground

MarcelHB:
## Description
On the SDL discourse, [I was asking](https://discourse.libsdl.org/t/performance-using-wayland-window/53600/1) around for performance tips esp. regarding Raspberry Pis, since I've found that it is kind of luck to find an environment that has somewhat acceptable _and_ predictable performance, and a low risk of side-effects, esp. input lags from hell that makes GemRB no fun to use.

A finding was: the upper bound of performance is roughly linearly bound to the number of pixels the RPi needs to refresh, so some simple operations at the same drawing buffer size will slow down simply because the window is maximized/fullscreen. VRAM gives something at max of 100FPS at full HD resolution.

SDL2 can do EGL + KMSDRM, so you can launch SDL2 applications directly from TTY and rule out some Voodoo by the window managers. But since old games are usually no more than 1024x768, so roughly 3x the upper bound, then opening a full HD window for this resolution will give us a penalty that makes us wait for `SDL_RenderPresent` more than we spend on the game (VSync disabled), like in the main menu:

perf_fullhd

Now, with this PR there I had the idea to adjust the window that preserves the aspect ratio while it's not full HD-upscaled fullscreen. So when you request fullscreen from TTY on 1920x1080 with the game setting on 1024x768, then the fullscreen window is 1366x768, so no stretching, the same visual result (upscaled fullscreen) and 25% more space for performance improvements by us, for comparison 5 vs 10ms in `SDL_RenderPresent`:

perf_subhd

_This should not affect any other setup_, i. e. using windows managers won't see this and probably cannot. The rest is some infrastructure changes to Tracy as it is recommended as best practice by the docs to have its own shared lib when using it across multiple ones.

## Checklist


gemrb/plugins/SDLVideo/SDL20Video.cpp
```diff
@@ -245,6 +272,26 @@ int SDL20VideoDriver::CreateSDLDisplay(const char* title, bool vsync)
...
+	}
+
+	// Only cover the case when width and height have the same alignment in both cases
+	if (!(mode->w/mode->h > 0 && screenSize.w/screenSize.h > 0)) {
+		return;
+	}
+
+	// e. g. display is 1920x1080, game is 1024x768 -> fullscreen window 1365x768
+	float displayRatio = mode->w * 1.0 / mode->h;
+	auto width = static_cast<int>(floorf(screenSize.h * displayRatio));
```
lynxlynxlynx, line 288: Won't this be problematic on windows? Last time it was `ceilf`. Perhaps use the replacement from there and just adjust in the other direction.

lynxlynxlynx:
Other than that, LGTM!

gemrb/plugins/SDLVideo/SDL20Video.cpp
MarcelHB: You mean what I did to fix by e73f53eeccbf3d525f770709b1407ec8e92f3e4d? The math is a little more here since I don't have the display ratio fix, i. e. what to use for `div`. I'll have look for a general approach.

lynxlynxlynx:
Thanks!

Issue #2137, opened 2024-09-18
IWD2 travel region incompatibility


lynxlynxlynx:
So this is what the save  game looks like in GemRB:
![iwd2_savegame_gemrb](https://github.com/user-attachments/assets/5231fff3-6a1e-4dde-9924-d7109edd6f1f)

And this is what happens when loaded in the original game:
![iwd2_savegame](https://github.com/user-attachments/assets/5f421db5-d688-49ca-b287-f37e134f36c3)

It's the same for every local travel point I've tested,

_Originally posted by @MarcelHB in https://github.com/gemrb/gemrb/issues/2125#issuecomment-2358060737_

MarcelHB:
Broke somewhere between v0.9.1 and v0.9.2 ... will dive deeper.

MarcelHB:
It's 11f146274742bc65596f9ab62d97b50259befbeb or later since this particular one makes the original crash.

MarcelHB:
Looks like override and alternate point don't get their Y values set.

PR #2138, opened 2024-09-18
AREImporter fix for IWD2

MarcelHB:
## Description
See #2137. 

So ARE1000 has _alternate_ trigger points set (2x 4B) but we did `ReadPoint` that reads 2x 2B. Thus the Y value got lost and trigger points didn't work.

IESDP should maybe be updated, too, since it's both incomplete and doesn't match NI naming.

## Checklist


lynxlynxlynx:
Thanks. Also updated IESDP.

Issue #2143, opened 2024-09-19
 IWD2: Morale recovery zero


MarcelHB:
#### Bug description
If one of the PCs has broken morale, I can't get them recovered. Ctrl+M says the following:

```
[Actor/DEBUG]: Debugdump of Actor Nummer Vier (Nummer Vier, Nummer Vier):
Scripts: <none> <none> <none> <none> <none> <none> <none> <none>
...
Morale:     1   current morale:1
Moralebreak:5   Morale recovery:0
```

Looks like recovery will never happen then: https://github.com/gemrb/gemrb/blob/b52ca94e41f2f251209b1900900b4473c1e11ffc/gemrb/core/Scriptable/Actor.cpp#L2947-L2950

In contrast, this is a healthy one, but also no positive recovery:
```
Debugdump of Actor Nummer Drei (Nummer Drei, Nummer Drei):
...
Morale:     96   current morale:96
Moralebreak:5   Morale recovery:0
```

I'd say that this values lacks an assignment somewhere. BG1 in contrast looks good on a new game.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

lynxlynxlynx:
charbase has it at 60, so something went wrong.

MarcelHB:
Yep, reading is skipped since refactoring there.

MarcelHB:
Hm, original sets 1 in the file whereas other games such as BG set 60 there. I'd consider multiplying this value by sixty to be consistent with the rest for now.

lynxlynxlynx:
Sounds fine, less complexity elsewhere.

MarcelHB:
Uncovers another bug that _sometimes_ doesn't put the PC out of the yellow circle, and prevents saving. Trying to dig through.

lynxlynxlynx:
There are codepaths where we set STATE_PANIC potentially permanently, if that's the thing that caused your yellow.

PR #2144, opened 2024-09-19
Morale fixes

MarcelHB:
## Description
See #2143, CREImporter fix.

## Checklist

### DIFF
--- a/gemrb/plugins/CREImporter/CREImporter.cpp
+++ b/gemrb/plugins/CREImporter/CREImporter.cpp
@@ -1553,10 +1553,9 @@ void CREImporter::GetActorIWD2(Actor *act)
 	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_CHR]);
 	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_MORALE]);
 	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_MORALEBREAK]);
-	//HatedRace is a list of races, so this is skipped here
-	str->Seek(1, GEM_CURRENT_POS);
-	//act->BaseStats[IE_HATEDRACE]=tmpByte;
 	str->ReadScalar<Actor::stat_t, ieByte>(act->BaseStats[IE_MORALERECOVERYTIME]);
+	act->BaseStats[IE_MORALERECOVERYTIME] *= 60; // guess since IWD2 uses 1 here
+	str->Seek(1, GEM_CURRENT_POS);
 	//No KIT word order magic for IWD2
 	str->ReadDword(act->BaseStats[IE_KIT]);
 	ReadScript(act, SCR_OVERRIDE);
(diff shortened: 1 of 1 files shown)


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -878,12 +878,14 @@ static void pcf_morale (Actor *actor, ieDword /*oldValue*/, ieDword /*newValue*/
 		// recover from panic, since morale has risen again
 		// but only if we have really just recovered, so panic from other
 		// sources isn't affected
-		if ((actor->Modified[IE_MORALE]-1 == actor->Modified[IE_MORALEBREAK]) || (actor->Modified[IE_MORALEBREAK] == 0) ) {
+		if ((actor->Modified[IE_MORALE]-1 >= actor->Modified[IE_MORALEBREAK]) || (actor->Modified[IE_MORALEBREAK] == 0) ) {
 			if (!third || !(actor->Modified[IE_SPECFLAGS]&SPECF_DRIVEN)) {
 				actor->SetBaseBit(IE_STATE_ID, STATE_PANIC, false);
+				actor->BaseStats[IE_CHECKFORBERSERK] = 0;
```
MarcelHB, line 884: Eh, not sure ... `SetCircleSize` actually looks at the `Modified` one. Can I simply set this here?

lynxlynxlynx: It's fine, matches the rest. The other one checks Modified, since also effects can set the stat.

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -878,12 +878,14 @@ static void pcf_morale (Actor *actor, ieDword /*oldValue*/, ieDword /*newValue*/
 		// recover from panic, since morale has risen again
 		// but only if we have really just recovered, so panic from other
 		// sources isn't affected
-		if ((actor->Modified[IE_MORALE]-1 == actor->Modified[IE_MORALEBREAK]) || (actor->Modified[IE_MORALEBREAK] == 0) ) {
+		if ((actor->Modified[IE_MORALE]-1 >= actor->Modified[IE_MORALEBREAK]) || (actor->Modified[IE_MORALEBREAK] == 0) ) {
```
lynxlynxlynx, line 881: I'm not so sure this is correct — won't it break non-morale induced panic? Effect based one should be fine if they don't have permanent timing, since the stat will get updated each tick, but we also permanently set it for turn undead and Panic actions, potentially only once.

MarcelHB: Hm, then we'd maybe need a panic cause marker or something like this to address this? I'm not sure, the whole thing isn't trivial.

lynxlynxlynx: IE_MORALERECOVERYTIME is not affected by effects, so we could zero it for the other cases, so normal passage of time won't affect morale. And add a check here to skip such cases, since there are other sources of morale.

MarcelHB: Ok, I do the following since there is probably more than I can figure out this way or fix correctly: I drop the 2nd commit so that the CREImporter is fixed and create the new ticket describing the remaining problems. Maybe you can have a look at this then?

lynxlynxlynx: Yeah, that's fine.

lynxlynxlynx:
Thanks!

PR #2148, opened 2024-09-22
Fix Haiku build

Peppersawce:
Haiku 0.9.3 build has issues with finding & using extend2da.py and cannot find SDL2 out of the box, as found while porting the engine to the platform: https://github.com/haikuports/haikuports/pull/11141

I'm submitting these small fixes so the engine builds on the system without need for patching.


admin/extend2da.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!python3
```
lynxlynxlynx: This is not portable enough, please use a patch downstream.
I guess `/usr/bin/env python` is also not sufficient on Haiku?

Peppersawce: Ah sorry, I botched that first commit, I meant to write it as: #!/usr/bin/python3
I fixed it in the second commit I think, it should be fine now :)

Btw Haiku accepts both #!python3 and #!/usr/bin/python3 so either syntax is fine for the systen

Peppersawce: Ah ok, I'll revert that change then

lynxlynxlynx:
Thanks!

admin/extend2da.py
lynxlynxlynx: For the cache, we only check that the dir is empty, so you don't lose any files. Beyond that standard permission limits apply.

Issue #2146, opened 2024-09-20
 Half-transparent actor shadows


MarcelHB:
#### Bug description
To be seen in #2122 or #2134, the original games have non-opaque shadows below actors, which makes the visual presentation a little smoother than these full black shadows we currently use.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

MarcelHB:
Weird, there is this https://github.com/gemrb/gemrb/blob/f875f732c2815f00ae8760569232a8a3955ddf48/gemrb/core/Scriptable/Actor.cpp#L7873-L7876

and we can modify RGB in whatever fashion, but A doesn't do anything to the result. Texture format is good and `u_rgba` ends up correctly in the shader. I suspect some palette SDLism has other ideas about alpha values here.

MarcelHB:
That's a little more complicated, as usual:

* For actors, `ALPHA_MOD` is always enabled. To me, it should be enabled only if there is a whole-actor alpha modification (e. g. invisibility).
* Most of the time it works since `tint.a` is also opaque and whole palette is fixed to be set opaque.
* But so is the shadow entry, that loses half-transparency by this. [1]
* If we make this `ALPHA_MOD` conditional, like when the alpha tint is below 255, shadows work, but no longer does the Cloak of Oscillating Colors (see https://github.com/gemrb/gemrb/issues/397)
* Even worse, without alpha mod, there is a code branch [2] that doesn't care about tint but then in [3], the palette isn't recognized as outdated any more, since we don't have tint value. For good reasons, a tint should invalidate a palette, but we originally don't have it here. Instead `CLOAKSC` has just modified colors but we won't notice in time (something else does, if we pause, then it paints slowly ...).

This needs to be defiddled somehow.

[1] https://github.com/gemrb/gemrb/blob/f875f732c2815f00ae8760569232a8a3955ddf48/gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp#L248-L252
[2] https://github.com/gemrb/gemrb/blob/f875f732c2815f00ae8760569232a8a3955ddf48/gemrb/plugins/SDLVideo/SDL20Video.cpp#L402-L407
[...]

Issue #897, opened 2020-08-05
Spell effect visible even if it souldn't be 


FrElvire:
#### Bug description
In this battle with draconis (dragon form), you can't see draconis as you are too far from him but you can always see the spell effects, very convenient to launch zone spell without being too close but I suspect you shouldn't not see them.

Evidently this case is just an example.

GemRB version: kmfrick los-slim-check branch

Tested with SDL1 and SDL2

#### Expected behavior
If you don't see a character as he is too far, you shouldn't see spell effect either

#### Screenshots
![spell_effect_far](https://user-images.githubusercontent.com/55502325/89396283-a373a300-d70e-11ea-8daf-ae87d29dca73.png)
![spell_effect_far_2](https://user-images.githubusercontent.com/55502325/89396326-b4241900-d70e-11ea-80aa-3451c10a9c4f.png)

lynxlynxlynx:
I could reproduce enough of this.

As part of trying to resolve it I've added more wallgroup occlusion, specifically to projectiles and scripted animations. It's not fully functional yet though. Besides what I noted for projectiles in 65a27542dc, the vvcs attached to the actor and blended standalone ones suffer from the same problem. I suspect it's the same reason we manage to appear to draw over the fog, even though fog gets rendered later.

@bradallred any ideas? I didn't manage to find out why `BlitFlags::ONE_MINUS_DST` is the problem, but interestingly making IE_VVC_BLENDED a noop (so the blit flag doesn't get set) actually didn't cause any obvious regressions — transparency still worked.

bradallred:
`IE_VVC_BLENDED` is a misnomer on our part. These things are always blended, the flags only control the mode of blending. See #1291.

Basically, if the resource has a black background (check in NI), then use `ONE_MINUS_DST` (or possibly `DST` or `SRC`). If we are adding `IE_VVC_BLENDED` somewhere programmatically, it's probably not the right thing to do. I wouldn't be surprised if we could completely remove `ScriptedAnimation::SetBlend`, it doesn't seem like something that would be contextual.

As for the stencils there isn't anything in the game logic that I know of making choices based on these flags or the `BlitFlags`.

The first thing to ensure is that the appropriate bit in `BlitFlags ::STENCIL_MASK` is set. things like `IE_VVC_NOCOVER` disable those.

If all that checks out then it's probably something going on in OpenGL:
[...]

PR #2151, opened 2024-09-24
Palettes, shadows and blending

### LINKED ISSUE #2150: Non-accelerated blending broken
#### Bug description

The non-accelerated SDL2 version of blending got worse over time. While somewhat OK at v0.9.1, first drawbacks at v0.9.2, v0.9.3 can be considered broken.

It's tested on both D3D and OpenGL. Also not an SDL problem since all tests were done with SDL v2.30.

#### Screenshots
v0.9.1
blending_091

v0.9.2
blending_092

v0.9.3
blending_093

- [x] master as of this issue
- [x] 0.9.3
- [x] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `OPENGL_BACKEND` enabled
- [x] SDL2 without `OPENGL_BACKEND` enabled

MarcelHB:
## Description
This fixes #2150, #2146 (and fallout) and also restores, again, infravision.

What does it do:
* Changes to `Palette` won't go unnoticed anymore which makes it easier to check if it needs to be updated to the surface.
* `SDLSurfaceSprite2D` has some easier to understand update-requirements for both palettes and surface.
* I've eliminated one manual blitting path, because I don't see where SDL2 is unable to do the same, so only one surface left to have.
* There is now an original palette + a tinted one when required, leaving the original untouched for rendering.
* Non-accelerated stencilling preserves the desired blend mode, so all code-paths look the same.

Tested all the known edge cases that now are easy to handle under less heuristics and state:
* PST: crystal room + Ignus spells
* IWD: Cloack of oscillating colors
* BG: Thalantyr's store room
* IWD2: infravision

## Checklist

### DIFF
--- a/gemrb/core/CharAnimations.cpp
+++ b/gemrb/core/CharAnimations.cpp
@@ -2801,6 +2801,8 @@ void CharAnimations::PulseRGBModifiers()
 
 static inline void applyMod(const Color& src, Color& dest, const RGBModifier& mod) noexcept
 {
+	dest.a = src.a;
+
 	if (mod.speed == -1) {
 		if (mod.type == RGBModifier::TINT) {
 			dest.r = ((unsigned int)src.r * mod.rgb.r)>>8;
--- a/gemrb/core/CharAnimations.cpp
+++ b/gemrb/core/CharAnimations.cpp
@@ -543,7 +543,7 @@ void CharAnimations::SetupColors(PaletteType type)
 		*/
 		for (int i = 0; i < colorcount; i++) {
 			const auto& pal32 = core->GetPalette32(static_cast<uint8_t>(Colors[i]));
-			PartPalettes[PAL_MAIN]->CopyColorRange(&pal32[0], &pal32[32], static_cast<uint8_t>(dest));
+			PartPalettes[PAL_MAIN]->CopyColors(dest, &pal32[0], &pal32[32]);
 			dest +=size;
 		}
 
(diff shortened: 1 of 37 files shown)


gemrb/core/CharAnimations.cpp
```diff
@@ -2801,6 +2801,8 @@ void CharAnimations::PulseRGBModifiers()
 
 static inline void applyMod(const Color& src, Color& dest, const RGBModifier& mod) noexcept
 {
+	dest.a = 0xFF;
```
lynxlynxlynx, line 2804: I guess this is just from the default init of Color? It'd be better if we fixed this at the source.
I see `BMPImporter::GetPalette`, PNGImporter and PVRZImporter already handle it properly, just BAMImporter is missing the same signature, since it's not an ImgMgr. But it has code in `BAMImporter::Import` that should take care of it — is the v2 alpha detection at the end failing us?

MarcelHB: The default init hasn't bothered be in this case, at least I'm not aware of one. It's this here: https://github.com/gemrb/gemrb/blob/f22f8625ee7676d4a8c32e6091534cbd782ee557/gemrb/plugins/FXOpcodes/FXOpcodes.cpp#L1351

gemrb/core/MurmurHash.h
```diff
@@ -0,0 +1,55 @@
...
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program; if not, write to the Free Software
+* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*
+*/
+
+#ifndef H_GEMRB_MURMUR_HASH
+#define H_GEMRB_MURMUR_HASH
```
lynxlynxlynx, line 21: What's the benefit over using our existing hashing capabilities, std?

gemrb/plugins/AREImporter/AREImporter.cpp
```diff
@@ -262,7 +262,7 @@ static TileProps MakeTileProps(const TileMap* tm, const ResRef& wedref, bool day
 		}
 		
 		if (hmit.clip.PointInside(pos)) {
-			b = hmpal->col[*hmit].r; // pick any channel, they are all the same
+			b = hmpal->operator[](*hmit).r; // pick any channel, they are all the same
```
lynxlynxlynx, line 265: Can you add a GetColorAt or somesuch getter, so all these accesses will look a bit cleaner?

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
```diff
@@ -103,27 +94,14 @@ void* SDLSurfaceSprite2D::LockSprite()
...
-{
-	// a 'version' implies tha palette was color modified
-	// however, the only caller her is RenderWithFlags
-	// where we already compare the version with the requested one
-	// the version includes the color modification so for now we dont worry about it
-	
-	const Holder<Palette>& pal = format.palette;
-	assert(pal);
-	return pal->GetVersion() != palVersion;
+	invalidate();
```
lynxlynxlynx, line 100: Why lowercase?

gemrb/core/MurmurHash.h
MarcelHB: std needs a specialization for this case, as far as I see, and thus some implementation. And since I'm no hashing export, this one is safe and fast to use for streaming 4-bytes.

gemrb/plugins/AREImporter/AREImporter.cpp
MarcelHB: Sure.

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
MarcelHB: Wasn't there some kind of private-vs-public kind of convention in some classes? Anyway, I can change it.

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
```diff
@@ -188,97 +167,122 @@ bool SDLSurfaceSprite2D::ConvertFormatTo(const PixelFormat& tofmt) noexcept
...
+	return format.palette && palVersion != format.palette->GetVersion();
 }
 
-void* SDLSurfaceSprite2D::NewVersion(version_t newversion) const noexcept
+bool SDLSurfaceSprite2D::needToReshade(const Color* tint, BlitFlags flags) const noexcept {
+	return flags != appliedBlitFlags
+		|| ((flags & BlitFlags::COLOR_MOD) && tint && (appliedTint.Packed() & 0xFFFFFF00) != (tint->Packed() & 0xFFFFFF00))
+		|| ((flags & BlitFlags::ALPHA_MOD) && tint && (appliedTint.Packed() & 0xFF) != (tint->Packed() & 0xFF))
+		|| needToUpdatePalette()
+		|| surfaceInvalidated;
```
lynxlynxlynx, line 209: This could be checked sooner, for a tiny speed increase.

MarcelHB: Sure, why not

gemrb/tests/core/Test_Palette.cpp
```diff
@@ -0,0 +1,84 @@
...
+	auto v3 = unit.GetVersion();
+	EXPECT_NE(v2a, v3);
+
+	unit.SetColor(0, c);
+	auto v4 = unit.GetVersion();
+	EXPECT_NE(v3, v4);
+}
+
+TEST(Palette_Test, CopyColors_Version) {
+Palette::Colors buffer;
```
lynxlynxlynx, line 62: misaligned

lynxlynxlynx:
Very nice, will have to recheck if it helps with #897 as well.

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
lynxlynxlynx: In some few, mostly in imported code, but definitely not for most. I think it would be better to unify rather than making it more diverse. The one valid exception are iterator accessors, for obvious reasons.

gemrb/core/MurmurHash.h
lynxlynxlynx: Did you forget to commit something? I don't see it actually being used anywhere.

gemrb/core/CharAnimations.cpp
lynxlynxlynx: These two opcodes [never pass an alpha](https://gibberlings3.github.io/iesdp/opcodes/bgee.htm#op8), so we can fix it in them. I suggest either doing it manually or by adding a  Color::FromBGR, where we force the alpha. There's several other calls that pass a custom Color that need to be updated.

MarcelHB: Ok, I'll take the `FromBGR` idea.

MarcelHB: I guess you meant all 4 methods on 0x8 and 0x9, right?

lynxlynxlynx: Yeah, those are fine, but aren't the other SetColorMod calls problematic too? I don't see anyone setting a non-zero alpha.

MarcelHB: This way I doesn't even work for my known case any more... Something eats the opaque value between opcode and `applyMod`.

lynxlynxlynx: It's not some other call resetting it to its value?

MarcelHB: At a second thought, I think the initial idea is better and the 2nd approach is fruitless: This `applyMod` isn't designed to work on any alpha value. Instead, it takes the original RGB, applies some mod and sets the destination's RGB. And yes, by any default init. of `Color`, alpha is 0, so is the palette's setup buffer.

So in my view, `applyMod` is the place to mark these colors as opaque.

lynxlynxlynx: It feels like an unnecessary side effect, since `dest` is passed by reference. `GlobalColorMod` does not use alpha, so you could fix its `rgb` member instead, it's just 4 places.

MarcelHB: That doesn't fix this since it is not copied on all code paths, on some it's just source.rgb x mod.rgb -> dest.rgb. So either we set dest.a = 0xFF (as initially) inside of `applyMod`, or we set buffer.a = dest.a = 0xFF outside of `applyMod`, or we change the color init., but I'm not sure about the impact by this.

MarcelHB: Or buffer.a = dest.a = src.a

MarcelHB: That's good we rule out unexpected side effects but also don't have to rewrite things or do another set-alpha-run.

lynxlynxlynx: What other case does that (I admit I forgot the one ColorMods.rgb setter in the same function)?

MarcelHB: Look at everything in `applyMod` that does not end in `dest = src;`: you will only see color modifications. So if `dest.a` is blank, it will remain regardless of source.a and modifier.a.

Issue #2179, opened 2024-10-02
 Actual ghosts around


MarcelHB:
#### Bug description
It's the second time I'm encountering something in IWD2 I can't explain so far or couldn't debug with some quick checks (like no cricles even when enforced for everyone, no cursor, no debug help, not occupying the search map).

There are sometimes floating pieces of something around that attack the party. The message tells me it's some enemy type that is around here:

![iwd2_ghost](https://github.com/user-attachments/assets/8fb43ad5-a58d-4acf-93b5-ffef77f41a03)

It's somewhat a problem since you can't kill them in regular or Ctrl-code ways.

#### Steps to reproduce

Take this IWD2 save game and watch for spawning enemies from the SW. Kill the obvious ones and watch out for these pieces.
[000000001-Quick-Save.zip](https://github.com/user-attachments/files/17235947/000000001-Quick-Save.zip)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

lynxlynxlynx:
It's actor chunks, I've hit it several times, more on reentry of areas. No idea how they persist despite everything we throw at them. My last change to stun them didn't appear to help, but we can just clear their script slots.

MarcelHB:
The fix didn't solve the problem. Some levels later and they are back.

lynxlynxlynx:
In new areas or ones you were in before?

MarcelHB:
AR6102, and this is after the fix. Wouldn't have played it without a clean version.

What if it's not the clones but the original actor? All three ghosts have animation ID 0x220 and that's what the splattered actor is being assigned right before.

lynxlynxlynx:
Cool, so you already found the reason. I think we can pull the teeth out of the source actor as well. Any story crucial actor will be immune to chunking, so the fact that they don't get one last tick to properly run scripts should be fine.

MarcelHB:
_Maybe_ it's the reason. There is also one copy getting 0x220, next to other values in the range between 0x200 and 0x240.

MarcelHB:
Also chances are good that the problem lies in loading? If they had survived, I'd not be able to save or had noticed it earlier.

lynxlynxlynx:
I don't know of anything else but scripting that could interfere with proper death/expiry.

MarcelHB:
[...]

PR #2180, opened 2024-10-03
Actor: clearing scripts when cloning

MarcelHB:
## Description
See #2179.

That's maybe just half of the fix since surviving chunks continue to be visible, by just spinning around locally and preventing saving when nearby. But it makes the game more playable again in this situation.

## Checklist


lynxlynxlynx:
Thanks, it should be enough for new cases. The other part of the problem is likely what you already reported, some incomplete deaths (unless it turns out to be a purely animation one).

Issue #2166, opened 2024-09-28
 Enemies always know where to find us


MarcelHB:
#### Bug description
Another things that contributes to increased difficulty is the fact that enemies can follow you across the map, if they have locked on one actor, even if you are faster and ran away or have teleported/Ctrl+J'd away and thus are way out of sight.

Enemies should stop following after (some time after) losing sight of us. It also contributes to some pathfinding-related CPU frame drops.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

lynxlynxlynx:
It all depends on how they're scripted, I'm not sure there's a universal answer to what's right.

MarcelHB:
For the general attack behavior, there is I guess: https://github.com/gemrb/gemrb/blob/a7019288236b3e088ec085031afe1ebcdf09f727/gemrb/core/GameScript/GSUtils.cpp#L1567-L1572

That looks looks a map-wide rubberband. I will do some more research.

lynxlynxlynx:
The fact that enemies can lock on to certain targets is exploited by players, a form of kiting. That's one of the reasons why there's not just an Attack action, but also AttackReevaluate, for example.

MarcelHB:
Sure, so do I, but it has some odd consequences. Let's look at this:

![iwd2_paths](https://github.com/user-attachments/assets/61e29233-7aa4-4e58-9162-4dd2e7fbdde5)

To the NW, there are a bunch of orcs. You may not see all of them but some but can reach them since the path is blocked, but apparently they have a lock on you. Then you go to the east where this cleric woman can lift the barriers.

Now the whole crows walks immediately the red path towards the party. If you also move, the path finder will drop frames considerably.

lynxlynxlynx:
Ok, that's ar2000. The T7_Firestarter_2 orc (first I opened) has 00t07m as his movement script:
```
IF
	Global("TEAM_7","MYAREA",1)
	Exists(NearestPC)
	!See(NearestPC,0)
THEN
	RESPONSE #100
		MoveToObject(NearestPC)
END
```
So first practical suspect would be the NearestPC object matching.

lynxlynxlynx:
Looking at NearestPC disassembly makes me think we already have a good implementation. I see it does the complicated see invisibility stuff, but there's no visual range check. So this requires further testing in the original.

See() does have a range check, but since it's inverted, it doesn't matter.

[...]

PR #2182, opened 2024-10-05
Three sound improvements

MarcelHB:
## Description
Looking at what has been fixed by 95d258a34a9b98f62f13fb4db325f634aa6ad818 and what was haunting me in https://github.com/gemrb/gemrb/issues/2166, I'd like to propose three sound improvements:

1. Walking sounds were not spatial but were vague all over the map, now they are. Unfortunately they are very loud this way compared to other sounds of the same volume setting. That's why I added some multiplier. Feel free to test and compare to the original.
2. Spatial sounds were not rolling off over distance. Now they do over while not turning fully mute technically (that's also a matter of the OpenAL distance model) but you can now hear things coming close and going away depending on the camera/listener's position. That's way more immersive hearing the enemies finding their way and coming closer. :slightly_smiling_face: 
3. OpenAL mostly doesn't support stereo data being spatialized, so many area affects stay as loud and centered no matter where you are on the map (see annoying thrall invisibility attempts). Now we can have these effects in the same way as everything else, following the official idea: splitting stereo into two mono channels and vector-play them.

Please feel free to test and give some feedback.

## Checklist

### DIFF
--- a/gemrb/core/Audio.cpp
+++ b/gemrb/core/Audio.cpp
@@ -24,6 +24,9 @@ namespace GemRB {
 
 const TypeID Audio::ID = { "Audio" };
 
+// walking/armor sounds are too loud relative to other noises
+static const EnumArray<SFXChannel, float> channelVolumeMultipliers { 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 0.3f, 0.3f, 0.1f };
+
 void Audio::UpdateChannel(const std::string& name, int idx, int volume, float reverb)
 {
 	channels[idx] = Channel(name);
@@ -50,7 +53,7 @@ int Audio::GetVolume(SFXChannel channel) const
 	if (channel >= SFXChannel::count) {
 		return 100;
 	}
-	return channels[channel].getVolume();
+	return channels[channel].getVolume() * channelVolumeMultipliers[channel];
 }
 
 float Audio::GetReverb(SFXChannel channel) const
--- a/gemrb/core/Audio.cpp
+++ b/gemrb/core/Audio.cpp
@@ -24,11 +24,42 @@ namespace GemRB {
 
 const TypeID Audio::ID = { "Audio" };
 
-// walking/armor sounds are too loud relative to other noises
-static const EnumArray<SFXChannel, float> channelVolumeMultipliers { 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 0.3f, 0.3f, 0.1f };
+static std::unordered_map<std::string, SFXChannel> channelEnumMap = {
+	{"NARRATIO", SFXChannel::Narrator},
+	{"AREA_AMB", SFXChannel::MainAmbient},
+	{"ACTIONS",  SFXChannel::Actions},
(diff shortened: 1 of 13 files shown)


lynxlynxlynx:
1. It's odd that this would be needed — we already take `sndchann.2da` factors into account. A further decrease by factors of 0.3 and 0.1 seem wild. 
Looking at iwd2-re, [at least at init](https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CBaldurChitin.cpp#L2191), there are no further factors applied to those values. The DirectSound calls [seem to pass negative values](https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CSound.cpp#L197) if that was correctly extracted.
Anyway, we could just provide overrides for the table. One or two mods may actually change it, but if even the default isn't working properly, they'd need to provide GemRB-specific tables anyway. So a hard override should be fine.

2. `SetupNewStream` uses quite different values for rolloff, I wonder if we could unify the approach. Ambients are not perfect yet either IIRC.

3. Annoying, but since the result for the user is better, let's go for it.

MarcelHB:
> 1. It's odd that this would be needed — we already take `sndchann.2da` factors into account. A further decrease by factors of 0.3 and 0.1 seem wild.
>        Looking at iwd2-re, [at least at init](https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CBaldurChitin.cpp#L2191), there are no further factors applied to those values. The DirectSound calls [seem to pass negative values](https://github.com/alexbatalov/iwd2-re/blob/4fa36bcd6c33e0d35c9e273561aa683aece88490/src/CSound.cpp#L197) if that was correctly extracted.
>        Anyway, we could just provide overrides for the table. One or two mods may actually change it, but if even the default isn't working properly, they'd need to provide GemRB-specific tables anyway. So a hard override should be fine.

Ok, I try the 2DA override approach.

> 2. `SetupNewStream` uses quite different values for rolloff, I wonder if we could unify the approach. Ambients are not perfect yet either IIRC.

That's even more suprising how it has worked so far. Given that you can have one distance model in OAL at the time and ambients  may have range settings. Best explanation seems to be this: https://stackoverflow.com/questions/68726623/java-openal-source-gain-of-zero-after-passing-the-max-distance.

It starts attenuating after ref. distance and stops attenuating after max. distance, leaving the sound at the last non-zero volume then. Roll-off describes the slope. So for what I did here was to have a higher rolloff and high max. distance to push the value towards inaudibility when far away from screen.

The doesn't exactly fit my impression how the ambients actually work (is there some hard distance check somewhere else? Haven't looked for it) but atm I'm not sure how much to unify here.

MarcelHB:
Please look at the 2DA approach.

lynxlynxlynx:
Are you testing with iwd2 or some other game?

MarcelHB:
I've tested IWD2, BG2 (where footsteps are esp. loud) and a bit of PST.

lynxlynxlynx:
They ship different values (iwd2 also has extra columns), so we need to be careful with a single override. BG2 for example already has:

> WALKINGC    25
> WALKINGM    25
> ARMOR       15
So in effect you're making walking sounds louder. IWD2 has them at 70.

MarcelHB:
No, not louder, because maybe the problem is actually something else: BG2's sndchann.2da has 20 channel items whereas we have 24 in the enum. I first copied gemrb/unharcdoded/shared/sndchann.2da and adjusted armor + walksounds down to what I had. It didn't work as expected. Then I added the four missing Char6 to 9 entries and things were good.

So maybe I just fix all these 2DAs?

lynxlynxlynx:
Yeah, `ReadSoundChannelsTable` doesn't do a string lookup, just reads by index. We already use the row name, so it's a simple fix to actually use it for everything. So I think you should fix that first and then recheck if bg2 values need any tweaking at all.

MarcelHB:
Changed the lookup into the name of the 2DA.

lynxlynxlynx:
Perfect, so no extra value tweaking is needed. :)

lynxlynxlynx:
Thanks!

Issue #2188, opened 2024-10-07
 Memory leak somewhere


MarcelHB:
#### Bug description

Playing around in IWD2 AR620x (probably unrelated), I started to notice that the process had reached 14GB of allocated RAM, growing steadily. Need to investigate later.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

MarcelHB:
Hm, various smaller stuff on Valgrind, but not the dimension I observed on Windows earlier. Will continue keeping an eye of this. Yet we should have a look at the things below.

```
==7207== 48 (40 direct, 8 indirect) bytes in 1 blocks are definitely lost in loss record 2,250 of 3,209
==7207==    at 0x4846FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==7207==    by 0x6D6F9AB: GetKitSpell(GemRB::FixedSizeString<8ul, &strncasecmp> const&, std::vector<SpellEntry*, std::allocator<SpellEntry*> >&) (CREImporter.cpp:431)
==7207==    by 0x6D6FC60: InitSpellbook() (CREImporter.cpp:451)
==7207==    by 0x6D6FD9D: GemRB::CREImporter::CREImporter() (CREImporter.cpp:457)
==7207==    by 0x6DB56C3: void std::_Construct<GemRB::CREImporter>(GemRB::CREImporter*) (stl_construct.h:119)
==7207==    by 0x6DB5428: construct<GemRB::CREImporter> (alloc_traits.h:661)
==7207==    by 0x6DB5428: std::_Sp_counted_ptr_inplace<GemRB::CREImporter, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<>(std::allocator<void>) (shared_ptr_base.h:604)
==7207==    by 0x6DB515A: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<GemRB::CREImporter, std::allocator<void>>(GemRB::CREImporter*&, std::_Sp_alloc_shared_tag<std::allocator<void> >) (shared_ptr_base.h:971)
[...]

PR #2189, opened 2024-10-10
Memory leak fixes

MarcelHB:
## Description
Of #2188 I fixed 3 problems by us (visible in different reports), two things seem to be related to STL and are known issues to show false positives, one looks too complicated and asks for some better solution. I'll open a follow-up.

Still no idea about Windows metrics, will do something if I find out what's wrong, if anything at all.

## Checklist

### DIFF
--- a/gemrb/plugins/CREImporter/CREImporter.cpp
+++ b/gemrb/plugins/CREImporter/CREImporter.cpp
@@ -43,16 +43,6 @@ static std::vector<ResRef> snglist; //IE_IWD2_SPELL_SONG
 static std::vector<ResRef> shplist; //IE_IWD2_SPELL_SHAPE
 static const ResRef EmptyResRef;
 
-struct LevelAndKit
-{
-	unsigned int level;
-	unsigned int kit;
-	
-	LevelAndKit(unsigned int level, unsigned int kit) noexcept
-	: level(level), kit(kit)
-	{}
-};
-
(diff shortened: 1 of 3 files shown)


gemrb/plugins/CREImporter/CREImporter.cpp
```diff
@@ -428,6 +428,12 @@ static void GetKitSpell(const ResRef& tableRef, std::vector<SpellEntry*>& list)
 			index = FindSpell(spellRef, splList);
 			assert (index != TableMgr::npos);
 		}
+
+		// IWD2 listdomn gives us dups here
```
lynxlynxlynx, line 432: spwi903 is the only one, at level 8 for Oghmatiques, at level 9 for CTE.
Anyway, wouldn't this way one of them not get the spell? I think the correct thing to do is to rerun an amended loop below if the entry already exists (overwrite values if they are 0). This is basically a data problem, the two adjacent lines should have been one.

MarcelHB: The `SpellEntry` uses a vector for all this, so overwriting using the existing interface won't work. Will look at this tomorrow.

MarcelHB: Minor code cleanup now.

lynxlynxlynx:
Thanks!

Issue #2196, opened 2024-10-16
 Overdrawing or bad blending for some spells or combinations


MarcelHB:
#### Bug description
Not sure if related to #2053. In the severed hand of IWD2, there are some red mages that may be hostile. After using their protection spells, they are way more sunlight than red.

#### Screenshots

![iwd2_overblending](https://github.com/user-attachments/assets/63cbb015-401b-4b5d-b474-cb2aaff3c052)

That's the debug dump:

```
Script name: gb_thayvianmage3
EFFECT QUEUE:
  0: 0x18e: SpellFocus (2, 8) S:SPLFOCUS
  1: 0x12d: ConcentrationModifier (4, 0) S:FEAT09
  2: 0x07: Color:SetPalette (27, 20) S:00STAF98
  3: 0x07: Color:SetPalette (23, 16) S:00STAF98
  4: 0x1ad: EffectsOnStruck (0, 0) S:SPWI228
  5: 0x8e: Icon:Display (0, 129) S:SPWI228
  6: 0x19d: VisualEffectIWD2 (0, 31) S:SPWI228
```

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Hmm, if that's it, then it must be from SPWI228 Death Armor and its use of fx_visual_effect_iwd2.

lynxlynxlynx:
Something's definitely fishy. The entry in overlay.2da has DARMORC, which is a purple plasma of sorts, not an egg.

MarcelHB:
Looking at the candidates, I'd say it's everything at once 😄

lynxlynxlynx:
It's easy to test if it's just the spell or if the staff confounds it.

MarcelHB:
It's just the spell. Easily reproducible.

PR #2197, opened 2024-10-17
Fix against wrong IE_SANCTUARY value

MarcelHB:
See #2196

## Checklist

### DIFF
--- a/gemrb/core/Scriptable/Actor.cpp
+++ b/gemrb/core/Scriptable/Actor.cpp
@@ -2453,6 +2453,11 @@ Actor::stat_t Actor::ClampStat(unsigned int StatIndex, stat_t Value) const
 		return Value;
 	}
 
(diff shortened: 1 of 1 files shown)


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -2453,6 +2453,11 @@ Actor::stat_t Actor::ClampStat(unsigned int StatIndex, stat_t Value) const
 		return Value;
 	}
 
+	// don't clamp, it's a bit field
+	if (StatIndex == IE_SANCTUARY) {
+		return Value;
+	}
+
```
lynxlynxlynx, line 2460: Wait, why is this a problem? Is just the `maximum_values` entry of the stat set too low?

lynxlynxlynx:
Ok, thanks!

PR #2200, opened 2024-10-21
Upgrade path storage

### LINKED ISSUE #880: Replace `Map::PathNode` with an `std::list<Point>`
I'm posting this here both to keep track of it myself and, mainly, to ask if I should wait until `subviews` is merged to do this.

`Map::PathNode` implements a doubly-linked list of (navmap) `Point`s and orientations, but: 
- it's 2020 and programmers don't have to write their own linked lists anymore in any sensible programming language; 
- the new pathfinder generates paths that have a much lower node count, so orientations can be just calculated in `Movable::DoStep` instead of `Map::FindPath`, saving a few function calls - not much of an optimization, but it'll contribute to make FindPath less of a behemoth both in terms of code and assembly;
- right now, `PathNode` doesn't store Points but coordinates, which makes it 4.8% more bothersome to pass it around.
- using the STL for this will also make it harder for memory leaks to appear in the future and simplify the code - one less custom data structure to keep track of.

lynxlynxlynx:
Want some more eyes on these commits, which seem to work fine. So more in the architectural / code sense. The current struct is a minimal implementation for what we were already using.
I also considered moving orientation determination to DoStep, but that was buggy, too much south-facing.

Fixes #880

### DIFF
--- a/gemrb/core/GameScript/Actions.cpp
+++ b/gemrb/core/GameScript/Actions.cpp
@@ -5727,17 +5727,16 @@ void GameScript::RandomWalkContinuous(Scriptable* Sender, Action* /*parameters*/
 	if (actor->BlocksSearchMap()) {
 		area->ClearSearchMapFor(actor);
 	}
-	const auto path = area->RandomWalk(actor->Pos, actor->circleSize, std::max<int>(5, actor->maxWalkDistance), actor);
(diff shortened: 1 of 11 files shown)


MarcelHB (approved):
Look good to me, tested it a bit.

lynxlynxlynx:
Alright, thanks!

PR #2201, opened 2024-10-26
Fix for `nullptr_t` confusion

MarcelHB:
## Description
Not sure why it has worked, but it no longer does on my MSYS2 gcc.

## Checklist


lynxlynxlynx:
Yeah, same here, but only after I started working on other things.

PR #2211, opened 2024-10-31
appveyor: bump image

lynxlynxlynx:
testing + maybe this fixes a msvc bug


lynxlynxlynx:
Eh, PR's don't cache, so since it wants to rebuild the deps, this will never work.

Issue #2176, opened 2024-09-30
 Some bad pixels in TIS graphics


MarcelHB:
#### Bug description
I only started to notice it in AR4100 of IWD2 that there are some black pixels around. Maybe something key color related.

#### Screenshots

![bad_tis_pixels](https://github.com/user-attachments/assets/3e505074-b16d-41b3-a9c1-a9af69055b80)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Looks fine here, also SDL2 GL:
![image](https://github.com/user-attachments/assets/c0f58726-9da7-44fe-9bb5-1b8fbf31a406)

lynxlynxlynx:
But it is visible during dusk and night:
![image](https://github.com/user-attachments/assets/6bc2c862-f683-4ab5-89d3-e32b34233d01)

lynxlynxlynx:
ar4000 also has some suspect bits, though nothing so bad:
![image](https://github.com/user-attachments/assets/0ac74440-297d-4187-860b-c6290fd9865a)

visible already in Targos docks:
![image](https://github.com/user-attachments/assets/62bd4700-3339-4a63-9809-61a10579f100)

So probably something on our end.
Btw, you can `ex('AdvanceTime(1800)')` to advance for 6h.

MarcelHB:
Looks like tinting then.

lynxlynxlynx:
Yeah, these areas don't have night weds.

PR #2227, opened 2024-11-24
Fix for some bad pixels in TIS sprites

MarcelHB:
## Description
See #2176

tis_pixels

## Checklist

- [] The proposed change builds also on our build bots (check after submission)

### DIFF
--- a/gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
+++ b/gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
@@ -178,8 +178,10 @@ void SDLSurfaceSprite2D::EnsureShadedPalette() const noexcept
 void SDLSurfaceSprite2D::ShadePalette(BlitFlags renderflags, const Color* tint) const noexcept
 {
 	Palette::Colors buffer;
+	buffer[0] = format.palette->GetColorAt(0);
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Nice, thanks!

gemrb/plugins/SDLVideo/SDLSurfaceSprite2D.cpp
```diff
@@ -178,8 +178,10 @@ void SDLSurfaceSprite2D::EnsureShadedPalette() const noexcept
 void SDLSurfaceSprite2D::ShadePalette(BlitFlags renderflags, const Color* tint) const noexcept
 {
 	Palette::Colors buffer;
+	buffer[0] = format.palette->GetColorAt(0);
 
-	for (size_t i = 1; i < 256; ++i) {
+	size_t startIndex = format.HasColorKey ? 1 : 0;
+	for (size_t i = startIndex; i < 256; ++i) {
 		buffer[i] = format.palette->GetColorAt(i);
```
Mingun, line 185: I seems, that this can simply be written as 
```cpp
for (size_t i = 0; i < 256; ++i) {
```

MarcelHB: No, the loop performs tinting and stuff. In case of color key, index 0 must not be touched.

PR #2234, opened 2024-12-16
Logger: Flush after write

Bubb13:
## Description
This pull request attempts to fix an issue I have been encountering with console output, where log statements often fail to get flushed to the console until a subsequent log occurs.

Example of the console window when launching an oBG2 game before / after the fix:
**Before:**
before
**After:**
after

## Checklist


lynxlynxlynx:
Could have used Flush, but this fine, thanks!

PR #2236, opened 2024-12-20
iwd2: fix object selector value order

Bubb13:
## Description
IWD2 parses object selectors in differing orders depending on whether it's working with .DLG resources (text) or .BCS resources (compiled). It seems that GemRB is using the dialog order to read .BCS files, which causes incorrect object selector resolution in scripts.

Here's the different orders:
- Dialog:
  ```
  [EA.GENERAL.RACE.SUBRACE.CLASS.SPECIFIC.GENDER.ALIGN.AVCLASS.CLASSMSK]
  ```
- BCS `OB` format:
  ```
  EA GENERAL RACE CLASS SPECIFIC GENDER ALIGNMNT SUBRACE filter1 filter2 filter3 filter4 filter5 [rect1.rect2.rect3.rect4] "scriptname" AVCLASS CLASSMSK
  ```

I've lightly tested the changes, and am opening this pull request as a draft for feedback as to whether the fix method is acceptable. The PR has softcoded the definition of dialog object selector parsing to `script.2da`.

Some uncertainties:
- The second-to-last object selector parameter, `AVCLASS`, is completely unused by IWD2. I've dummied out the relevant filter function to ignore this value.
- IWD2's dialog parser has incorrectly hardcoded `ALIGN` instead of `ALIGNMNT` as to where it pulls symbols from for that value. While the value functionally works off of `ALIGNMNT`, I've altered the listing for dialogs in `script.2da` to pull from the erroneous IDS to maintain compatibility with the original engine.
- IWD2's dialog parser additionally has a bug related to reading the `CLASSMSK` value where it truncates it to a byte. I have **not** replicated that behavior in GemRB.

## Checklist

### DIFF
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -1302,7 +1302,7 @@ static const L* FindLink(StringView name, const L2& names)
 
 static const IDSLink* FindIdentifier(const std::string& idsname)
 {
-	int len = static_cast<int>(idsname.size());
+	int len = static_cast<int>(idsname.size() + 1);
 	for (int i = 0; idsnames[i].Name; i++) {
 		if (!strnicmp(idsnames[i].Name, idsname.c_str(), len)) {
 			return idsnames + i;
--- a/gemrb/core/GameScript/GameScript.cpp
+++ b/gemrb/core/GameScript/GameScript.cpp
@@ -1691,6 +1691,44 @@ static void InitializeObjectIDS(const AutoTable& objNameTable)
 	ObjectFieldsCount = ObjectIDSCount - ExtraParametersCount;
 }
 
+// IWD2 parses object selectors in differing orders depending on whether it's working with .DLG resources (text) or .BCS resources (compiled)!
+//   Dialog: [EA.GENERAL.RACE.SUBRACE.CLASS.SPECIFIC.GENDER.ALIGN.AVCLASS.CLASSMSK]
+//   BCS: EA GENERAL RACE CLASS SPECIFIC GENDER ALIGNMNT SUBRACE filter1 filter2 filter3 filter4 filter5 [rect1.rect2.rect3.rect4] "scriptname" AVCLASS CLASSMSK
+static void InitializeDialogObjectIDS(AutoTable& objNameTable)
+{
+	TableMgr::index_t idsRow = objNameTable->GetRowIndex("DIALOG_IDS_COUNT");
+
+	if (idsRow != TableMgr::npos) {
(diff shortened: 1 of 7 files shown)


gemrb/core/GameScript/GameScript.cpp
```diff
@@ -1302,9 +1302,8 @@ static const L* FindLink(StringView name, const L2& names)
 
 static const IDSLink* FindIdentifier(const std::string& idsname)
 {
-	int len = static_cast<int>(idsname.size());
 	for (int i = 0; idsnames[i].Name; i++) {
-		if (!strnicmp(idsnames[i].Name, idsname.c_str(), len)) {
+		if (!stricmp(idsnames[i].Name, idsname.c_str())) {
```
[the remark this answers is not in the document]
Bubb13: The idea here is that the `strnicmp` call ends up matching `idsname.c_str()` against the **beginning** of `idsnames[i].Name`, which under the pre-PR `idsnames` array caused the `class20` entry to match `class`. I've altered the fix to simply add one to the check length instead.

gemrb/core/GameScript/GameScript.cpp
```diff
@@ -1659,6 +1659,49 @@ static void SetupSavedTriggers()
...
+		const IDSLink* poi = FindIdentifier(idsName.c_str());
+		if (poi == nullptr) {
+			idtargets[i] = nullptr;
+		} else {
+			idtargets[i] = poi->Function;
+		}
+		ObjectIDSTableNames[i] = ResRef(idsName);
+	}
+
+	MaxObjectNesting = ReadNamedRowColumnIndexSigned(objNameTable, "MAX_OBJECT_NESTING", 0);
```
lynxlynxlynx, line 1694: Why not just pass the column by name ("DATA")? That way it looks like you could avoid the extra functions.

lynxlynxlynx:
Other than that it looks fine. Thanks for splitting it up!

gemrb/core/GameScript/GameScript.cpp
Bubb13: Good call - I have a bit of a tendency towards micro-optimization. I've replaced these calls with standard named lookups.

lynxlynxlynx: We actually prefer string-based table lookups, since then it's much easier to change the structure of 2das (or data type) if needed. It will be relevant again for some EE tables like clasiskl.2da (duplicates part of our classes.2da/clskills.2da), since we'll want to unify the approach / layouts.

lynxlynxlynx:
Thanks!

Issue #2229, opened 2024-12-09
 Grey box around some spells


DarthRevan501st:
#### Bug description

So when I cast some spells they have grey boxes around them.

#### Steps to reproduce
1. Use a fresh copy of GemRB with BG1 from GOG( Obviously OG and not EE.) on Linux or Windows( Not sure about other platforms.)
2. The only change I made to the config file was my game path.
3. Start a new game, click the mission pack, and load the default mission pack save.
4. When you get sent to the island just start attacking enemies with your full party.

#### Expected behavior
When certian members of the party attack you will see the grey box in the picture below around some spells.

#### Screenshots
![Capture](https://github.com/user-attachments/assets/64b1294a-2134-4a97-a70c-0e5eaad2f520)

GemRB version (check as many as you know apply)
- [ ] master as of this issue
- [x] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1
Not sure about older versions as I haven't used them.

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [ ] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Thanks, this might be the same thing as #2210. Can reproduce on master, but not with wolves in ar1009 (what would cause the dispel?), I get it when getting spit on by the ankhegs in the center and left of the area.

lynxlynxlynx:
Yeah, it's fixed now.

DarthRevan501st:
> Yeah, it's fixed now.

Hey thanks! I look forward to the next release.

Issue #2210, opened 2024-10-31
 Transparency issues on some mirrored sprites


MarcelHB:
#### Bug description
Looks as if there is some coloring/transparency problem when sprites are mirrored to the right half of the clock. Will have a look at it.

#### Screenshots

![sprites_green](https://github.com/user-attachments/assets/9b72e4e1-2dcc-4c51-a331-b9529703f2e1)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
I noticed this once or twice, but couldn't pin it down. Eg. only once with the wand of lightning, which I've used maybe 20 times overall.

MarcelHB:
Probably one of the more annoying bugs. So far I was 1. not able to fix the bug through things I had in mind or 2. fix the bug at the expense of other bugs/performance issues I did not understand so far. Will continue to look for solutions.

PR #2240, opened 2024-12-24
Fixes for some transparency issues

MarcelHB:
## Description

Rarely have so little code lines kept me so busy. Little surprise since only some digging into SDL code made things a little more clear. See #2210 and maybe #2229, but I couldn't reproduce the latter in the same way as described in the issue.

## Checklist


lynxlynxlynx:
Cool, thanks, I'll give the other bug a retry after merging.

Issue #2241, opened 2024-12-24
Check ambient rolloff works


lynxlynxlynx:
It seems we currently just cut off ambients:
https://www.gibberlings3.net/forums/topic/38983-setup-ambient-sounds/#comment-344795

So we need to:
- [x] check how the originals behaved (IOW do we need a game flag or can we just improve?)
- [x] add/fix our rolloff regardless
- [x] check the attached wav problem
- [x] check the attached ogg problem

MarcelHB:
There was always a cut off for ambients as far as I remember. I smoothed the rolloff somewhat, but yes, that's not optimal yet and requires to work with some limitations of OpenAL.

OGG has no stereo-spatial workaround yet. Not sure what's wrong with WAV file.

MarcelHB:
In PST and BG2 original, it looks like a linear roll-off, with the radius roughly reflecting the pixels away from the camera center until it goes silent, i. e. something at the horizontal edge with radius 400px at 1024x768 is almost inaudible. So unless somebody spots a significant difference, one model should fit them all.

I'll review what we are doing with the params for OAL here.

lynxlynxlynx:
IIRC OpenAL uses a non-linear model, something we opt-in with a setting. Doesn't appear our parameters give a gradual enough function though.

MarcelHB:
While testing my settings, I found another bug: _Stereo_ ambients, like the crystal sound in AR1202 in PST never get their spatial flag set.

MarcelHB:
One side-question because it kept me busy for some time. Is this is a good idea? https://github.com/gemrb/gemrb/blob/757558d4ce5507ac07046d889deae26f7d3b0a6a/gemrb/core/ResourceManager.cpp#L189-L190

in contrast to swapping these two lines? I was trying to load an OGG file for replacing an existing ambient file (WAV) by putting it into `override`. Problem is now that some other reader associated to the current `type` comes first (`ACMReader`), and will find some suitable WAV file in the `chitin.key`. Even renaming the OGG file to `.wav` and having precedence didn't help much since `ACMReader` refuses to ingest this correctly and again ends up in the shipped version.

lynxlynxlynx:
I think we can swap them without much worry. I suspect it might cause more disk lookups, but I doubt that's significant.

MarcelHB:
Ok, but I'll leave like this for now, I don't want to introduce unforeseen side effects just for some testing shortcuts.

PR #2244, opened 2024-12-25
Some more sound improvements and fixes

MarcelHB:
## Description
Essentially everything sound-related discussed in #2241.

Regarding the volume roll-off of ambients: Good and easy candidates to compare are the crystal in AR1202 of PST, or the buzzing machine very close to where BG2 starts in the dungeon.

It's not that easy and good to directly try to reflect the original since the sound model there is primitive, i. e.  no spatial magic, but more to somewhat get volume levels roughly between the loudest point (center) and the edge before the ambient mgr stops playback due to radius. Can't do too steep roll-off since that would go low too early.

So please have a test and feel free to suggest different params.

## Checklist

### DIFF
--- a/gemrb/core/AmbientMgr.cpp
+++ b/gemrb/core/AmbientMgr.cpp
@@ -314,7 +314,7 @@ tick_t AmbientMgr::AmbientSource::Enqueue() const
 {
 	if (stream < 0) return -1;
 	// print("Playing ambient %s, %d/%ld on stream %d", ambient->sounds[nextref], nextref, ambient->sounds.size(), stream);
-	return core->GetAudioDrv()->QueueAmbient(stream, ambient->sounds[nextref]);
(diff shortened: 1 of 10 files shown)


lynxlynxlynx:
Looks good, thanks, except for the minor details that you left an unconditional Log line in OGGReader::ReadSamplesIntoChannels. 

Will try later, currently stuck in wallgroup debugging.

MarcelHB:
You are right, one of my activity trace makers left over. Cleaned up.

lynxlynxlynx:
Sounds fine as well, that bg2 area top ambient is a really nice test, which also makes it clear stereo is working. 👍🏾 

bg1 start tavern ambient is super low volume, even at max setting. It has 30% set, so a data fault — it's the same in the original.

Also noticed iwd2 ar6000 undead footstep sounds are hilariously bad, but that is not new, nothing changed.

Issue #2247, opened 2024-12-27
 GUIMG: Cannot remove spell


MarcelHB:
#### Bug description
When trying to remove a mage spell from the book, after confirming, there is a Python error saying:

```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "D:/Sources\gemrb\gemrb\GUIScripts/GUIMG.py", line 304, in RemoveSpell
[Python/ERROR]:     OnMageRemoveSpell(btn)
[Python/ERROR]:   File "D:/Sources\gemrb\gemrb\GUIScripts/GUIMG.py", line 372, in OnMageRemoveSpell
[Python/ERROR]:     GemRB.RemoveSpell (pc, spelltype, level, index)
[Python/ERROR]: TypeError: 'NoneType' object cannot be interpreted as an integer
```

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

lynxlynxlynx:
Ah, cool, so this isn't limited to ees. index is None there, the lambda expecting to automatically get a button. Should be trivial to fix.

PR #2249, opened 2024-12-27
#2247 Update GUIMG.py when attempting to remove a spell from the mage spellbook, issue caused passing 'none' value as spell

Mariihmp:
Fix: pass correct spell index, when removing spells from mage spellbook

## Checklist


lynxlynxlynx:
Thanks; like Sonar says, you have a typo that breaks this code further instead of fixing it.

Since RemoveSpell always runs on the same button, we indeed can't use its value. But that also means that we can remove the btn parameter to the two functions involved.

lynxlynxlynx:
Hey @Mariihmp , do you plan to finish the PR?

lynxlynxlynx:
Alright, fixed this jumble and manually merged.

PR #2252, opened 2024-12-29
OGG & OpenAL changes

### LINKED ISSUE #2251: Error when ambient sound is switched from a 2-channel source to a 1-channel source
Ambients support several resources and play them in a queue. If you use two different files as sources for the ambient, one with 2 channels and one with a single channel, then when one is finished and the next one is started, the console displays:

> Unable to queue buffer: 0xa004 - Invalid Operation

Only going from 2 to 1 causes problems, the other way is fine.

Test case here:
https://www.gibberlings3.net/forums/topic/38983-setup-ambient-sounds/#findComment-344859

MarcelHB:
## Description
I did three things here:

1. OAL: restricted increased source gain to local ambients, globals were measurably too loud this way.
2. OAL: some little refactoring
3. OGG: changed the way samples are counted to match the WAV reading, fixes #2251 

## Checklist

### DIFF
--- a/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
+++ b/gemrb/plugins/OpenALAudio/OpenALAudio.cpp
@@ -865,7 +865,7 @@ int OpenALAudioDriver::SetupNewStream(int x, int y, int z,
 
 	alSourcef(source, AL_PITCH, 1.0f);
 	alSourcei(source, AL_LOOPING, 0);
-	alSourcef(source, AL_GAIN, 0.03f * gain);
+	alSourcef(source, AL_GAIN, 0.01f * gain);
 	alSourcei(source, AL_REFERENCE_DISTANCE, REFERENCE_DISTANCE);
 	alSourcei(source, AL_ROLLOFF_FACTOR, 0);
 	alSourcei(source, AL_SOURCE_RELATIVE, !point);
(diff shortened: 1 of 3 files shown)


MarcelHB:
To satisfy Sonar, we have to refactor `read_samples` and surroundings.

gemrb/plugins/OpenALAudio/OpenALAudio.cpp
```diff
@@ -504,12 +502,13 @@ std::pair<ALuint, ALuint> OpenALAudioDriver::loadSound(StringView ResRef, tick_t
 		alBufferData(buffers[0], format, channel1.data(), actualSamples * 2, samplerate);
 		alBufferData(buffers[1], format, channel2.data(), actualSamples * 2, samplerate);
 	} else {
-		short* memory = (short*) malloc(totalBytesPerChannel);
-		//multiply always with 2 because it is in 16 bits
-		unsigned int cnt1 = acm->read_samples(memory, numSamples) * 2;
-		//it is always reading the stuff into 16 bits
-		alBufferData(buffers[0], GetFormatEnum(channels, 16), memory, cnt1, samplerate);
-		free(memory);
+		std::vector<char> memory;
```
lynxlynxlynx, line 505: Why not just use a char array? You don't use the extra functionality of a vector after all and it would remove the need for the reinterpret_cast.

MarcelHB: You mean `std::array`? That needs the size at compile time. And to my knowledge, var-length arrays only work when `new`'d, and then need to be deleted explicitly. I just try to keep unmanaged cases low.

lynxlynxlynx: I meant a plain one, but yes, that would require manual memory management.
How about just using vectors of shorts and adjusting the size?

lynxlynxlynx:
Thanks, I'll just merge — PR builds can't update the cache, so rerunning the appveyor build wouldn't do anything.

PR #2274, opened 2025-01-17
GUIScript: GemRB_ExecuteString() can use object under cursor as script context

Bubb13:
## Description
This pull request allows `ExecuteString()` to use the object under the cursor as the script context, which is a pretty nifty feature lifted from the EE's `Eval()` debug console command.

## Checklist


lynxlynxlynx:
Thanks! Those sorting commits are also fine btw, feel free to open a separate PR.

Issue #2276, opened 2025-01-18
MVEPlayer: audio underrun in Black Isle movie


MarcelHB:
#### Bug description
While debugging progress for #2264, I noticed that the BG2 videos of Black Isle and WotC have their sound start to crackle badly  at some point. Other videos look good, as far as I have looked at them.

The reason is that there is a buffer underrun, i. e. the playback source turns into `AL_STOPPED` because there is nothing left to play, and needs to be revived by `alSourcePlay` on the next chunk. On the OAL backend, this causes broken sound.

I suspect that the reason is somewhere in the MVE decoder or player not supplying all audio data, or not fast enough, since neither libavcodec players nor NI suffer from the same.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

MarcelHB:
Looking at this (NI): https://github.com/Argent77/NearInfinity/blob/0bee49d8363e90c07039e349434d6d5e20553f3a/src/org/infinity/resource/video/MveAudioDecoder.java#L193-L204

vs. GemRB:
```cpp
	if (stream_mask & MVE_DEFAULT_AUDIO_STREAM) {
		// ...
	} else {
		/* alternative audio stream, which we don't care about */
	}
```
Could be worth an investigation, since that "don't care" branch gets hit frequently.

MarcelHB:
That's not the problem.

MarcelHB:
Frame waiting in `MVEPlayer::next_frame` isn't good: If the frame wait time is reduced by the decoding time and some additional margin, the problem is gone, but that's some playing around on this machine and possibly not the solution we want, so maybe look at the other players to see the what's fitting.

MarcelHB:
Major contribution to this on Windows: `sleep_for(x)` over microseconds is like x µs + ~14ms. ... Will try a workaround for this one.

Issue #2264, opened 2025-01-05
Better sound profile setup for different use cases


MarcelHB:
The issues in the past and present around the soundscape are keeping us busy with playing around both in actual settings in the backend but also looking for flags and params somewhere across the code. Some things suboptimal:

* Distance cut offs are a bad hack between external managers (`AmbientMgr`) and the backend model, maybe not working well at all
* Three places alone in the OpenAL backend to tackle source settings.
* Params from the user side need to know too much about the backend technicalities (`REFERENCE_DISTANCE`)
* Variety of params in functions in `Audio.h` (e. g. `Point` vs. `x, y, z`, `spatial` vs. `point`)

So similar to what we now have by `PixelFormat`, I propose to switch over to something like `SourceConfig`, offering a bunch of (param'ed) presets, like for music, game voices, dialog voices, static ambients (fixed range) and whatever sounds we want to use, e. g. to distinguish footsteps from spell effects. So roughly, it covers the things like

* spatial settings, EFX settings
* setting/channel name to use for volume multiplier
* max. audible distance in terms of pixels (ambients) or otherwise resolution-relative units (more on the screen, more to listen to, and close by ...)

This should get us some cleanup in the interfaces and backends, and if we ever come up with something better, I hope it keeps the refactoring low.

I'd even propose to switch the OAL model to something linear because 1. everything else is actually hard to calculate, and even harder to test in all kinds of environments and 2. we should get a free cut-off by this, even if there is no manager around.

lynxlynxlynx:
Sounds good, I like the idea of a simplified interface via presets.

PR #2277, opened 2025-01-23
Sound refactoring

MarcelHB:
## Description

Everything was nice and beautiful and even lead to negative code count, until I started noticing the details, and current bugs of our SDLAudio that asked for some creativity. 

So the general idea here is:
* Try to the make the backends slimmer, e. g. by moving away/deduping things like cache, settings and music playback.
* Cleanup of interfaces and wrapping everything into two types of sources, and data buffers, and have new classes do the resource management (like `Audio/Playback.cpp` or `Audio/MusicLoop.cpp`).
* Sound presets to have all kinds of settings and features independent of the backend, see #2264 
* Linear roll-off everywhere, so both backends sound similar spatially and we can much better calculate and play with settings.

Some todos:

- [x] (SDL) Fixing a segfault when terminating the game, it's a race over the music buffer callback with their thread
- [x] (SDL) Sometimes a crash between videos, something lock in rt lib.
- [x] (SDL) Something doesn't work out together with AmbientMgr yet
- [x] (SDL) Looping ambients ignore last volume setting on next turn
- [x] AmbientMgr has some time mgmt problem, minor gap between loops
- [x] (SDL) Quitting the game to the main menu may keep ambients playing
- [x] (SDL) Volume seems to jump sometimes when switching to UI and back (related to above)
- [x] (SDL) Music slider in settings seems to have no effect.
- [x] Small but audible gap between some music playlist items.
- [x] Recheck all the settings menus.
- [x] Evaulate both backends on slower machines for unnoticed performance impacts.
- [x] More testing of course, Valgrind runs.
- [x] Sonar, code formatting maybe; currently not working on all of my platforms.

But so far I'm quite happy how well it works. So please have a first impression and feel free to test. I apologize for one big commit by I was reiterating over the whole thing multiple times.

## Checklist


gemrb/core/Audio/AmbientMgr.cpp
```diff
@@ -309,12 +302,37 @@ tick_t AmbientMgr::AmbientSource::Tick(tick_t ticks, Point listener, ieDword tim
...
 
+BufferCacheEntry AmbientMgr::AmbientSource::GetBuffer(ResRef resource, const AudioPlaybackConfig& config)
+{
+	auto entry = bufferCache.get().Lookup(resource);
+	if (entry) {
+		return *entry;
+	}
+
+	ResourceHolder<SoundMgr> acm = gamedata->GetResourceHolder<SoundMgr>(resource);
+	auto length = ((acm->get_length() / acm->get_channels()) * 1000) / acm->get_samplerate();
```
lynxlynxlynx, line 313: Is there any chance for this to SIGFPE?

MarcelHB: Not sure, I have taken this from the previous code. Will check.

gemrb/core/Audio/Backend.h
```diff
@@ -0,0 +1,103 @@
...
+ * GNU General Public License for more details.
+
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#ifndef H_AUDIO_BACKEND
+#define H_AUDIO_BACKEND
```
lynxlynxlynx, line 21: I propose the filenames be renamed to AudioBackend.* as well, not just to match the class and guard, but to be less generic.

gemrb/core/Audio/MusicLoop.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+void MusicLoop::Load(ResourceHolder<SoundMgr> music)
+{
+	std::lock_guard<std::recursive_mutex> l(mutex);
+
+	currentFormat.bits = 16;
+	currentFormat.channels = 2; // get_channels may report garbage
+	currentFormat.sampleRate = music->get_samplerate();
+
+	loadBuffer.resize(currentFormat.GetNumBytesForMs(125) / 2);
```
lynxlynxlynx, line 55: Is there any particular significance to 125? I suggest you add a comment why we use 1/8s, so it will be easier to reason about if we ever need to revisit this.

gemrb/core/Audio/Playback.cpp
```diff
@@ -0,0 +1,184 @@
...
+	if (cacheEntry) {
+		return *cacheEntry;
+	}
+
+	ResourceHolder<SoundMgr> acm = gamedata->GetResourceHolder<SoundMgr>(resource);
+	if (!acm) {
+		return {};
+	}
+
+	auto length = ((acm->get_length() / acm->get_channels()) * 1000) / acm->get_samplerate();
```
lynxlynxlynx, line 161: It's the same here, perhaps it's worth it to create another function in SoundMgr.

gemrb/core/Audio/MusicLoop.cpp
MarcelHB: By a recent talk by Valve 😄 It was mentioned how they manage sound, such that 125ms was seen as a balance between a decent memory buffer amount that is usually long enough to cover a high percentile of I/O latency spikes on low machines without interruption. And since I was out of competing models ...

Unless an Raspberry Pi starts to struggle here, I'd commit it like "educated guess".

gemrb/core/Audio/Backend.h
lynxlynxlynx: Same for Settings.h.

gemrb/core/Audio/MusicLoop.cpp
lynxlynxlynx: Would be fine as a comment, it being somewhat of an industry standard is very good justification.

gemrb/core/GameScript/Actions.cpp
```diff
@@ -2625,7 +2633,7 @@ void GameScript::ToggleDoor(Scriptable* Sender, Action* parameters)
 			door->AddTrigger(TriggerEntry(trigger_failedtoopen, actor->GetGlobalID()));
 
 			//playsound unsuccessful opening of door
-			core->PlaySound(door->IsOpen() ? DS_CLOSE_FAIL : DS_OPEN_FAIL, SFXChannel::Actions, *p, GEM_SND_SPATIAL);
+			core->GetAudioPlayback().PlayDefaultSound(door->IsOpen() ? DS_CLOSE_FAIL : DS_OPEN_FAIL, SFXChannel::Actions);
```
lynxlynxlynx, line 2636: This should remain spatial, unfortunately not all default sounds are the same. For example the next one about xp shouldn't be.

gemrb/core/GameScript/GSUtils.cpp
```diff
@@ -542,21 +540,9 @@ void DisplayStringCore(Scriptable* const Sender, ieStrRef Strref, int flags, con
...
-		}
 
 		Actor* actor = Scriptable::As<Actor>(Sender);
-		// Spatial unless PC, cutscene or dialog
-		if (actor && !actor->InParty && !core->InCutSceneMode() && !core->GetGameControl()->InDialog()) {
-			pos = Sender->Pos;
-			soundFlags |= GEM_SND_SPATIAL;
-		}
-
-		if (flags & DS_QUEUE) soundFlags |= GEM_SND_QUEUE;
```
lynxlynxlynx, line 558: To anyone else reviewing: we don't pass DS_QUEUE through any caller.

gemrb/core/Interface.cpp
```diff
@@ -913,15 +916,15 @@ bool Interface::ReadDamageTypeTable()
...
 
-bool Interface::ReadSoundChannelsTable() const
+bool Interface::ReadSoundChannelsTable()
 {
 	AutoTable tm = gamedata->LoadTable("sndchann");
 	if (!tm) {
 		return false;
 	}
 
+	// reverb is currently ignored
```
lynxlynxlynx, line 926: any plans with this? I know we currently just stored it.
Btw, at least some of the games also provided sndenvrn.2da and sndresrf.2da, which look like a subset of our reverbs.2da.

gemrb/core/Interface.cpp
```diff
@@ -2116,9 +2141,9 @@ int Interface::PlayMovie(const ResRef& movieRef)
 		}
 	}
 
-	Holder<SoundHandle> sound_override;
+	Holder<PlaybackHandle> soundOverride;
 	if (!sound_resref.empty()) {
-		sound_override = AudioDriver->Play(sound_resref, SFXChannel::Narrator);
+		soundOverride = GetAudioPlayback().Play(sound_resref, audioSettings.ConfigPresetDialog(SFXChannel::Narrator));
```
lynxlynxlynx, line 2146: Shouldn't this use ConfigPresetMovie?

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -3722,34 +3723,42 @@ sound as if it was said by that PC (EAX).\n\
...
-		if (channel_name) {
-			channel = core->GetAudioDrv()->GetChannel(channel_name);
+		if (channelName) {
+			channel = AudioSettings::GetChannelByName(StringView { channelName });
 		}
 
-		if (pyref == Py_None) {
-			core->GetAudioDrv()->Play("", channel, pos, flags);
+		auto config = core->GetAudioSettings().ConfigPresetScreenAction(channel);
+		if (pos.x != 0 || pos.y != 0) {
```
lynxlynxlynx, line 3749: You can use the cleaner `!p.IsZero()`.

gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -7873,7 +7882,12 @@ static PyObject* GemRB_UpdateVolume(PyObject* /*self*/, PyObject* args)
 {
 	int type = 3;
 	PARSE_ARGS(args, "i", &type);
-	core->GetAudioDrv()->UpdateVolume(type);
+	if (type == 1 || type == 3) {
+		core->GetMusicLoop().UpdateVolume();
+	}
+	if (type == 2 || type == 3) {
+		core->GetAmbientManager().UpdateVolume();
+	}
```
lynxlynxlynx, line 7890: You could use bit comparisons here to simplify. But this made me see a bug: the parameter is not optional any more, so we either need to fix the doc or use `"|i"`

lynxlynxlynx:
Wow, it turned it into quite an effort! I guess it also fixes #2276?

Architecturally it looks sound, pun intended, I'm just not satisfied with Play ergonomics:
`core->GetAudioPlayback().Play(sound, core->GetAudioSettings().ConfigPresetEnvVoice(channel));`
It's much more verbose than before and stepping through will be more annoying. How about we add an overload, so one could also use it, for example, like:
`core->GetAudioPlayback().Play(sound, SoundPreset::VoiceGlobal, channel);`
It might also remove the need for some includes.

gemrb/core/Interface.cpp
MarcelHB: Not right now. Also there is other values around somewhere (I noticed dampening) but I couldn't make a model out of it within a few seconds, so maybe at some later point.

gemrb/core/GameScript/Actions.cpp
MarcelHB: I will think of something.

gemrb/core/Interface.cpp
MarcelHB: Good idea.

MarcelHB:
Regarding `Play` & co, yeah. Hm, I'm thinking of something with symbols mapping to the calls, or templated calls maybe since the arguments vary a little.

gemrb/core/GameScript/Actions.cpp
lynxlynxlynx: For some context, these are sounds that were hardcoded in the originals. Some are GUI related like the XP etc., but not all. The GUI ones should be global, things like opening and closing sounds spatial.

lynxlynxlynx:
Btw, why is it called ConfigPreset **EnvVoice**? SpatialVoice is immediately clear to me and this is just the non-spatial variant. Maybe GlobalVoice would be better, to avoid misuse? We almost need a glossary.

MarcelHB:
So far:

* Spatial voice: voice is located on the map and subject to EFX, like existence sounds.
* Env voice: voice is on screen but subject to EFX, mostly selection sounds of the PCs
* Dialog voice: nothing of that

lynxlynxlynx:
In GemRB_UpdateVolume the `type & 3` should be `type & 1` or 2 updates both volumes.

lynxlynxlynx:
That's better, thanks.

gemrb/core/Audio/BufferCache.h
```diff
@@ -0,0 +1,51 @@
...
+#define H_AUDIO_BUFFER_CACHE
+
+#include "AudioBackend.h"
+#include "LRUCache.h"
+
+namespace GemRB {
+
+struct BufferCacheEntry {
+	Holder<SoundBufferHandle> handle;
+	time_t length;
```
lynxlynxlynx, line 30: Please default init it.

gemrb/core/Audio/Playback.h
```diff
@@ -0,0 +1,77 @@
...
+	time_t GetLengthMs() const;
+	bool IsPlaying() const;
+	void SetPosition(const Point& p);
+	void Stop();
+	void StopLooping();
+
+private:
+	Holder<SoundSourceHandle> source;
+	time_t length;
+	int32_t height;
```
lynxlynxlynx, line 49: same here

gemrb/core/Audio/BufferCache.h
```diff
@@ -0,0 +1,51 @@
...
+
+	BufferCacheEntry() {};
+	BufferCacheEntry(Holder<SoundBufferHandle> handle, time_t length)
+		: handle(std::move(handle)), length(length)
```
lynxlynxlynx, line 34: there's probably already a sonar warning for it, but can this be made `explicit`? Same for some in PlayBack.h and AudioPoint.

MarcelHB:
I have discovered some WTF around SDLAudio: https://github.com/gemrb/gemrb/blob/d74d18b80cadd14d2b536a1ce7c4fce7c31b816c/gemrb/plugins/SDLAudio/SDLAudio.cpp#L38-L44

It started to noticed that one of the Windows or Linux builds was always wrong – mutually exclusive, i. e. moving sound to the opposite channel direction of the viewport movement.

I simplified that piece of code a little and made sure that `angle` is 0 when to the north, 90 when to the east etc. (docs: https://wiki.libsdl.org/SDL2_mixer/Mix_SetPosition).

My current observation is that the mixer appears to interpret the angle different, depending on the platform. So what is now formally correct, so I hope, sounds wrong on my Linux lib of the same version. Funnily, aside from left-right-confusion, Linux sounds much smoother spatially than Windows on SDL where it performs slighly crackling jumps when moving around a sound source.

MarcelHB:
In the SDL mixer code, I found some suspicious things around the number of channels https://github.com/libsdl-org/SDL_mixer/blob/a37e09f85d321a13dfcf0d4432827ee09beeb623/src/effect_position.c#L2100-L2112

On Windows, my hardware returns 6 channels (5.1) for SDL. I later have to check what is there on Linux, and if this needs some extra treatment.

MarcelHB:
So if somebody has headphones virtualization on (5+1 channels) with SDL, the mixer will do something different re. positioning than on stereo, so it was reproducible and allows some workaround.

All in all this SDLAudio backend was somewhat painful and worse than anticipated, yet it was also buggy for me before, mostly for strangely distorted music. Anyway, I had to spent more time in the mixer's code than the docs to figure out how and why things (don't) work.

From my perspective I have fixed everything I noticed during testing, and tested everything I suspected to be affected by these changes. So feel free to have an ear into this, although the only noticable difference is supposed to be the linear roll-off in OpenAL now.

Btw. I don't exactly get why the Github clang-format action sometimes does not like the code for some details, as I have prepared it with clang-format locally.

lynxlynxlynx:
IIRC the SDLAudio backend was initially added for the android port as a more lightweight alternative. Now that it is closer to feature parity, I can't imagine there being much of a performance difference. But it's still useful for portability purposes and in fact the mac builds currently rely on it.

Re clang-format: perhaps you have a different version? I think it only enforces a minimum.

I'll give it a test soon, still in the middle of action bar customization.

MarcelHB:
After 2h of playing PST on OAL, I only noticed a jump/gap when transitioning between playlist items any more. Can hopefully fix it tomorrow.

MarcelHB:
Whew, no big deal just some oversight. And another detour to the mixer code.

lynxlynxlynx:
Sounds fine for what I tried. :) I would decrease the walk sound range though, it sounds too far from my testing. How did you get the 3 in `config.muteDistance = width * 3;`?

MarcelHB:
> Sounds fine for what I tried. :) I would decrease the walk sound range though, it sounds too far from my testing. How did you get the 3 in `config.muteDistance = width * 3;`?

The original idea as a bit like [They're coming](https://www.youtube.com/watch?v=kyevhryWKHk) in LOTR, purely artistical and due to a lack of better reference :)

But yeah, in my first iteration I had half-width screen factors. And then I didn't notice it much since PST doesn't have any walk sounds, yet some orc hordes in IWD2 in otherwise ambient-less maps are a good way to test if you like this. I've fixed it that it it fades in closer, but still maybe a little off-screen.

lynxlynxlynx:
Will push manually, since I have it locally already. Thanks again!

Issue #2310, opened 2025-02-01
PST: No extra proficiency for TNO


MarcelHB:
#### Bug description
TNO is a fighter for me r/n, and Porphiron in AR0100 (and some others) should grant weapon class proficiencies, and for every few levels, IIRC, TNO earns such an extra proficieny point to spend – but I don't know the exact mechanism.

So there is this dialog check `ExtraProficiencyGt(Protagonist,0)` that my TNO has never fulfilled for a single time until level 9. A quick look at the scripts don't show any uses of `IncrementExtraProficiency`, and a quick search through our code doesn't show any calls to these methods by hardcoding, so probably the LU lacks this?

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

MarcelHB:
Ah, thanks. So when I look at the bytes in NI, some of the "unused proficiency" bytes should ne non-zero, 0x2 by now if I interpret the `charprof.2da` correctly. But aside from these initial proficiencies, nothing is there:

![Image](https://github.com/user-attachments/assets/9dc1cd54-2d04-4ed1-9e49-02456a6cf5b3)

lynxlynxlynx:
Best to just step through the level up process, it's probably something simple.

MarcelHB:
Looks like `CREImporter` never writes this stat (`IE_FREESLOTS`).

Issue #2288, opened 2025-01-27
PST: tooltip delay setting only knows on/off


MarcelHB:
#### Bug description
When I adjust PST's tooltip setting, there is only two steps r/n:

1. left-most slider setting: show instantly
2. everything else: never? I haven't discovered any tooltips even with just a little bit below instantly

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

lynxlynxlynx:
Immediately looks suspicious:
```py
gemrb/GUIScripts/GUIOPT.py:327: GemRB.SetTooltipDelay (GemRB.GetVar ("Tooltips") * TOOLTIP_DELAY_FACTOR//10)
gemrb/GUIScripts/pst/GUIOPT.py:215:     GemRB.SetTooltipDelay (GemRB.GetVar ("Tooltips") )
```

Issue #2287, opened 2025-01-27
PST: very high precison clock tooltip


MarcelHB:
#### Bug description

Good to know that ...

![Image](https://github.com/user-attachments/assets/b8bd1fce-1306-4457-bb7c-255378c27566)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

Issue #2281, opened 2025-01-26
PST: wrong values or displaying of bonuses


MarcelHB:
#### Bug description

Right after starting the game, open Morte's GUIREC right after the beginning of PST. The bonuses (armor + attributes) are said to be 65534.

#### Screenshots

![Image](https://github.com/user-attachments/assets/4eeb74f0-3f73-4acb-957e-619c7407c6bb)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

lynxlynxlynx:
That's clearly just an underflow and we should display the negative value.

Warphi:
Hey, I can take this on.

lynxlynxlynx:
Cool! You can start by looking where we display the text in `gemrb/GUIScripts/pst/GUIREC.py`. Is it bad already when read into python etc.

PR #2323, opened 2025-02-04
PST: Some fixes

MarcelHB:
## Description

See

* #2281
* #2287
* #2288
* #2310 

## Checklist

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -435,9 +435,7 @@ Interface::Interface(CoreSettings&& cfg)
 	// the originals used double ticks for haste handling
 	Time.ticksPerSec = maxRefreshRate / 2;
 
-	// set up the tooltip delay which we store in milliseconds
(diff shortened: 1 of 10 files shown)


gemrb/core/Interface.cpp
```diff
@@ -4135,4 +4133,22 @@ float Interface::GetAnimationFPS(const ResRef& anim) const
 	return ANI_DEFAULT_FRAMERATE;
 }
 
+void Interface::ApplyTooltipDelay() const
+{
+	ieDword tooltipDelay = vars.Get("Tooltips", 0);
+
+	// PST: [0, 119] -> [0, 2500]
+	// everything else: [0, 10] -> [0, 2500]
+	// with the highest value meaning "never"
```
lynxlynxlynx, line 4142: There's just one thing I'm not sure about — wasn't the factor of 10 used to match what is stored in the ini? Eg. in my icewind.ini Tooltips=30.

MarcelHB: I see ... will change it.

PR #2331, opened 2025-02-16
Fix detection of RaspberryPi OS

draghan:
Old check is not sufficient in the new RaspberryPi OS, which can lead to building a broken build, as the default compiler on this OS is currently GCC 12.
Extended file check to additionally look for '/etc/apt/sources.list.d/raspi.list'.


lynxlynxlynx:
Thanks!

Issue #2314, opened 2025-02-02
PST: screen tints shouldn't tint UI


MarcelHB:
#### Bug description
It mostly happens during spell casting or working of spells that the screen turns a little darker, or gray (as mentioned in #2306)

While it's the style of the original to tint the world a little, the UI elements should remain unaffected. I find this full-screen effect somewhat annonying since you need to continue to operate during this time.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
I don't think this is particularly easy to fix. It's done by the fx_tint_screen opcode, but the problem is we do all the fading globally (GlobalTimer, WindowManager). So I imagine quite some refactoring would be needed or much of GlobalTimer duplicated to support this just for a single window/view.

MarcelHB:
Yeah, I had a feeling like this ...

PR #2333, opened 2025-02-22
Screen tinting

MarcelHB:
## Description
See #2314. I'm aware of PST's `FadeToBlack` and also the opcode 0xC3 as far as it concerns some spells. For the fading, it now matches the original that leaves the UI presented, and the spells also only touch the game world. So is there some case where this wouldn't help or needs more attention?

## Checklist


lynxlynxlynx:
How do the fades in other games look now? For example when you use the portal in Chateau Irenicus.

MarcelHB:
It's the same as in the original.

lynxlynxlynx:
Sounds fine then. :)

Issue #1849, opened 2023-07-05
KeyEvents can be split between Windows


bradallred:
`GameControl::OnKeyRelease` appears to be the only override for the key release event. I don't know why we don't just handle everything in `GameControl::OnKeyPress` for consistency and to avoid bugs arising from key events becoming split between windows.

> we do track mouse down across focus changes via `trackingWindow`. we _could_ do something similar with keystrokes, tho im not sure we should.

I feel a lot better about doing this knowing that nothing else is using `OnKeyRelease`. Of course now I'm also questioning why we use `OnKeyPress` over `OnKeyRelease` because that is both what feels natural to me and would avoid having issues arise from a key down event triggering a window to open/close and then the subsequent release going to a different window.

_Originally posted by @bradallred in https://github.com/gemrb/gemrb/issues/1661#issuecomment-1622060742_

bradallred:
It's worth asking what the originals did, as well as what the EEs do.

lynxlynxlynx:
It's all firing on kb key press in the games. Don't have a strong opinion whether we should respect that or not.

PR #2338, opened 2025-02-24
GameControl: Implement smooth keyboard scrolling

### LINKED ISSUE #2316: BG2EE: Scrolling using arrow keys
#### Bug description
Scrolling on the game screen using the arrow keys causes the screen to "jump"/scroll with a lag, instead of being smooth.

Scrolling diagonally using arrow keys not working.

Scrolling by moving the cursor to the edge of the game screen/window works normally.

- [ ] master as of this issue
- [x] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

CalamariDetective:
## Description
This resolves #2316. Previous behavior is that on pressing an arrow key the viewport would jump in the direction of the arrow key, repeating if key was held down. Diagonal key scrolling did not work. This commit changes to a smooth scroll with diagonal support that matches original IE behavior. Tested with IWD2 and PST.

This is my first PR ever, so any feedback is appreciated. Thanks!

## Checklist

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -436,8 +436,14 @@ void GameControl::WillDraw(const Region& /*drawFrame*/, const Region& /*clip*/)
 		}
 	}
 
+	// Prevent key scrolling from continuing when another window opens
+	if (scrollKeysActive && !IsReceivingEvents()) {
+		scrollKeysDown = scrollKeysActive = 0;
+		vpVector.reset();
+	}
+
 	if (!vpVector.IsZero() && MoveViewportTo(vpOrigin + vpVector, false)) {
-		if ((Flags() & IgnoreEvents) == 0 && core->GetMouseScrollSpeed() && !screenFlags.Test(ScreenFlags::AlwaysCenter)) {
+		if ((Flags() & IgnoreEvents) == 0 && !scrollKeysActive && core->GetMouseScrollSpeed() && !screenFlags.Test(ScreenFlags::AlwaysCenter)) {
 			orient_t orient = GetOrient(Point(), vpVector);
 			// set these cursors on game window so they are universal
 			window->SetCursor(core->GetScrollCursorSprite(orient, numScrollCursor));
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks, will take a look!

gemrb/core/GUI/GameControl.cpp
```diff
@@ -1117,6 +1132,24 @@ bool GameControl::OnKeyRelease(const KeyboardEvent& Key, unsigned short Mod)
...
+		case GEM_UP:
+		case GEM_DOWN:
+		case GEM_LEFT:
+		case GEM_RIGHT:
+			{
+				unsigned int releasedKey = 1 << (Key.keycode - GEM_LEFT);
+				scrollKeysDown &= ~releasedKey;
+				scrollKeysActive &= ~releasedKey;
+
+				// If the opposite arrow key is still held down, switch to that direction:
```
lynxlynxlynx, line 1144: Why is this needed, won't the next OnKeyPress set it properly?

gemrb/core/GUI/GameControl.cpp
```diff
@@ -1117,6 +1132,24 @@ bool GameControl::OnKeyRelease(const KeyboardEvent& Key, unsigned short Mod)
...
+				unsigned int oppositeKey = releasedKey & 1 ? releasedKey << 1 : releasedKey >> 1;
+				if (scrollKeysDown & oppositeKey) {
+					scrollKeysActive |= oppositeKey;
+				}
+
+				ApplyKeyScrolling();
```
lynxlynxlynx, line 1150: This looks redundant, we should scroll on either press or release. In general, the originals did it on press, but I don't remember if I checked scrolling (#1849).

CalamariDetective: I just checked PST: if you hold left and press right, it will scroll left then right. If you let go of right it will scroll left again. If you hold right, then press left, it ignores the left key and keeps scrolling right. Right overrides left and down overrides up. Releasing down/right makes it scroll up/left if they are still held. Not sure if they're doing down/up events or just checking the key press state of those keys each tick.

gemrb/core/GUI/GameControl.cpp
CalamariDetective: You only get repeat OnKeyPress events for the last key pressed. So, for example, if you press left and up, you'll get one event for left and repeated events for up. If you release up and keep left held down, you don't get any OnKeyPress events for left anymore.

gemrb/core/GUI/GameControl.cpp
lynxlynxlynx: So the oppositeKey stuff should be enough, we don't need to call ApplyKeyScrolling, right?

CalamariDetective: ApplyKeyScrolling updates vpVector to match the keys that are currently pressed, so it needs to be called in OnKeyPress to start scrolling in a direction, and in OnKeyRelease to stop scrolling in that direction. I added some comments to clarify this.

Regarding the discussion on not relying on OnKeyRelease because another window might open:  in the initial coding I did find that OnKeyRelease isn't received if another window opens (for example, pressing "i" to open the inventory while scrolling). I addressed that in WillDraw: if key scrolling is active it stops scrolling if IsReceivingEvents() is false.

lynxlynxlynx:
Thanks!

Issue #2334, opened 2025-02-22
BG2: Wrong menu music in ToB


MarcelHB:
#### Bug description
When in the menu where choosing what to play, entering ToB starts the SoA music whereas the original plays the ToB theme.

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.4
- [ ] 0.9.3
- [ ] 0.9.2

lynxlynxlynx:
It's all in Start*.py, dependant on PlayMode and/or oldgame.

PR #2340, opened 2025-02-24
BG2 menu: audio & video fixes

MarcelHB:
## Description
See #2334. Also the video flags didn't work as expected any more.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg2/Start.py
+++ b/gemrb/GUIScripts/bg2/Start.py
@@ -21,32 +21,32 @@ import GemRB
 import GameCheck
 from GUIDefines import SV_SAVEPATH
 
-skip_videos = False
+skipVideos = False
 
 def RunStart2(isTOB):
-	global skip_videos
-	
+	global skipVideos
+
+	skipIntro = GemRB.GetVar ("SeenIntroVideos")
 	if isTOB:
-		GemRB.SetMasterScript("BALDUR25","WORLDM25")
-		GemRB.SetVar("oldgame",0)
-		if not skip_videos and not skip_videos&2:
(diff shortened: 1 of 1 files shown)


gemrb/GUIScripts/bg2/Start.py
```diff
@@ -21,32 +21,31 @@
 import GameCheck
 from GUIDefines import SV_SAVEPATH
 
-skip_videos = False
+skipVideos = False
+skipIntro = 0
```
lynxlynxlynx, line 25: This is for the game-specific intro, but now we have two very similar vars and approached. I proposed you either also use skipIntro in OnLoad or use SeenIntroVideos in RunStart2. Either can be treated as a bitfield and both persist.

MarcelHB: I now use `SeenIntroVideos` for the game specific ones, too.

gemrb/GUIScripts/bg2/Start.py
```diff
@@ -21,32 +21,32 @@
...
 
-skip_videos = False
+skipVideos = False
 
 def RunStart2(isTOB):
-	global skip_videos
-	
+	global skipVideos
+
+	skipIntro = GemRB.GetVar ("SeenIntroVideos")
```
lynxlynxlynx, line 29: Won't this be None on first run, breaking the anding lower?

MarcelHB: No, the `OnLoad` sets this to 1 before getting here, and it's not executed when all intro videos are skipped.

lynxlynxlynx:
Thanks!

Issue #2169, opened 2024-09-29
 IWD2: Show usability in item description


MarcelHB:
#### Bug description
In IWD2, there is a distinct section in the item's description that lists if the item is (un)usable by some classes/attitude/race. This is helpful since it's sometimes not apparent through the explanation. And in contrast to other games, this list isn't part of it.

![iwd2_inv_help](https://github.com/user-attachments/assets/0a0aa6c1-5d9d-41c6-a0bd-ba7bbcd7cd6a)

#### Screenshots
![iwd2_inv_help_ni](https://github.com/user-attachments/assets/358976bc-3acf-4bd6-b1ca-91d8e6778367)

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.3
- [ ] 0.9.2
- [ ] 0.9.1

lynxlynxlynx:
Ah, interesting. We can handle this on the python side, we just need to first expose `item->UsabilityBitmask`, `item->KitUsability` in `GemRB_GetItem`.

PR #2341, opened 2025-02-26
IWD2 item restriction description

MarcelHB:
## Description
See #2169. I have added this feature in a simplified way, i. e. it doesn't fully match the original, for two reasons: 1. I was too lazy to get it all from the decompiled disassembly and even more 2. the choice between displaying who can and who cannot was bad for the user, or at least not following an apparent pattern, and also hiding information technically.

So for simplicity I decided to always list who _cannot_ use an item:

1. There are unusability bits set.
2. Check if any such bits are set inside of each group (class, race, alignment, stance), and if so show what's excluded in this group.
3. For the kit classes, check if there is more discrinimation between kits, and if so, also show what's excluded in detail.

## Checklist


gemrb/plugins/GUIScript/GUIScript.cpp
```diff
@@ -8989,6 +8989,8 @@ PyDoc_STRVAR(GemRB_GetItem__doc,
   * 'Tooltips'           - the item tooltips\n\
   * 'Locations'          - the item extended header's ability locations\n\
   * 'Spell'              - the spell's strref if the item is a copyable scroll\n\
+  * 'UsabilityBitmask'   - class usability bitmask (IWD2)\n\
+  * 'KitUsability'       - kit usability bitmask (IWD2)\n\
```
lynxlynxlynx, line 8993: These are not iwd2 specific btw.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -242,6 +242,23 @@ def GetItemDescription (item, itemtype):
 	if (itemtype & 2):
 		text = item["ItemDesc"]
 
+	if GameCheck.IsIWD2():
+		# restrictions apply, we only list them
+		usabilityMask = item["UsabilityBitmask"]
+		if usabilityMask > 0:
+			text = GemRB.GetString(text) + "\n\n"
+			text += f"[color=ffff00ff]{GemRB.GetString(31530)}[/color]\n  "
```
lynxlynxlynx, line 250: f-strings were introduced in python 3.6, while we still advertise 3.3 as the minimum, so please use .format() instead.

MarcelHB: Wow, I didn't realize how old Python 3 already is ...

MarcelHB:
> Races, kits and classes as well: we already have all this info in their respective tables (races.2da, classes.2da), so it doesn't need to be duplicated. This way the function also won't have to be changed if more entries are added via mods.

Except maybe classes, I did not find a good way to match the order of bits to these tables without writing something similar or more confusing. But if you have an idea, let me know.

> And I'd bet there are items that are only restricted from certain kits, not their base classes.

Yes, there is some Tempus-exclusive book (don't have the list r/n) but it worked out as expected to me.

lynxlynxlynx:
Then just add another another column to the tables with the masks.

MarcelHB:
Alright, I will think of something.

MarcelHB:
I added something for classes and races, `USABILITY` was ready to use, just didn't notice it in the first turn.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -253,6 +272,97 @@ def GetItemDescription (item, itemtype):
...
+	text = ""
+	n = CommonTables.Races.GetRowCount()
+
+	for i in range(n):
+		rowName = CommonTables.Races.GetRowName(i)
+		raceMask = CommonTables.Races.GetValue(rowName, "USABILITY")
+		nameRef = CommonTables.Races.GetValue(rowName, "NAME_REF")
+
+		if raceMask == 0x7FFFFFFF: # GetValue gives as signed value
+			raceMask = 0x80000000
```
MarcelHB, line 307: A little out of ideas how to better handle this for now.

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Why the revert?

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -253,6 +272,97 @@ def GetItemDescription (item, itemtype):
...
+	return text
+
+def GetIWD2RestrictedStances(mask):
+	text = ""
+	if mask & 0x20000: # neutral
+		text += GemRB.GetString(31535) + "\n  "
+	if mask & 0x10000: # lawful
+		text += GemRB.GetString(29372) + "\n  "
+	if mask & 0x1000: # chaotic
+		text += GemRB.GetString(31536) + "\n  "
```
lynxlynxlynx, line 293: Looks like you misassigned some strings. My dump has:

> @31530 = ~Not Usable By:~
> @31531 = ~Usable By:~
> @31532 = ~Good~
> @31533 = ~Lawful~
> @31534 = ~Neutral~
> @31535 = ~Chaotic~
> @31536 = ~Evil~

lynxlynxlynx: Alternatively you could just dump them into a new aligns.2da-like table for clean lookups.

gemrb/GUIScripts/InventoryCommon.py
MarcelHB: I messed up with git ...

gemrb/GUIScripts/InventoryCommon.py
MarcelHB: Thanks, the search of NI runs somewhat async or something, I have spent too much time on finding the strings correctly.

MarcelHB: re. new 2DA: is it fine to only create one of IWD2 for now? Don't want to go through this again if not required. 🥲

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -242,6 +243,24 @@ def GetItemDescription (item, itemtype):
...
+			text = GemRB.GetString(text) + "\n\n"
+			text += "[color=ffff00ff]" + GemRB.GetString(31530) + "[/color]\n  "
+
+			if usabilityMask & 0xE000: # alignment
+				text += GetIWD2RestrictedAlignments(usabilityMask)
+			if usabilityMask & 0x31000: # stance
+				text += GetIWD2RestrictedStances(usabilityMask)
+			if usabilityMask & 0x3F800000: # race
+				text += GetIWD2RestrictedRaces(usabilityMask)
+			if usabilityMask & 0x7FF: # classes
```
lynxlynxlynx, line 260: I suggest, just for future proofing, instead of 0x7ff: 0xfff to give space for another class like the shaman.

But isn't the check too weak, since it covers only the class range? 00bard02 has no class bits set, only kit ones.

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Of course, we don't need it in other games.

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Nothing much better comes to mind.

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: Sorry, scrap that last part, I was looking at the wrong field.

MarcelHB: > I suggest, just for future proofing, instead of 0x7ff: 0xfff to give space for another class like the shaman.

Can't say much about this. It's _Unknown_ in IESDP and since IWD2 has no EE, I guess there won't be one any soon.

> But isn't the check too weak, since it covers only the class range? 00bard02 has no class bits set, only kit ones.

This item has class bits set for me, all but the bard class, and all kit ones, so this should fit.

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -253,6 +272,97 @@ def GetItemDescription (item, itemtype):
...
+		rowName = CommonTables.Classes.GetRowName(i)
+		parentClass = CommonTables.Classes.GetValue(rowName, "CLASS")
+		mask = CommonTables.Classes.GetValue(rowName, "USABILITY")
+		nameRef = CommonTables.Classes.GetValue(rowName, "NAME_REF")
+		kitMask = ClassKitUsability.get(i + 1, 0)
+
+		if parentClass != 0:
+			continue
+
+		if classesMask & mask or (kitMask and kitsMask & kitMask):
```
lynxlynxlynx, line 347: Can kitMask really every not be numeric?

gemrb/GUIScripts/InventoryCommon.py
lynxlynxlynx: We already support the shaman and a 13th class in core, so this would makes it implementable through mods without touching our code.

MarcelHB: Alright

gemrb/GUIScripts/InventoryCommon.py
```diff
@@ -253,6 +271,96 @@ def GetItemDescription (item, itemtype):
...
+	showStances = mask & 0x31000
+
+	text = ""
+	for i in range(n):
+		rowName = AlignmentStancesTable.GetRowName(i)
+		isStance = AlignmentStancesTable.GetValue(rowName, "STANCE")
+		useMask = AlignmentStancesTable.GetValue(rowName, "USABILITY")
+		nameRef = AlignmentStancesTable.GetValue(rowName, "NAME_REF")
+
+		if ((isStance and showStances) or (not isStance and showAlignments)) and useMask & mask:
```
lynxlynxlynx, line 290: Maybe I'm being dumb, but why do you need to distinguish alignment axes at all? They will be checked in table order, so alignments will be properly grouped already.

MarcelHB: Yeah, serves no need any more.

lynxlynxlynx:
Alright, thanks!

PR #2354, opened 2025-06-22
Pathfinder improvements: traversability cache.

draghan:
Implemented TraversabilityCache for caching positions of all Actors on the navmap. The reasoning behind this is to unburden the execution hotpath in FindPath method, so that it won't call `Actor* Map::GetActor(const Point& p, int flags, const Movable* checker) const`, which is not cheap operation. Instead just make one memory access to the properly maintained cache, which stores information about each navmap cell: which Actor (if any) occupies the cell and whether it's bumpable or not.

The layout of data in the TraversabilityCache class and all its components was carefully chosen to be friendly to the data prefetcher and CPU data cache, validated by benchmarks on three different test benches representing different class of hardware:
- bench 0: desktop with AMD Ryzen 9 5950x, fairly modern (2020), highend desktop chip, x86-64, 16C 32T, boost up to 4.9 GHz, cache L1 64 KB per core, L2 512 KB per core, L3 64 MB. Ubuntu 24.04, g++ 13.3.0
- bench 0w: same HW as bench0, but on Windows 10, msvc 19.38.33143
- bench 1: laptop Dell Latitude E7240 with Intel Core i5-4300U, older mobile chip (2013), x86-64, 2C 4T, boost up to 2.9 GHz, cache: L1 64 KB per core, L2 256 KB per core, L3 3 MB. Mint 21.1, g++ 11.4.0
- bench 2: Raspberry Pi 4B Rev 1.2 with Broadcom BCM2711 SoC (2019), Quad core Cortex-A72 (ARM v8) 64-bit, 4C 4T, boost up to 1.5 GHz, cache: L1 32 KB data + 48 KB instruction per core, L2 1 MB. Debian 12 aarch64, clang++ 14.0.6

The cache is updated once a frame, if and only if anyone called FindPath - there is no point in updating the cache if nobody wants to use it. Updating once a frame means we're introducing a slight change of actors' perception when comes to finding path: previously they had immediate access to the newest actors' positions and if any changes in positions were done in the current frame, all the other actors knew about it right away. Now this information propagation is one frame delayed. There is no functional difference in behavior, bit there is a slight technical difference: this results in rare need of recalculation path without blocking actors where previously it wasn't needed, but also vice versa: no need of recalculation, where previously it was required. It's visible sometimes when measuring against original FindPath performance: speedup from 15ms to 200µs or (slowdown the other way around) means we encountered scenario where this difference manifests.

Additionally made some minor improvements in the FindPath method: 
1. Removed check for the usage of the LazyThetaStar - the less branches in the hotpath, the better. 
2. Made most data on which the algorithm works static - no need to make new allocations each call.

Those changes should yield improvements to the FindPath execution time across the board: on average it's about 2x faster. 
Detailed benchmark reports from multiple test cases and multiple machines are attached to the pull request: 
[traversabilityCacheFinal.zip](https://github.com/user-attachments/files/21559697/traversabilityCacheFinal.zip)

Test cases:
- case 0: BG2, AR0602, "Irenicus's Dungeon 1st Floor": party of four, dungeon, no random movement of NPCs, no fighting, typical exploration scenario with mixed (short, medium, long and impossible) paths to be calculated. Going from library to the Cambion room, going back and forth inside the cambion room 4 times, going back and forth between the Cambion room and the Weapons room 2 times, going into the tight corridor above the Weapons room, triggering impossible path via targeting beyond closed door 2x. High repeatability of test case.

- case 1: BG2, AR0700, "Waukeen's Promenade": party of four, outdoor, lots of random movement of NPCs, no fighting. Living city example scenario with mostly short paths to be calculated for NPCs, but also some long ones, across the whole area, for the party. Test case: standing in near the entrance to the Temple of Ilmater for 5 minutes (with trapped NPCs inside the party, which seems to generate lots of FindPath requests), then go to south exit and back 2 times. Exact repeatability impossible.

- case 2: BG1: AR0146, "Thieves' Maze": party of six, no random movement of NPCs, no fighting. Stress test for the pathfinding algorithm but also for the path scheduling behavior (frequently losing the target and stopping, frequent bumping into each other). Mix of short/medium and long paths to be calculated. Requesting some paths to the neighbouring corridor but also to the other side of the maze. Medium repeatability, due to high rate of unpredictable dropping of paths by the party actors.
Important note: apart from performance, I think the pathfinding is functionally broken in its core right now, as in this test case the calculated path was often leading actors straight to the wall; properly functioning algorithm would never have done that, unless there is some other factor which causes this behavior.

- case 3: IWD2: AR3001, "Fortress Approach - Goblin Warrens Level 1": party of six, dungeon. Lots of hostile NPCs, big fight in a tight spot: luring three groups of enemies (31 NPCs) to a narrow corridor near Pondmuk, where they all can't reach the party. Standing there for 5 minutes. High stress test to the pathfinding algorithm. Mix of short and impossible paths. Exact repeatability impossible.

Brief summary of results:
- bench0:
  - case0: long(avg=2.00x, med=1.99x), med(avg=1.95x, med=1.96x), short(avg=1.90x, med=1.76x)
  - case1: long(avg=2.03x, med=2.04x), med(avg=2.03x, med=2.02x), short(avg=1.72x, med=1.67x)
  - case2: long(avg=1.65x, med=1.65x), med(avg=1.67x, med=1.62x), short(avg=1.31x, med=1.31x)
  - case3: long(avg=2.33x, med=2.21x), med(avg=2.33x, med=2.30x), short(avg=1.76x, med=1.96x)
- bench0w:
  - case0: long(avg=1.77x, med=1.77x), med(avg=1.65x, med=1.77x), short(avg=1.64x, med=1.53x)
  - case1: long(avg=1.79x, med=1.80x), med(avg=1.66x, med=1.68x), short(avg=1.12x, med=1.03x)
  - case2: long(avg=1.66x, med=1.66x), med(avg=1.60x, med=1.60x), short(avg=1.28x, med=1.31x)
  - case3: long(avg=1.98x, med=1.97x), med(avg=1.86x, med=1.87x), short(avg=1.44x, med=1.46x)
- bench1:
  - case0: long(avg=2.20x, med=2.22x), med(avg=1.97x, med=1.97x), short(avg=1.84x, med=1.69x)
  - case1: long(avg=2.23x, med=2.24x), med(avg=2.24x, med=2.17x), short(avg=2.50x, med=2.53x)
  - case2: long(avg=1.72x, med=1.72x), med(avg=1.63x, med=1.62x), short(avg=1.44x, med=1.38x)
  - case3: long(avg=2.34x, med=2.35x), med(avg=2.70x, med=2.71x), short(avg=2.26x, med=2.31x)
- bench2:
  - case0: long(avg=1.93x, med=1.95x), med(avg=1.54x, med=1.49x), short(avg=1.36x, med=1.31x)
  - case1: long(avg=2.01x, med=2.01x), med(avg=1.82x, med=2.01x), short(avg=1.28x, med=1.21x)
  - case2: long(avg=1.69x, med=1.70x), med(avg=1.46x, med=1.47x), short(avg=1.15x, med=1.12x)
  - case3: long(avg=2.19x, med=2.16x), med(avg=2.08x, med=2.19x), short(avg=1.15x, med=1.17x)


gemrb/core/TraversabilityCache.h
```diff
@@ -0,0 +1,221 @@
...
+#define TRAVERSABILITY_CACHE_H
+
+//#include "Map.h"
+#include "Scriptable/Actor.h"
+
+
+namespace GemRB {
+
+
+struct FitRegion {
```
lynxlynxlynx, line 31: This needs an explanation justifying its existence vs just using Region.

gemrb/core/Scriptable/Actor.h
```diff
@@ -445,6 +445,17 @@ using vvcDict = std::multimap<ResRef, ScriptedAnimation*>;
...
+	bool IsTraversable() const
+	{
+		if (!ValidTarget(GA_NO_DEAD | GA_NO_UNSCHEDULED)) {
+			return true;
+		}
+		if (ValidTarget(GA_ONLY_BUMPABLE)) {
+			return true;
+		}
+		return false;
+	}
```
lynxlynxlynx, line 457: This looks unused and now should also go.

gemrb/core/PathFinder.cpp
```diff
@@ -329,37 +360,16 @@ Path Map::FindPath(const Point& s, const Point& d, unsigned int size, unsigned i
...
-						distFromStart[smptChildIdx] = newDist;
-					}
-				}
-
-				if (distFromStart[smptChildIdx] < oldDist) {
-					PQNode newNode(nmptChild, getHeuristic(smptChild, smptChildIdx));
-					open.emplace(newNode);
-				}
-			} else {
+			{
```
lynxlynxlynx, line 372: I appreciate this making the diff less noisy to review, but in the final version the extra scope should be removed. Can be an extra commit to maintain readability.

gemrb/core/PathFinder.cpp
```diff
@@ -259,15 +270,35 @@ Path Map::FindPath(const Point& s, const Point& d, unsigned int size, unsigned i
...
+	FibonacciHeap<PQNode> open; // todo: remove this data structure or rewrite its implementation - too much allocations there!
+	// make most data storage for this algorithm static, to avoid memory allocations;
+	// each run we just clear the storage, which is keeping the underlying allocated memory at hand
+	static std::vector<bool> isClosed(mapSize.Area(), false);
+	static std::vector<NavmapPoint> parents(mapSize.Area(), Point(0, 0));
+	static std::vector<unsigned short> distFromStart(mapSize.Area(), std::numeric_limits<unsigned short>::max());
+
+	// resize if needed (in case of a map change; probably can be done once, when new map is loaded)
+	if (isClosed.size() != mapCellsCount) {
+		isClosed.resize(mapCellsCount);
```
lynxlynxlynx, line 284: Isn't this redundant? You always resize it below.

gemrb/core/PathFinder.cpp
```diff
@@ -259,15 +270,35 @@ Path Map::FindPath(const Point& s, const Point& d, unsigned int size, unsigned i
...
+	// number of elements, while memset performs better for large vectors
+	memset(static_cast<void*>(parents.data()), 0, sizeof(Point) * mapCellsCount);
+	memset(static_cast<void*>(distFromStart.data()), std::numeric_limits<unsigned short>::max(), sizeof(unsigned short) * mapCellsCount);
+
+	// begin algo init
 	distFromStart[smptSource.y * mapSize.w + smptSource.x] = 0;
 	parents[smptSource.y * mapSize.w + smptSource.x] = nmptSource;
 	open.emplace(PQNode(nmptSource, 0));
 	bool foundPath = false;
-	static bool usePlainThetaStar = gamedata->GetMiscRule("LAZY_THETA_STAR") == 0;
```
lynxlynxlynx, line 270: This then needs to be removed from miscrule.2da too.

lynxlynxlynx:
Thanks! Overall I worry about maintaining proper cache state, since script actions and effects can also change all three cached actor states. Some of it is predictable, e.g. effects run before actions, but both can trigger both, sometimes in the same frame, so the order is not really deterministic. And it's not just about a potential 1 frame delay, since `FindPath` is ideally ran just once per trip. So e.g. if a trap script summoned a bunch of ai-less monsters in the way, the actor could walk right through, since nothing triggered the cache update.

So I think it would be better to update the cache directly when changes happen. For position we have `Scriptable::SetPos` (only one or two places still modify Pos directly), for being dead `Actor::Die` (PS: corpses can move, but never impede), for unscheduled it's tricker (maybe loop over everyone without a full-day schedule in `Game::AdvanceTime` + `pcf_avatarremoval`), for bumpable it's just in several places. The combat check could be done in `FindPath` once, since it's just checking one number and is not really tied to an actor state. The animation one can be ignored, IE_EA has `pcf_ea`, while the "IsMoving" part may require more work, since I'm pretty sure we set the stance directly in several places. But perhaps it could be simplified for the cache and we'd update when giving the Movable a path instead.

If all that works out, there's no need for a blanket `TraversabilityCache::Update` or worrying about correct invalidation. Perhaps it will reduce the speedups a bit, but the main gains of avoiding the repeated lookups should persist. Of the cached components only a few are bound to change often: position, is moving, and to a lesser extent (combat yes/no) and IE_EA (e.g. when starting combat with neutral actors).

draghan:
I'm [attaching](https://github.com/user-attachments/files/20868322/6b6e47d28.zip) benchmarks' results: raw logs + detailed reports for the following matrix of (test cases X test benches).

Results grouped by test benches, in each bench directory there are logs+reports.

Test cases:
- case 0: BG2, AR0602, "Irenicus's Dungeon 1st Floor": party of four, dungeon, no random movement of NPCs, no fighting, typical exploration scenario with mixed (short, medium, long and impossible) paths to be calculated. Going from library to the Cambion room, going back and forth inside the cambion room 4 times, going back and forth between the Cambion room and the Weapons room 2 times, going into the tight corridor above the Weapons room, triggering impossible path via targeting beyond closed door 2x. High repeatability of test case.

- case 1: BG2, AR0700, "Waukeen's Promenade": party of four, outdoor, lots of random movement of NPCs, no fighting. Living city example scenario with mostly short paths to be calculated for NPCs, but also some long ones, across the whole area, for the party. Test case: standing in near the entrance to the Temple of Ilmater for 5 minutes (with trapped NPCs inside the party, which seems to generate lots of FindPath requests), then go to south exit and back 2 times. Exact repeatability impossible.

- case 2: BG1: AR0146, "Thieves' Maze": party of six, no random movement of NPCs, no fighting. Stress test for the pathfinding algorithm but also for the path scheduling behavior (frequently losing the target and stopping, frequent bumping into each other). Mix of short/medium and long paths to be calculated. Requesting some paths to the neighbouring corridor but also to the other side of the maze. Medium repeatability, due to high rate of unpredictable dropping of paths by the party actors. 
_Important note: apart from performance, I think the pathfinding is functionally broken in its core right now, as in this test case the calculated path was often leading actors straight to the wall; properly functioning algorithm would never have done that, unless there is some other factor which causes this behavior._

- case 3: IWD2: AR3001, "Fortress Approach - Goblin Warrens Level 1": party of six, dungeon. Lots of hostile NPCs, big fight in a tight spot: luring three groups of enemies (31 NPCs) to a narrow corridor near Pondmuk, where they all can't reach the party. Standing there for 5 minutes. High stress test to the pathfinding algorithm. Mix of short and impossible paths. Exact repeatability impossible.

Hardware:
- bench 0: desktop with AMD Ryzen 9 5950x, fairly modern (2020), highend desktop chip, x86-64, 16C 32T, boost up to 4.9 GHz, cache L1 64 KB per core, L2 512 KB per core, L3 64 MB.
- bench 1: laptop Dell Latitude E7240 with Intel Core i5-4300U, older mobile chip (2013), x86-64, 2C 4T, boost up to 2.9 GHz, cache: L1 64 KB per core, L2 256 KB per core, L3 3 MB.
- bench 2: Raspberry Pi 4B Rev 1.2 with Broadcom BCM2711 SoC (2019), Quad core Cortex-A72 (ARM v8) 64-bit, 4C 4T, boost up to 1.5 GHz, cache: L1 32 KB data + 48 KB instruction per core, L2 1 MB.

gemrb/core/Scriptable/Actor.h
draghan: Good catch, thanks.

gemrb/core/PathFinder.cpp
draghan: Eagle eye! Thanks.

gemrb/core/PathFinder.cpp
draghan: Sure, it's left exactly for this purpose. It'll be removed.

draghan:
> Overall I worry about maintaining proper cache state, since script actions and effects can also change all three cached actor states. Some of it is predictable, e.g. effects run before actions, but both can trigger both, sometimes in the same frame, so the order is not really deterministic. And it's not just about a potential 1 frame delay, since `FindPath` is ideally ran just once per trip. So e.g. if a trap script summoned a bunch of ai-less monsters in the way, the actor could walk right through, since nothing triggered the cache update.

@lynxlynxlynx I'm pretty sure the `Update()` method will catch it - we detect all relevant changes (position, bumpable and alive states) to all Actors on the map (for currently cached, but also added to the map and removed from it since the last frame), given of course that the spawned monsters will appear in the `map->actors` vector. The original implementation also relied on the `map->actors`, so I think it shouldn't introduce any functional difference.

Also a side note about the delay: I see your stance that it shouldn't be a big deal, but if that **would be** a potential problem, I think we could remove the 'once per frame' condition and make it run on every `FindPath` request: in my benchmarks the vast majority of calls is including the cache update - it looks like it's rare for more actors to request for new path in the same frame (apart from the actors from the party when the group move is scheduled directly by the player or some combat scenarios).
I didn't think it would be an issue, even in modern multiplayer games the 1-frame delay in NPC's perception is no big deal (and TBF, here we're not even delaying their "perception"), so limiting it to 'once per frame' was in my eyes just free performance when the player moves 2+ actors at the time.

> So I think it would be better to update the cache directly when changes happen. 

We could update on the fly and it has it's own pros (mainly - only change what's really needed, no pulling of all actors' state necessary, which I like!), but it'd require putting manually all the guards in all possible places of influence to the relevant data, as you said. I feel it's introducing maintainability issues and potential for hard to catch bugs.

Please think about it in the context of what I said and let me know whether to make changes to this part or not.

gemrb/core/PathFinder.cpp
lynxlynxlynx: Nope.

gemrb/core/TraversabilityCache.h
lynxlynxlynx: It should be a comment until we address the larger point, decide whether it's worth to have two implementations or less convenience. Or something in the middle, like having Region subclass FitRegion. Did you find a lot of users of the shortcut references?

gemrb/core/TraversabilityCache.cpp
```diff
@@ -0,0 +1,293 @@
...
+	} else {
+		ResetIsAlive(new_idx);
+	}
+	region.push_back(CalculateRegion(inActor));
+	return new_idx;
+}
+
+Region TraversabilityCache::CachedActorsState::CalculateRegion(const Actor* inActor)
+{
+	// code from Selectable::DrawCircle, will it be always correct for all NPCs?
```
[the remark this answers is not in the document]
lynxlynxlynx: We have an open issue to better match the blocked shape for bigger actors, but this approach should be fine either way.

lynxlynxlynx:
> @lynxlynxlynx I'm pretty sure the Update() method will catch it - we detect all relevant changes (position, bumpable and alive states) to all Actors on the map (for currently cached, but also added to the map and removed from it since the last frame), given of course that the spawned monsters will appear in the map->actors vector. The original implementation also relied on the map->actors, so I think it shouldn't introduce any functional difference.

I agree that `Update` would catch it, since it's complete. But the problem is, if I read the code correctly, that it just isn't ran often enough (regardless of that once-per-frame guard).

If you send a lone actor to some easy destination, it will call `FindPath` only once, however long the walking will take. So potentially a long while of a stale cache if nobody else triggers an update.
Or consider just a bunch of NPCs assigned to wander randomly (`RandomWalk`) — those single-step paths are calculated without `FindPath`. If the space is small enough or you wait long enough, the lack of cache updates should enable them to walk through each other and avoid some empty spots.
Or consider a cutscene teleporting someone (almost the same as the trap example).

Correctness is more important than speed. Running `Update` each Map tick would get you much closer, but still wouldn't be rigorous. And one tick of a difference can mean that a cutscene breaks (for example from different/impossible pathing); it has happened several times before.

I know it's annoying, looking all those places up showed it'd be quite some work. But I don't see a way to ensure correctness without it, there's too much interplay (also time isn't always monotonous). And I think it's easier to reason about it upfront than to deal with later circumstantial heisenbugs. We can improve the maintainability with good design, but also warning comments. I can start preparing and cleaning these regions where we'd need to do singular updates from, so it'll be easier for you to hook them in in the PR. Just give me some time and maybe work on something else in the interim.

lynxlynxlynx:
Ok, did some minor cleanup and verified it will actually be much simpler to update piecemeal:
- `Scriptable::SetPos` for position
- `Map::ClearSearchMapFor` and `Map::BlockSearchMapFor` ... instead of the dead and unscheduled checks
- bumpability:
  - `core->GetGame()->CombatCounter` can be done once in `FindPath` (once per frame/tick is enough); we could also add a convenience function
  - `pcf_ea` if oldValue != newValue
  - `IsMoving`: `Movable::SetStanceDirect`
 
I left some comments in `ValidTarget` for the above.

draghan:
> I know it's annoying, looking all those places up showed it'd be quite some work. 

I don't really care if it's "annoying" or not, if that must be done to ensure correctness, it will be done, don't worry about that. I believe that with your help we can track down all necessary places. The thing is: I don't think we're introducing any regression in this patch (besides potential differences with the one frame guard, but let's ignore that for a second).

I think conclusions from your scenarios are not correct. Let's dive into them a little bit further. 
Below, by "current code" I'll mean the code without this patch.

> If you send a lone actor to some easy destination, it will call FindPath only once, however long the walking will take. So potentially a long while of a stale cache if nobody else triggers an update.

Yes, that part is correct. But the follow up question to get to the meat: who cares if the cache is stale, when nobody wants to use it?

This cache is >>accessed<< **only** when calculating a new path. Not when actors are already moving along their assigned paths.

So the observed result will be exactly the same as in the current code: in this patch we just swapped `look for potential actor from this navmap point` to `look into the prepared actors state cache from that navmap point`, but we did it solely in the `FindPath` method - and the cache is always updated before its use in this method.

Current code is not calling more `FindPath` in this scenario besides at the beginning of the journey, so a stale cache will not affect literally anything.

>  Or consider a cutscene teleporting someone (almost the same as the trap example).

Any actor which will ask for a path via `FindPath` will know the new position of the teleported actor, because the `Update` is triggered from `FindPath`. Any actor which doesn't rely on `FindPath` to go somewhere will have their behavior unchanged.

> Or consider just a bunch of NPCs assigned to wander randomly (RandomWalk) — those single-step paths are calculated without FindPath. If the space is small enough or you wait long enough, the lack of cache updates should enable them to walk through each other and avoid some empty spots.

Again: no `FindPath`, no change in the behavior.


There are pros and cons of each approach and as I mentioned, I suspect updating on the fly only the changed data may be perhaps better in terms of performance (it's hard to tell in advance, as in this approach we must take into consideration also the scale: updates will be smaller, but they will be constantly pushing changes, multiple times each frame).
I was thinking about it yesterday and found one argument against: this will introduce more synchronization points between threads, once we move it to a separate one. This will be the definition of "annoying" in my book. :)

gemrb/core/TraversabilityCache.h
lynxlynxlynx: We don't; I know both gcc and clang support `__attribute__((always_inline))`, but apparently msvc's `__forceinline` doesn't actually force it in several scenarios, the most relevant one being debug builds. So annoying for testing, but not for regular users.

But I like the general approach, just the names can be shorter (X, Y, W, H).

lynxlynxlynx:
Hmm, you're right, nothing else can happen between the cache update and its use. A bonus of doing piecemeal updates would be that it enables reuse of the cache elsewhere. `DoStep` has two such candidates and the `ValidTarget` implementation could be simplified (kinda neutralizing the other changes it would require). Maybe there's more. Maybe none are important wrt performance.

The choice is between 5 places for small updates vs 1 for big. If the pathfinding is moved to its own thread, then the cache would be local to it, right, no need for any interthread comms. In the other case, wouldn't you just do the locking in the update function they would call, again being thread-local?

draghan:
I'm attaching [663254955.zip](https://github.com/user-attachments/files/20967583/663254955.zip) with new benchmarks' results: raw logs + detailed reports for the following matrix of (test cases X test benches).

I have reimplemented making measurements, to accommodate possibility of tracking distributed operations.
I've ditched usage of nanobench: it wasn't possible to track operations from different places and comparing the accumulated time from those places to the original run.
It was also doing some funny batching of runs, even when I explicitly set the warmup to none and iterations to 1, which rendered previous measurements more artificial than I'd like, as it was averaging a batch of few runs of `FindPath` and `FindPathOriginal` per request - this was masking all the real variability, including hiding better improvements but also bigger cache update impact.

Now I'm measuring only one run of `FindPath` and `FindPathOriginal` per request, to test real world scenarios, with the resolution of 1 microsecond. The statistical significance should be derived from test cases themselves providing enough samples, not from repeating measurements which is directly affecting the results.

I've added also results from Windows on my main rig, to validate also against Microsoft's compiler.

Results grouped by test benches, in each bench directory there are logs+reports from 4 test cases. Data points are categorized into 3 groups of operations, by their original time duration: short <1ms, medium 1-15ms, long >15ms.

Test cases:


- case 0: BG2, AR0602, "Irenicus's Dungeon 1st Floor": party of four, dungeon, no random movement of NPCs, no fighting, typical exploration scenario with mixed (short, medium, long and impossible) paths to be calculated. Going from library to the Cambion room, going back and forth inside the cambion room 4 times, going back and forth between the Cambion room and the Weapons room 2 times, going into the tight corridor above the Weapons room, triggering impossible path via targeting beyond closed door 2x. High repeatability of test case.

- case 1: BG2, AR0700, "Waukeen's Promenade": party of four, outdoor, lots of random movement of NPCs, no fighting. Living city example scenario with mostly short paths to be calculated for NPCs, but also some long ones, across the whole area, for the party. Test case: standing in near the entrance to the Temple of Ilmater for 5 minutes (with trapped NPCs inside the party, which seems to generate lots of FindPath requests), then go to south exit and back 2 times. Exact repeatability impossible.

- case 2: BG1: AR0146, "Thieves' Maze": party of six, no random movement of NPCs, no fighting. Stress test for the pathfinding algorithm but also for the path scheduling behavior (frequently losing the target and stopping, frequent bumping into each other). Mix of short/medium and long paths to be calculated. Requesting some paths to the neighbouring corridor but also to the other side of the maze. Medium repeatability, due to high rate of unpredictable dropping of paths by the party actors.
    _Important note: apart from performance, I think the pathfinding is functionally broken in its core right now, as in this test case the calculated path was often leading actors straight to the wall; properly functioning algorithm would never have done that, unless there is some other factor which causes this behavior._

- case 3: IWD2: AR3001, "Fortress Approach - Goblin Warrens Level 1": party of six, dungeon. Lots of hostile NPCs, big fight in a tight spot: luring three groups of enemies (31 NPCs) to a narrow corridor near Pondmuk, where they all can't reach the party. Standing there for 5 minutes. High stress test to the pathfinding algorithm. Mix of short and impossible paths. Exact repeatability impossible.

Hardware:
- bench 0: desktop with AMD Ryzen 9 5950x, fairly modern (2020), highend desktop chip, x86-64, 16C 32T, boost up to 4.9 GHz, cache L1 64 KB per core, L2 512 KB per core, L3 64 MB. Ubuntu 24.04, g++ 13.3.0
- bench 0w: same HW as bench0, but on Windows 10, msvc 19.38.33143
- bench 1: laptop Dell Latitude E7240 with Intel Core i5-4300U, older mobile chip (2013), x86-64, 2C 4T, boost up to 2.9 GHz, cache: L1 64 KB per core, L2 256 KB per core, L3 3 MB. Mint 21.1, g++ 11.4.0
- bench 2: Raspberry Pi 4B Rev 1.2 with Broadcom BCM2711 SoC (2019), Quad core Cortex-A72 (ARM v8) 64-bit, 4C 4T, boost up to 1.5 GHz, cache: L1 32 KB data + 48 KB instruction per core, L2 1 MB. Debian 12 aarch64, clang++ 14.0.6

Brief summary (in general providing average results, providing median where there were some variability or outliers, for details please take a look at attached reports):
- bench0:
  - case0: long and medium paths typically 2x, short vary from over 3x to 0.2x, with 1.46x on avg
  - case1: typically 2x (even over, up to 2.4x) for long ops, with more variability in medium (1.5x-2.8x) 2.25x avg, with short ops vary from 0.02x(for extremely short, where path calculation was <=1us and cache update taking 54us) to 5.8x, avg 1.75x
  - case2: avg long 1.65x, medium 1.96x, short 1.0x (0.16x-2.3x)
  - case3: Long median 2.2x, medium avg 2.27x, short 1.32x
- bench0w:
  - case0: long and medium consistent avg/median  1.75x, short avg 1.53x vary from 0.49x to 3.8x
  - case1: long and medium very consistent 1.8x and 1.72x respectively, short vary but 1.33x on avg
  - case2: long avg 1.67x, medium median 1.6x, short avg 1.38x
  - case3: long median 2x, medium avg 1.87x, short avg 1.42x
- bench1:
  - case0: long avg 2.2x, medium avg 1.75x, short avg 1.63x
  - case1: long avg 2.24x, medium avg 2.44x, short avg 2.4x
  - case2: long avg 1.74x, medium avg 1.57x, short vary, avg 1.0x
  - case3: long median 2.3x, medium avg 2.66x, short vary, 1.87x on avg
- bench2:
  - case0: long median 1.96x, medium avg 1.59x, short vary, on avg 1.05x
  - case1: long avg 1.96x, medium 1.57x median, short avg 1.26x
  - case2: long avg 1.7x, medium 1.51x median, short avg 0.94x
  - case3: long median 2.16x, medium avg 1.96x, short median 2.14x

draghan:
> bumpability:
     - core->GetGame()->CombatCounter can be done once in FindPath (once per frame/tick is enough); we could also add a convenience function
      - pcf_ea if oldValue != newValue
     - IsMoving: Movable::SetStanceDirect

I think I need some more guidance here @lynxlynxlynx .

1) If `core->GetGame()->CombatCounter` is non-zero, then what should happen? Should I do bulk update for all actors on the map, should they turn non-traversable (non-bumpable)? And then revert back when it switches to zero?

2) `Pcf-ea`: if for a particular actor old value is different than new value, what should happen to this actor's traversability? I assume you are talking about this one, it that right? https://github.com/gemrb/gemrb/blob/4facc6ca07d46ccfb790760149a29c5026f98cfa/gemrb/core/Scriptable/Actor.cpp#L953

3) `Movable::SetStanceDirect`: what should I do to traversability for particular StanceIDs?

lynxlynxlynx:
These are all conditions from `Actor::ValidTarget`, so just mirror that.
1. In this case nobody can bump or get bumped, so the other bumpability checks are redundant.
2. In that case check the new value with the same check as in `ValidTarget` — enemies can't get bumped.
3. Just toggle with `IsInMovingStance`.

draghan:
I've implemented first part of updates on the fly, just the position changes for now, treating all actors as non-bumpable (so the results can have some outliers comparing to valid representation, but it should give close enough results to evaluate).
I've run first benchmarks on bench0 to see where we stand. 
Attached [firstUpdatesOnTheFly.zip](https://github.com/user-attachments/files/21090987/firstUpdatesOnTheFly.zip).

Performance-wise it already looks terrible, it basically butchers gains for short and medium paths and sometimes even long ones.

Position update from one actor on my main rig takes ~5-15µs, but there are dozens and sometimes hundreds of them accumulating when actors are moving, resulting in more frame time wasted than saved.

I don't even implemented full updates yet, which will only worsen the results.

This does not look good and TBH I don't have much hope for this approach.

lynxlynxlynx:
Can you push the diff (compared to this PR) somewhere?

draghan:
I've pushed that to this branch: https://github.com/draghan/gemrb/tree/pathfinderNoBulkUpdate
Check today's commits.
It's dirty atm, as it was not intended for reviewing right now.

lynxlynxlynx:
While it could be improved further, I didn't think of the main problem before: `FindPath` is usually called rarely, but `DoStep` updates the position all the time when moving, every few ticks depending on the movement speed. Even if that did some cheaper differential update of the cache, a cost would be there for each step for each actor. So it's not reconcilable, at least not with this type of cache (not that I can think of a better scheme).

So I agree, let's go with the initial idea.

draghan:
I don't discard this yet, I do have an idea how to make this a bit smarter. This won't change the frequency characteristics of updates on the fly, but maybe can marginalize the impact.
Can the blocking shape change for an actor over time, like I don't know, maybe for shapeshifters?

lynxlynxlynx:
Rarely, but yes, exactly in such cases (polymorph, shapechange); basically the animation type changes and that means the associated data is all new. The challenge is that this is done via effects, so the animation id changes every tick. It doesn't mean we'd need to update the cache every tick though, since we also store the value from the previous tick in PrevStats and that can used in pcf_anim_id to detect real changes.

draghan:
> I do have an idea how to make this a bit smarter. This won't change the frequency characteristics of updates on the fly, but maybe can marginalize the impact.

From the first quick evaluation, I think I additionally halved the cache update time by removing re-evaluation of the blocking shape from the cache update and one additional change in cache representation, sneak peak from bench0 case0:
```
baseline $ update:
    Mean extra time per measurement: 76.88 μs
    Median extra time per measurement: 78.00 μs

improved $ update:
    Mean extra time per measurement: 30.35 μs
    Median extra time per measurement: 31.50 μs
```

This is not solving the updates' on the fly tho, they are just simply too frequent; but for the bulk update it lands nicely.

Once I'll test this a bit more and solve any edgecases I could find and perform complete suite of benchmarks, I'll update the PR.

gemrb/core/TraversabilityCache.h
bradallred: I would prefer to use `Region` and if there is a demonstrable issue then open an issue for that and address it. We dont _need_ those fields obviously. If all our compilers support the anonymous structures extension then just use that instead. Otherwise we can discuss options in that issue and if need be remove those fields.

bradallred: something like this should work if we can use the "GNU" c++ dialect.

```diff
 public:
 	// WARNING: we reinterpret_cast this struct!
 	// Point and Size must be first and in this order
-	Point origin;
-	Size size;
+	union {
+		Point origin;
+		struct {
+			// these must be the same type and order as Point
+			int& x;
+			int& y;
+		};
+	};
 	
-	// unfortunately anonymous structs are an extension in C++...
-	int& x = origin.x;
-	int& y = origin.y;
-	int& w = size.w;
-	int& h = size.h;
+	union {
+		Size size;
+		struct {
+			// these must be the same type and order as Size
+			int& w;
+			int& h;
+		};
+	};
 
 	Region() noexcept = default;
 	Region(int x, int y, int w, int h) noexcept;
```

draghan: I don't know about that GNU dialect, but AFAIK in standard C++ it's an UB to use plain union for type punning, and considering the portability of the project - we really don't want to do that, do we?

Separate issue for that is fine by me, I have kept somewhere measurements with vs without use of this approach to back it up (and to validate if it even does anything in the first place).
But I would really like to keep this FitRegion here, if not for final merge, just for the sake going through this PR: the measurements I've done so far are including this change; ditching it now rises a need to re-do the benchmarks to have comparable data and those are taking a few hours of my time to complete.

lynxlynxlynx: For the current users there's no need for a `Point` or `Size` alternative, the FitRegions are populated in one place from actor data and then only their data members get accessed.

But yeah, the other question remains too — would anything else benefit from a leaner `Region`? Unless we find it in another hot path, it's probably not worth the bother.

draghan: Getting rid of excessive fields is good not only for time savings, but also obviously reduces memory footprint.
For this patch I don't really care about that aspect (we're trading here shorter execution times for bigger runtime footprint anyway), but the engine will overall benefit from this reduction in general, when applied globally.

What I meant previously - if that's bothering you, I can remove this `FitRegion` before merging this PR, with the assumption we will address this right after; but I'd like to finish work on this patch with this `FitRegion` present on my end as is, if that's not a problem.

> At the end of the day, I'm personally more concerned with cleaner code than squeezing performance with like this. I suspect there are lots of better targets for performance gains. 

Of course they are, they're waiting "just around the corner", I briefly discussed my plans in the discord channel: this patch was meant to be only an appetizer but it already yields surprisingly good gains (perhaps because I'm paying attention to little details like this one :).
Next on the list is tackling the priority queue implementation (I have a funny feeling that even brute force linear search of the whole open nodes' space every iteration would be more performant than the current implementation, but who know, only the benchmarks will tell) and reducing branches in the pathfinding algorithm.
As for the clean code... I have some thoughts about this topic and most of them are not very positive. :) Anyway, it's very subjective term, I would rather talk objective numbers.
I'm here just for the sole reason of targeting performance problems in the pathfinder. Reducing weight of the `Region` instances resulted in (variable between test cases and test benches, but visible) couple of percentage points of improvements in short and medium paths, so here it is.

> However, if the motivation behind the two rects is performance based we should be aware of the penalties we may incur from creating new Point or Size structs that would come alongside such a change.

Hmm, I don't think I understand the point. What new Point or Size do you have in mind?

draghan:
Updated version, a bit more than I initially wanted to do, but it paid off. I don't think I could pull off any meaningful improvements to this part.

Major change is using the traversability cache cell value as a collection of tokens instead as the absolute mark; previous version wasn't working for tracking multiple actors on the same navmap point, current version solves it practically without any overhead (an `add` instead of a `write`) and it also reduces some previous overhead of re-drawing neighbouring actors.

Added cache for blocking shape for actors, this means we won't constantly call `actor->IsOver()` in the cache updates. This neatly stabilizes the short paths and pulls their improvements closer to the long and medium ones.

Updated benchmark results are attached [74dded91f.zip](https://github.com/user-attachments/files/21335323/74dded91f.zip).

Brief dump of stats excluding outliers, for details - as always - please look at the reports ("extra" here means extra tracked time, and we've tracked only cache updates):
- bench0:
  - case0: long(avg=2.00x, med=2.00x), med(avg=1.94x, med=1.99x), short(avg=1.79x, med=1.70x), extra(avg=37.58 μs, med=39.00 μs, total=4.47 ms, count=120)
  - case1: long(avg=1.98x, med=2.04x), med(avg=2.14x, med=2.08x), short(avg=1.70x, med=1.82x), extra(avg=14.71 μs, med=7.00 μs, total=10.78 ms, count=741)
  - case2: long(avg=1.66x, med=1.67x), med(avg=1.64x, med=1.61x), short(avg=1.42x, med=1.37x), extra(avg=32.19 μs, med=32.00 μs, total=4.96 ms, count=155)
  - case3: long(avg=2.23x, med=2.22x), med(avg=2.32x, med=2.29x), short(avg=1.77x, med=1.70x), extra(avg=9.30 μs, med=6.00 μs, total=18.71 ms, count=2281)
- bench0w:
  - case0: long(avg=1.81x, med=1.76x), med(avg=1.65x, med=1.76x), short(avg=1.69x, med=1.52x), extra(avg=35.48 μs, med=37.00 μs, total=3.41 ms, count=97)
  - case1: long(avg=1.76x, med=1.76x), med(avg=1.68x, med=1.66x), short(avg=1.56x, med=1.62x), extra(avg=5.48 μs, med=4.00 μs, total=5.38 ms, count=1060)
  - case2: long(avg=1.66x, med=1.67x), med(avg=1.57x, med=1.58x), short(avg=1.35x, med=1.37x), extra(avg=32.62 μs, med=31.00 μs, total=3.52 ms, count=109)
  - case3: long(avg=1.97x, med=1.99x), med(avg=1.93x, med=1.94x), short(avg=1.54x, med=1.53x), extra(avg=16.81 μs, med=14.00 μs, total=53.33 ms, count=3290)
- bench1:
  - case0: long(avg=2.15x, med=2.18x), med(avg=1.99x, med=1.96x), short(avg=2.00x, med=1.85x), extra(avg=67.43 μs, med=69.00 μs, total=6.34 ms, count=95)
  - case1: long(avg=2.23x, med=2.21x), med(avg=2.35x, med=2.37x), short(avg=2.29x, med=2.26x), extra(avg=21.93 μs, med=11.00 μs, total=8.93 ms, count=410)
  - case2: long(avg=1.73x, med=1.73x), med(avg=1.76x, med=1.60x), short(avg=1.29x, med=1.29x), extra(avg=66.17 μs, med=67.00 μs, total=8.14 ms, count=124)
  - case3: long(avg=2.32x, med=2.29x), med(avg=2.67x, med=2.69x), short(avg=2.13x, med=2.28x), extra(avg=34.73 μs, med=18.00 μs, total=35.25 ms, count=1196)
- bench2:
  - case0: long(avg=1.97x, med=1.97x), med(avg=1.57x, med=1.48x), short(avg=1.20x, med=1.21x), extra(avg=164.64 μs, med=159.00 μs, total=16.14 ms, count=99)
  - case1: long(avg=2.15x, med=2.15x), med(avg=1.86x, med=2.06x), short(avg=1.31x, med=1.34x), extra(avg=54.73 μs, med=26.00 μs, total=42.69 ms, count=798)
  - case2: long(avg=1.69x, med=1.69x), med(avg=1.42x, med=1.45x), short(avg=1.18x, med=1.12x), extra(avg=153.14 μs, med=140.50 μs, total=18.07 ms, count=119)
  - case3: long(avg=2.18x, med=2.16x), med(avg=2.02x, med=2.13x), short(avg=1.15x, med=1.17x), extra(avg=286.05 μs, med=110.00 μs, total=138.45 ms, count=497)

lynxlynxlynx:
Nice to see also the short distances cases pick up. 
I don't have any architectural comments to the last few changes. Something else remains that we already discussed on Discord — the unnecessary and wrong use of `sizeFactor`. The `FitRegion` comment is most elegantly resolved with a dependency-free struct. Other than that, I think you can start with the final cleanup.

draghan:
I should have some time this week to work on that.

Few questions @lynxlynxlynx :
1) What is the version of clang format you use on the CI? My local version is clearly not impressing your CI. ^^'
2) What would be the final decision for the `FitRegion`? I could remove it, as per Brad's request, this isn't an issue as long as we plan to address it soon.

About the `sizeFactor` - I'm not sure yet how to approach this; when I was looking at that few weeks ago, the calculations of `IsOver` seemed to be independent of `sizeFactor` at a glance, yet removing it and replacing with some integer multiple of radius was giving not optimal shapes - too big or too small, depending on the circle size, while using the `sizeFactor` was giving the perfect boundaries. There has to be some hidden connection. I must understand better what are the dependencies there under the hood.

lynxlynxlynx:
1. It was created with clang-format 17, but I've used later versions as well, so I'm surprised you'd get many differences. Especially since we're explicit as possible in the config.
2. Like I said, I'd just go for simplistic `struct { int, int, int, int }` (plus the smaller set of methods, of course). Region itself can be looked at later.
3. `sizeFactor` is only used for drawing the actual selection ellipse on master and I don't see your code changing that. Typically it has a value of 1, which only changes if you hover over the actor or its portrait (or it's talking). Then it oscillates between 1 and 1.2. So even if it mattered, it's not likely you've often hit it in tests. So I'd say you're seeing ghosts; perhaps from one of your interim versions?

lynxlynxlynx:
Running it manually over the whole codebase barely changes anything here. So my guess would be our .clang-format is not being picked up or overriden by your editor:
```sh
$ clang-format --version
clang-format version 20.1.8
$ clang-format -i gemrb/**/*.[ch] gemrb/**/*.cpp
$ git diff
diff --git a/gemrb/core/Game.h b/gemrb/core/Game.h
index bb8402131..4aa5beb4e 100644
--- a/gemrb/core/Game.h
+++ b/gemrb/core/Game.h
@@ -55,9 +55,9 @@ enum BlitFlags : uint32_t;
 #define SX_COMBAT 4 //combat xp, adjusted by difficulty
 
 //joinparty flags
-#define JP_JOIN    1 //refresh join time
-#define JP_INITPOS 2 //init startpos
-#define JP_SELECT  4 //select the actor after joining
+#define JP_JOIN     1 //refresh join time
+#define JP_INITPOS  2 //init startpos
+#define JP_SELECT   4 //select the actor after joining
 #define JP_OVERRIDE 8 // replace actor, if any
 
 //protagonist mode
diff --git a/gemrb/core/Interface.cpp b/gemrb/core/Interface.cpp
index 079b73282..ca86232b3 100644
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -4155,7 +4155,7 @@ void Interface::ApplyTooltipDelay() const
                                              tooltipDelay * 21;
        } else {
                delay = tooltipDelay >= 100 ? std::numeric_limits<int>::max() :
-                                            tooltipDelay * 25;
+                                             tooltipDelay * 25;
        }
 
        WindowManager::SetTooltipDelay(delay);
```

draghan:
> Like I said, I'd just go for simplistic struct { int, int, int, int } (plus the smaller set of methods, of course). Region itself can be looked at later.

Ok, sure. If that's the case, I will minimize it further, we don't even need 4x32 bits there.

> sizeFactor is only used for drawing the actual selection ellipse on master and I don't see your code changing that. Typically it has a value of 1, which only changes if you hover over the actor or its portrait (or it's talking). Then it oscillates between 1 and 1.2. So even if it mattered, it's not likely you've often hit it in tests. So I'd say you're seeing ghosts; perhaps from one of your interim versions?

Yeah, you're totally right. I have no idea what that was about, I must have had broken something in the solution or perhaps I've hallucinated (how can I be sure whether I am still a human at this point??). I'll remove references to `sizeFactor`.

As for the style checks - I don't know. I used standalone clang-format. I'll re-format with the newer version, we'll see.

Another question: do you know by any chance if there are bigger (in context of blocking shape) creatures than dragons in any of supported games?

lynxlynxlynx:
Dragons are the biggest, nobody has a circle size greater than 7.

As for the format, 17 was the chosen base, since that's what the runner had at the time. So PR's are checked specifically against clang-format-17. Considering the results with 20, I doubt there were back-and-forth changes in versions between the two.

gemrb/core/TraversabilityCache.h
draghan: As established with Jaka, I have left `FitRegion` here. Moved it to inner scope, added an explaining comment.

gemrb/core/PathFinder.cpp
draghan: Done.

lynxlynxlynx:
Ok, the appveyor failure can be ignored, it's not fixable in this PR. A new master build would have to be triggered for its cache to update, then this here synced ... tedious, since we can't freeze the whole image.
I'll look at the final diff tomorrow, thank you very much!

lynxlynxlynx:
Wohoo!

PR #2359, opened 2025-07-24
bg1 GUIRECCommon.py: fixed character record when bonus priest spells are present

mark-314159:
## Description
In BG1, opening up character record for a non-priest character, then selecting a cleric/druid (e.g. Jaheira) results in the following Python stack trace:

[Python/ERROR]: TypeError: an integer is required (got type str)
[Python/ERROR]: 
The above exception was the direct cause of the following exception:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUIRECCommon.py", line 829, in TypeSetStats
[Python/ERROR]:     res.append (GemRB.GetString (strref) + ': ' + str (val) + stattype)
[Python/ERROR]: SystemError: <built-in function GetString> returned a result with an error set
[Python/ERROR]: 
During handling of the above exception, another exception occurred:
[Python/ERROR]: TypeError: an integer is required (got type tuple)
[Python/ERROR]: 
The above exception was the direct cause of the following exception:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUICommonWindows.py", line 670, in SelectionChanged
[Python/ERROR]:     SelectionChangeHandler ()
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUICommonWindows.py", line 569, in <lambda>
[Python/ERROR]:     selectionHandler = lambda win=window, fn=selectionHandler: fn(win)
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUIREC.py", line 214, in UpdateRecordsWindow
[Python/ERROR]:     Text.SetText (GetStatOverview (pc))
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUIREC.py", line 272, in GetStatOverview
[Python/ERROR]:     outputtext+= GUIRECCommon.GetBonusSpells (pc)
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUIRECCommon.py", line 768, in GetBonusSpells
[Python/ERROR]:     return TypeSetStats (stats, pc)
[Python/ERROR]:   File "/usr/local/share/gemrb/GUIScripts/GUIRECCommon.py", line 847, in TypeSetStats
[Python/ERROR]:     res.append (GemRB.GetString (s))
[Python/ERROR]: SystemError: <built-in function GetString> returned a result with an error set

Issue is that GetBonusSpells() on line 764 resolves the strref string with what looks like the SPELLLEVEL token, then passes in the stattype: 'r'. So stattype 'r' wasn't covered and GemRB.GetString was being called on a string and failing.

### Other issues observed - probably for a different PR:

Original game:
Original BG1 char record

GemRB (with fix):
GemRB BG1 char record with fix

Newline formatting appears to be different in GemRB, but I haven't worked out why yet. The [p]...[/p] tags seem to be created correctly, but leaving for another time.

Additionally, the "Reaction" is +3 in the original game, but 11 in GemRB, again for reasons that currently elude me.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/GUIRECCommon.py
+++ b/gemrb/GUIScripts/GUIRECCommon.py
@@ -761,7 +761,7 @@ def GetBonusSpells(pc, expand = True):
 			if GameCheck.IsPST ():
 				stats.append ((4239, bonus, 'p'))
 			else:
-				stats.append ((GemRB.GetString (10345), bonus, 'r'))
+				stats.append ((GemRB.GetString (10345), bonus, 'p'))
 
 	if expand:
 		stats.append ("\n")
@@ -801,10 +801,14 @@ def TypeSetStats (stats, pc, recolor = False):
 			elif stattype == 'd': # strref is an already resolved string
 				res.append (strref)
 			elif stattype == 'p': # a plus prefix if positive
+				# special handling to cover bonus priest spells where string is resolved earlier
+				resolvedString = strref
+				if isinstance (strref, int):
+					resolvedString = GemRB.GetString (strref)
 				if val > 0:
-					res.append (GemRB.GetString (strref) + ': +' + str (val))
+					res.append (resolvedString + ': +' + str (val))
 				else:
-					res.append (GemRB.GetString (strref) + ': ' + str (val))
+					res.append (resolvedString + ': ' + str (val))
 			elif stattype == 's': # both base and (modified) stat, but only if they differ
 				base = GB (val)
 				stat = GS (val)
@@ -840,7 +844,7 @@ def TypeSetStats (stats, pc, recolor = False):
 					if res:
 						res[-1] += s
 				else:
-					res.append (s);
+					res.append (s)
 			elif recolor:
 				res.append (won + GemRB.GetString (s) + woff)
 			else:


lynxlynxlynx:
Thanks, but it would be better to just use the 'p' mode instead, it will simplify everything.

mark-314159:
Would you prefer something more like this then?

			elif stattype == 'p': # a plus prefix if positive
				# special handling to cover bonus priest spells where string is constructed earlier
				resolved_string = strref
				if isinstance (strref, int):
					resolved_string = GemRB.GetString (strref)
				if val > 0:
					res.append (resolved_string + ': +' + str (val))
				else:
					res.append (resolved_string + ': ' + str (val))

Adding in a new 'r' mode seemed more specific, and would continue to error out if there were other genuine cases that haven't been accounted for. But I did modify the 'p' case first, then thought the alternative looked better, but happy to change it to whatever style you prefer.

lynxlynxlynx:
No, it's just that this was the last use of r that somehow got missed when cleaning up. There's no need to modify p, just change the caller to not resolve the string itself. You can look at the other case in the branching — they'll be identical save for the different strref value.

mark-314159:
The "Level 1: +2" then gets reported as "Level 7: +2" as I think the call to GemRB.SetToken ("SPELLLEVEL", str (level + 1)) affects the GemRB.GetString() method?
Screenshot from 2025-07-24 16-32-32

lynxlynxlynx:
GetString is what actually resolves the embedded token, but I see now why this is different. It's because by using p the resolution is deferred and by that time all the loop iterations have passed, so all the tokens will resolve to the same, last, value. Unfortunate, but we do need to handle this one case separately.

So yeah, please use the second approach, but change resolved_string to camel case.

lynxlynxlynx:
Perfect, thanks!

Issue #2361, opened 2025-08-10
Audio occlusion


lynxlynxlynx:
Apparently the games supported audio occlusion, some sort of naive muffling that was used together with roll-off on supported cards. 

To quote @skellytz, who is actively researching it:

> The original BG2 supported EAX 2.0 and made use of its [occlusion effects](https://www.youtube.com/watch?v=it5AcT3BuTU). Back when we used compatible sound cards, many sounds farther away from the center of the screen were muffled like this:
> Attached File  [bg2-occlusion.ogg](https://www.shsforums.net/index.php?app=core&module=attach&section=attach&attach_id=31473) 
>
> This effect is different from environmental presets, reverb, or fading. At the time, most other games used occlusion for physical obstacles like walls or underwater effects ([example](https://www.youtube.com/watch?v=it5AcT3BuTU)). BG2 used it for sound sources farther away from the center of the screen, even without any obstacles, which I felt was kind of an unpolished feature, to be honest. Still, it did process sounds farther away better than without occlusion enabled.

MarcelHB:
Looks somewhat difficult to model this for an iso perspective without incurring other oddities.

Simply taking the distance without having any obstacle to take into account (as in the example) isn't exactly natural sound, at least not for distances we are looking at here.

Taking the center of the screen again, and using in-line-of-sight to any source as a condition could possibly result in weird audio when PCs and sources are scattered around the screen, also when thinking about the many caves and scrolling over no-go areas on such maps (we may be in the void, so what do we hear, nothing, everything, ...?).

The best thing maybe is to check whether none of the PCs has a LOS of at least screen size distance, or maybe more, and then considering the source as muffled, e. g. for being behind a door in a room that currently nobody is in. But that's also modelling sound that is inaccurate when everybody is just around the corner, although I think I've heard this kind of implementation in many games.

Something that could _mildly_ help here without doing so much math on our own is OpenAL's directional source and cone feature, e. g. specifying the orientation's 120° cone of a caster which should made the sound muffle when facing the edge of the screen, though I haven't tried it out as far as I can remember.

lynxlynxlynx:
While we could experiment with using doors, wallgroups and/or the searchmap (for something closer to the base concept), I think it's easiest to just wait for more information.

MarcelHB:
A simple, physically oversimplified but cheap and forgiving approach could be to check whether the source is below fog or in an undiscovered area, since this could simply match the impression of not having anything in sight.

[...]

Issue #383, opened 2019-11-12
We need individual speech channels


lynxlynxlynx:
Our audio is either queued or it hard stops whatever is playing and starts the new sound. This latter approach is used for the so called speech channel, since you don't want the actor to continue yapping if he received a new order or to immediately react to taking damage.

But we have only one global speech channel, so it can happen that when things are busy, actors cut each other off. It's visible in some bg2 cutscenes, for example #376. So it appears we need individual channels instead, however openal also has some limits or problems with high channel counts (IIRC), so we must think it through. Perhaps something in between will be the right approach, with a pool of channels, not one per actor. And maybe we overuse it at certain times.

MarcelHB:
I don't have the overview over all channels we use right now, but there are two simple options that we can combine:

* Doing a hard cut over the number of sources because everything above is actually annoying - let's say 16.
* Having pools (let's say GUI, music, effects, voices) and doing first-come-first-served per pool as it looks difficult to reliably query the actual number of concurrently playable sources in OpenAL. On the other hand, if there are too many for the implementation, playback will fail in some nth call - so bad luck.

Anyway, my presumably hardware-backend Creative OAL keeps the CPU notably busy already.

bradallred:
I think music already has a dedicated channel so that is covered.

A first-come-first-serve pool is fine, but you still need a way to associate a source (actor, projectile, etc) with a channel because often you need to stop it before the sound finishes. You would need to do something like return a monotonic id (just returning the channel id is not good enough) when we play a sound so the source can hang onto it and stop it if it is still playing without stomping on somebody else. The driver would be able to tell if the passed id is still valid.

How do we handle this for sound fx? certainly there are even more of those going on over the top of each other than voices. is this a problem for those too? if not then couldn't this problem be solved by the same method used for fx?

lynxlynxlynx:
We have sound handles for this, so that's covered. The only special thing about speech is that it stops the previous sound.

lynxlynxlynx:
IESDP also confirms that speech channels were per-actor:

> WaitForVoiceChannel() Variants: [BGEE/PSTEE]
> 	
>   This action delays subsequent script actions until the voice channel of the active creature is ready to play a new sound.

lynxlynxlynx:
Doing some digging in the repo it shows it used up to 21 channels in some sort of a priority queue, reusing the last one if exhausted. But this the overall system, not limited to speech. 

[...]

PR #2363, opened 2025-08-13
Audio: directional sound for spell casting

MarcelHB:
## Description

A minor experiment after reading #2361. It's even dampening (as in the example) and makes use of something OpenAL provides and that we can use umambiguously.

directional_sound

So having the screen at the same position, the direction determines some interpolated loss of gain when outside of the purple cone. Sounds like this when considering the three directions (E, S, W) on the screenshot.
[out.webm](https://github.com/user-attachments/assets/6e09410e-6f46-4b31-b47f-334a4af39f5f)

## Checklist

### DIFF
--- a/gemrb/core/Audio/AudioSettings.cpp
+++ b/gemrb/core/Audio/AudioSettings.cpp
@@ -143,6 +143,24 @@ AudioPlaybackConfig AudioSettings::ConfigPresetDialog(SFXChannel channel) const
 	return config;
 }
 
+// Directional (casting voices)
+AudioPlaybackConfig AudioSettings::ConfigPresetDirectional(SFXChannel channel, const Point& p, orient_t orientation) const
+{
+	auto config = ConfigPresetSpatialVoice(channel, p);
+	config.directional = true;
+	config.cone = 120;
+
+	using t = std::underlying_type_t<orient_t>;
+	// 0 is south, and 90 deg turn is west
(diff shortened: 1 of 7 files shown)


MarcelHB:
Let me try something more I found in the docs ...

MarcelHB:
Nice, turns out we get a low pass filter for free on directional sound in the EFX, too. Updated the example above, sounds even better now.

lynxlynxlynx:
Maybe I'm deaf, but the only difference I hear between the first two is that the first one is more pronounced on the left channel (as expected), while the right one *is* muffled. Is that a testing artifact, I'd expect the same muffle left and right?

MarcelHB:
No, this is how it's supposed to be. Screen remains centered and so the listener position, caster is to the far western edge of the screen, so generally the sound always is dominant on the left channel:

* sample 1: caster casts towards the center (eastwards)
* sample 2: caster casts towards the south, so there is a 90° difference in the direction, and just a slight fade
* sample 3: caster casts away from center, towards the west -- high freqs are muffled a little

Looks like this:
directional_sound_freq

lynxlynxlynx:
Oh, from the picture I thought you were always casting towards the south and then measuring on different sides. Clearly not, all is fine.

lynxlynxlynx:
Thanks! If you're itching for more audio improvements, I suggest #383, so it gets cut off less often.

PR #2366, opened 2025-08-15
WEDImporter: prevents crash when there are no door tiles

MarcelHB:
## Description
When there are no door tiles, using `[0]` now raises some assertion error for me.

## Checklist


lynxlynxlynx:
Thanks!

PR #2370, opened 2025-08-15
More OOB fixes

MarcelHB:
## Description
After encoutering another assertion failure, I looked through the code for static indices to containers that 

1. are dynamically sized, potentially empty, and
2. not guarded otherwise already (like `i == 0`, or `.empty()`),
3. or use an OOB index as termination pointer

Of course, there is maybe more around but that slipped through my search pattern for now. Anyway, please double check.

## Checklist


gemrb/plugins/TLKImporter/TLKImporter.cpp
```diff
@@ -364,7 +364,7 @@ String TLKImporter::GetString(ieStrRef strref, STRING_FLAGS flags)
 		str->ReadDword(StrOffset);
 		str->ReadDword(l);
 
-		if (type & 1) {
+		if (type & 1 && l > 0) {
```
lynxlynxlynx, line 367: Doesn't this potentially cause similar problems in ResolveTags later, since it's accessing [0] of an empty string? It would be cleaner to just immediately return if l == 0, since the rest of the function can't do anything useful any more.

MarcelHB: Yes, that looks bad on its own, but there is only this use here. Then I'll make it drop out with an empty string instead.

lynxlynxlynx:
Thanks!

PR #2379, opened 2025-09-07
Implement IE rolling methodology and fix +/- ability buttons in bg1 and iwd

dolio:
## Description
This PR makes two changes to character generation.

1. In BG1, BG2 and IWD, I implemented the actual rolling method from the infinity engine. The way this works is as follows:
   - For each score, roll 3d6 _repeatedly_, with modifiers, until the result falls within the allowed range
   - If the total scores are less than 75, reroll all scores as above
 
   BG2 was already doing the 75 part while the other two weren't. Previously gemrb was just setting low scores to the minimum. This makes a big difference in some cases. For instance, paladins have a minimum charisma of 17. In the actual infinity engine, you have a 1/4 chance to get an 18, because there are 3 ways to roll 17 and 1 way to roll 18. In gemrb you'd have 1/125 chance of an 18.

   I also switched back to 3d6 instead of 3d5+3, because the latter would bias the distribution _upwards_ while doing the actual IE stuff. Previously, with 3d5+3 and _without_ rerolling until 75, the average gemrb BG/IWD roll would be worse than the minimum actual IE roll, I think.
2. I ported the way the +/- ability score buttons work in BG2 back to BG1 and IWD. The latter had a problem where if you press-and-hold on one of those buttons, it will actually modify the _previously clicked_ ability score. So, if you click +strength, then hold down the -dexterity button, points would be taken out of strength.

I didn't touch IWD2, because I know less about it, and my goal was to fix the first part (the second part was just something I encountered while trying things out).

I tested locally by overriding the GUIScripts directory in some .cfg files to copies with my changes. I think that should be fine, since these are all python script changes.

## Checklist

### DIFF
--- a/gemrb/GUIScripts/bg1/GUICG4.py
+++ b/gemrb/GUIScripts/bg1/GUICG4.py
@@ -172,12 +172,12 @@ def OnLoad():
 
 		Button = AbilityWindow.GetControl(i*2+16)
 		Button.OnPress (LeftPress)
-		Button.SetVarAssoc("Ability", i )
+		Button.SetValue(i)
 		Button.SetActionInterval (200)
 
 		Button = AbilityWindow.GetControl(i*2+17)
 		Button.OnPress (RightPress)
-		Button.SetVarAssoc("Ability", i )
+		Button.SetValue(i)
 		Button.SetActionInterval (200)
 
 	TextAreaControl = AbilityWindow.GetControl(29)
@@ -191,10 +191,10 @@ def OnLoad():
 	AbilityWindow.ShowModal(MODAL_SHADOW_GRAY)
 	return
 
-def RightPress():
+def RightPress(Button):
 	global PointsLeft
 
-	Abidx = GemRB.GetVar("Ability")
+	Abidx = Button.Value
 	Ability = GemRB.GetVar("Ability "+str(Abidx) )
 	CalcLimits(Abidx)
 	GemRB.SetToken("MINIMUM",str(Minimum) )
@@ -223,10 +223,10 @@ def JustPress():
 	TextAreaControl.SetText(AbilityTable.GetValue(Abidx, 1) )
 	return
 
-def LeftPress():
+def LeftPress(Button):
 	global PointsLeft
 
-	Abidx = GemRB.GetVar("Ability")
+	Abidx = Button.Value
 	Ability = GemRB.GetVar("Ability "+str(Abidx) )
 	CalcLimits(Abidx)
 	GemRB.SetToken("MINIMUM",str(Minimum) )
--- a/gemrb/GUIScripts/iwd/CharGen.py
+++ b/gemrb/GUIScripts/iwd/CharGen.py
@@ -1247,13 +1247,13 @@ def AbilitiesPress():
 		AbilitiesPlusButton = AbilitiesWindow.GetControl (16 + i * 2)
 		AbilitiesPlusButton.SetState (IE_GUI_BUTTON_ENABLED)
 		AbilitiesPlusButton.OnPress (AbilitiesPlusPress)
-		AbilitiesPlusButton.SetVarAssoc ("AbilityIndex", i + 1)
+		AbilitiesPlusButton.SetValue(i)
 		AbilitiesPlusButton.SetActionInterval (200)
 
 		AbilitiesMinusButton = AbilitiesWindow.GetControl (17 + i * 2)
 		AbilitiesMinusButton.SetState (IE_GUI_BUTTON_ENABLED)
 		AbilitiesMinusButton.OnPress (AbilitiesMinusPress)
-		AbilitiesMinusButton.SetVarAssoc ("AbilityIndex", i + 1)
+		AbilitiesMinusButton.SetValue(i)
 		AbilitiesMinusButton.SetActionInterval (200)
 
 	AbilitiesStoreButton = AbilitiesWindow.GetControl (37)
@@ -1322,11 +1322,11 @@ def AbilitiesLabelPress():
 	AbilitiesTextArea.SetText (AbilitiesTable.GetValue (AbilityIndex, 1) )
 	return
 
-def AbilitiesPlusPress():
+def AbilitiesPlusPress(btn):
 	global AbilitiesWindow, AbilitiesTextArea
 	global AbilitiesMinimum, AbilitiesMaximum
 
-	Abidx = GemRB.GetVar ("AbilityIndex") - 1
+	Abidx = btn.Value
 	AbilitiesCalcLimits(Abidx)
 	GemRB.SetToken ("MINIMUM", str(AbilitiesMinimum) )
 	GemRB.SetToken ("MAXIMUM", str(AbilitiesMaximum) )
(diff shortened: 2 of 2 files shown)


lynxlynxlynx:
Thanks, taking a look. IWD2 and PST use point buy, so it doesn't matter there.

gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -1415,28 +1415,37 @@ def AbilitiesRerollPress():
...
+
+			total += Value
+
+			GemRB.SetVar ("Ability" + str(i + 1), Value)
+			Label = AbilitiesWindow.GetControl (0x10000003 + i)
+
+			if i==0 and HasStrExtra and Value==18:
+				Label.SetText("18/"+str(e) )
+			else:
+				Label.SetText(str(Value) )
```
lynxlynxlynx, line 1450: It's the same code in all three files. Please make a new function in gemrb/GUIScripts/GUICommonWindows.py and just use it instead.

gemrb/GUIScripts/bg1/GUICG4.py
```diff
@@ -87,21 +87,33 @@ def RollPress():
...
+	while total < 75:
+		total = 0
+
+		for i in range(6):
+			CalcLimits(i)
+
+			v = 0
+			# roll until the result falls in the allowed range
+			while v < Minimum or v > Maximum:
+				v = GemRB.Roll(dice, size, Add)
```
lynxlynxlynx, line 105: Both while loops could loop (seemingly) indefinitely if unlucky or bad data was provided. Perhaps add a guard to be on the safe side? E.g. break out after 10 iterations.

gemrb/GUIScripts/bg1/GUICG4.py
```diff
@@ -167,11 +167,13 @@ def OnLoad():
 
 		Button = AbilityWindow.GetControl(i*2+16)
 		Button.OnPress (LeftPress)
+		Button.SetValue(i)
 		Button.SetVarAssoc("Ability", i )
```
lynxlynxlynx, line 171: Using both SetValue and SetVarAssoc does not make sense and they also modify the same data. If the vars are not needed any more, remove the association calls.

gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -1350,11 +1352,11 @@ def AbilitiesPlusPress():
 			AbilitiesDoneButton.SetState (IE_GUI_BUTTON_ENABLED)
 	return
 
-def AbilitiesMinusPress():
+def AbilitiesMinusPress(btn):
 	global AbilitiesWindow, AbilitiesTextArea
 	global AbilitiesMinimum, AbilitiesMaximum
 
-	Abidx = GemRB.GetVar ("AbilityIndex") - 1
+	Abidx = btn.Value - 1
```
lynxlynxlynx, line 1359: These -1s look wrong, don't you end up with [-1,4] now?

dolio: That's what I thought, but when I tried not subtracting the buttons ended up always modifying the wrong thing. So `GemRB.GetVar("AbilityIndex")` must be actually returning the index, and the rest of the code is actually written for the `[-1,4]` range.

I could try fixing that as well.

gemrb/GUIScripts/iwd/CharGen.py
lynxlynxlynx: It seems feasible for the whole rolling process, even though globals are involved. But I suggest you first fix up this PR and then implement DRY as a separate commit.

gemrb/GUIScripts/iwd/CharGen.py
lynxlynxlynx: AbilityIndex is set up to be always +1 in iwd, unlike the other games — it's nicely visible in the diff. And then below this line where Abidx is used.

gemrb/GUIScripts/bg1/GUICG4.py
dolio: I've added tests to ensure that infinite loops don't happen with bad data.

However, I don't think it worth worrying about unlucky situations. For one, no, you cannot just break out after 10 iterations. That will still give incorrect statistics in some cases. In the paladin case, 10 iterations is only 17% likely to roll either a 17 or 18, so 83% of the time the number will not actually be sampled correctly. Even 100 iterations would mean that ~15% of the time the number is not being sampled correctly.

But, I've clicked the reroll button on a paladin many times while doing this, and never saw a perceptible delay. I tried putting an extra loop around this one (the individual stat roll) that repeats it 10,000 times, and there is still almost never a perceptible delay. And the way things are actually set up for these games, the iterations spent within the loop make it more likely that the outer loop will finish, because higher minimums make the rolls more likely to add up to 75 or more.

Basically, people don't get unlucky enough for that aspect to be a problem.

dolio:
When I factored out the label code, I used the one that will actually display `18/01`, since some would instead display `18/1`. I also tweaked it to show `18/00` instead of `18/100`, since the former is what the original games show (and is the D&D terminology).

gemrb/GUIScripts/bg1/GUICG4.py
lynxlynxlynx: We have to be defensive because these tables are something mods change.

gemrb/GUIScripts/iwd/CharGen.py
lynxlynxlynx: How come you split out just the label bit? You could pass `globals()` along if you don't want too many parameters.

dolio: Because I don't know what you actually want split out. This review comment shows the 4 lines I split out. When I asked, you said not to split out the entire rolling process yet. Also, the prior line appears to reference a variable that is unique to each file, and I'm not familiar enough with python to know what to do about that.

I didn't even write this part of the code, just indented it in two places. Which parts do you want factored out?

lynxlynxlynx: Then we had a misunderstanding, that's all. I asked for other comments to be resolved first, since it could impact the deduplication and it's always better to do that separately, so it doesn't hide any interim changes to logic. Now the third commit is more like something that should be split up, the fixes merged into the original commits, while the separation stays. Or I can just call it a day and squash merge everything together.

The whole rolling could be shared. It's true that IWD would need to be first changed to use "Ability " instead of "Ability" and 0 based indexing if we want to do it properly, but there are two shortcuts:
- two more parameters (var name, value offset)
- use GameCheck.IsIWD1() to ifdef the different behaviour for now (IMHO cleaner)

BG1 and BG2 already look identical. You can ignore IWD if this is too hard, I can fix it later.

lynxlynxlynx:
Do you need any help, did you plan on finishing this PR?

dolio:
Sorry, I've been a bit busy on other stuff.

I rebased the commits to fold the changes into the original two commits.

However, to be honest, I don't know how to simply factor out the rolling code. It's true that it's essentially the same in all the modules. However, it uses _several_ global variables that get set in other procedures. I don't see how the rolling code could be factored out without also factoring out functions that set the globals. But that includes e.g. `OnLoad` for the BG modules, but it is rather different for IWD, because everything is in the single CharGen module for that.

So, this could all be reworked to either not use globals, or to pick out all the global parts involved in ability rolling and move them to the common file. But that seems pretty significantly beyond the scope of what I was trying to fix here.

lynxlynxlynx:
Ok, thanks, I'll take over from there.

lynxlynxlynx:
It's done, all three games now use the same code for the whole abilities process, split out into a shared GUICG4 module. It's all the same besides the little prev/next differences that are chosen based on game type.

PR #2401, opened 2025-11-10
Improved UI performance on IWD2

MarcelHB:
## Description
I noticed that on slow devices IWD2 UI drawing performance remains rather low, like making a difference from 20-30fps to 40-50fps when hiding the UI. I guess it's the UI's structure that is constantly dirty with the pause button animation keeping everything busy in the hierarchy.

Yet, it shouldn't be this way when pausing, but the debug view shows red even if the animation has stopped:

iwd2_pause_btn

Here I just added that a button shouldn't report `IsAnimated` when the underlying animation has stopped, e. g. because of being in a paused state.

## Checklist

### DIFF
--- a/gemrb/core/GUI/Button.cpp
+++ b/gemrb/core/GUI/Button.cpp
@@ -342,7 +342,7 @@ void Button::SetState(State state)
 
 bool Button::IsAnimated() const
 {
-	if (animation) {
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Nice, thanks!

PR #2405, opened 2025-11-23
BG1 EE MVP

MarcelHB:
## Description
Just to check how far we can go. It's something, now, at least we can load a save game (GUICG needs some fixes and changes to UI, races, classes ...) and much of the ordinary stuff in-game looks good, too:

bg1ee1

bg1ee2

## Checklist

### DIFF
--- a/gemrb/GUIScripts/GameCheck.py
+++ b/gemrb/GUIScripts/GameCheck.py
@@ -76,7 +76,7 @@ def IsTOB ():
 	return GemRB.HasResource ("worldm25", RES_WMP) and GemRB.GetVar("oldgame") == 0
 
 def HasTOB ():
-	return GemRB.HasResource ("worldm25", RES_WMP)
+	return not IsBGEE() and GemRB.HasResource ("worldm25", RES_WMP)
 
 def HasHOW ():
 	return GemRB.HasResource ("expmap", RES_WMP)


lynxlynxlynx:
Hmm, that's an odd choice, why not just reuse the bg2 GUIScripts? The two new engines were the same after all.

MarcelHB:
So you'd try to patch the bg2 scripts here instead?

lynxlynxlynx:
Yes, exactly. Besides the Start scripts (somehow I've seen two different main menus before), I don't expect much need for change. Going the bg1 route would require patching all of its non-shared stuff.

MarcelHB:
Then it may look like this. The game control buttons apparently still need attention, due to some incompatibilities in `bg2/Game.py`, and something in GUICG.

gemrb/GUIScripts/bg2/Start.py
```diff
@@ -94,6 +136,8 @@ def OnLoad():
 		if GameCheck.IsBG2EE ():
 			GemRB.PlayMovie ("logo", 1)
 			GemRB.PlayMovie ("intro", 1)
+		elif GameCheck.IsBGEE ():
+			GemRB.PlayMovie ("logo", 1)
```
lynxlynxlynx: You can merge this with the above check and use `IsAnyEE()`. All the ees have the same two movies, except for pstee that uses introh. But handling iwdee and pstee will be a harder challenge anyway.

lynxlynxlynx:
Thanks. The unhardcoded files were copied from bg2ee?

MarcelHB:
> Thanks. The unhardcoded files were copied from bg2ee?

No, I took the BG1 files, but a custom `guils.chu`, a modified `gemrb.ini` (some on EE effects set to 1) and `avatars.2da` from BG2EE.

lynxlynxlynx:
I think that will be problematic too. Remember, they took the bg2 engine, upgraded it and ported bg1 to it to get bgee. So besides the graphics, it's all mostly the same when compared to bg2ee. And closer to bg2 than bg1. Things like the walking timestep definitely need to be from bg2.

gemrb/GUIScripts/bg2/LoadScreen.py
```diff
@@ -34,7 +34,24 @@ def StartLoadScreen ():
...
+	tmp = Table.GetRowCount ()
+	tmp = GemRB.Roll (1, tmp-1, 0)
+	HintStr = Table.GetValue (tmp, 0)
+
+	Bar = LoadScreen.GetControl (0)
+	Bar.AddAlias("LOAD_PROG")
+	Bar.SetVarAssoc ("Progress", 0)
+	Bar.OnEndReached (lambda: EndLoadScreen(HintStr))
+
+	# original has no load screen, just use own simple one simple instead
```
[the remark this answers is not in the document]
MarcelHB: Oh yeah. I don't see any and also no suitable spare assets. I built a mostly empty window with a title and a progress bar.

MarcelHB:
> I think that will be problematic too. Remember, they took the bg2 engine, upgraded it and ported bg1 to it to get bgee. So besides the graphics, it's all mostly the same when compared to bg2ee. And closer to bg2 than bg1. Things like the walking timestep definitely need to be from bg2.

Ok, then I copy this one over, too. But aside from this case, I noticed that there are different 2DAs required when it comes to classes so I took the conservative approach regarding the assets.

lynxlynxlynx:
Understandable, since you started with the bg1 scripts ...

gemrb/GUIScripts/bg2/Start.py
lynxlynxlynx: Actually my copy of bgee shows both back to back and then I'm thrown to a different start screen:
image

I knew I remembered something was odd with the start screens — there are several options.

MarcelHB: I've never seen this ... mine is v2.6.6.0, if that's the missing detail.

MarcelHB:
I now copied over the unhardcoded from BG2EE except one line for the window title in `gemrb.ini` and my `guils.chu`.

GUICG still lacks half-orcs and breaks at classes, but I guess that's the general state there right now?

lynxlynxlynx:
Any uses of IsBG2EE need to be rechecked and most will have to be converted to IsAnyEE. That's one thing, but I'm surprised halforcs don't work, as that should all be dictated by data, not code.

MarcelHB:
Half-orcs is just a strref matter. I'll have a look at the rest of the game checks.

lynxlynxlynx:
BreakableWeapons should be set to 1.

MarcelHB:
There are too many things looking wrong, even when naively poking them by trying `IsAnyEE()`. Like wrong strrefs, bad panel IDs, or whatever I cannot identify just by looking at the effect.

I'd just put the few simple things I can address here, the rest just probably be done GUI by GUI.

lynxlynxlynx:
It's a shame the strrefs don't match, but yeah, that's /just/ about rechecking all the ones in 2das and perhaps some hardcoded ones.

lynxlynxlynx:
Ah, sorry, turns out they unhardcoded BreakableWeapons, so it should be at 0:
> each normal weapon has a 1% chance to fire a spell of the same name (e.g. sw1h01.spl for bastard swords) that does the actual breakage. The spells use a chapter-based 318 to block them starting in chapter seven.

lynxlynxlynx:
Thanks!

Issue #821, opened 2020-06-15
Decomissioning plan for sdl1


lynxlynxlynx:
Initially when discussing #441 last year I was sceptical of just dumping sdl1 for several reasons:
- it was the more mature backend on master
- it seemed work upstream continued still
- it would provide a nice base to test sdl2 experimentation against
- it'd be nice to warn people before completely dropping it

Focusing more on subviews with sdl1 again, I see we have several bugs, some whack-a-moly, so perhaps it's better to dump it sooner than later. Nobody is doing 1.2.16 either, commits are just piling up.

Considering everything, I propose we do the anniversary release from master, which is just shy of getting the finished pathfinder rewrite and with few regressions, announce the EOL of our sdl1 backend at the same time and then proceed to rip it out as soon as it is convenient. I don't think we can get subviews to release quality in a month and a half, otherwise I'd propose it for that.

Thoughts?

bradallred:
we might as well wait till subviews is merged. The SDL 2 driver on master is horrible performance wise and doesn't offer much in terms of benefit. I also don't see a point in removing it over just letting it bit rot and accept contributions from anybody willing to submit them a la SDL itself is doing.

That said it certainly shouldn't be default once subviews is merged (and should probably issue a deprecated warning when building). There is at least one platform (Mac) where the performance of SDL 1.2 is completely unplayable and I expect this situation to continue to degrade (for all platforms) in more ways than over time.

MarcelHB:
No general objection.

We should only find some way to make SDL2 and subviews code work without the special OpenGL stuff of #441, even at the cost of some visual artifcats, in a release distribution since this will otherwise exclude some comparably young platforms from running (#810), given the age of the original games.

Other than that: Easier CMake, no dual-maintenance, so I welcome that.

bradallred:
> We should only find some way to make SDL2 and subviews code work without the special OpenGL 

that's not going to happen without writing a shader for every graphics backend you want to support. its simply not possible to do in a way that doesn't both require tons of code and rewriting of things and also at the significant expense of performance. At that level of work you might as well introduce a new backend that extends the SDL 2 backend with SDL_gfx (assuming that actually is capable of what we need) or another cross platform abstraction layer with the capabilities required.

MarcelHB:
No, I just think of opt-out'ing the shader code at runtime if the alternative is not being able to play at all, at the cost of living with things like #775.

bradallred:
[...]

PR #2406, opened 2025-11-30
Zooming

MarcelHB:
## Description
Inspired from a feature I've been using in the EE frequently (due to general overview or better micro managing), I don't see a reason to not generally introduce this to all games: using the mouse wheel to zoom in and out. Currently two steps in, two steps out:

zoom_out

zoom_in

Things I've tested in particular:
* Weather
* Shaking on small rooms (Oswald's vessel)
* Dialogs
* Overlays
* Movement speeds (using the edges and dragging the cursor)
* SDL v1.2, but it mostly gets swallowed by other issues, see https://github.com/gemrb/gemrb/issues/821#issuecomment-3593000903

Performance increases on slow device when zooming in, or slows down when zooming out, so this is some more ad-hoc way to tune this, e. g. in some combats.

## Checklist

### DIFF
--- a/gemrb/core/GUI/GameControl.cpp
+++ b/gemrb/core/GUI/GameControl.cpp
@@ -2863,9 +2863,12 @@ int GameControl::GetOverheadOffset() const
 	return 0;
 }
 
-unsigned int GameControl::GetZoomLevel() const
+void GameControl::SetScalePercent(unsigned int level)
 {
-	return zoomLevel;
+	auto value = Clamp(level, 25u, 160u);
+	zoomLevel = (value - 20) / 5;
+
+	ScaleViewport(100);
 }
 
 unsigned int GameControl::GetScalePercent() const
--- a/gemrb/core/GUI/GameControl.h
+++ b/gemrb/core/GUI/GameControl.h
@@ -232,7 +232,7 @@ public:
 	void ToggleAlwaysRun();
 	int GetOverheadOffset() const;
 	void TryDefaultTalk() const;
-	unsigned int GetZoomLevel() const;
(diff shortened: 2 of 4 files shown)


lynxlynxlynx:
Cool, LGTM and it works fine with my touchpad too (at least when the wayland scaling doesn't trigger, since it takes over).

Is there a particular reason you store the zoom in each window? For savegame and actions compatibility, we'll need to have a global setting. We already read and write it from the GAM file:
https://github.com/search?q=repo%3Agemrb%2Fgemrb%20zoom&type=code

There was already a working zoom here, but since it's entangled with 4K support and not yet polished, I wouldn't say this was wasted effort:
https://github.com/Goddard/gemrb/tree/pie4k

MarcelHB:
> Is there a particular reason you store the zoom in each window? For savegame and actions compatibility, we'll need to have a global setting. We already read and write it from the GAM file: https://github.com/search?q=repo%3Agemrb%2Fgemrb%20zoom&type=code

Mainly because I was looking for where the draw buffer handling is done. Anyway, `gameWin` is the only, the global user for this, so that's fine.

I wasn't aware of the GAM file, I'll have a look at it for the EEs.

MarcelHB:
I now mapped the 27 zoom steps of the original EE.

Anyway, I found something in handling SDL that is visible in some literal edge cases of interior maps. Will look at this tomorrow.

MarcelHB:
Testing savegames in the original EE looks broken before this already, but other than this I'm good for now.

lynxlynxlynx:
Thanks!

burner1024:
This is amazing! Thank you!

PR #2414, opened 2025-12-10
Improvement to drawing speed

MarcelHB:
With zooming and something like FullHD settings, we can exert some more pressure to the renderer and see some potential bottlenecks. What you see here is the profiling of one single map tile:

tracy_perf_transparency

On a fast computer, the drawing is faster than a call to `HasTransparency` that takes around 50% of map drawing time, even falling below my display Hz on debug builds.

We don't need this call to `SDL_GetColorKey`, we have all we need to know in the format descriptor already, and there is no change to this from outside the sprite class, resulting in some 80 to 100% improvement in some circumstances.

## Checklist


lynxlynxlynx:
Nice, thanks!

Issue #2411, opened 2025-12-08
Rare assertion crash in path finding


MarcelHB:
#### Bug description
It's hard to reproduce but in about one of five loads of a BG1 candle keep (prologue) save game, the game hits an assertion violation.

#### Screenshots
```
C:/MSYS2/mingw64/include/c++/15.2.0/array:210: std::array<_Tp, _Nm>::value_type& std::array<_Tp, _Nm>::operator[](size_t
ype) [with _Tp = float; long long unsigned int _Nm = 512000; reference = float&; size_type = long long unsigned int]: As
sertion '__n < this->size()' failed.

Thread 1 hit Breakpoint 1.2, 0x00007ff87c17f1e7 in msvcrt!abort () from C:\Windows\System32\msvcrt.dll
(gdb) bt
#0  0x00007ff87c17f1e7 in msvcrt!abort () from C:\Windows\System32\msvcrt.dll
#1  0x00007ff84649ab7b in ?? () from C:\MSYS2\mingw64\bin\libstdc++-6.dll
#2  0x00007ffff274f6f3 in std::array<float, 512000ull>::operator[] (
    this=0x7ffff2d607a0 <GemRB::Map::FindPath(GemRB::Point const&, GemRB::Point const&, unsigned int, unsigned int, int,
 GemRB::Actor const*)::open+4096000>, __n=512000) at C:/MSYS2/mingw64/include/c++/15.2.0/array:210
#3  0x00007ffff2682392 in GemRB::BucketPriorityQueue::CostPointBuckets::PushPoint (
    this=0x7ffff29787a0 <GemRB::Map::FindPath(GemRB::Point const&, GemRB::Point const&, unsigned int, unsigned int, int,
 GemRB::Actor const*)::open>, bucketIdx=5119, point=..., cost=5159.40283)
    at D:/Sources/gemrb/gemrb/core/BucketPriorityQueue.h:136
#4  0x00007ffff26825d5 in GemRB::BucketPriorityQueue::Push (
    this=0x7ffff29787a0 <GemRB::Map::FindPath(GemRB::Point const&, GemRB::Point const&, unsigned int, unsigned int, int,
 GemRB::Actor const*)::open>, point=..., cost=5159.40283) at D:/Sources/gemrb/gemrb/core/BucketPriorityQueue.h:48
#5  0x00007ffff25950f9 in GemRB::Map::FindPath (this=0x50917690, s=..., d=..., size=0, minDistance=0, flags=5,
[...]

PR #2415, opened 2025-12-14
Fix pathfinding crash in Candlekeep.

draghan:
Fix for #2411. Path requested from a seagull had max costs oscillating around 5500, which exceeded our buckets availability.

Extended bucket count by 1024 and added an extra storage at the end of the buckets' arrays, to accommodate any potential overshoots.


lynxlynxlynx:
LGTM, thanks!

lynxlynxlynx:
It makes me wonder how come it was possible though. Seagulls fly, so there's effectively no searchmap obstacles for them except for map borders. Looking at the code it's `GameScript::RandomFly` which is just a series of moves in straight lines, no obvious `FindPath` involved.

PR #2435, opened 2026-02-21
Implement gamepad controls

draghan:
Both axis are working, all wired buttons are working, double click is working, dragging is working, text input via D-pad is working.
Additionally fixed an issue with map scrolling via keys, where the up+right diagonal movement was being cancelled by not wrong bit pattern assumption.


lynxlynxlynx:
Thanks!

PR #2449, opened 2026-03-24
Add v0.9.5 release to `metainfo.xml`.

fpiesche:
## Description
Simple as that; the `metainfo.xml` is used by e.g. Flathub to determine the version number for the package, and by graphical package managers to display release information.

## Checklist


lynxlynxlynx:
Thanks!

PR #2468, opened 2026-04-13
2446: Deduplicate the portrait image selection code

rsantellan:
## Description
I have created a new class **gemrb/GUIScripts/GUIPortraitCommon.py** that has the basic for the portrait movement, selection, custom selection.
I also fix an issue with the custom portrait in iwd2 that is not opening.

What I couldn't do is try to replicate the behavior of the class: **GUICG13.py** that calls everything on the load because I keep the methods that make different logic in each game so the class is more simple and not making to pass everything as parameter.

I have used snake case on the function names, but maintained the style of the code that I was copying from.

## Checklist


lynxlynxlynx:
The last commit should be reverted, all the files should use tabs for indentation.
Why did you go with snake case — you can see everything is using camel case. That's not maintaining style.
The OnLoads and what iwd is using are still duplicating a lot of code. If you deal with that you'll see the end result will be closer to what the code looked like before and you also won't need so many indirection functions.

rsantellan:
> The last commit should be reverted, all the files should use tabs for indentation. Why did you go with snake case — you can see everything is using camel case. That's not maintaining style. The OnLoads and what iwd is using are still duplicating a lot of code. If you deal with that you'll see the end result will be closer to what the code looked like before and you also won't need so many indirection functions.

Tonight I will revert it, I have to change that the IDE use tabs instead of spaces. I will change the names to camel case and look at the OnLoads!

lynxlynxlynx:
What's up with the unrelated spacing regressions in gemrb/GUIScripts/iwd/CharGen.py?

rsantellan:
> What's up with the unrelated spacing regressions in gemrb/GUIScripts/iwd/CharGen.py?

I don't see any spacing on that file, all are tabs. Is there a place where I can look up the linter used ? So I can use the same one, the IDE mess up withe the spaces and tabs first and now this.

lynxlynxlynx:
In the diff you can see random spaces were removed before opening parenthesis `(`.
We don't have a prescribed linter for python. The style checker on PRs only checks a few minor things. It's best to disable any autoformatting.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+	GemRB.SetVar ("Gender",0) #scrapping the gender value
+	return
+
+def PortraitSetPicture(PortraitButton, PortraitsTable, LastPortrait):
+	picture_size = "L"
+	if GameCheck.IsBG1OrEE() or GameCheck.IsIWD1():
+		picture_size = "G"
+	PortraitName = PortraitsTable.GetRowName (LastPortrait)+picture_size
+	if GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		PortraitButton.SetPicture (PortraitName, "NOPORTLG")
```
lynxlynxlynx, line 397: the backup bam (both) is used in several places. I suggest you resolve it only once to simplify —  you can do it in the global scope. It should be used everywhere, so this if check will become redundant, the call is currently missing a parameter.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+	if GameCheck.IsIWD2():
+		next_script = "CharGen"
+	GemRB.SetNextScript (next_script)
+	GemRB.SetVar ("Gender",0) #scrapping the gender value
+	return
+
+def PortraitSetPicture(PortraitButton, PortraitsTable, LastPortrait):
+	picture_size = "L"
+	if GameCheck.IsBG1OrEE() or GameCheck.IsIWD1():
+		picture_size = "G"
```
lynxlynxlynx, line 394: There's still plenty of snake_case in new code ...

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: Similarly for the large suffix

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+	Button.OnPress(custom_done)
+	if GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		Button.SetDisabled(True)
+	else:
+		Button.SetState(IE_GUI_BUTTON_DISABLED)
+
+	Button = Window.GetControl(7)
+	Button.SetText(15416)
+	if GameCheck.IsIWD2():
+		Button.MakeEscape()
```
lynxlynxlynx, line 349: this doesn't need to be conditional

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+	if CustomWindow:
+		CustomWindow.Close()
+	if AppearanceWindow:
+		if GameCheck.IsBG1OrEE():
+			AppearanceWindow.ShowModal (MODAL_SHADOW_GRAY) # narrower than CustomWindow, so borders will remain
+		elif GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+			AppearanceWindow.ShowModal (MODAL_SHADOW_NONE) # narrower than CustomWindow, so borders will remain
+
+# This is for applying
+def PortraitApplySelection(AppearanceWindow, LastPortrait):
```
lynxlynxlynx, line 220: no need to pass AppearanceWindow around, since it's a global. This is true for some other functions as well

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+		else:
+			next_func()
+
+# Button abort function
+def PortraitButtonCustomAbort():
+	PortraitCustomAbort(AppearanceWindow)
+
+# Button custom press, must pass function because of BG1
+def PortraitButtonCustomPress(custom_done_func = None):
+	PortraitCustomPress(PortraitsTable, LastPortrait, lambda: PortraitButtonCustomDone(custom_done_func), PortraitButtonCustomAbort)
```
lynxlynxlynx, line 155: If you move the game-specific logic into this module, this and similar jumping through hoops will not be necessary.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+				LastPortrait += 1
+				if LastPortrait >= PortraitsTable.GetRowCount():
+					LastPortrait = 0
+					if flag:
+						PortraitSetPicture(PortraitButton, PortraitsTable, LastPortrait)
+						break
+					flag = True
+
+	if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		AppearanceWindow.Focus()
```
lynxlynxlynx, line 121: probably doesn't need to be conditional

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+				if PortraitsTable.GetValue(LastPortrait, 0) == Gender:
+					PortraitSetPicture(PortraitButton, PortraitsTable, LastPortrait)
+					break
+				LastPortrait += 1
+				if LastPortrait >= PortraitsTable.GetRowCount():
+					LastPortrait = 0
+					if flag:
+						PortraitSetPicture(PortraitButton, PortraitsTable, LastPortrait)
+						break
+					flag = True
```
lynxlynxlynx, line 118: Isn't the iwd2 logic enough here? LastPortrait is set already at the start of the function

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
+# SPDX-FileCopyrightText: 2011 Contributors to the GemRB project <https://gemrb.org>
```
lynxlynxlynx, line 1: 2026

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,400 @@
...
+	# Load the Gender
+	Gender = GetGenderFunc()
+	
+	# Optional extra setup
+	if ExtraSetupFunc:
+		ExtraSetupFunc(AppearanceWindow)
+
+	# Load the Portraits Table
+	PortraitsTable = GemRB.LoadTable("PICTURES")
+	if not GameCheck.IsIWD1():
```
lynxlynxlynx, line 34: this is fine for iwd1 too, randomization is welcome everywhere

gemrb/GUIScripts/bg1/GUICG12.py
```diff
@@ -2,215 +2,20 @@
...
-LastPortrait = 0
-Gender = 0
-
-def SetPicture ():
-	global PortraitsTable, LastPortrait
-
-	PortraitName = PortraitsTable.GetRowName (LastPortrait)+"G"
-	PortraitButton.SetPicture (PortraitName)
-	return
+from GUIPortraitCommon import *
```
lynxlynxlynx, line 9: Please use normal imports, star imports are only used for files with constants.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: Have searched and remove all of the snake_case vars

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: @lynxlynxlynx do you mean the `Button.MakeEscape()` ? It only appears on IWD 2

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: Living on the past... Fixing it.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: If I leave only the iwd2 logic on the other when you select a female character the first portrait is from a male.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: I though about that but didn't want to import all the:
```
from ie_stats import *
import CharGenCommon
import CharOverview
``` 
And iwd1 is different on the calls that make. Should I send everything to that class and not leave anything on the outside modules ?

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: Yes. It just binds the escape key to the button.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: Yes, ideally everything would be here. I think you'll only need to do extra imports for bg1 with the back/next callback. Maybe bg2 for the CharGenState stuff.

Later things can be further simplified by storing the gender in the character stat immediately, I see not all games do that now, hence you have so many different GetGender functions instead of one.

Longterm we want to unify the character generation processes, since now almost every game has a different implementation, even though, except for iwd2 and pst, they're the same. Extracting common logic like this is a good and simpler step before doing any major refactoring.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: Hmm, that needs more looking into — the randomization above looks fine, is it missing something? Look at pictures.2da for the game you're testing with.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: The get gender part was the easy part, the problem is that BG1 and BG2 use CharGenCommon in different way, and I can't import both of them on the created class. I try some hacks searching on internet and they don't work and are really horrible.

lynxlynxlynx: The trick is to not import them at global scope, but in their own individual ones only when needed. There won't be conflicts that way. Look at LUSpellSelection.py for how next() is handled, as an example.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: @lynxlynxlynx I search and couldn't find the missing parameter. You mean for SetPicture or for another method ?

lynxlynxlynx: Line 399 is missing it.

rsantellan: But on gemrb/GUIScripts/iwd2/CGPortrait.py line 20 and gemrb/GUIScripts/bg1/GUICG12.py line 24 there is not a second parameter ? Should I unify all the calls with the second parameter ?

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: I leave like IWD2 and now is working, but I have two weird behaviors. 
- If I choose male, cancel, choose female sometimes it leaves the first picture as male. 
- One time it give this error:
```
[ResourceManager]: Searching for 'G'...
[ResourceManager/WARN]: Couldn't find 'G'... Tried G.bmp G.pvrz G.mos G.png 
[GUIScript/ERROR]: Runtime Error:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "~/gemrb/gemrb/GUIScripts/MetaClasses.py", line 45, in handler
    return f(*args, **kwargs)
[Python/ERROR]: RuntimeError: Picture resource not found!
[Python/ERROR]: 
During handling of the above exception, another exception occurred:
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "~/gemrb/gemrb/GUIScripts/bg1/GUICG12.py", line 10, in OnLoad
    SetupAppearanceWindow()
    ~~~~~~~~~~~~~~~~~~~~~^^
[Python/ERROR]:   File "~/gemrb/gemrb/GUIScripts/GUIPortraitCommon.py", line 106, in SetupAppearanceWindow
    PortraitSetPicture()
    ~~~~~~~~~~~~~~~~~~^^
[Python/ERROR]:   File "~/gemrb/gemrb/GUIScripts/GUIPortraitCommon.py", line 422, in PortraitSetPicture
    PortraitButton.SetPicture (PortraitName)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
[Python/ERROR]:   File "~/gemrb/gemrb/GUIScripts/MetaClasses.py", line 47, in handler
    raise type(e)(str(e) + "\nMethod Docs:\n" + str(f.__doc__))
[Python/ERROR]: RuntimeError: Picture resource not found!

Method Docs:
===== Button_SetPicture =====

**Metaclass Prototype:** SetPicture (PictureResRef[, DefaultResRef])

**Description:** Sets the Picture of a Button Control from a BMP file or a Sprite2D.

**Parameters:**
  * PictureResRef - the name of the picture (a .bmp resref) or a Sprite2D object
  * DefaultResRef - an alternate bmp should the picture be nonexistent

**Return value:** N/A

**See also:** [Button_SetPicture](Button_SetPicture.md), [Button_SetPLT](Button_SetPLT.md), [Button_SetSprites](Button_SetSprites.md), [Button_SetPictureClipping](Button_SetPictureClipping.md), [Window_SetPicture](Window_SetPicture.md)
```
None of the two error can reproduce it.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: > the backup bam (both) is used in several places. I suggest you resolve it only once to simplify — you can do it in the global scope. It should be used everywhere, so this if check will become redundant, the call is currently missing a parameter.

I don't understand what do you want to do with this one.

lynxlynxlynx: Unify? Yes, a fallback bam should always be provided.

The second comment: Now you have to use GameCheck almost every time you use SetPicture because the fallback bam differs between the games. If you immediately figure out which bam is correct when the module loads, you won't have to check every time.

gemrb/GUIScripts/iwd2/CGPortrait.py
```diff
@@ -2,211 +2,10 @@
...
-	PortraitTable = GemRB.LoadTable ("pictures")
-	PortraitName = PortraitTable.GetRowName (LastPortrait )
-	GemRB.SetToken ("SmallPortrait", PortraitName+"S")
-	GemRB.SetToken ("LargePortrait", PortraitName+"L")
-	GemRB.SetVar ("PortraitIndex", LastPortrait)
-	GemRB.SetNextScript ("CharGen2") #Before race
-	return
+def OnLoad():
+	
+	SetupAppearanceWindow()
```
lynxlynxlynx: Nice, see how now 3 files are now the same? This means they could be shared outside the per-gametype dirs, so there's 1 file instead of 3. But if it remains like this, where SetupAppearanceWindow doesn't need any parameters, even further simplification is possible. If you add this same OnLoad function to GUIPortraitCommon, the games that currently use these duplicate files can just call SetNextScript("GUIPortraitCommon") instead of SetNextScript("GUICG2")  (or similar) and the individual files removed altogether.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: The error is because you're not passing any parameters, at least in the committed version.

gemrb/GUIScripts/iwd2/CGPortrait.py
rsantellan: I can't seem to remove the callback for iwd1, my attempts are failing. I remove the other ones but can't figure out how to remove that one.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: > > the backup bam (both) is used in several places. I suggest you resolve it only once to simplify — you can do it in the global scope. It should be used everywhere, so this if check will become redundant, the call is currently missing a parameter.
> 
> I don't understand what do you want to do with this one.

@lynxlynxlynx still don't know what to do with this.

lynxlynxlynx: This part should be in the global scope and then EmptyPortrait used everywhere possible. Probably there need to be two or three vars for all the portrait sizes.
```
	if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo() or GameCheck.IsIWD1():
		EmptyPortrait = "NOPORTMD"
	else:
		EmptyPortrait = "NOPORTLG"
```
In other words, no function should have any direct mention of NOPORT*.

gemrb/GUIScripts/iwd2/CGPortrait.py
lynxlynxlynx: It doesn't work if you import CharGen and then do the calls with it?
```
	CharGen.RaceButton.MakeDefault()
	CharGen.CharGenState = 1
	CharGen-SetCharacterDescription()
```

rsantellan: Ohh I didn't think about that, I was trying to import the functions that I needed and wasn't working. That works, so now there is none callback function anymore. I have clean up a little bit the class removing unused methods.
You said that **even further simplification is possible** what do you mean by that ?

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: @lynxlynxlynx do you have any example that I can look into ?

gemrb/GUIScripts/iwd2/CGPortrait.py
lynxlynxlynx: That's what I was saying, don't use "from imports" unless it's for constants only. I think you snuck in some more.

The further simplification you already did by calling GUIPortraitCommon directly.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx, line 137: isn't this param now redundant?

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: I don't know what more I can say, I already gave you part of the code (it's existing code btw). Maybe reread this whole thread?

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -15,14 +15,17 @@
 PortraitsTable = 0
 LastPortrait = 0
 Gender = 0
+EmptyPortrait = "NOPORTMD"
```
lynxlynxlynx: maybe EmptyPortraitMedium would be clearer, since that's what it is

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -15,14 +15,17 @@
...
-	SetupAppearanceWindow()
-
-def SetupAppearanceWindow():
 	global AppearanceWindow, PortraitButton, PortraitsTable, LastPortrait
-	global Gender
+	global Gender, EmptyPortrait, EmptyPortraitBig
 
+	if GameCheck.IsBG1OrEE():
+		EmptyPortrait = EmptyPortraitBig
+
```
lynxlynxlynx: do this above, so it's all in one place.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -235,12 +233,7 @@ def PortraitApplySelection():
 def PortraitCommonLargeCustom():
 	global PortraitList1, PortraitList2
 	global RowCount1
-	global CustomWindow
-
-	if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo() or GameCheck.IsIWD1():
-		EmptyPortrait = "NOPORTMD"
-	else:
-		EmptyPortrait = "NOPORTLG"
+	global CustomWindow, EmptyPortrait
```
lynxlynxlynx: You don't need to mention global vars if you don't modify them.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -348,23 +341,20 @@ def PortraitCustomPress(CustomDone):
 	LargeSuffix = "L"
```
lynxlynxlynx: these l/m/s/g suffixes should get the same treatment

rsantellan: The G what stand for ?

- l is Large
- m is Medium
- s is Small
- g is ?

lynxlynxlynx: maybe giant, I don't know. You can look at the files for hints. It's used further down this module.

Also, it became even more complicated with the ees, so we'll probably need more. See discussion here:
https://www.gibberlings3.net/forums/topic/29946-tutorial-and-library-universal-portrait-packs/

So it's probably best if you use a dictionary instead, then EmptyPortrait["small"] and so on.

gemrb/GUIScripts/iwd/CharGen.py
```diff
@@ -19,6 +19,7 @@
 import LUSkillsSelection
 import LUCommon
 import PaperDoll
+from GUIPortraitCommon import SetupAppearanceWindow
```
lynxlynxlynx: this is now redundant

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+	# Optional extra setup
+	if GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		# HACK
+		import CharGenCommon
+		CharGenCommon.PositionCharGenWin(AppearanceWindow, -6)
+
+		TextAreaControl = AppearanceWindow.GetControl(7)
+		if TextAreaControl:
+			TextAreaControl.SetText("")
+	if GameCheck.IsIWD2():
```
lynxlynxlynx: ok, just nitpicking left, please bear with me. :)
This can be an elif, a game can only be of one type.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		import CharOverview
+		CharOverview.PositionCharGenWin(AppearanceWindow)
+		
+	# Load the Portraits Table
+	PortraitsTable = GemRB.LoadTable("PICTURES")
```
lynxlynxlynx, line 54: just for easier searching, please lowercase the table name.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+	PortraitsStart = PortraitsTable.FindValue(0, 2)
+	FemaleCountAdd = 1
+	LastPortraitOffset = 0
+	if GameCheck.IsIWD2():
+		FemaleCountAdd = 0
+		LastPortraitOffset = -1
+	FemaleCount = PortraitsTable.GetRowCount() - PortraitsStart + FemaleCountAdd
+	print(FemaleCount)
```
lynxlynxlynx: can go

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		LastPortrait = GemRB.Roll(1, FemaleCount, PortraitsStart - 1)
+	else:
+		LastPortrait = GemRB.Roll(1, PortraitsTable.GetRowCount() - FemaleCount, LastPortraitOffset)
+	# Controls
+	PortraitButton = AppearanceWindow.GetControl(1)
+	PortraitButton.SetFlags(IE_GUI_BUTTON_PICTURE | IE_GUI_BUTTON_NO_IMAGE, OP_SET)
+	PortraitButton.SetState(IE_GUI_BUTTON_LOCKED)
+
+	LeftButton = AppearanceWindow.GetControl(2)
+	if GameCheck.IsIWD1():
```
lynxlynxlynx: are these two iwd2 checks hurtful in other games? I imagine we don't need to ifdef the button enablement.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		RightButton.SetState(IE_GUI_BUTTON_ENABLED)
+		RightButton.SetFlags(IE_GUI_BUTTON_RADIOBUTTON, OP_OR)
+		
+	BackButton = AppearanceWindow.GetControl(5)
+	if GameCheck.IsIWD1():
+		BackButton.SetText(13727)
+	else:
+		BackButton.SetText(15416)
+	
+	if GameCheck.IsIWD2() or GameCheck.IsIWD1():
```
lynxlynxlynx: this can be the default cancel everywhere, just like done is the default accept.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		BackButton.MakeEscape()
+		
+	CustomButton = AppearanceWindow.GetControl(6)
+	CustomButton.SetText(17545)
+	if GameCheck.IsIWD1():
```
lynxlynxlynx: doesn't need to be conditional

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		CustomButton.SetState(IE_GUI_BUTTON_ENABLED)
+
+	DoneButton = AppearanceWindow.GetControl(0)
+	if GameCheck.IsIWD2():
+		DoneButton.SetText(36789)
+	else:
+		DoneButton.SetText(11973)
+		if GameCheck.IsIWD1():
```
lynxlynxlynx: same here

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+			DoneButton.SetState(IE_GUI_BUTTON_ENABLED)
+	DoneButton.MakeDefault()
+
+	# Events
+	RightButton.OnPress(PortraitButtonRightPress)
+	LeftButton.OnPress(PortraitButtonLeftPress)
+	if GameCheck.IsBG1OrEE():
+		# HACK
```
lynxlynxlynx: please remove all these hack comments. I know some were from the original files, but these are not really hacks.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+def GetGender():
+	if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		return GemRB.GetVar("Gender")
+	else:
+		MyChar = GemRB.GetVar("Slot")
+		return GemRB.GetPlayerStat(MyChar, IE_SEX)
+	
+# Button right press function	
+def PortraitButtonRightPress():
+	global PortraitsTable, LastPortrait
```
lynxlynxlynx: PortraitsTable doesn't need to be on the list

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+
+# Button left press function	
+def PortraitButtonLeftPress():
+	global LastPortrait
+	LastPortrait = PortraitPrev()
+	PortraitSetPicture()
+
+# Button custom done
+def PortraitButtonCustomDone():
+	global AppearanceWindow
```
lynxlynxlynx: no need for this global

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+	PortraitPictureButtonName = PortraitCustomDone()
+	if GameCheck.IsBG1OrEE():
+		# HACK
+		import CharGenCommon
+		CharGenCommon.next()
+	if GameCheck.IsIWD1():
```
lynxlynxlynx: elif

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+
+# Button next press
+def PortraitButtonNextPress():
+	global PortraitsTable, LastPortrait
+	PortraitPictureButtonName = PortraitApplySelection()
+	if GameCheck.IsBG1OrEE():
+		# HACK
+		import CharGenCommon
+		CharGenCommon.next()
+	if GameCheck.IsIWD1():
```
lynxlynxlynx: elif

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+		CGPortraitChangeToRace(PortraitPictureButtonName)
+	
+# This if for moving to next portrait
+def PortraitNext():
+	global PortraitsTable, LastPortrait, Gender
```
lynxlynxlynx: only LastPortrait is needed

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+	Button.OnPress(PortraitCustomAbort)
+	
+	Button = Window.GetControl(0)
+	PortraitName = PortraitsTable.GetRowName(LastPortrait) + PortraitSuffix["large"]
+
+	if GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+		if GemRB.HasResource(PortraitName, RES_BMP, 1) or GemRB.HasResource(EmptyPortraitMedium, RES_BMP, 1):
+			Button.SetPicture(PortraitName, EmptyPortraitMedium)
+	else:
+		Button.SetPicture(PortraitName, EmptyPortraitMedium)
```
lynxlynxlynx: This game check is also redundant. The fallbacks are not likely to be missing, but still, we can just use the inner if block everywhere.

It was probably added since SetPicture throws a runtime error if both files don't exist. I think we can just print a warning instead and handle it nicely. So you can just use the last line here and I'll fix the c++ side.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -0,0 +1,390 @@
...
+	GemRB.SetVar ("Gender",0) #scrapping the gender value
+	return
+
+def PortraitSetPicture():
+	global PortraitButton, PortraitsTable, LastPortrait, EmptyPortraitBig
+	PortraitName = PortraitsTable.GetRowName (LastPortrait)+PortraitSuffix["set"]
+	PortraitButton.SetPicture(PortraitName, EmptyPortraitBig)
+	return
+
+def CGPortraitChangeToRace(portrait_picture_button_name):
```
lynxlynxlynx: ssssssss

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: I assume that this part:
```
	if GameCheck.IsIWD2():
		FemaleCountAdd = 0
		LastPortraitOffset = -1
```
Not the FemaleCount var. Is that right ?

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: You mean the iwd1 ? I set it to all of them and are working. Or should I remove it from all of them ?

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: no, just the print

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: It looks like it was needed in iwd, like they start disabled. If they don't, then you can remove it whole. Otherwise just enable them everywhere.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: which is pictures ...

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: this remains

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: this remains

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: this remains

rsantellan:
@lynxlynxlynx I now have an issue on the commit 9aee96e221e74fdfe05de9d9ca391076fbc8dbbd that now the custom portrait is not working anymore.

```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "~/gemrb-git/gemrb/GUIScripts/GUIPortraitCommon.py", line 142, in PortraitButtonCustomDone
    PortraitPictureButtonName = PortraitCustomDone()
[Python/ERROR]:   File "~/gemrb-git/gemrb/GUIScripts/GUIPortraitCommon.py", line 188, in PortraitCustomDone
    PortraitLarge = PortraitList1.QueryText()
                    ^^^^^^^^^^^^^
[Python/ERROR]: NameError: name 'PortraitList1' is not defined
```
But I can't figure out why the previous commit works and this one do not, I have tested all games and in all is the same.

lynxlynxlynx:
It's only set up in `PortraitCustomPress`. Can the other function be called without that happening first?

rsantellan:
@lynxlynxlynx I now have an issue on the commit 9aee96e221e74fdfe05de9d9ca391076fbc8dbbd that now the custom portrait is not working anymore.

```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "~/gemrb-git/gemrb/GUIScripts/GUIPortraitCommon.py", line 142, in PortraitButtonCustomDone
    PortraitPictureButtonName = PortraitCustomDone()
[Python/ERROR]:   File "~/gemrb-git/gemrb/GUIScripts/GUIPortraitCommon.py", line 188, in PortraitCustomDone
    PortraitLarge = PortraitList1.QueryText()
                    ^^^^^^^^^^^^^
[Python/ERROR]: NameError: name 'PortraitList1' is not defined
```
But I can't figure out why the previous commit works and this one do not, I have tested all games and in all is the same.

> It's only set up in `PortraitCustomPress`. Can the other function be called without that happening first?

Found the error, I called the set portrait on the first button. Silly me, put the carriage before the horses...

lynxlynxlynx:
Ok, does everything work now?

rsantellan:
> Ok, does everything work now?

Yes, tested on all the games.

lynxlynxlynx:
Nice. Also by providing custom portraits?

rsantellan:
> Nice. Also by providing custom portraits?

Yes, the only thing that it a little weird is on iwd1 it happens once that the male and female portrait where mixed, the first one, then it get fixed. But can't replicate.

lynxlynxlynx:
Ok good, thanks! 
You didn't resolve everything, but let's make a clean cut. I'll merge this and now that you know the code intimately, you can also look at the copy of this functionality in GUIRECCommon. It is using a different window, so you'll need to add at least one parameter to the main function, so the module knows whether it's in character generation or in a running game.

PR #2475, opened 2026-04-25
Projectile: Fix invalid return value from iterator

Bubb13:
## Description
This PR fixes the following assert when casting a spell on the MSVC debug build:

```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Assertion Failed!

Program: ...ames\Infinity-Engine\Installations\GemRB_Debug\gemrb_core.dll
File: C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC\14.50.35717\include\list
Line: 147

Expression: cannot dereference end list iterator
```

The bug seems to stem from the function attempting to return the projectile it inserted by dereferencing `iter` – which actually points to the entry after the inserted element. The assert is thrown when `iter` points to the ending sentinel node of the list.

## Checklist


lynxlynxlynx:
Thanks!

Issue #2446, opened 2026-03-21
Deduplicate the portrait image selection code


lynxlynxlynx:
GUICG12, GUIRECCommon, CGPortrait and others in various games all duplicate the same functionality. Rip it out into a shared python module (GUIRECCommon and PaperDoll being examples of that) and then use that instead.

rsantellan:
@lynxlynxlynx I will try to take a shot at this.

lynxlynxlynx:
Thanks, I suggest you take it in steps, for example one commit to:
- extract the portrait code from bg2 character generation into a new file
- look at bg1's character generation, see what upgrades the new file needs to be used (will definitely need a GameCheck.IsBG1 check for what happens on pressing next/back) and do it
- same for iwd
- same for iwd2

Once character generation is done, the customization from in-game can be tackled. That's probably all in GUIRECCommon and perhaps another file for iwd2.

rsantellan:
@lynxlynxlynx I found a bug on IW2 when trying to use a custom portrait on character generation:

```
[Python/ERROR]: Traceback (most recent call last):
[Python/ERROR]:   File "~/gemrb-git/gemrb/GUIScripts/iwd2/CGPortrait.py", line 168, in CustomPress
    CustomWindow = Window = GemRB.LoadWindow (18)
                            ~~~~~~~~~~~~~~~~~^^^^
[Python/ERROR]: TypeError: function takes at least 2 arguments (1 given)
```
If I put "GUICG" as the BG1 and BG2 it works correctly, but don't know if that is ok. 

I'm working on this branch: https://github.com/rsantellan/gemrb/tree/2446-branch
The thing that I don't know how to do it on the back button on IW2.

I will start looking at IW1 to see how extract the logic, I believe is more complicated than the other 3.

lynxlynxlynx:
Yes, that fix is correct.

[...]

PR #2476, opened 2026-04-25
2446 branch guireccommon

rsantellan:
## Description
This is the continuation of the issue #2446 I'm tackling the class **GUIRECCommon** 

## Checklist


gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
 	"set": "G" if (GameCheck.IsBG1OrEE() or GameCheck.IsIWD1()) else "L",
 }
 
-def OnLoad():
+def OnLoad(GUIRECCommon = False):
```
lynxlynxlynx: Don't use the same name, there would be a conflict if we ever needed to import that module and it somewhat interferes with grepping.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
-	global Gender
-
+	global Gender, IsGUIRECCommon
+	
+	IsGUIRECCommon = GUIRECCommon
 	# Window
-	AppearanceWindow = GemRB.LoadWindow(11, "GUICG")
+	if GUIRECCommon:
+		AppearanceWindow = GemRB.LoadWindow(18, "GUIREC")
+		AppearanceWindow.AddAlias("SUB_WIN", 0)
```
lynxlynxlynx: redundant

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
+		# Load the Gender
+		Gender = GemRB.GetPlayerStat(Pc, IE_SEX)
+	else:
+		AppearanceWindow = GemRB.LoadWindow(11, "GUICG")
+		# Load the Gender
+		Gender = GetGender()
 	
-	# Load the Gender
-	Gender = GetGender()
+	if not GUIRECCommon:
```
lynxlynxlynx: this line can go

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
+		LastPortraitOffset = 0
+		if GameCheck.IsIWD2():
+			FemaleCountAdd = 0
+			LastPortraitOffset = -1
+		FemaleCount = PortraitsTable.GetRowCount() - PortraitsStart + FemaleCountAdd
+		if Gender == 2:
+			LastPortrait = GemRB.Roll(1, FemaleCount, PortraitsStart - 1)
+		else:
+			LastPortrait = GemRB.Roll(1, PortraitsTable.GetRowCount() - FemaleCount, LastPortraitOffset)
 	else:
```
lynxlynxlynx: if not x
else

is an antipattern, since it is a double negation and thus usually requires more effort to read

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: I don't have to use the name: OnLoad or the GUIRECCommon ?

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
+	global Gender, IsGUIRECCommon
+	
+	IsGUIRECCommon = GUIRECCommon
 	# Window
-	AppearanceWindow = GemRB.LoadWindow(11, "GUICG")
+	if GUIRECCommon:
+		AppearanceWindow = GemRB.LoadWindow(18, "GUIREC")
+		AppearanceWindow.AddAlias("SUB_WIN", 0)
+		Pc = GemRB.GameGetSelectedPCSingle()
+		PortraitName = GemRB.GetPlayerPortrait(Pc, 0)["ResRef"]
```
lynxlynxlynx: I'd move this to where it's actually used for the first time.

And gender-fetching should be in GetGender(), so you can call it unconditionally.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: The **AddAlias** part ?

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: If I remove that if then it will be calling the CharGenCommon and CharOverview of bg2 and iwd2, is that ok ?

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: and the whole block should be moved to its own function, now that it0s so big. All they do is in the end is set LastPortrait, so this will be easy to isolate. Or better yet, make two functions — in chargen we want randomness, in game the current index.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
+			if PortraitName == PortraitsTable.GetRowName(i).upper():
+				LastPortrait = i
+				break
 
 	# Controls
-	PortraitButton = AppearanceWindow.GetControl(1)
+	PortraitButtonNumber = 1
+	if GUIRECCommon:
+		PortraitButtonNumber = 0
+	PortraitButton = AppearanceWindow.GetControl(PortraitButtonNumber)
```
lynxlynxlynx: Please either use a dictionary or set all the ids up at the start (like the LU* files do).

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,93 +30,134 @@
...
+		CancelButton.SetState(IE_GUI_BUTTON_ENABLED)
+		CancelButton.OnPress(AppearanceWindow.Close)
+		CancelButton.SetText(13727)
+		CancelButton.MakeEscape()
+	else:
+		BackButton = AppearanceWindow.GetControl(5)
+		if GameCheck.IsIWD1():
+			BackButton.SetText(13727)
+		else:
+			BackButton.SetText(15416)
```
lynxlynxlynx: this is also available in iwd1, so we can just use this branch. Even if the original did use "cancel", nobody will notice.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: Calling it their I only call one time: 
**Pc = GemRB.GameGetSelectedPCSingle()**
And is all used a few lines below.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +182,41 @@ def PortraitButtonLeftPress():
...
+			import CharGenCommon
+			CharGenCommon.next()
+		elif GameCheck.IsIWD1():
+			CGPortraitChangeToRace(PortraitPictureButtonName)
+	else:
+		PortraitName = PortraitPictureButtonName[:-1]
+		pc = GemRB.GameGetSelectedPCSingle()
+		# eh, different sizes
+		if GameCheck.IsBG2OrEE():
+			GemRB.FillPlayerInfo(pc, PortraitName + "M", PortraitName + "S")
```
lynxlynxlynx: you already have vars for these ...

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +182,41 @@ def PortraitButtonLeftPress():
...
+		else:
+			GemRB.FillPlayerInfo(pc, PortraitName + "L", PortraitName + "S")
+		#GemRB.GetView("SUB_WIN", 0).Close()
+		#AppearanceWindow.Close()
```
lynxlynxlynx: remove

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: and there's duplication with PortraitApplySelection

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: GUIRECCommon, there's no OnLoad.py

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: yes, you're not using it anywhere. In case it's not obvious, every comment is for the line it is opened under. Or the last for multiline comments, but what it applies to is the first thing that's written in the box anyway.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: that's cheap, no need to worry about it

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: why would you?

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -63,90 +80,52 @@ def OnLoad(GUIRECCommon = False):
...
-	PortraitButton = AppearanceWindow.GetControl(PortraitButtonNumber)
+	PortraitButton = AppearanceWindow.GetControl(WindowButtonPosition[ButtonDictionaryPrefix+"portrait"])
 	PortraitButton.SetFlags(IE_GUI_BUTTON_PICTURE | IE_GUI_BUTTON_NO_IMAGE, OP_SET)
 	PortraitButton.SetState(IE_GUI_BUTTON_LOCKED)
 
-	LeftButtonNumber = 2
-	if GUIRECCommon:
-		LeftButtonNumber = 1
-	LeftButton = AppearanceWindow.GetControl(LeftButtonNumber)
+	LeftButton = AppearanceWindow.GetControl(WindowButtonPosition[ButtonDictionaryPrefix+"left"])
```
lynxlynxlynx: This is not much of a simplification, rather redefine the dict depending on the mode.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -30,30 +30,47 @@
...
+	"modification_left": 1,
+	"right": 3,
+	"modification_right": 2,
+	"custom": 6,
+	"modification_custom": 5,
+	"done": 0,
+	"modification_done": 3,
+	"back": 5,
+	"modification_cancel": 4,
+	"custom_portrait": 6,
```
lynxlynxlynx: ssssss

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -182,41 +198,36 @@ def PortraitButtonLeftPress():
...
 def PortraitButtonCustomDone():
-	if not IsGUIRECCommon:
-		PortraitPictureButtonName = PortraitCustomDone()
-		if GameCheck.IsBG1OrEE():
-			import CharGenCommon
-			CharGenCommon.next()
-		elif GameCheck.IsIWD1():
-			CGPortraitChangeToRace(PortraitPictureButtonName)
-	else:
+	if not IsPortraitModification:
```
lynxlynxlynx: you have a double negation here too

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,102 +30,159 @@
...
+	RightButton.OnPress(PortraitButtonRightPress)
+	
+	if IsPortraitModification:
+		CancelButton = AppearanceWindow.GetControl(WindowButtonPosition["Cancel"])
+		CancelButton.SetState(IE_GUI_BUTTON_ENABLED)
+		CancelButton.OnPress(AppearanceWindow.Close)
+		CancelButton.SetText(13727)
+		CancelButton.MakeEscape()
 	else:
+		BackButton = AppearanceWindow.GetControl(WindowButtonPosition["Back"])
```
lynxlynxlynx: Why do you need two different keys? It doesn't look like you use back/cancel anywhere else. Unify it (and the button name) and then you can move the GetControl, SetState and MakeEscape outside the conditional.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,102 +30,159 @@
...
+	"CustomPortrait": 6,
+	"CustomDone": 6,
+	"CustomCancel": 7,
+	"CustomPortraitList": 4,
+}
 
-	# Window
-	AppearanceWindow = GemRB.LoadWindow(11, "GUICG")
+def OnLoad(PortraitModification = False):
+	global AppearanceWindow, PortraitButton, PortraitsTable, LastPortrait
```
lynxlynxlynx: Is LastPortrait still needed here?

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -28,102 +30,159 @@
...
+		Pc = GemRB.GameGetSelectedPCSingle()
+		return GemRB.GetPlayerStat(Pc, IE_SEX)
 	else:
-		MyChar = GemRB.GetVar("Slot")
-		return GemRB.GetPlayerStat(MyChar, IE_SEX)
+		if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
+			return GemRB.GetVar("Gender")
+		else:
+			MyChar = GemRB.GetVar("Slot")
+			return GemRB.GetPlayerStat(MyChar, IE_SEX)
```
lynxlynxlynx: In the end of the refactoring process this function will not be needed any more. To make it more obvious how to get there, I suggest you do it like:

```
	if GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo():
		return GemRB.GetVar("Gender")
	if IsPortraitModification:
		pc = GemRB.GameGetSelectedPCSingle()
	else:
		pc = GemRB.GetVar("Slot")
	return GemRB.GetPlayerStat(pc, IE_SEX)
```

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +198,36 @@ def PortraitButtonLeftPress():
...
 	PortraitPictureButtonName = PortraitApplySelection()
-	if GameCheck.IsBG1OrEE():
-		import CharGenCommon
-		CharGenCommon.next()
-	elif GameCheck.IsIWD1():
-		CGPortraitChangeToRace(PortraitPictureButtonName)
+	if IsPortraitModification:
+		PortraitName = PortraitPictureButtonName[:-1]
+		pc = GemRB.GameGetSelectedPCSingle()
+		GemRB.FillPlayerInfo(pc, PortraitName + PortraitSuffix["large"], PortraitName + PortraitSuffix["small"])
```
lynxlynxlynx: how come this branch doesn't close the windows?

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +198,36 @@ def PortraitButtonLeftPress():
...
+		if GameCheck.IsBG1OrEE():
+			import CharGenCommon
+			CharGenCommon.next()
+		elif GameCheck.IsIWD1():
+			CGPortraitChangeToRace(PortraitPictureButtonName)
 
 # Button next press
 def PortraitButtonNextPress():
 	global PortraitsTable, LastPortrait
 	PortraitPictureButtonName = PortraitApplySelection()
```
lynxlynxlynx: I suggest you just name it PortraitName, so you don't need two vars. And it's a better name anyway, it has nothing to do with buttons.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +198,36 @@ def PortraitButtonLeftPress():
...
-		CGPortraitChangeToRace(PortraitPictureButtonName)
+	if IsPortraitModification:
+		pc = GemRB.GameGetSelectedPCSingle()
+		GemRB.FillPlayerInfo(pc, PortraitList1.QueryText(), PortraitList2.QueryText())
+		if CustomWindow:
+			CustomWindow.Close()
+		if AppearanceWindow:
+			AppearanceWindow.Close()
+	else:
+		PortraitPictureButtonName = PortraitCustomDone()
```
lynxlynxlynx: You can call PortraitCustomDone in both cases and then don't need to manually close the windows. You could even move the FillPlayerInfo inside it, since all you need is there.

And please rename PortraitCustomDone to SaveCustomPortrait, so it's clearer what it's about and why it needs to exist besides this similarly named PortraitButtonCustomDone.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: and similarly to above, you can move the setting to PortraitApplySelection.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -302,7 +382,50 @@ def PortraitCustomPress():
...
+		ButtonDone.SetState(IE_GUI_BUTTON_DISABLED)
+
+	ButtonCancel = Window.GetControl(WindowButtonPosition["CustomCancel"])
+	if IsPortraitModification:
+		ButtonCancel.SetText(13727)
+		ButtonCancel.MakeEscape()
+		ButtonCancel.SetState(IE_GUI_BUTTON_ENABLED)
+	else:
+		ButtonCancel.SetText(15416)
+	if GameCheck.IsIWD2():
```
lynxlynxlynx: this can go

lynxlynxlynx:
Ok, that's better already, the first iteration had a positive line count difference. Besides the comments above, it looks like PortraitCommonSmallCustom and PortraitCommonLargeCustom can easily be merged.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: > how come this branch doesn't close the windows?

It close it on PortraitApplySelection

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -252,26 +236,32 @@ def PortraitPrev():
...
-	if not IsPortraitModification and (GameCheck.IsIWD2() or GameCheck.IsBG2OrEE() or GameCheck.IsBG2Demo()):
-		GemRB.SetNextScript ("CharGen2")
-	return PortraitLarge
+		
+	if IsPortraitModification:
+		Pc = GemRB.GameGetSelectedPCSingle()
+		GemRB.FillPlayerInfo(Pc, PortraitList1.QueryText(), PortraitList2.QueryText())
+		return
+	else:
+		PortraitLarge = PortraitList1.QueryText()
```
lynxlynxlynx: You could have left these two outside, since both blocks need them.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -139,22 +194,24 @@ def PortraitButtonLeftPress():
...
-		CharGenCommon.next()
-	elif GameCheck.IsIWD1():
-		CGPortraitChangeToRace(PortraitPictureButtonName)
+	PortraitName = PortraitApplySelection()
+	if not IsPortraitModification:
+		if GameCheck.IsBG1OrEE():
+			import CharGenCommon
+			CharGenCommon.next()
+		elif GameCheck.IsIWD1():
+			CGPortraitChangeToRace(PortraitName)
```
lynxlynxlynx, line 203: Now it is obvious these two use the same follow-up, which just calls for another function.

There's also the bigger question whether SaveCustomPortrait and PortraitApplySelection can be merged, since they're also very similar.

rsantellan: Sorry but I don't see how to merge them. They have some similarities but a lot of differences.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -276,11 +275,12 @@ def PortraitCustomAbort():
 
 # This is for applying
 def PortraitApplySelection():
-	global AppearanceWindow, LastPortrait
+	global AppearanceWindow, LastPortrait, PortraitsTable
 	if AppearanceWindow:
 		AppearanceWindow.Close ()
-	PortraitTable = GemRB.LoadTable("pictures")
-	PortraitName = PortraitTable.GetRowName(LastPortrait)
+	if PortraitsTable == 0:
```
lynxlynxlynx: how could this ever be true? The only way into this module sets it unconditionally.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: then at least do the first part

rsantellan:
Hi @lynxlynxlynx with all the changes in the file now I'm at a loss of what is missing in this.

lynxlynxlynx:
Look at the few remaining unresolved comments (easy on the website) and my general reply from above:

> Ok, that's better already, the first iteration had a positive line count difference. Besides the comments above, it looks like PortraitCommonSmallCustom and PortraitCommonLargeCustom can easily be merged.

lynxlynxlynx:
Also, I've now pushed a small cleanup for chargen, which should enable you to always look up by stat in GetGender. You can just inline the whole function too afterwards.

rsantellan:
> Also, I've now pushed a small cleanup for chargen, which should enable you to always look up by stat in GetGender. You can just inline the whole function too afterwards.

@lynxlynxlynx Do I just do a merge from master to my branch and apply the changes ? Or do a cherry pick ? Don't know really much is that will mess the history

lynxlynxlynx:
Yes, a merge is fine, a rebase even better.

rsantellan:
@lynxlynxlynx now it says that it has problems with the clang format, even though I didn't touch it and it should be alright because was merged on master.

lynxlynxlynx:
Yeah, ignore it. You did an odd merge, so the commits appear as new and so it got retriggered.

rsantellan:
> Yeah, ignore it. You did an odd merge, so the commits appear as new and so it got retriggered.

I did a git rebase to master.
```
 2003  git checkout master 
 2004  git pull
 2005  git checkout 2446-branch-guireccommon
 2006  git rebase master
 2007  git status
 2008  git pull
 2009  git rebase
 2010  git status
 2011  git push
```
That is my history of the command line

lynxlynxlynx:
The first 4 lines are fine and should be enough (assuming master is tracking upstream master, not the one in your fork). I don't know the default, but I'd specify --rebase next time on the first pull already; now if it can't fast forward, it will create a merge commit. The second pull is probably what caused the commit duplication. If it bothers you, you can look at `git reflog` and retry. 

I'm surprised github is displaying it as such though; both that it can apply cleanly, but also not recognizing the commits as no-ops.

rsantellan:
Hi @lynxlynxlynx after a little impasse from personal issues I'm back with this.  

> Also, I've now pushed a small cleanup for chargen, which should enable you to always look up by stat in GetGender. You can just inline the whole function too afterwards.

I think this is done, what will be the next step ?

lynxlynxlynx:
Well, same as before, look at the few (2) open comments that remain. And as already stated twice: PortraitCommonSmallCustom and PortraitCommonLargeCustom can easily be merged.

gemrb/GUIScripts/GUIPortraitCommon.py
rsantellan: @lynxlynxlynx this is no more.

lynxlynxlynx: you still have two if blocks with the same condition one after the other.

rsantellan: Sorry, but I don't see where. This is outdated now.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -282,84 +279,48 @@ def PortraitApplySelection():
 		GemRB.FillPlayerInfo(Pc, PortraitName + PortraitSuffix["large"], PortraitName + PortraitSuffix["small"])
 	return PortraitName + PortraitSuffix["large"]
 
-# This is for the large custom portrait
-def PortraitCommonLargeCustom():
+# This is for the large and small custom portrait
+def PortraitCommonCustom(type = 'small'):
```
lynxlynxlynx: `type` is a python keyword, so please use a different name like "size".

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -282,84 +279,48 @@ def PortraitApplySelection():
...
+	global RowCount1, RowCount2
 	global CustomWindow
 
 	Window = CustomWindow
 
-	Portrait = PortraitList1.QueryText()
+	Portrait = PortraitList1.QueryText() if 'medium' == type else PortraitList2.QueryText()
 
+	row_var = "Row1" if type == "medium" else "Row2"
+	row_count = RowCount1 if type == "medium" else RowCount2
```
lynxlynxlynx: please no snake case, the style should always be consistent with the rest of the code.

gemrb/GUIScripts/GUIPortraitCommon.py
```diff
@@ -282,84 +279,48 @@ def PortraitApplySelection():
...
+def PortraitCommonCustom(type = 'small'):
 	global PortraitList1, PortraitList2
-	global RowCount1
+	global RowCount1, RowCount2
 	global CustomWindow
 
 	Window = CustomWindow
 
-	Portrait = PortraitList1.QueryText()
+	Portrait = PortraitList1.QueryText() if 'medium' == type else PortraitList2.QueryText()
```
lynxlynxlynx: You have many lines with inline if statements, so I suggest you use a normal block to define variables first, so the rest of the code is not littered with these differences.

gemrb/GUIScripts/GUIPortraitCommon.py
lynxlynxlynx: See new line 52 vs 71.

lynxlynxlynx:
Thanks, I've untangled the mess and pushed it. Tried bg2 chargen and ingame modification and it still works. Did a few other minor cleanups later, so combined with this merge hell, please just create a new branch after you update to a clean upstream.

PR #2481, opened 2026-04-28
ResourceManager: BIFs by language

### LINKED ISSUE #2479: EE: language setting ignored for speech files
#### Bug description
When I set `GameLanguagePath` to something, I get the text loaded in that language and the main PC's voice, but nothing else, so all other voiced audio remains English. Looks like this is not used:

#### Screenshots
Image

- [x] master as of this issue
- [ ] 0.9.5
- [ ] 0.9.4
- [ ] 0.9.3

MarcelHB:
A working approach how to tackle #2479, by looking for BIF files from the language's `data/` and replacing them by what is specified in the key file.

## Checklist

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -384,6 +384,12 @@ Interface::Interface(CoreSettings&& cfg)
 		ThrowException("The path must point to a game directory with a readable chitin.key file.");
 	}
 
+	// Inject language specific data BIFs
+	auto languageDataPath = PathJoin(config.GamePath, config.GameLanguagePath, "data");
+	if (DirExists(languageDataPath)) {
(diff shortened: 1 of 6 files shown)


lynxlynxlynx:
Too bad the chitin hardcodes the folder, otherwise simply adding the path earlier would suffice.

MarcelHB:
> Too bad the chitin hardcodes the folder, otherwise simply adding the path earlier would suffice.

The main issue I encountered trying this, was that the files are accessed by index from the resource in question: https://github.com/gemrb/gemrb/blob/676f692714fd65f856e553c289cca4863ef39403/gemrb/plugins/KEYImporter/KEYImporter.cpp#L187-L194

lynxlynxlynx:
Thanks!

Issue #2478, opened 2026-04-28
EE: encoding display issues


MarcelHB:
#### Bug description
In an English EE, _Fearûn_ is displayed badly, but in German it's getting wilder due to umlaute, but also other things like quotes or bullet points, e. g. around the term _Zurück_, with three artifacts and just one being for _ü_.

#### Screenshots

<img width="291" height="66" alt="Image" src="https://github.com/user-attachments/assets/e49021c1-9d30-485e-993a-86390226efcb" />

GemRB version (check as many as you know apply)
- [x] master as of this issue
- [ ] 0.9.5
- [ ] 0.9.4
- [ ] 0.9.3

Video Driver (check as many as you know apply)
- [ ] SDL1.2
- [x] SDL2 built with `OPENGL_BACKEND` enabled
- [ ] SDL2 without `OPENGL_BACKEND` enabled

lynxlynxlynx:
Sounds similar / could be the same as #1929 and #1281.

MarcelHB:
Probably yes, it's just everywhere not only in item descriptions.

MarcelHB:
Funny, while it's safe to hardwire the EEs to UTF-8 input and see a lot of things work, it also makes some text sections disappear for now, maybe related to this underscore mystery mentioned in #1281

<img width="290" height="256" alt="Image" src="https://github.com/user-attachments/assets/cf663a78-8050-4683-9ada-972cc7ddb076" />

<img width="224" height="261" alt="Image" src="https://github.com/user-attachments/assets/250d75f5-5cb1-49b2-acb7-47d144166540" />

lynxlynxlynx:
We don't detect that they're using unicode in the tlk? For sure it could be hardcoded if needed, but I'm surprised.

Issue #1929, opened 2023-10-17
Unicode issue in EE item description - wrong dash


lynxlynxlynx:
While there are general TTF font issues still (see #1895), we now do use them for the normal font (#1924). However, a different character is printed (â?). Compare:
https://forums.beamdog.com/uploads/editor/1m/i9eqxahips3l.png
![image](https://github.com/gemrb/gemrb/assets/121515/066d5cc9-5187-4baa-ae50-468bdce8658b)

Issue #1281, opened 2021-06-11
console: inputting _ turns the following characters invisible 


lynxlynxlynx:
It's just as bizarre as it sounds. I didn't find any other special character that would have the same result.

bradallred:
maybe it depends on the font, cuz I dont see this happening

lynxlynxlynx:
yeah, it's in our demo.

lynxlynxlynx:
You can type it in an existing string and anything following it will vanish (and reappear on removal).

PR #2484, opened 2026-04-29
UTF-8 for EEs and text rendering resilience

MarcelHB:
Two patches on #2478 #1281 #1929

1. If we see an EE, we can probably skip any further encoding determination logic and set UTF-8 regardless, as BG1/2EE have all language TLKs in UTF-8.
2. Previously, when there was no suitable glyph found, the rendering aborted the rest. I changed that to skipping the character.

This makes us have this for now which should be an improvement, although still no proper dashes/bullet points:

ee_utf8_nobreak

Non-latin chars of course still don't work but that's a different task.

## Checklist


lynxlynxlynx:
Thanks! I'm surprised the glyph problm happens with TTF fonts, which have basic characters like that.

PR #2491, opened 2026-05-02
TTF unless BAM is preferred

### LINKED ISSUE #2490: EE: TTF fonts not used
#### Bug description
Looking at the mildly ugly font of the EEs and surrounding issues, one thing quickly emerges: we don't even use TTF.

https://github.com/gemrb/gemrb/blob/1ac834d320cf03dc6f87412bbf2379112ec21b68/gemrb/core/Interface.cpp#L1269

Although there is no preference for `NORMAL` being a BAM font, there is nothing making it load the TTF instead.

Manually changing 0 to the TTF type then gives all the EE style and many visual fixes, but this seems like the wrong approach at this point of code -- or we simply check for EE and change this behavior.

- [x] master as of this issue
- [ ] 0.9.5
- [ ] 0.9.4
- [ ] 0.9.3

MarcelHB:
See #2490.

## Checklist

### DIFF
--- a/gemrb/core/Interface.cpp
+++ b/gemrb/core/Interface.cpp
@@ -1266,7 +1266,7 @@ void Interface::LoadFonts()
 
 		Holder<Font> fnt;
 		ResourceHolder<FontManager> fntMgr;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #2496, opened 2026-05-09
Fixes around libfmt and GCC v16

MarcelHB:
## Description
When using GCC 16.1 of MSYS2, a bunch of errors around `consteval` start to appear. This shouldn't happen as it's C++20 or newer. So I

* updated libfmt to v12.1.0, although this did not fix anything
* set `FMT_USE_CONSTEVAL=0`, which did fix it, so I assume there is some internal check going wrong with newer GCC
* dropped the vendored source code in favor of CMake loading

## Checklist


lynxlynxlynx:
Sorry, this won't do. Packagers really dislike when an internet connection is required to build something. We need to make it optional, so they can override it with their local/packaged version if we don't want to bundle it any more. Use the system one if found, otherwise fetch fresh.

And it would fit better into CONFIGURE_FMT_INCLUDES, perhaps renamed.

MarcelHB:
libfmt-dev is a little behind (v10.1), even on Ubuntu 26.04, so this could raise other issue. Ok, will try something else.

lynxlynxlynx:
Is overriding FMT_USE_CONSTEVAL enough? I see fmt tries itself a bunch of checks in base.h already in our version.

MarcelHB:
I guess it's sufficient then.

lynxlynxlynx:
I'm asking, on linux gemrb compiled fine after one -Wmaybe-unitialised change.

MarcelHB:
No, I only get this kind of error (haven't tested Linux r/n since I need a dockerized setup for this):
```
D:/Sources/gemrb/gemrb/core/Strings/CString.h: In instantiation of 'bool GemRB::FixedSizeString<LEN, CMP>::Format(ARGS&& ...) [with ARGS = {const char (&)[3], GemRB::FixedSizeString<8, strncmp>&}; long long unsigned int LEN = 8; int (* CMP)
(const char*, const char*, size_t) = strncmp]':
D:/Sources/gemrb/gemrb/tests/core/Strings/Test_CString.cpp:32:11:   required from here
   32 |         fu.Format("{}", src);
      |         ~~~~~~~~~^~~~~~~~~~~
D:/Sources/gemrb/gemrb/core/Strings/CString.h:140:44: error: call to consteval function 'fmt::v12::fstring<GemRB::FixedSizeString<8, strncmp>&>((* & std::forward<const char (&)[3]>((* & args#0))))' is not a constant expression
  140 |                 auto ret = fmt::format_to_n(begin(), LEN + 1, std::forward<ARGS>(args)...);
      |                            ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from D:/Sources/gemrb/gemrb/includes/fmt/format.h:41,
                 from D:/Sources/gemrb/gemrb/core/Strings/StringView.h:8,
                 from D:/Sources/gemrb/gemrb/core/Strings/CString.h:10:
D:/Sources/gemrb/gemrb/core/Strings/CString.h:140:30:   in 'constexpr' expansion of 'fmt::v12::fstring<GemRB::FixedSizeString<8, strncmp>&>((* & std::forward<const char (&)[3]>((* & args#0))))'
D:/Sources/gemrb/gemrb/includes/fmt/base.h:2744:53:   in 'constexpr' expansion of 'fmt::v12::basic_string_view<char>(((const char*)s))'
 2744 |     if (FMT_USE_CONSTEVAL) parse_format_string<char>(s, checker(s, arg_pack()));
      |                            ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
D:/Sources/gemrb/gemrb/includes/fmt/base.h:544:5: error: 'args#0' is not a constant expression
  544 |     while (*s++) ++len;
      |     ^~~~~
```

Disabling `FMT_USE_CONSTEVAL` fixes that, but only starting at v12.1.0, apparently: https://github.com/fmtlib/fmt/commit/a2fd48e039c9f1cdfc7e4b7e8d0c601ae558dbee so maybe there is another fix, or we need to upgrade this either way.

lynxlynxlynx:
Yeah, let's just bump it then and add that define.

MarcelHB:
I can do this in the evening, but feel free if you find time until then.

lynxlynxlynx:
Thanks, that would be appreciated. I'm still in the middle of the resave diff reduction weeds.

PR #2497, opened 2026-05-10
libfmt v12.1.0

MarcelHB:
Together with that unset `FMT_USE_CONSTEVAL`.

I skipped formatting the code, so that will be red here.

## Checklist


cmake/Helpers.cmake
```diff
@@ -134,6 +134,8 @@ FUNCTION(CONFIGURE_COMPILER)
 
 		ADD_FLAG_IF_SUPPORTED(FLAG "-Wimplicit-fallthrough" VALUE "2")
 		ADD_FLAG_IF_SUPPORTED(FLAG "-Woverloaded-virtual" VALUE "0")
+		# prevents some GCC v16.1 errors
+		ADD_DEFINITIONS("-DFMT_USE_CONSTEVAL=0")
```
lynxlynxlynx: Why not put this in CONFIGURE_FMT_INCLUDES?

MarcelHB: Because the name made me think it's just about includes :) ... I'll make that clear, too.

lynxlynxlynx:
Thanks!

Issue #2455, opened 2026-03-30
Mac (Source Ports build) crash


lynxlynxlynx:
@tomkidd a crash was reported against the new Mac Source Ports build. Could you try rebuilding with the patch below?
Btw, I couldn't find where your recipes are stored — the [website entry](https://www.macsourceports.com/sourceport/gemrb) doesn't link to them and the two copies I found on the github organization were stale.

> I think we have the same issue on macOS ( version 0.9.5 with IWD 2):
> 
>  . . . 
> [ResourceManager]: Found 'STARTN.mos' in 'chitin.key'.
> Assertion failed: (ref), function GemRB_View_AddSubview, file GUIScript.cpp, line 1030.
> Abort trap: 6
> logout
>  . . . 

 _Originally posted by @svetlyo in [#2074](https://github.com/gemrb/gemrb/issues/2074#issuecomment-4151645274)_

```diff
diff --git a/gemrb/CMakeLists.txt b/gemrb/CMakeLists.txt
index e5de3e0c2..7213e4457 100644
--- a/gemrb/CMakeLists.txt
+++ b/gemrb/CMakeLists.txt
@@ -1,6 +1,6 @@
 SET(CMAKE_CXX_VISIBILITY_PRESET "hidden")
 # FreeBSD clang still has symbol visibility issues with the GUIScript plugin
-IF(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+IF(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
        SET(CMAKE_CXX_VISIBILITY_PRESET "default")
 ENDIF()
 SET(CMAKE_VISIBILITY_INLINES_HIDDEN TRUE)
```
or if not using clang, also deleting the extra check in that line.

tomkidd:
Try the latest build (same link, just new build). It should work now. There was also a packaging issue which I resolved. 

Thanks for the heads up, this patch seems to have fixed it.

lynxlynxlynx:
Ok, I'll apply a cleaned up version then, thanks!

Issue #2074, opened 2024-04-12
 Assertion failed: (delref), function GemRB_RemoveView, with clang on FreeBSD


traveler-gemrb:
#### Bug description

gemrb-git build as Release (-O3 -DNDEBUG) fails here with 

[ResourceManager]: Found 'GTRBPSK.mos' in 'chitin.key'.
Assertion failed: (delref), function GemRB_RemoveView, file /usr/home/Kuba/GemRBGit/gemrb/gemrb/plugins/GUIScript/GUIScript.cpp, line 2015.

after  (used git bisect)

commit 605519eddd7f105ef595dc492b98aacd91b5d6ff
Author: Jaka Kranjc <[lynxlupodian@users.sourceforge.net](mailto:lynxlupodian@users.sourceforge.net)>
Date: Tue Mar 26 11:07:12 2024 +0100

cmake: propagate compiler flag changes to parent scope

fixes last known regression and hopefully the odd crashes we've been
having #2063

cmake/Helpers.cmake | 5 ++++-
gemrb/core/Geometry.cpp | 2 +-
2 files changed, 5 insertions(+), 2 deletions(-)

Env:

FreeBSD 14.0-STABLE #0 stable/14-e069c451f amd64

FreeBSD clang version 17.0.6 (https://github.com/llvm/llvm-project.git llvmorg-17.0.6-0-g6009708b4367)

traveler-gemrb:
To reproduce, one need to load game (BG1) or just load menu (IWD1).

lynxlynxlynx:
Well please try what we talked about.

traveler-gemrb:
Hello, 

I've already tested default Release, that is (-O3 -DNDEBUG), without any custom -flto / -march=native etc. I think I've also tried -O2 (same result).

traveler-gemrb:
Thanks for a clarification, how can I get back to previous style? I think that was intended look.

lynxlynxlynx:
I don't know which version you have, but from above reverting d47c4cf2 should do it (or copying the file).

traveler-gemrb:
FWIW, 

Same with 14.1-STABLE #0 stable/14-91df7d335 amd64 and FreeBSD clang version 18.1.5

lynxlynxlynx:
@bsdcode can you do a quick check if it still runs for you? It's odd that you two would have so different experiences on FreeBSD so close in time.

bsdcode:
[...]

PR #2501, opened 2026-06-01
include OpenBSD in fix for CXX visibility preset

rfht:
## Description
OpenBSD is also affected by https://github.com/gemrb/gemrb/issues/2074 and the solution in https://github.com/gemrb/gemrb/issues/2455. This adds OpenBSD to the cases, and fixes the bug:
```
assertion "ref" failed: file "gemrb/plugins/GUIScript/GUIScript.cpp", line 1030, function "GemRB_Vie
w_AddSubview"
```

## Checklist

- [ ] I extended the documentation, if necessary **N/A**


lynxlynxlynx:
Thanks and too bad!

PR #2502, opened 2026-06-22
Populate EE damage feedback tokens

### LINKED ISSUE #2492: EE: Damage message placeholders are not fully populated
#### Bug description
On suffering from a hit, EEs have this kind of message, for example strref 3805:
```
erleidet <AMOUNT> Schaden (<TYPE>) von <DAMAGER>
```

Unfortunately `<TYPE>` is empty (and treated as `unknown` in the GemRB console), and also `<DAMAGER>` remains blank.

There is probably more messages to check here, like resistance information.

#### Screenshots

Image

- [x] master as of this issue
- [ ] 0.9.5
- [ ] 0.9.4
- [ ] 0.9.3

hyperspace9k1:
Fixes #2492.

This populates EE damage feedback placeholders for BGEE/BG2EE.

- Move damage type lookup before the detailed-message branch so BG2-style EE messages can use `TYPE`.
- Populate `TYPE` in the BG2 damage feedback branch.
- Populate `DAMAGER` directly for EE game types, while preserving classic BG2's existing colored speaker-prefix behavior.
- Map BGEE/BG2EE `DMG_*` rows to verified game TLK strrefs.

The BGEE and BG2EE elemental/magic damage strrefs differ because the games have independent TLKs. English and German TLKs were spot-checked; other locales rely on the standard TLK invariant that strref numbering is consistent across translations.

Validation:
- `git diff --check`
- GemRB pre-commit hook with clang-format 22.1.8
- `cmake --build build --target gemrb_core -j4`


gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4600,8 +4600,13 @@ void Actor::DisplayCombatFeedback(unsigned int damage, int resisted, int damaget
 		} else { //bg2
 			//<DAMAGER> did <AMOUNT> damage to <DAMAGEE>
 			tokens["DAMAGEE"] = GetName();
-			// wipe the DAMAGER token, so we can color it
-			tokens["DAMAGER"] = String {};
+			tokens["TYPE"] = std::move(type_name);
+			// classic bg2 keeps the attacker in the colored speaker prefix, while the EE strings use the token directly
+			if (core->config.GameType.substr(core->config.GameType.size() - 2, 2) == "ee") {
```
Mingun: I'm sure, there should be a better way to check if you play EE game

hyperspace9k1: Good catch — switched it to `core->HasFeature(GFFlags::HAS_EE_EFFECTS)`, which is the canonical EE check used elsewhere in the engine (e.g. `DisplayMessage.cpp`). It's set only for bgee/bg2ee in their `gemrb.ini`, so it cleanly distinguishes EE from classic. Done in 46fd04036.

hyperspace9k1:
Verifying against the actual EE game data turned up three things:

**1. The attacker's name shows twice.** BG2EE `Damage2` (3805) is `takes <AMOUNT> <TYPE> damage from <DAMAGER>`, with no subject, so it relies on the speaker-name prefix. But the EE branch passes `hitter` as the speaker *and* fills `DAMAGER` with the attacker, giving "**Attacker** - takes 5 fire damage from **Attacker**". The speaker should be `this`. (Pre-existing; filling `DAMAGER` just exposed it.)

**2. BGEE isn't fixed.** `DAMAGE_STR2` stays `3805`, but BGEE's strref 3805 is unrelated text ("…release the prisoners."), so the tokens have nothing to fill. BGEE's actual damage strings are `1256` / `24124`. (The `DMG_*` type mappings are correct.)

**3. The `detailed` branch may be a better fit.** `Actor.cpp:4559-4584` already handles this format: it fills the tokens, covers the resist/bonus variants (1254/1255, also in #2492), and speaks as `this`. It's only disabled for EE (`DAMAGE_DETAIL* = -1`). Setting those rows (BG2EE 3805/1254/1255, BGEE 1256/1254/1255) and dropping the `Damage2` edit would fix both games at once.

I can take either route: speaker fix + BGEE strref, or move EE onto the `detailed` branch. Preference?

gemrb/core/Scriptable/Actor.cpp
```diff
@@ -4600,8 +4600,13 @@ void Actor::DisplayCombatFeedback(unsigned int damage, int resisted, int damaget
 		} else { //bg2
 			//<DAMAGER> did <AMOUNT> damage to <DAMAGEE>
 			tokens["DAMAGEE"] = GetName();
-			// wipe the DAMAGER token, so we can color it
-			tokens["DAMAGER"] = String {};
+			tokens["TYPE"] = std::move(type_name);
+			// classic bg2 keeps the attacker in the colored speaker prefix, while the EE strings use the token directly
+			if (core->HasFeature(GFFlags::HAS_EE_EFFECTS)) {
+				tokens["DAMAGER"] = damager->GetName();
```
lynxlynxlynx: I'm confused:
- the diff suggests a different string is used in the EEs, but the table update doesn't seem to touch this; where does TYPE actually get used?
- damager could still be null here. When it's not, don't you get a duplicated name? It would work for strings like 1255, which is from the POW of the damagee.
- but that's also the iwd mode above, so why not just modify those strrefs as needed and get more correct results at the same time?

hyperspace9k1:
Took your route 3: moved the EEs onto the detailed branch instead of patching `Damage2`. Data-only now, no engine change.

- `TYPE`: now used by the detailed branch, whose strings actually contain it (the old 3805 path didn't, esp. for BGEE).
- Duplicate/null damager: detailed speaks as the damagee and fills `DAMAGER` only when the hitter is an actor, so neither happens.
- `DAMAGE_DETAIL1/2/3` set per game, `DMG_*` mapped to verified TLK strrefs.
- BGEE `DAMAGE_IMMUNITY` fixed: 5199 was unrelated text, now 24683. (BGEE's 3805 was wrong text too, which is why the old approach never worked there.)

Trade-off: the EEs have no no-damager detail strings, so `DAMAGE_STR2/3` fall back to plain "Damage Taken". Trap/self damage loses the "(amount)" it showed before. Can keep it by duplicating resist/bonus logic into `Damage2` instead, if you prefer.

Verified in-engine on both EEs: renders as `<damagee> - Takes <n> <type> damage from <attacker>`, single name.

lynxlynxlynx:
Ok, that's better, thanks.

lynxlynxlynx:
It doesn't look like you tested this fully, crushing is using a titlecased string, while it should 32431. Probably the rest in the 11k range have the same problem.

PR #2504, opened 2026-07-22
 BAMImporter: fix uncompressed V1 frame data loss + encumbrance label overlap

BobbyBicc:
## Summary
While tracking down garbled text in the inventory/loot weight overlay
("18:" rendering as illegible noise), found three separate issues:

1. **Root cause**: `BAMImporter::GetAnimationFactory()` read all
   uncompressed V1 frame data in one bulk Seek+Read into a shared
   buffer, then located each frame via pointer arithmetic. For
   BG2:EE's NUMBER.bam this silently dropped pixel data for several
   frames while leaving others intact, even though the frame table
   itself (offsets/dimensions/RLE flags) parsed correctly.

   To find a working reference implementation, compared against Near
   Infinity's `BamV1Decoder.java`: its frame-table parsing (the
   compression-flag high bit, the offset mask) is identical logic to
   GemRB's, ruling out a header-parsing bug - but its pixel read is
   much simpler, seeking directly to each frame's own absolute file
   offset with no bulk-buffer/relative-offset indirection at all.
   Porting that same approach into GemRB fixed the corruption
   completely, confirmed via direct pixel-level before/after
   comparison.

2. Numeric BAM fonts assume frame 10 is always a ':' glyph
   (frame N -> '0'+N). BG2:EE's NUMBER.bam frame 10 is actually
   unrelated "lbs" unit-suffix art, roughly double a digit's width.
   Added a sanity check to skip registering implausibly-wide "extra"
   frames instead of rendering whatever's there.

3. `Control.CreateLabel()` sized every label to the whole host
   button regardless of alignment, so paired top/bottom labels (eg.
   current/max encumbrance on the bag icon) rendered on nearly the
   same rows as illegible overlapping text. Now splits the button
   frame between TOP/BOTTOM-aligned label pairs.

## Before / after

1-before-ui-garbled
<br>
image

3-before-bam-frames-blank
4-after-bam-frames-fixed

## Test plan
- [x] Verified via direct pixel-level comparison (before/after) that
      all 11 frames of NUMBER.bam decode correctly post-fix
- [x] Verified in-game: encumbrance overlay in Container.py and
      GUIINV.py renders clean digits instead of garbled text
- [x] Would appreciate a check that the RLE-compressed BAM paths
      (untouched by this fix) still work as expected
- [ ] Need to do a check with BG and BG2 original.
- [ ] Need to do a check with Torment
- [ ] Need to do a check with IWD

### DIFF
--- a/gemrb/plugins/BAMImporter/BAMFontManager.cpp
+++ b/gemrb/plugins/BAMImporter/BAMFontManager.cpp
@@ -75,6 +75,8 @@ Holder<Font> BAMFontManager::GetFont(unsigned short /*ptSize*/, FontStyle /*styl
 	auto pal = af->GetFrameWithoutCycle(0)->GetPalette();
 	auto fnt = MakeHolder<Font>(std::move(pal), lineHeight, baseLine, background);
 
+	ieWord digitWidth = isNumeric ? af->GetFrame(0)->Frame.w : 0;
+
 	std::map<Sprite2D*, ieWord> tmp;
 	for (ieWord cycle = 0; cycle < af->GetCycleCount(); cycle++) {
 		for (ieWord frame = 0; frame < af->GetCycleSize(cycle); frame++) {
@@ -84,6 +86,9 @@ Holder<Font> BAMFontManager::GetFont(unsigned short /*ptSize*/, FontStyle /*styl
 			ieWord chr = '\0';
 			if (isNumeric) {
 				chr = frame + '0';
+				if (frame >= 10 && spr->Frame.w > digitWidth * 3 / 2) {
+					continue;
+				}
 			} else {
 				chr = ((frame << 8) | (cycle & 0x00ff)) + 1;
 			}
--- a/gemrb/plugins/BAMImporter/BAMFontManager.cpp
+++ b/gemrb/plugins/BAMImporter/BAMFontManager.cpp
@@ -77,12 +77,11 @@ Holder<Font> BAMFontManager::GetFont(unsigned short /*ptSize*/, FontStyle /*styl
 	auto pal = af->GetFrameWithoutCycle(0)->GetPalette();
 	auto fnt = MakeHolder<Font>(std::move(pal), lineHeight, baseLine, background);
 
-	ieWord digitWidth = isNumeric ? af->GetFrame(0)->Frame.w : 0;
-
 	if (isNumeric) {
-		ieWord spaceWidth = digitWidth / 2;
-		void* pixels = calloc(spaceWidth * lineHeight, 4);
-		Holder<Sprite2D> spaceSpr = VideoDriver->CreateSprite(Region(0, 0, spaceWidth, lineHeight), pixels, PixelFormat::ARGB32Bit());
+		ieWord spaceWidth = af->GetFrame(0)->Frame.w / 2;
+		void* pixels = calloc(spaceWidth * lineHeight, 1);
+		PixelFormat fmt = PixelFormat::Paletted8Bit(af->GetFrameWithoutCycle(0)->GetPalette(), true, 0);
+		Holder<Sprite2D> spaceSpr = VideoDriver->CreateSprite(Region(0, 0, spaceWidth, lineHeight), pixels, fmt);
 		fnt->CreateGlyphForCharSprite(' ', spaceSpr);
 	}
 
--- a/gemrb/plugins/BAMImporter/BAMFontManager.cpp
+++ b/gemrb/plugins/BAMImporter/BAMFontManager.cpp
@@ -7,6 +7,8 @@
 #include "BAMImporter.h"
 #include "Sprite2D.h"
 
+#include "Video/Video.h"
+
 using namespace GemRB;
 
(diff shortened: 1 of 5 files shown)


gemrb/plugins/BAMImporter/BAMImporter.cpp
```diff
@@ -152,8 +152,14 @@ Holder<Sprite2D> BAMImporter::GetFrameInternal(const FrameEntry& frameInfo, bool
 		if (frameInfo.RLE) {
 			pixels = DecodeRLEData(dataBegin, rgn.size, CompressedColorIndex);
 		} else {
+			// Read this frame's pixels directly from its own absolute offset rather than
+			// slicing it out of the single bulk-read buffer built in GetAnimationFactory();
+			// that buffer + pointer-arithmetic approach was silently dropping data for some
+			// frames of at least one real-world uncompressed BAM (see #<issue-number>).
+			// This matches how Near Infinity's BamV1Decoder reads frame data.
```
MarcelHB: 1. Please don't let the Claude-isms, or whatever this was, creep into here. What _was_ is Ok to have in the commit description, but this is pointless and even incomplete bloat.
2. Shortly looking at NI, the RLE should not be any different, i. e. using an absolute position, too, but feel free to verify.
3. If so, this whole V1 handling looks a little awkward regardless and appears to have survived some refactorings already, so I'd suggest to improve this altogether, e. g. by getting rid of this `data` parameter.

lynxlynxlynx:
First, it looks fine in bg2, so changing a whole format handling should be done carefully to not cause regressions.
image

It also shows the label change shouldn't be needed. The alignment stuff should take care of alignment, so it doesn't matter that they occupy the same bounding box.

The lbs symbol is needed in the originals, the hack is not appropriate.

The EEs used TTF fonts in the end, perhaps for NUMBER{,2} too?
EDIT: this is a screenshot from bgee:
image

BobbyBicc:
> First, it looks fine in bg2, so changing a whole format handling should be done carefully to not cause regressions. image
> 
> It also shows the label change shouldn't be needed. The alignment stuff should take care of alignment, so it doesn't matter that they occupy the same bounding box.
> 
> The lbs symbol is needed in the originals, the hack is not appropriate.
> 
> The EEs used TTF fonts in the end, perhaps for NUMBER{,2} too?

Ah I used BG EE  and BG 2 EE when I tested I will remember to do a check with BG, IWD and Torment and see so it does not break things. Using TTF could work but guess we would lose that nostalgic number look unless we builds a font similar to it.

lynxlynxlynx:
If they used a TTF font, it's already part of the shipped game data. Out of all places I'm in Ukraine right now — without bg2ee with me, so I can't check for you.

MarcelHB:
There is no TTF for `NUMBER`.

gemrb/plugins/BAMImporter/BAMFontManager.cpp
```diff
@@ -75,6 +77,17 @@ Holder<Font> BAMFontManager::GetFont(unsigned short /*ptSize*/, FontStyle /*styl
 	auto pal = af->GetFrameWithoutCycle(0)->GetPalette();
 	auto fnt = MakeHolder<Font>(std::move(pal), lineHeight, baseLine, background);
 
+	ieWord digitWidth = isNumeric ? af->GetFrame(0)->Frame.w : 0;
```
lynxlynxlynx: this can be merged into the if block, since it's only used there.

gemrb/plugins/BAMImporter/BAMFontManager.cpp
```diff
@@ -75,6 +77,17 @@ Holder<Font> BAMFontManager::GetFont(unsigned short /*ptSize*/, FontStyle /*styl
...
+
+	if (isNumeric) {
+		ieWord spaceWidth = digitWidth / 2;
+		void* pixels = calloc(spaceWidth * lineHeight, 1);
+		PixelFormat fmt = PixelFormat::Paletted8Bit(af->GetFrameWithoutCycle(0)->GetPalette(), true, 0);
+		static std::vector<Holder<Sprite2D>> spaceGlyphSprites;
```
lynxlynxlynx: what's the benefit of keeping them around, considering there is no deduplication attempt?

BobbyBicc: Fixed, the spaceSpr does not need to live outside the if statement.

lynxlynxlynx:
The ees and pst are the only ones that ship with a number font that doesn't have the space already embedded in the lbs frame. But the ees deliberately don't want it, so only pst is different in gemrb. This means that the changes in that file are mostly wrong. As for pst, instead of readding space-frame generation, we can keep it clean and instead generate another label for the lb frame on the gui side.

lynxlynxlynx:
Do you plan to finish this?

BobbyBicc:
will try to get some time, works been busy.
Sorry for the delay.

lynxlynxlynx:
No rush, just wondering. The BAMImporter stuff looks fine at a glance, but the rest is not mergeable. You can ignore the pst difference too.

Issue #1244, opened 2021-04-29
worldmap icon pulsation-on-mouse over missing 


lynxlynxlynx:
Split from #1152 to keep that subviews-only.

On hover, it should pulse to red. This is actually data driven in bg2 (and bg1 only) which has all the color info in the mapicons.bam, including the fade in-out of the red overlay at the end. The previous frames are for individual states (normal, visited and so on).

bradallred:
BG1 is actually missing these frames AFICT, but just using a `ColorAnimation` on it produces good results. just a matter of dialing in the colors.

bradallred:
bah, reverence the wrong issue in c7326e9

lynxlynxlynx:
Migrating the relevant info from the other bug, which is now fixed.

bg1 visited, potential yet unvisited (bluish tint) and unvisited (sepia?):
![image](https://user-images.githubusercontent.com/121515/121421109-1de40380-c96e-11eb-923c-125e2876bc7b.png)

....
The colors are a bit too strong in bg1 and unvisited areas are miscolored (this looks to behave differently than bg2 — only known, border accessible areas unvisited areas are blue), but overall it works:
![image](https://user-images.githubusercontent.com/121515/121537154-f8a1d480-ca03-11eb-8149-f77956f8f552.png)

bradallred:
the uncommitted changes I have look decent. the highlight color here is just pure blue 0x0000ffff
<img width="103" alt="image" src="https://user-images.githubusercontent.com/1126744/121561523-cda08c00-c9d5-11eb-803b-30c6e92af3f0.png">

unvisited+unreachable
<img width="96" alt="image" src="https://user-images.githubusercontent.com/1126744/121562283-89fa5200-c9d6-11eb-8ceb-f4f64bd3159c.png">

unfortunately, we dont have a way to pass render flags to text so thats a problem.

lynxlynxlynx:
Can you make the unvisited one even redder? It looks like you just made it translucent.

bradallred:
<img width="92" alt="image" src="https://user-images.githubusercontent.com/1126744/121577654-fbda9780-c9e6-11eb-8a70-544e35b271e7.png">

how's that?

lynxlynxlynx:
Closer, but is it impossible to make it look like in the original? It was lighter there still (see above).

bradallred:
[...]

PR #2505, opened 2026-07-22
WorldMapControl: fix overlapping click hit-regions

### LINKED ISSUE #2498: BG2 EE: Can't leave Waukeen's Promenade
#### Bug description
When attempting to leave Waukeen's Promenade, none of the other districts of Athkatla is accessible. Clicking any location will return to the game. No Python errors.

#### Screenshots

Image

- [x] master as of this issue
- [ ] 0.9.5
- [ ] 0.9.4
- [ ] 0.9.3

BobbyBicc:
Fixes #2498.

## Problem

`OnMouseOver()` grew each area's clickable region to include its
caption text, but the growth didn't match how the caption is actually
drawn: it took `max(icon, text)` for both width and height, anchored
at the icon's own top-left corner. The caption is centered under the
icon (extending both left and right when wider) and drawn below it,
not overlapping it - so a long caption only widened the region
rightward, and never extended it downward past the icon at all.
Neither matched the caption's real footprint, and the resulting
overlap between adjacent areas' regions is what let a click meant for
one district silently register as a neighboring one (eg. "Waukeen's
Promenade").

## Fix

- Grow the region to the union of the icon's rect and the caption's
  actual rect: widen symmetrically around the icon's center when the
  caption is wider (matching the existing centering math used to draw
  it), and extend the height downward by the caption's height plus
  one extra line-height of padding, since the caption sits below the
  icon rather than overlapping it.
- Regions can still legitimately overlap for closely-packed areas even
  with correct sizing, so when the cursor is inside more than one,
  pick whichever area's icon is closest to it instead of whichever
  happens to come first in the world map's area table.

## Screenshots

1-before-hitbox-overlap

image

### DIFF
--- a/gemrb/core/GUI/WorldMapControl.cpp
+++ b/gemrb/core/GUI/WorldMapControl.cpp
@@ -9,6 +9,7 @@
 
 #include "DisplayMessage.h"
 #include "Game.h"
+#include "Geometry.h"
 #include "Interface.h"
 #include "WorldMap.h"
 
@@ -204,6 +205,8 @@ bool WorldMapControl::OnMouseOver(const MouseEvent& me)
 	Area = nullptr;
 
 	unsigned int ec = worldmap->GetEntryCount();
+	WMPAreaEntry* closest = nullptr;
+	unsigned int closestDist = 0;
 	for (unsigned int i = 0; i < ec; i++) {
 		WMPAreaEntry* ae = worldmap->GetEntry(i);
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Interesting. What needs checking is the original — did it have an active region over the label as well? If so, was it similar to ours, but centered, so the overlap was minimal?

The second commit looks suspect though. Multi-frame WMP icons have separate frames, an animation if you will for feedback. The rest were simpler or computed a color, see #1244 for what was dug out so far. So best to keep it out of this PR.

BobbyBicc:
image

The hitbox actually hit on the text to in the orig.
image
 when only centering was applies the overlap was quite small but maybe the fix should be the centering and the font size?
 
 The hover animation is similar maybe some color adjusting is needed but your right it should probably be moved out to it's own PR, I'll work on that and do some test with the font sizes.
 The original fonts used seems quite compact not sure what was used.

BobbyBicc:
image

Seems to be some overlaps in the original, I'll do some test to mimic the result.

lynxlynxlynx:
Ok, perhaps there are some differences (if any) just from the order of the icons getting set up.

So what's with the font size? For me in bg2, it looks normal; it matches the rendering size — it's a fixed size font anyway.

BobbyBicc:
Made a change so it picks the "closest" from the cursor and it the original they seem to have some padding for their hitboxes under the text so added some padding 1x font size height.

image

It's quite hard to verify without the hover feature but if you include the code from that it might be easier to verify the behaviour.
I think it behaves quite similar to BG 2.

gemrb/core/GUI/WorldMapControl.cpp
```diff
@@ -222,23 +224,33 @@ bool WorldMapControl::OnMouseOver(const MouseEvent& me)
...
+				rgn.x -= (ts.w - rgn.w) / 2;
 				rgn.w = ts.w;
+			}
+			rgn.h += ts.h + ftext->LineHeight;
 		}
 		if (!rgn.PointInside(mapOff)) continue;
 
+		long dx = mapOff.x - ae->pos.x;
+		long dy = mapOff.y - ae->pos.y;
+		long dist = dx * dx + dy * dy;
```
lynxlynxlynx: You can use the Distance or SquaredDistance functions instead.

gemrb/core/GUI/WorldMapControl.cpp
```diff
@@ -222,23 +224,33 @@ bool WorldMapControl::OnMouseOver(const MouseEvent& me)
...
+		if (!closest || dist < closestDist) {
+			closest = ae;
+			closestDist = dist;
+		}
+	}
+
+	if (closest) {
```
lynxlynxlynx, line 243: I think this conditional can go and we instead check and shortcircuit much earlier if we don't have any icons to work with. It's only relevant for the `ec == 0` case now.

gemrb/core/GUI/WorldMapControl.cpp
BobbyBicc: SquaredDistance is now used instead.

lynxlynxlynx:
Verified it works as advertised without problems in other games. The little cleanup can be done later, thanks.

PR #2506, opened 2026-07-23
WorldMapControl: hover feedback for the click target (depends on #2505)

### LINKED ISSUE #1244: worldmap icon pulsation-on-mouse over missing

BobbyBicc:
Depends on and is stacked on #2505 - please review/merge that first.

## Problem

Only single-frame icons had any hover feedback (a color pulse gated
on `HighlightSelected()`); multi-frame icons had none, making it easy
to misjudge which district's hitbox the mouse was actually over -
the same underlying confusion behind #2498.

## Fix

Extend the pulse to every icon on hover, mark the control dirty when
the hovered area changes so it redraws right away, and override
`IsAnimated()` so it keeps redrawing while hovered instead of only on
the next unrelated redraw.

## Screenshot

image

### DIFF
--- a/gemrb/core/GUI/WorldMapControl.cpp
+++ b/gemrb/core/GUI/WorldMapControl.cpp
@@ -99,7 +99,7 @@ void WorldMapControl::DrawSelf(const Region& rgn, const Region& /*clip*/)
 			if (!(worldmap->Flags & 1)) flags ^= BlitFlags::COLOR_MOD;
 			if (worldmap->Flags == 2) flags |= BlitFlags::GREY;
 		}
-		if (m == Area && m->HighlightSelected()) {
+		if (m == Area) {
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
bg1:
- text also got recolored (but doesn't cycle)
- bigger color range, probably why it feels faster
- less of a pause on loop, which makes it looks smoother, not jagged

bg2:
- text also got recolored (but doesn't cycle)
- the icon border needs recoloring as well, now it remains blue on unvisited areas and is missing on others

iwd1 and iwd2 had no highlight, so this is a regression.

pst is unaffected, which is fine.

PR #2512, opened 2026-08-14
Fix clang compilation error

draghan:
Restore `Font`'s virtual destructor, to avoid:
`error: destructor called on non-final 'GemRB::Font' that has virtual functions but non-virtual destructor`


lynxlynxlynx:
Thanks!

PR #2516, opened 2026-08-14
Fix dropping paths in tight places

draghan:
When ordering to move the party while being in a thight spot, the ordered movement was often dropped after a while, before the characters arrive at their target goals.

This was a result of 2 compounding and kind-of similar bugs, both tied to ignoring actors.

1) Actor footprints were painted onto wall tiles.
`PaintIfPassable` accepted any tile that was not `IMPASSABLE`. Any actor standing next to the wall stamped `ACTOR` bits onto it. Then, if anyone reading the `TileProps` and passing the predicate for the `ACTOR` will see a hole in the wall.
Fixed that: allow painting actors only on walkable terrain.

2) `IsWalkableTo` were ignoring walls, if there was an actor in the line (if told to ignore actors).
Fixed that by checking the geometry first, before ignoring actors: if there is a wall on the line, it's obviously not walkable.

----

### State before the fix:
<details><summary>BG2, needed 5 clicks to complete this epic journey</summary>

https://github.com/user-attachments/assets/7858d2bf-add4-4266-a3ce-a1f01ece6eaf

</details>

<details><summary>The best Maze, cannot even go to neighboring corridor</summary>

https://github.com/user-attachments/assets/449b5744-89b8-4ddd-89f1-475b1399d891

</details>

### State with the fix:
<details><summary>BG2, now player will be bored I guess</summary>

https://github.com/user-attachments/assets/3f97f062-88b7-4d18-9ff0-a7a1af081965

</details>

<details><summary>The best Maze, could they go to the other side in one go?</summary>

https://github.com/user-attachments/assets/38cb7f1d-21ea-4fb0-aac0-ae7cbb337574

</details>

### DIFF
--- a/gemrb/core/Map.cpp
+++ b/gemrb/core/Map.cpp
@@ -2326,8 +2326,7 @@ bool Map::IsWalkableTo(const Point& s, const Point& d, bool actorsAreBlocking, c
 bool Map::IsWalkableTo(const SearchmapPoint& s, const SearchmapPoint& d, bool actorsAreBlocking, const Actor* caller) const
 {
 	PathMapFlags ret = PathFinder::GetBlockedInLineTile(tileProps, s, d, true, caller);
-	PathMapFlags mask = PathMapFlags::PASSABLE | (actorsAreBlocking ? PathMapFlags::UNMARKED : PathMapFlags::ACTOR);
-	return bool(ret & mask);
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Excellent, this will greatly improve the player experience, thanks!

PR #2517, opened 2026-08-15
Pathfinding tests: insfrastructure and basics

draghan:
Implement basic infrastructure for pathfinding-related tests, so that they won't need any prepared game-ready data, the test cases could be defined in-place and stay readable.

Add basic tests covering the infra, the walkable and LOS, and also the actors' painting.

### DIFF
--- a/gemrb/tests/core/SearchMapBuilder.h
+++ b/gemrb/tests/core/SearchMapBuilder.h
@@ -5,13 +5,9 @@
 #ifndef TESTS_SEARCHMAPBUILDER_H
 #define TESTS_SEARCHMAPBUILDER_H
 
-#include "../../core/Holder.h"
 #include "../../core/Region.h"
-#include "../../core/Sprite2D.h"
 #include "../../core/TileProps.h"
 
-#include <cstdlib>
-#include <cstring>
 #include <gtest/gtest.h>
 #include <initializer_list>
 #include <string>
@@ -51,7 +47,7 @@ namespace test {
 	/**
 	 * A map drawing, one string per row.
 	 */
-	using MapRows = std::initializer_list<const char*>;
+	using MapRows = std::initializer_list<std::string>;
 
 	inline bool IsActorGlyph(const char c)
 	{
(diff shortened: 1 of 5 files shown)


gemrb/tests/core/SearchMapBuilder.h
```diff
@@ -0,0 +1,258 @@
...
+				return PathMapFlags::PASSABLE | PathMapFlags::TRAVEL;
+			case 'D':
+				return PathMapFlags::PASSABLE | PathMapFlags::DOOR_IMPASSABLE;
+			case 'O':
+				return PathMapFlags::PASSABLE | PathMapFlags::DOOR_IMPASSABLE | PathMapFlags::DOOR_OPAQUE;
+			default:
+				// actors stand on plain floor; their footprint is painted afterwards
+				if (IsActorGlyph(c)) return PathMapFlags::PASSABLE;
+				// anything else is a typo in the literal, not terrain
+				abort();
```
Mingun: Probably need to print a character to std::cerr before aborting.

lynxlynxlynx: & Would a cleaner `ASSERT_TRUE(false)` not work in this context?

gemrb/tests/core/SearchMapBuilder.h
```diff
@@ -0,0 +1,258 @@
...
+	 *
+	 * Owns the pixel buffer behind the TileProps. Sprite2D free()s the buffer it is handed,
+	 * so it has to come from malloc().
+	 */
+	class TestSearchMap {
+	public:
+		TestSearchMap(const MapRows inMapRows)
+		{
+			height = static_cast<int>(inMapRows.size());
+			width = height ? static_cast<int>(strlen(*inMapRows.begin())) : 0;
```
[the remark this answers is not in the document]
lynxlynxlynx: Just use a std::string for storage instead?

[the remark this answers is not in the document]
draghan: Oh, sure. I was under the impression that `std::initializer_list` won't allow for implicit conversion and we will end up with over-explicit bloat on the call site:
```cpp
	const TestSearchMap map {
		std::string{"........"},
		std::string{"..1..a.."},
		std::string{"........"}
	};

```
instead of just:
```cpp
	const TestSearchMap map {
		"........",
		"..1..a..",
		"........"
	};
```
...but I was wrong, it works fine. Thanks.

gemrb/tests/core/SearchMapBuilder.h
draghan: Sure, aborting without explanation is not the best handling.
`ASSERT_*` won't work, they are designed for use in test's definition.
I didn't want it coupled with the test framework, but I guess it is meant just for tests, so it won't be that bad, no?
Changed it to `ADD_FAILURE`.
Thanks!

draghan:
Changed to use the fresh `OwningTileProps`, to silence all the sonar's complaints about raw allocs.

lynxlynxlynx:
Thanks!

PR #2518, opened 2026-08-15
Refactor: extract `IsOverCircle` and `LineStepper`

draghan:
Prep work for more pathfinder tests. Introduced in a separate PR, as it touches a logic in the engine.

Extract `IsOverCircle` static method, so that it can be used without valid `Selectable` instance.

Extract `LineStepper` - code for stepping in a line between 2 points, which was duplicated in 2 places already. Adding this logic to tests would duplicate it 3rd time.


lynxlynxlynx:
Thanks!

PR #2520, opened 2026-08-16
Implement more pathfinding tests

draghan:
Add GEM_EXPORT to TraversabilityCache, to use it in tests.

Add Glyphs as named constants.

Implement traversability cache mock.

Add ability to define actors from the glyps.

Add tests for asserting the infra added above behaves correctly.

Add tests for actual FindPath.

### DIFF
--- a/gemrb/tests/core/SearchMapBuilder.h
+++ b/gemrb/tests/core/SearchMapBuilder.h
@@ -5,8 +5,11 @@
 #ifndef TESTS_SEARCHMAPBUILDER_H
 #define TESTS_SEARCHMAPBUILDER_H
 
+#include "../../core/PathFinder.h"
 #include "../../core/Region.h"
+#include "../../core/Scriptable/Selectable.h"
 #include "../../core/TileProps.h"
+#include "../../core/TraversabilityCache.h"
 
 #include <gtest/gtest.h>
 #include <initializer_list>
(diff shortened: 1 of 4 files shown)


gemrb/tests/core/SearchMapBuilder.h
```diff
@@ -204,21 +318,219 @@ namespace test {
...
+			if (tiles.empty() || !(tiles.back() == tile)) tiles.push_back(tile);
+		};
+
+		Point p = from;
+		appendIfNotPresent(SearchmapPoint { p });
+		for (size_t i = 0; i < path.Size(); ++i) {
+			PathFinder::LineStepper<NavmapPoint> walk { p, path.GetStep(i).point };
+			while (walk.Step()) {
+				if (walk.Current() == p) {
+					// a step should always moves, bail out if that ever stops holding
```
lynxlynxlynx:
```suggestion
					// a step should always move, bail out if that ever stops holding
```

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -185,6 +290,65 @@ TEST(SearchMapBuilderTest, TerrainRoundTrips)
...
+	const Point where = map.ActorPosOf(0);
+
+	// the centre carries a bumpable actor's single token ...
+	EXPECT_EQ(traversability.StateAt(where), TraversabilityCache::TraversabilityCellValueActor);
+	// ... and somewhere well outside the circle carries none
+	EXPECT_EQ(traversability.StateAt(TestSearchMap::Nav(1, 1)), TraversabilityCache::TraversabilityCellValueEmpty);
+
+	// a second, non-bumpable actor on the same spot pushes it over the always-blocking mark
+	traversability.AddActor(where, 1, false);
+	EXPECT_GE(traversability.StateAt(where), TraversabilityCache::TraversabilityCellValueActorNonTraversable);
```
lynxlynxlynx, line 417: It's not relevant for this specific test, but in practice we should always avoid actors having the same position — we try to bump them to the nearest available space even when data asks otherwise. The originals did the same, in fact it was used for speedrunning records, since it could make you jump walls. :)

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -326,22 +530,321 @@ TEST(PathFinderTest, ActorBlocksOnlyWhenActorsAreBlocking)
...
+
+	// intending to bump: the blocker is transparent, so the route runs straight through it
+	const Path bumping = test::CallFindPath(map, traversability, from, to, self, 1, PF_SIGHT);
+	EXPECT_FALSE(bumping.Empty()) << "a bumpable actor must not stop a route that would bump it";
+
+	// treating actors as solid: there is no way past in a one tile corridor
+	const Path yielding = test::CallFindPath(map, traversability, from, to, self, 1,
+						 PF_SIGHT | PF_ACTORS_ARE_BLOCKING);
+	EXPECT_TRUE(yielding.Empty()) << "a blocking actor plugs the only corridor, so there is no route";
+}
```
lynxlynxlynx, line 644: This one would also be useful in a copy with "a1" input, since currently bumping only considers one actor at a time (I'm pretty sure the originals did the same).

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -326,22 +530,321 @@ TEST(PathFinderTest, ActorBlocksOnlyWhenActorsAreBlocking)
...
+	const test::TestTraversability traversability { map };
+	const Path path = test::CallFindPath(map, traversability, start, dest, nullptr, map.ActorCircleSizeOf(0));
+
+	ASSERT_FALSE(path.Empty());
+	EXPECT_TRUE(test::PathAvoidsWalls(map, start, path));
+
+	// it has to come round the open bottom row, so it has to get more >1 waypoint
+	EXPECT_GT(path.Size(), 1u);
+
+	// it must never set foot on the barrier
```
lynxlynxlynx: Isn't this sufficiently covered by the PathAvoidsWalls test above?

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -326,22 +530,321 @@ TEST(PathFinderTest, ActorBlocksOnlyWhenActorsAreBlocking)
...
+		"#.........#.........#",
+		"#.........#.........#",
+		"#####################"
+	};
+
+	constexpr size_t smallActor = 0;
+	constexpr size_t bigActor = 1;
+	ASSERT_EQ(map.Actors().size(), 2);
+	ASSERT_EQ(map.ActorCircleSizeOf(smallActor), 1);
+	ASSERT_EQ(map.ActorCircleSizeOf(bigActor), 4);
```
lynxlynxlynx: These three would better fit into ParsesActorCharacters or their own test.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -326,22 +530,321 @@ TEST(PathFinderTest, ActorBlocksOnlyWhenActorsAreBlocking)
...
+TEST(FindPathTest, RespectsActorSize)
+{
+	// two halls joined by a single one tile slit at (10,5), which is the only way across
+	const TestSearchMap map {
+		"#####################",
+		"#.........#.........#",
+		"#.........#.........#",
+		"#.........#.........#",
+		"#.........#.........#",
+		"#..1..4........E....#",
```
lynxlynxlynx: I suggest you use 3, so it's not an illegal position. Or move 1 one to the left.

draghan: Can't this scenario happen temporarily when bumped?

gemrb/tests/core/Test_PathFinder.cpp
draghan: Do you mean 2 actors (PC and NPC) standing next to each other, like that?
```cpp
	const TestSearchMap map {
		"##########",
		"#S..a1..E#",
		"##########"
	};
```

gemrb/tests/core/Test_PathFinder.cpp
draghan: Ah yes, it is. It's a leftover from earlier version. Removed, thanks.

gemrb/tests/core/Test_PathFinder.cpp
lynxlynxlynx: Yeah, could be "aa" to be simpler.

gemrb/tests/core/Test_PathFinder.cpp
lynxlynxlynx: I don't think so, BumpAway gets called before the new position is set in DoStep.

gemrb/tests/core/Test_PathFinder.cpp
draghan: Added 2 separate SearchMapBuilderTest tests.

gemrb/tests/core/Test_PathFinder.cpp
draghan: We will verify that when we will have actual movement tests (it's a bit tricky to do, but I think I got infra almost ready for this).
For now, I'll change this to 3, per your suggestion.

gemrb/tests/core/Test_PathFinder.cpp
draghan: Do we expect `FindPath` to fail in this case? Because it's not failing, works exactly as original test.
We can't test actual bump mechanic right now, it will come with movement tests.

lynxlynxlynx: On first thought I would expect a failure, but on the second, a success is probably a better expectation, since we have a back-off mechanic.

draghan: You call `FindPath` and it either ignores all the actors or not, depending on `PF_ACTORS_ARE_BLOCKING` flag. So there can't be different result here.

lynxlynxlynx: So there's no conflict.

lynxlynxlynx:
Thanks!

PR #2521, opened 2026-08-16
Add infrastructure for movement tests

draghan:
Added test data to demo:
- avatars.2da - 6 rows giving circle sizes 3–8, reusing CHMB1 art,

Added GEM_EXPORT to PathFinderScheduler so tests can call Start/Stop.

New test infrastructure:
- LiveGameFixture.h - headless Interface + demo game; GameControl, unpause, pathfinder scheduler in immediate mode
- TestGameMap.h - real Map from an ASCII drawing, actors spawned from glyphs, TestGameLoop::RunFrame()
- Test_Movement.cpp - 6 basic movement tests: walking, wall-crossing per frame, sizes, party membership, no stale footprint

Changed test infrastructure:
- SearchMapBuilder.h - ActorPainting enum so we can skip painting actors 'by hand'
- Test_Map.cpp - migrated to use LiveGameTest,
- CMakeLists.txt - registers Test_Movement.cpp.

### DIFF
--- a/gemrb/tests/core/Test_Movement.cpp
+++ b/gemrb/tests/core/Test_Movement.cpp
@@ -1,0 +1,230 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the GemRB project <https://gemrb.org>
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Movement tests: the real per-frame loop, with bumping, backoff and repathing.
+// The live-game scaffolding lives in TestGameFixture.h; what is here is movement specific.
+
+#if defined(USE_OPENGL_BACKEND) || (!defined(__APPLE__) && !defined(WIN32))
+
+	#include "TestGameMap.h"
+
+	#include <gtest/gtest.h>
+
+namespace GemRB {
+
+
+class MovementTest : public GameMapTest {
+};
+
+// === helpers ===
+
+// The invariant that catches tunnelling. One frame moves the actor a good fraction of a tile,
+// so it can stand on open floor before the frame and on open floor after it and still have
+// passed straight through a wall in between. Speed 0 samples the segment as finely as the
+// engine ever steps, so a wall cannot hide between two samples.
+static testing::AssertionResult StepStayedOffWalls(const TestSearchMap& drawn, const Point& from, const Point& to)
+{
+	constexpr int noSpeed = 0;
+	constexpr int noCircle = 0;
+	const PathMapFlags crossed = PathFinder::GetBlockedInLine(drawn.Props(), from, to, false, noSpeed, noCircle);
+	if (bool(crossed & (PathMapFlags::SIDEWALL | PathMapFlags::DOOR_IMPASSABLE))) {
+		return testing::AssertionFailure()
+			<< "step (" << from.x << ',' << from.y << ") -> (" << to.x << ',' << to.y << ") crosses a wall";
+	}
+	return testing::AssertionSuccess();
+}
+
+// Runs frames until the actor stops moving, checking every single step against a wall. Returns
+// how many frames it took, or the budget if it never arrived.
+static int WalkUntilStopped(const TestGameMap& live, Actor* actor, int frameBudget = 200)
+{
+	int frames = 0;
+	for (; frames < frameBudget && actor->InMove(); ++frames) {
+		const Point before = actor->Pos;
+		TestGameLoop::RunFrame();
+		EXPECT_TRUE(StepStayedOffWalls(live.Drawing(), before, actor->Pos)) << "on frame " << frames;
+	}
+	return frames;
+}
+
+// === TestGameMap infrastructure tests ===
+
+// A drawn map can carry a real actor, and the real pathfinder answers for it.
+TEST_F(MovementTest, DrawnMapCarriesARealActor)
+{
+	// '2' is a party member of circle size 2, standing where the glyph is
+	const TestGameMap live {
+		"##########",
+		"#........#",
+		"#..2....E#",
+		"#........#",
+		"##########"
+	};
+	const TestSearchMap& drawn = live.Drawing();
+
+	ASSERT_EQ(live.ActorCount(), 1);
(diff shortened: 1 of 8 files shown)


gemrb/tests/core/Test_Movement.cpp
```diff
@@ -0,0 +1,230 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the GemRB project <https://gemrb.org>
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Movement tests: the real per-frame loop, with bumping, backoff and repathing.
+// The live-game scaffolding lives in TestGameFixture.h; what is here is movement specific.
+
+#if defined(USE_OPENGL_BACKEND) || (!defined(__APPLE__) && !defined(WIN32))
```
draghan, line 8: I'm not sure about this conditional compilation. I copied it over from Test_Map, but I don't know if it should be still kept.

lynxlynxlynx: It's fine, it was added due to buildbot failures (opengl is enabled only on release builds, since runners can't do full tests otherwise). And for the vast majority of tests its enough to be ran once.

gemrb/tests/core/Test_Movement.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+	// the live actors reach the cache the pathfinder consults, each owning its own cell
+	live.RefreshTraversability();
+	EXPECT_GT(live.StateAt(drawn.ActorPosOf(0)), TraversabilityCache::TraversabilityCellValueEmpty);
+	EXPECT_EQ(live.ActorAt(drawn.ActorPosOf(0)), live.ActorOf(0));
+	EXPECT_EQ(live.ActorAt(drawn.ActorPosOf(1)), live.ActorOf(1));
+}
+
+// Every size the glyph alphabet allows, has a creature behind it.
+// Sizes 3 to 8 are the test-only creatures in demo/override.
```
lynxlynxlynx: this was obviated

gemrb/tests/core/Test_Movement.cpp
```diff
@@ -0,0 +1,230 @@
...
+	Actor* actor = live.ActorOf(0);
+	ASSERT_NE(actor, nullptr);
+
+	actor->WalkTo(drawn.End(), 0, 0);
+	ASSERT_FALSE(actor->GetPath().Empty()) << "the walk needs a path to follow";
+
+	const int frames = WalkUntilStopped(live, actor);
+	EXPECT_LT(frames, 200) << "the walk has to finish inside the frame budget";
+	EXPECT_GT(frames, 1) << "arriving in one frame would mean it teleported, not walked";
+	EXPECT_EQ(actor->Pos, drawn.End()) << "and it has to end up where it was sent";
```
lynxlynxlynx, line 164: How does this work? Cells denote the middle of a searchmap rect, so an actor of size 2 should end up on the border between two cells. Don't know which way we'd round, but because of the wall, I'd expect either way for the path to end one cell sooner than E.

draghan: From what I understand, we never try to place an actor between tiles.

There are 2 "footprints" related to the actor's circle size.
First is the footprint blocking other actors. Defined here:
https://github.com/gemrb/gemrb/blob/5891a414cdd1757ee53dd5f055d4a225a97658c6/gemrb/core/TileProps.cpp#L140-L180

Second is the standing footprint, `FindPath` consults to decide "can this actor stand here?". Defined here:
https://github.com/gemrb/gemrb/blob/5891a414cdd1757ee53dd5f055d4a225a97658c6/gemrb/core/PathFinder.cpp#L536-L576

Blocking other actors is bigger, the standing one is smaller.
An actor's position always resolves to exactly one tile, and the standing footprint is centred on that tile.
Illustration of the footprints:

```cpp
=== size 1 ===  standing (S)      blocking (P)
                #############     #############
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #.....S.....#     #.....P.....#
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #############     #############

=== size 2 ===  standing (S)      blocking (P)
                #############     #############
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #....PPP....#
                #.....S.....#     #....PPP....#
                #...........#     #....PPP....#
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #...........#
                #############     #############

=== size 3 ===  standing (S)      blocking (P)
                #############     #############
                #...........#     #...........#
                #...........#     #...........#
                #...........#     #....PPP....#
                #....SSS....#     #...PPPPP...#
                #....SSS....#     #...PPPPP...#
                #....SSS....#     #...PPPPP...#
                #...........#     #....PPP....#
                #...........#     #...........#
                #...........#     #...........#
                #############     #############

=== size 4 ===  standing (S)      blocking (P)
                #############     #############
                #...........#     #...........#
                #...........#     #....PPP....#
                #....SSS....#     #...PPPPP...#
                #...SSSSS...#     #..PPPPPPP..#
                #...SSSSS...#     #..PPPPPPP..#
                #...SSSSS...#     #..PPPPPPP..#
                #....SSS....#     #...PPPPP...#
                #...........#     #....PPP....#
                #...........#     #...........#
                #############     #############
```


For actor's circle size of 2, standing footprint has radius=0, which gives one tile only - and this will be the exact tile we command it to move onto.
For any bigger actor, he won't fit with the center tile against the wall - there is new test added just for that.

I'm not sure I answered your concern. ^^'

gemrb/tests/core/Test_Movement.cpp
draghan: removed

gemrb/tests/core/Test_Movement.cpp
lynxlynxlynx: Yes, thanks.

PR #2522, opened 2026-08-23
Extend test coverage in Test_PathFinder

draghan:
Added also `AsRanges` helper to `String.h`.

### DIFF
--- a/gemrb/tests/core/SearchMapBuilder.h
+++ b/gemrb/tests/core/SearchMapBuilder.h
@@ -513,13 +513,13 @@ namespace test {
 	 * A distance in whole searchmap tiles, as navmap pixels.
 	 *
 	 * A tile is 16x12, so there is no single pixel count for a tile; FindPath() measures with
-	 * Distance() on tile coordinates scaled by SEARCHMAP_SQUARE_DIAGONAL, which is 20. That is
+	 * Distance() on tile coordinates scaled by SEARCHMAP_SQUARE_DIAGONAL. That is
 	 * the number a minDistance argument is really in, so this is what a test means by "n tiles
 	 * away".
 	 */
-	constexpr unsigned int Tiles(const unsigned int count) noexcept
+	constexpr unsigned int Tiles(const unsigned int tilesCount) noexcept
 	{
-		return count * 20;
+		return tilesCount * SEARCHMAP_SQUARE_DIAGONAL;
 	}
 
 	/**
--- a/gemrb/tests/core/SearchMapBuilder.h
+++ b/gemrb/tests/core/SearchMapBuilder.h
@@ -5,14 +5,20 @@
 #ifndef TESTS_SEARCHMAPBUILDER_H
 #define TESTS_SEARCHMAPBUILDER_H
 
+#include "../../core/GameData.h"
+#include "../../core/Geometry.h"
 #include "../../core/PathFinder.h"
 #include "../../core/Region.h"
 #include "../../core/Scriptable/Selectable.h"
+#include "../../core/Strings/String.h"
 #include "../../core/TileProps.h"
 #include "../../core/TraversabilityCache.h"
 
+#include <cmath>
+#include <cstdlib>
 #include <gtest/gtest.h>
 #include <initializer_list>
+#include <memory>
 #include <string>
 #include <vector>
 
@@ -242,6 +248,28 @@ namespace test {
 		TileProps& Props() noexcept { return props; }
 		const TileProps& Props() const noexcept { return props; }
 
+		/**
+		 * Rewrites one tile's terrain, as the glyph of the same name would have set it.
+		 * For a test that has to change the map between two runs - opening a door, or taking
+		 * an obstacle away - so both runs can be stated against the same drawing.
+		 * Actor glyphs are not accepted: they imply a footprint and a cache entry, which a
+		 * single tile write cannot deliver.
+		 */
+		void SetTile(const int x, const int y, const char glyph)
+		{
+			if (IsActorGlyph(glyph)) {
+				ADD_FAILURE() << "SetTile() cannot place actor '" << glyph
+					      << "': draw it on the map instead";
+				return;
+			}
+			if (x < 0 || y < 0 || x >= width || y >= height) {
+				ADD_FAILURE() << "SetTile() called outside the map, at (" << x << ',' << y << ')';
+				return;
+			}
+			props.SetTileProp(SearchmapPoint(x, y), TileProps::Property::SEARCH_MAP,
+					  uint8_t(ParseSearchMapChar(glyph)));
+		}
+
 		int Width() const noexcept { return width; }
 		int Height() const noexcept { return height; }
 
@@ -339,6 +367,15 @@ namespace test {
 		 * list of coordinates, and shows at a glance how much of it the pathfinder chose.
 		 */
 		testing::AssertionResult MatchesWithPath(const Point& from, const Path& path, MapRows expected) const;
+
+		/**
+		 * As Matches(), with the actor mark on every tile the standing footprint of an actor of
+		 * that circle size covers, when it stands on `centre`.
+		 *
+		 * The ground the actor needs for itself, which GetBlockedInRadiusTile() asks about.
+		 */
+		testing::AssertionResult MatchesStandingFootprint(const SearchmapPoint& centre, uint16_t circleSize,
+								  MapRows expected) const;
 
 		// centre of a tile in navmap coordinates, which is what the Point-taking overloads want
 		static Point Nav(int x, int y) noexcept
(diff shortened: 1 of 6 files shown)


draghan:
@lynxlynxlynx please review all the `DISABLED_` tests with special care, check if I got the expectations correctly.
If so, we have few defects to fix.

I expect to remove the `DISABLED_` status gradually when fixing the discovered bugs.

gemrb/tests/core/SearchMapBuilder.h
```diff
@@ -444,25 +481,45 @@ namespace test {
...
-				    map.ActorIdentityOf(index), map.ActorCircleSizeOf(index), flags);
+				    map.ActorIdentityOf(index), map.ActorCircleSizeOf(index), flags,
+				    minDistance);
+	}
+
+	/**
+	 * A distance in whole searchmap tiles, as navmap pixels.
+	 *
+	 * A tile is 16x12, so there is no single pixel count for a tile; FindPath() measures with
+	 * Distance() on tile coordinates scaled by SEARCHMAP_SQUARE_DIAGONAL, which is 20. That is
```
lynxlynxlynx: I suggest you move SEARCHMAP_SQUARE_DIAGONAL to the header, so it can also be used here directly.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+	};
+
+	const Point from = map.Start();
+	const Point to = map.End();
+	const test::TestTraversability traversability { map };
+
+	// with no distance to keep, the walk ends on the goal itself
+	const Path onto = test::CallFindPath(map, traversability, from, to);
+	ASSERT_FALSE(onto.Empty());
+	EXPECT_EQ(SearchmapPoint(onto.GetStep(onto.Size() - 1).point), SearchmapPoint { to });
```
lynxlynxlynx: I suggest you add `Path::GetLastStep()`, since you use it 17 times just in this PR.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+
+	// asked to stop five tiles out, it ends within that of the goal but not on it
+	constexpr unsigned int keepAway = test::Tiles(5);
+	const Path shortOf = test::CallFindPath(map, traversability, from, to, nullptr, 1, PF_SIGHT, keepAway);
+	ASSERT_FALSE(shortOf.Empty());
+	EXPECT_TRUE(test::PathIsSane(map, from, shortOf));
+
+	const Point last = shortOf.GetStep(shortOf.Size() - 1).point;
+	EXPECT_NE(SearchmapPoint(last), SearchmapPoint { to }) << "it was told to keep its distance";
+	EXPECT_LT(unsigned(Distance(last, to)), keepAway) << "and to come at least that close";
```
lynxlynxlynx: at most?

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+					     noSpeed, noCircle))
+		<< "with PF_SIGHT it may only stop where it can see the goal";
+
+	// which costs it the walk around the divider
+	EXPECT_GT(test::PathLength(from, seeing), test::PathLength(from, blind));
+}
+
+// Telling a walker to keep its distance has to shorten the walk, wherever the click landed.
+//
+// DISABLED: the early stop checks its line of sight against the click point rather than against
```
lynxlynxlynx, line 990: a bug

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+	// encumbrance at half speed doubles it, so the usable range across the supported games is roughly
+	// 0-1500. The sweeps below run past the games' shipped values.
+	constexpr int slowestSpeed = 2000;
+
+}
+
+// How fast the observer walks has nothing to do with whether a wall is in the way, so every wall
+// on the line has to block sight at every speed, wherever it stands and however thick it is.
+//
+// DISABLED: The tile space line walk borrows NormalizeDeltas(), a navmap function whose
```
lynxlynxlynx, line 1074: as discussed, PathFinder::GetBlockedInLineTile should ignore speed, also making this test (and a couple later ones) redundant.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+	ASSERT_FALSE(fromEast.Empty());
+	EXPECT_TRUE(test::PathIsSane(map, east, fromEast));
+	EXPECT_TRUE(test::PathAvoidsWalls(map, east, fromEast));
+	EXPECT_EQ(SearchmapPoint(fromEast.GetStep(fromEast.Size() - 1).point), SearchmapPoint(6, 2))
+		<< "and the east caller against the east face";
+}
+
+// A click into rock walks up to the face of the wall the caller is standing on.
+//
+// DISABLED: the directed adjustment only reaches one tile out, so against a wall two tiles thick
```
lynxlynxlynx, line 1182: hmm, this is more of a design question. The lookback distance could always be increased, but it's also true that in bg1 one can get arbitrarily big parties, so formation rotation can result in goals very deep in walls. Doing our best (like the test expects) even for inaccessible areas seems like the most obvious thing to do.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+
+	for (const bool bumpable : { true, false }) {
+		const test::TestTraversability traversability { map, bumpable };
+		const Path path = test::CallFindPath(map, traversability, from, onto, nullptr, 1,
+						     PF_SIGHT | PF_ACTORS_ARE_BLOCKING);
+		ASSERT_FALSE(path.Empty()) << "bumpable = " << bumpable;
+		EXPECT_TRUE(test::PathIsSane(map, from, path)) << "bumpable = " << bumpable;
+		const SearchmapPoint last { path.GetStep(path.Size() - 1).point };
+		EXPECT_NE(last, targetTile) << "bumpable = " << bumpable << ": the route ended on the target";
+		EXPECT_LE(std::abs(last.x - targetTile.x), 1) << "bumpable = " << bumpable;
```
lynxlynxlynx, line 1252: EXPECT_EQ? When could it not be 1?

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+
+	ASSERT_FALSE(backing.Empty());
+	EXPECT_EQ(test::CountBackAwayOrients(from, backing), 0u) << "the corners here are all square";
+	EXPECT_EQ(test::PathOrients(backing), test::ForwardOrients(from, backing));
+}
+
+// An actor backing away faces the threat it is retreating from, not the way it is going.
+//
+// DISABLED: Theta* collapses a straight run to a single waypoint, and FindPath() always gives its
+// first built node a forward orientation. That node is the only one, so on the plainest retreat
```
lynxlynxlynx: this was mentioned several times already, but it should be fixed either way. The exception was added for a very specific case, which can be revisited after other fixes — perhaps it will end up being obviated.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -930,6 +899,1368 @@ TEST(FindPathTest, PathUTurn)
...
+	const Point from = map.Start();
+	const test::TestTraversability traversability { map };
+	const Path path = test::CallFindPath(map, traversability, from, map.End());
+
+	ASSERT_FALSE(path.Empty()) << "the maze has exactly one way through and it must be found";
+	EXPECT_TRUE(test::PathIsSane(map, from, path));
+	EXPECT_TRUE(test::PathAvoidsWalls(map, from, path));
+
+	// a range rather than a number: the corner count is a tuning detail, but a route which has
+	// collapsed to one or two legs cannot have gone round the bends
```
lynxlynxlynx, line 1577: isn't this robustly enough caught by PathAvoidsWalls?

lynxlynxlynx:
Overall very nice and it will be handy not just for resolving the bugs the disabled tests show, but also when fixing the circle size geometries and diagonal pathing (this one will need more tests).

gemrb/tests/core/SearchMapBuilder.h
draghan: Much better, thanks.

gemrb/tests/core/Test_PathFinder.cpp
draghan: Nice, thanks. Done.

gemrb/tests/core/Test_PathFinder.cpp
draghan: Reworded this msg.

gemrb/tests/core/Test_PathFinder.cpp
draghan: good, thanks for confirmation.

gemrb/tests/core/Test_PathFinder.cpp
draghan: I'd be reluctant to remove this test, after all it spells out a valid and important invariant.

gemrb/tests/core/Test_PathFinder.cpp
draghan: I didn't think yet about how to fix this. This will be helpful, thanks. I've put it directly to the `DISABLED` reasoning, so it won't get lost forgotten here.

gemrb/tests/core/Test_PathFinder.cpp
draghan: I thought _any_ tile adjacent to the target should be fine here, so also a tile above and below, no?

gemrb/tests/core/Test_PathFinder.cpp
draghan: expanded this comment, so that anyone that will pick up this later will have the context

gemrb/tests/core/Test_PathFinder.cpp
draghan: For such constructed map - yes, should be enough, but 'avoid walls' is a by-product of what we want to assure here. We spell out assertion about a different property here: we expect certain number of waypoints.

draghan:
> Overall very nice and it will be handy not just for resolving the bugs the disabled tests show, but also when fixing the circle size geometries and diagonal pathing (this one will need more tests).

Thanks! And thank you for the review, I know it's a lot of code to read. 

TBH I think the pathfinding tests are kind of boring, actual movement tests will be more interesting, with asserting all the intermittent states. I want to catch the tunneling I saw in the BG1 Maze - after my recent changes fixing stability of the movement I haven't seen it yet, but I'm sure it's hidden somewhere in the shadows. ಠ_ಠ

lynxlynxlynx:
You don't think that one was just the same painting bug with actors adjacent to walls clearing them on repathing with non-blocking actors?

gemrb/tests/core/Test_PathFinder.cpp
lynxlynxlynx: Ah, right, optimality doesn't matter here.

draghan:
Yes, I think it could have been a separate issue. Laying out a path is a separate logic from actual changing position of the movable. I saw clearly that pathfinder was actually pathing through walls, but actors trying to go along that path just stopped on the walls, like they should. But actual yeeting to the other side must have something to do with maybe bumping?

lynxlynxlynx:
Thanks!

PR #2524, opened 2026-08-30
Fix compilation under g++ 11.5

draghan:
The g++ 11.5 can't compile CREImporter.cpp:

```
88:/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: error: ‘char* __builtin_strncpy(char*, const
char*, long unsigned int)’ offset [0, 7] is out of the bounds [0, 0] [-Werror=array-bounds]
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘GemRB::FixedSizeString<LEN, CMP>& GemRB::FixedSizeString<LEN, CMP>::operator=(const STR&) [
with STR = char [5]; typename std::enable_if<((((! std::is_enum<T>::value) && (! std::is_fundamental<STR>::
value)) && (! std::is_pointer<T>::value)) || std::is_same<STR, const char (&)[1]>::value), int>::type
<anonymous> = 0; long unsigned int LEN = 8; int (* CMP)(const char*, const char*, size_t) = strncasecmp]’ at
/tmp/opencode/gemrb-master/gemrb/core/Strings/CString.h:122:10,
    inlined from ‘virtual GemRB::Actor* GemRB::CREImporter::GetActor(unsigned char)’ at /tmp/opencode/gemrb-
master/gemrb/plugins/CREImporter/CREImporter.cpp:883:25:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: error: ‘char* __builtin_strncpy(char*, const
char*, long unsigned int)’ offset [0, 7] is out of the bounds [0, 0] [-Werror=array-bounds]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Trigger line is CREImporter.cpp:883: `act->SmallPortrait = "NONE";` ("NONE" = char[5], target ResRef=FixedSizeString<8>).

The new `CopyFrom` helper first measures the actual length of the source range, then copies only `min(sourceLength, LEN)` characters into the fixed buffer and fills the rest of the buffer with zeros. 
Because the copy never asks to read more bytes than the source actually contains, GCC's bounds check passes. 
The buffer is still always zero-padded and NUL-terminated, the behavior is unchanged.

### DIFF
--- a/gemrb/core/Strings/CString.h
+++ b/gemrb/core/Strings/CString.h
@@ -20,6 +20,8 @@
 #include <cstdio>
 #include <cstring>
 #include <cwctype>
+#include <iterator>
+#include <memory>
 
 #ifndef WIN32
 	#define stricmp  strcasecmp
@@ -87,6 +89,22 @@ class FixedSizeString {
 	static_assert(LEN < 255, "Cannot create FixedSizeString larger than 255 characters");
 	char str[LEN + 1] { '\0' };
 
+	template<typename STR>
+	void CopyFrom(const STR& s) noexcept
+	{
+		const auto src = std::begin(s);
+		const auto srcLen = static_cast<size_t>(std::distance(src, std::end(s)));
+		if (srcLen == 0) {
+			std::fill(begin(), bufend(), '\0');
+		} else {
+			size_t n = std::min(srcLen, LEN);
+			std::copy_n(std::addressof(*src), n, str);
+			if (n < LEN) {
+				std::fill_n(str + n, LEN - n, '\0');
+			}
+		}
+	}
+
 public:
 	using size_type = uint8_t;
 	using iterator = char*;
@@ -113,13 +131,13 @@ public:
 	template<typename STR, ENABLE_CHAR_RANGE(STR)>
 	FixedSizeString(const STR& s) noexcept
 	{
-		strncpy(str, &s[0], LEN);
+		CopyFrom(s);
 	}
 
 	template<typename STR, ENABLE_CHAR_RANGE(STR)>
 	FixedSizeString& operator=(const STR& s) noexcept
 	{
-		strncpy(str, &s[0], LEN);
+		CopyFrom(s);
 		return *this;
 	}
 
--- a/gemrb/tests/core/Strings/Test_CString.cpp
+++ b/gemrb/tests/core/Strings/Test_CString.cpp
@@ -7,6 +7,7 @@
 #include "Strings/CString.h"
 
 #include <gtest/gtest.h>
+#include <string>
 
 namespace GemRB {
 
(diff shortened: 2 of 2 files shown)


lynxlynxlynx:
What am I missing, why is this a valid complaint? An empty FixedSizeString still has the same buffer size.

draghan:
It's not really about the empty case, it's about using  `strncpy` in a "dangerous" manner: `dest` has length of 8, `source` has length of 5, and we are requesting to copy from `source` up to 8 characters.

I see this as some false positive in this particular case - string literal `"NONE"` is null terminated and nothing bad will happen, but to be honest, it's not entirely wrong alarm. The original code has potential to UB:
>  The behavior is undefined (...) if the size of the array pointed to by src is less than count and it does not contain a null character
>
>> // source: https://en.cppreference.com/c/string/byte/strncpy

I'm not sure why only GCC 11 complains about it, perhaps newer versions can properly see that we assign null-terminated array?

draghan:
Hmm, I accidentally looked up the C version of the doc. The C++ version of the doc I quoted doesn't have that UB wording, but I'm pretty sure it will be also illegal.
https://en.cppreference.com/cpp/string/byte/strncpy

MarcelHB:
> I'm not sure why only GCC 11 complains about it, perhaps newer versions can properly see that we assign null-terminated array?

Some years ago, some GCC version, maybe it was 11, was riddled with these findings, but the ones after appear to be fixed, or don't check this under the current settings.

lynxlynxlynx:
The code is careful enough to copy one character less, so even a non-null-terminated source string will end up null terminated just by virtue of the buffer initialization. But it could hypothetically end up with parts of the old string for the difference.

However this test passes just fine for me:
```c++
TEST(CStringTest, Assignment)
{
    auto src = FixedSizeString<8> { "spwi103" };
    src = "1";
    EXPECT_STREQ(src.c_str(), "1");

    src = FixedSizeString<8> { "s" };
    src = "123";
    EXPECT_STREQ(src.c_str(), "123");

    src = FixedSizeString<8> { "spwi103" };
    src = "1234567890";
    EXPECT_STREQ(src.c_str(), "12345678");

    src = FixedSizeString<8> { "spwi103" };
    char noNull[] = { "1234567890" };
    memset(&noNull, 1, sizeof(noNull));
    src = noNull;
    EXPECT_STREQ(src.c_str(), "\x1\x1\x1\x1\x1\x1\x1\x1");
}
```

draghan:
This is all about a scenario like this:
```cpp
    char dest[8];
    char srcNoNull[5];
    memset(&srcNoNull, 1, 5);
    strncpy(dest, srcNoNull, 8); // here implementation of strncpy will try to read `srcNoNull` out of bounds
```

lynxlynxlynx:
Me wanting to ensure the memory ahead was predictable broke the lower size requirement. Fixing that confirms both the problem and fix, thanks.

PR #2528, opened 2026-09-12
Use Amanatides & Woo's algorithm in line of sight checks

draghan:
This PR introduces Amanatides & Woo's algorithm to be used in line of sight (LOS) checks.

## The bug

Every line query in the pathfinder - `GetBlockedInLine`, `IsVisibleLOS`, `IsWalkableTo` - answered its question by simulating a walk along the line with `LineStepper`, the same stepper `Movable::DoStep` uses to move an actor one frame at a time. Reusing a movement stepper for a geometry question caused two bugs:

- **Sampling rate followed actor speed.** `LineStepper`'s step size scales with speed, so a fast actor could stride over a one-tile wall between two samples and never see it. Whether a wall   blocked you depended on how fast you were walking.
- **The walk didn't follow the line.** `LineStepper`'s rounding forces at least one pixel of   progress per axis per step, so a shallow line got walked at 45 degrees until one axis ran out, then straight. This happened regardless of speed, even at speed 0, and on one map put the walk to tiles off the real segment.

fig1_walk_vs_segment

fig2_speed_dependence

fig3_missed_gap

## The fix: GridRayCast

Replaced the stepper-based walk with `GridRayCast`, an integer grid traversal (Amanatides & Woo's "A Fast Voxel Traversal Algorithm for Ray Tracing"). It enumerates exactly the tiles a straight segment crosses, in order, using only integers - no floats, no division, no square root and no actor speed in the equation.

Consequences of dropping speed from the query:

- `actorSpeed` is gone from `GetBlockedInLine`, `GetBlockedInLineTile`, `IsWalkableTo`, and `IsVisibleLOS`.
- `IsVisibleLOS` also loses its `Actor*` parameter. Sight depends on the two endpoints and the map, nothing about who's looking.
- `FindPath` no longer reads `actorContext.speed`.
- Callers were updated in `Map.cpp`, `EffectQueue.cpp`, `GSUtils.cpp`, `Matching.cpp`.

### Corner case

When a ray passes exactly through the point where four tiles meet (`tMaxX == tMaxY`), both axes step at once - the walk is diagonal and never actually inside either tile beside the diagonal.
`GridRayCast` just reports this (`CutACorner()`, `CornerBesideX/Y()`) and leaves the decision to the caller:

- Sight ignores it - a ray has zero width, it can thread the joint between two diagonal walls.
- Walkability blocks only if both corner tiles are blocked - an actor can round one convex corner, but can't thread the joint between two blocked ones.

fig5a_corner_joint_blocked

fig5b_corner_diagonal_allowed

## Fallout: Movable::DoStep couldn't follow straight paths

Once `FindPath` started handing out genuinely straight legs, `DoStep` couldn't walk them, because it moved with the same rounding `NormalizeDeltas()` the line queries used to use. Two separate bugs there:

1. **Rounding killed direction.** `NormalizeDeltas` ceil'd each axis to at least one pixel per frame, so a step of `(1.96, 0.31)` became `(2, 1)` - forcing 45-degree movement on shallow legs. Fixed by adding `ScaleDeltas()`, which scales without rounding. `Movable` now carries the leftover fraction of a pixel across frames and clamps each frame's move to what's left of the leg, so the actor still lands exactly on the waypoint.

2. **Direction was wrong even before rounding.** The old code normalized in pixel space then scaled `dy` by 0.75, which rotates the vector. The actor re-aimed every frame from wherever that  error left it, the drift accumulated: about 1.2 px of drift per walked tile, unbounded with leg length (~25 px drift over 20 tiles, ~48 px over 40). Fixed by normalizing in tile space - one scalar from `hypot(dx/16, dy/12)` applied to both axes, which keeps the direction exact. Drift on the same legs is now flat and under 1 px regardless of length.

`NormalizeDeltas` is kept only for `CalculateRunAwayPoint`/`CalculateRandomWalkPoint`, which must terminate and have nowhere to store a fraction. It's now documented as unsafe to reuse for sampling geometry or for moving an actor along an already-validated path.

Trajectories below are measured frame-by-frame in the engine:

fig4_movement_trajectory

fig6_bend

fig7_bend_leaves_corridor

fig8_curved_paths

## Net effect on pathfinding

Ran 1000 `FindPath` requests on a mixed map, before and after: same 958/958 valid paths found. What changes is the legs making up those paths - 4300 legs before, 4292 after - because the pathfinder now follows actual straight lines instead of the old zig-zag approximation.

## Performance work

`GridRayCast` is on the hot path for every walkability and LOS check. The algorithm was implemented in a branchless manner, with additional few tricks to yield maximum performance.

Aside from the algorithm itself, included a few follow-up commits optimizing for performance in the LOS neighborhood:

- `PathFinder::GetBlockedTile` and `TileProps::QuerySearchMap` moved to inline definitions in their headers. Both classes are `GEM_EXPORT`, and under GCC's `-fPIC` handling an out-of-line method on an exported class is subject to symbol interposition, so GCC refuses to inline through it even from a call site in the same file (confirmed via objdump: a bare PLT jump). Clang doesn't have this restriction. Moving the bodies into the header sidesteps the rule. Added `GEM_LIKELY`/ `GEM_UNLIKELY` (wrapping `__builtin_expect`) to mark `QuerySearchMap`'s bounds check as unlikely - without the hint, once inlined into a walk loop, GCC lays out the never-taken bounds-check path as the fall-through, measured 18% slower on the line query.
- `AccumulateAlongTheLine` replaced a loop-invariant function pointer (dispatching between big/small actor tile lookups) with a bool decided once and a lambda - neither GCC nor Clang folds a call  through a function pointer into a direct call even when the target never changes, so this let the  common (small-actor) path inline all the way to the tile fetch.
- The tile-flag fixup in `GetBlockedTile` (TRAVEL implies PASSABLE, DOOR flags clear/force other flags) collapsed from three dependent branches to a single 256-entry lookup table, since `PathMapFlags` is a `uint8_t` and there are only 256 possible inputs.
- `GridRayCast::Step()` was recomputing its comparison term (`errX * ay - errY * ax`, two 64-bit multiplies) every iteration. Since `errX`/`errY` only ever grow by fixed amounts, the comparison value is now carried forward with a masked add instead.

### Benchmarks

Measured with a synthetic LOS query and path-request microbenchmark, on four machines:

- **bench0**: AMD Ryzen 9 5950X, Linux, gcc
- **bench0w**: AMD Ryzen 9 5950X, Windows, MSVC
- **bench1**: Intel Core i5-4300U, Linux, gcc
- **bench2**: Raspberry Pi 4 (aarch64), Linux, clang

`x` columns are the speedup multiplier vs. the previous row (>1 faster, <1 regression). "Baseline" is the pre-`GridRayCast` vendored `LineStepper` walk.

| SHA | Commit | LOS ns/query (bench0/0w/1/2) | Path us/req (bench0/0w/1/2) | LOS x (bench0/0w/1/2) | Path x (bench0/0w/1/2) |
|---|---|---|---|---|---|
| c634259f9 | **Baseline** (pre-GridRayCast, vendored algorithm) | 3596.5 / 3089.2 / 6704.5 / 10793.8 | 411.5 / 790.0 / 940.3 / 1787.3 | - | - |
| 8477e47ca | Introduce GridRayCast | 183.4 / 203.1 / 464.3 / 904.0 | 97.74 / 158.95 / 221.78 / 420.54 | 19.61 / 15.21 / 14.44 / 11.94 | 4.21 / 4.97 / 4.24 / 4.25 |
| 57f841e2c | GetBlockedTile inline in header | 176.6 / 196.9 / 457.9 / 917.5 | 86.76 / 158.72 / 220.76 / 426.82 | 1.04 / 1.03 / 1.01 / 0.99 | 1.13 / 1.00 / 1.00 / 0.99 |
| 26e084966 | Tile-flag fixup: branches→lookup | 176.1 / 168.8 / 421.6 / 809.3 | 86.14 / 156.52 / 219.47 / 426.87 | 1.00 / 1.17 / 1.09 / 1.13 | 1.01 / 1.01 / 1.01 / 1.00 |
| 848c18c65 | AccumulateAlongTheLine: dispatch fn removed | 167.5 / 166.7 / 392.0 / 750.2 | 87.06 / 158.74 / 214.08 / 418.09 | 1.05 / 1.01 / 1.08 / 1.08 | 0.99 / 0.99 / 1.03 / 1.02 |
| e4bbd21c3 | Inline QuerySearchMap into header | 172.0 / 151.7 / 366.4 / 651.7 | 83.95 / 156.46 / 211.64 / 423.48 | 0.97 / 1.10 / 1.07 / 1.15 | 1.04 / 1.01 / 1.01 / 0.99 |
| 7f750675a | Carry Step()'s comparison term | 216.2 / 138.0 / 311.7 / 530.2 | 83.98 / 155.27 / 206.44 / 410.73 | 0.80 / 1.10 / 1.18 / 1.23 | 1.00 / 1.01 / 1.03 / 1.03 |
| - | **Total vs baseline** | - | - | **16.64 / 22.39 / 21.51 / 20.36** | **4.90 / 5.09 / 4.55 / 4.35** |

## Tests and note on Bresenham's algorithm

Existing pathfinder and movement tests were updated for the new signatures and to accommodate fixed straight LOS and walk.

As a bonus resulting from a side check comparison versus Bresenham's algorithm, added tests that place walls at the specific tiles where a naive Bresenham-style raster line would disagree with the true segment and assert `GridRayCast` gets the geometrically correct answer in every case.

A Bresenham's algorithm for drawing raster line isn't a hypothetical alternative implementation - it's similar rounding bug the old `LineStepper` walk had on shallow lines, so it's worth showing concretely why it's wrong even though it looks at a glance like a reasonable algorithm for LOS checks:

fig9_bresenham_missed_walls


gemrb/core/PathFinder.cpp
```diff
@@ -194,7 +194,6 @@ Path PathFinder::FindPath(const TraversabilityCache::Data_t& traversabilityCache
 
 	const unsigned int actorCircleSize = actorContext.circleSize;
 	const Movable* const actorIdentity = actorContext.identity;
-	const int actorSpeed = actorContext.speed;
```
lynxlynxlynx, line 197: If I'm reading it correctly, this can now be removed from ActorPathContext struct too.

gemrb/tests/core/Test_PathFinder.cpp
```diff
@@ -880,13 +880,14 @@ TEST(FindPathTest, PathUTurn)
...
 		"#####################",
 		"#.........#....@....#",
-		"#.........#...**....#",
-		"#..**.....#..**.....#",
-		"#...**....#.**......#",
-		"#....******@*.......#",
+		"#.........#..**.....#",
+		"#..**.....#.**......#",
+		"#...**....#.*.......#",
+		"#....******@........#",
```
lynxlynxlynx: Hmm, this is not intuitive, since it's not a straight diagonal except directly after the waypoint.

draghan: If you correct for the perspective ratio (16:12):

image

the leg on the right side looks ok-ish. There are still 2 questionable tiles, but to a degree of suspending the doubts. 
But the leg on the left side is clearly wrong, there is no way one leg should stay in that "L" shape, it was the whole point of this PR to get rid of that bend...
I don't know where the issue is, I confirmed the new algo is correct on any path I threw at it.

draghan: oh, I know what's wrong. The tests are still using the old algo to draw between the waypoints ᗒ ͟ʖᗕ

draghan: Corrected.
Both tests where the '*' matcher was used are corrected:
image

image

lynxlynxlynx: Is there any use in keeping LineStepper? Now it's only used to test itself.

draghan: Good point, removed.

gemrb/core/PathFinder.cpp
draghan: Another good point, removed that too. Thanks.

lynxlynxlynx:
Thanks!

PR #2532, opened 2026-09-22
Actor: fix format_error crash when logging an out-of-range stat index

KomendantRitter:


MarcelHB:
Looks like a valid finding though.

lynxlynxlynx:
Yeah, I recreated it, just haven't pushed yet.

